PR #6, opened 2024-04-19
chore: add PR testing workflows

Datron:
## Problem
Superposition doesn't have testing for every PR setup

## Solution
add that using github actions

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,4 @@
-DATABASE_URL=postgres://postgres:docker@dockerdns:5432/config?sslmode=disable
+DATABASE_URL=postgres://postgres:docker@localhost:5432/config?sslmode=disable
 RUST_LOG=debug
 AWS_ACCESS_KEY_ID=test
 AWS_SECRET_ACCESS_KEY=test
@@ -8,7 +8,7 @@ DB_USER=postgres
 DB_HOST=dockerdns:5432
 DB_NAME=config
 APP_ENV=DEV
-AWS_REGION_ENDPOINT=http://dockerdns:4566
+AWS_REGION_ENDPOINT=http://localhost:4566
 ALLOW_SAME_KEYS_OVERLAPPING_CTX=true
 ALLOW_DIFF_KEYS_OVERLAPPING_CTX=true
 ALLOW_SAME_KEYS_NON_OVERLAPPING_CTX=true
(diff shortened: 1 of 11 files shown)


crates/service_utils/src/db/utils.rs
```diff
@@ -11,9 +11,14 @@ use urlencoding::encode;
 
 pub async fn get_database_url() -> String {
     let db_user: String = get_from_env_unsafe("DB_USER").unwrap();
-    let kms_client = kms::new_client();
-    let db_password_raw = kms::decrypt(kms_client, "DB_PASSWORD").await;
-    let db_password = encode(db_password_raw.as_str()).to_string();
+    let app_env: String = get_from_env_or_default("APP_ENV", "TEST".into());
+    let db_password: String = if app_env.as_str() == "TEST" {
```
ShubhranshuSanjeev, line 15: any reason to deviate from the KMS setup for testing?

makefile
```diff
@@ -131,32 +126,14 @@ run: kill build
 		do echo "waiting for postgres, localstack bootup"; \
 		sleep 0.5; \
 		done
-	sed -i 's/dockerdns/$(DOCKER_DNS)/g' ./.env
 	make superposition -e DOCKER_DNS=$(DOCKER_DNS)
 
-ci-test: cleanup ci-setup
```
ShubhranshuSanjeev, line 137: why remove cleanup from here?

.env.example
```diff
@@ -1,4 +1,4 @@
-DATABASE_URL=postgres://postgres:docker@dockerdns:5432/config?sslmode=disable
+DATABASE_URL=postgres://postgres:docker@localhost:5432/config?sslmode=disable
```
ShubhranshuSanjeev, line 1: changing from `dockerdns` to `localhost` as we won't have docker-in-docker, since we are moving away from Jenkins?

crates/service_utils/src/db/utils.rs
Datron: Why do we do this in development environments? Makes easy setups complex. So I wanted to add in an exception for this. Also, KMS encryption is only for juspay, we should remove this from this repo

makefile
Datron: We don't need docker cleanup anymore, github actions automatically cleans up containers

crates/service_utils/src/db/utils.rs
ShubhranshuSanjeev: Sure, make sense.

mahatoankitkumar (approved):

pratikmishra356 (approved):

ShubhranshuSanjeev (approved):

PR #10, opened 2024-04-24
add snapshot while context , default configs updates

pratikmishra356:
## Problem
Record for config's history after every cac_config changes,
we also encountered problem regarding generating snowflake_id, we were not dropping the acquired mutex lock,
and when generating another snowflake_id , it gets stuck in deadlock kind of situation

## Solution
Created config_versions table which will store the config's snapshot after every updates which affect the configs.
Fetch contexts and default_configs and insert into config_Versions table after any updates within the same transaction.
we are explicitly dropping acquired lock after snowflake_id generation
Also accepting tags parameters in every update request so that user can tag every versions
NA

## Pre-deployment activity
Create db changes 

## Post-deployment activity
NA

## API changes
get_config apis we will now provide priority inside context objects
adding optional tags field in every context, default_config and experiments updates
accepting **version** queryparam in every get_config apis

NA


crates/context_aware_config/migrations/2024-04-22-122806_config_verions/up.sql
```diff
@@ -0,0 +1,9 @@
+-- Your SQL goes here
+-- Name: functions; Type: TABLE; Schema: public; Owner: -
+--
+CREATE TABLE public.config_versions (
+    id text PRIMARY KEY,
+    config json,
```
mahatoankitkumar, line 6: I think this should be `Not Null`

crates/context_aware_config/src/db/schema.rs
```diff
@@ -601,6 +601,14 @@ diesel::table! {
 diesel::joinable!(default_configs -> functions (function_name));
 diesel::joinable!(dimensions -> functions (function_name));
 
+diesel::table! {
+    config_versions (id) {
+        id -> Text,
+        config -> Json,
```
mahatoankitkumar, line 607: Currently if you don't do `NOT NULL`, it makes this
`config -> Nullable<Json>,`, Can you please check this?

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/config/types.rs
```diff
@@ -1,16 +1,17 @@
...
     pub overrides: Map<String, Value>,
     pub default_configs: Map<String, Value>,
 }
 
-#[derive(Serialize, Clone)]
+#[derive(Serialize, Clone, Deserialize)]
 pub struct Context {
     pub id: String,
     pub condition: Value,
+    pub priority: i32,
```
mahatoankitkumar, line 15: Are okay with serving `priority` in the `getConfig` response?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -98,24 +99,64 @@ fn is_not_modified(max_created_at: Option<NaiveDateTime>, req: &HttpRequest) ->
     max_created_at.is_some() && parsed_max <= last_modified
 }
 
-async fn generate_cac(
+pub fn generate_config_from_version(
+    version: Option<Value>,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+) -> superposition::Result<Config> {
+    let config_version = match version {
```
mahatoankitkumar, line 106:
```suggestion
    let config = match version {
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -313,7 +356,8 @@ async fn get_filtered_config(
                 .map_or_else(|_| json!(value), |int_val| json!(int_val)),
```
mahatoankitkumar, line 356: I think we can now deprecate this handler.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -276,30 +279,35 @@ fn put(
...
-        })
+    let version_id = generate_snowflake_id(&state)?.to_string();
+    db_conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+        let put_response = put(req, transaction_conn, &user)
+            .map(|resp| Json(resp))
+            .map_err(|err: superposition::AppError| {
+                log::info!("context put failed with error: {:?}", err);
+                err
+            })?;
+        add_config_version(version_id, transaction_conn)?;
```
mahatoankitkumar, line 295: Just a query, should we fail the context/default_config generation, if the updation of `config_version` fails?

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/context/handlers.rs
pratikmishra356: yes

crates/context_aware_config/src/api/config/mod.rs
```diff
@@ -1,4 +1,4 @@
-mod handlers;
+pub mod handlers;
```
Datron, line 1: Why make this public?

crates/context_aware_config/src/api/config/types.rs
Datron: We can, but this shouldn't break the API contract we have with clients @pratikmishra356

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -247,25 +252,23 @@ fn get_put_resp(ctx: Context) -> PutResp {
...
 ) -> superposition::Result<PutResp> {
     use contexts::dsl::contexts;
-    let new_ctx = create_ctx_from_put_req(req, conn, user)?;
+    let new_ctx = create_ctx_from_put_req(req, transaction_conn, user)?;
 
-    if already_under_txn {
-        diesel::sql_query("SAVEPOINT put_ctx_savepoint").execute(conn)?;
-    }
-    let insert = diesel::insert_into(contexts).values(&new_ctx).execute(conn);
+    diesel::sql_query("SAVEPOINT put_ctx_savepoint").execute(transaction_conn)?;
```
Datron, line 265: Why did we change this? Are we saying that all calls to this function are within a transaction. If not, this code will fail.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -442,37 +445,45 @@ async fn delete_context(
...
-        Ok(0) => Err(not_found!("Context Id `{}` doesn't exists", ctx_id)),
-        Ok(_) => {
-            log::info!("{ctx_id} context deleted by {}", user.get_email());
-            Ok(HttpResponse::NoContent().finish())
-        }
-        Err(e) => {
-            log::error!("context delete query failed with error: {e}");
-            Err(unexpected_error!("Something went wrong."))
+    let mut snowflake_generator = state.snowflake_generator.lock().unwrap();
+    let version_id = snowflake_generator.real_time_generate();
```
Datron: Can we generate snowflake ID inside add_config_version function?

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -139,25 +147,29 @@ async fn create(
...
-            "message": "DefaultConfig created/updated successfully."
-        }))),
-        Err(e) => {
-            log::info!("DefaultConfig creation failed with error: {e}");
-            Err(unexpected_error!(
-                "Something went wrong, failed to create DefaultConfig"
-            ))
-        }
-    }
+    let version_id = generate_snowflake_id(&state)?;
```
Datron, line 150: Can we generate version_id inside of add_config_version instead of repeating this code in each handler?

Datron (changes requested):

Datron:
Can we squash all commits into one?

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -139,25 +147,29 @@ async fn create(
...
+                "message": "DefaultConfig created/updated successfully."
+            }))),
+            Err(e) => {
+                log::info!("DefaultConfig creation failed with error: {e}");
+                Err(unexpected_error!(
+                    "Something went wrong, failed to create DefaultConfig"
+                ))
+            }
+        }?;
+        add_config_version(version_id, version_type, transaction_conn)?;
```
mahatoankitkumar, line 170: Do this before doing a match on upsert.

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/config/types.rs
pratikmishra356: yes , its not breaking api contract with clients , since we are adding a field

crates/context_aware_config/src/api/config/handlers.rs
pratikmishra356: why?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -182,7 +193,31 @@ async fn get(
         );
     }
 
-    let mut config = generate_cac(&mut conn).await?;
+    let mut config_version = json!({});
+    if let Some(Value::String(version_id)) = query_params_map.get("version") {
```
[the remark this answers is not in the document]
pratikmishra356: @Datron this api can have dynamic query params fields for dimensions , that is why we cant have specific type for this

crates/context_aware_config/src/api/context/handlers.rs
pratikmishra356: yes @Datron all call to this function are within a transaction

Datron (changes requested):
What happens when someone tries to update tags? Is that possible?

crates/context_aware_config/src/api/config/mod.rs
Datron: What about this?

crates/superposition/src/main.rs
```diff
@@ -134,6 +135,8 @@ async fn main() -> Result<()> {
         return view! { <App app_envs=routes_ui_envs.clone()/> };
     });
 
+    let snowflake_generator = Arc::new(Mutex::new(SnowflakeIdGenerator::new(1, 1)));
```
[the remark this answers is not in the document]
Datron: @pratikmishra356

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -177,12 +221,13 @@ async fn get(
         query_params_map.insert(
             key,
             value
-                .parse::<i32>()
+                .parse::<i64>()
```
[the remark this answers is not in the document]
pratikmishra356: @Datron expecting version_id in query params, which is snowflake_id and it is i64

crates/context_aware_config/src/db/schema.patch
```diff
@@ -0,0 +1,19 @@
...
++++ b/crates/context_aware_config/src/db/schema.rs
+@@ -2,13 +2,13 @@
+ 
+ diesel::table! {
+     config_versions (id) {
+         id -> Int8,
+         config -> Json,
+         config_hash -> Text,
+-        tags -> Nullable<Array<Nullable<Varchar>>>,
++        tags -> Nullable<Array<Varchar>>,
```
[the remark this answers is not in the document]
pratikmishra356: We can concatenate tags with comma and store it as string , but then the DB filter query will not be optimised
https://diesel.rs/guides/configuring-diesel-cli.html
https://diesel.rs/guides/migration_guide.html

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -18,11 +19,12 @@ use diesel::{
...
-use service_utils::service::types::DbConnection;
-use service_utils::{bad_argument, db_error, unexpected_error};
+use serde_json::{from_value, json, Map, Value};
+use service_utils::{
+    bad_argument, db_error, result as superposition, service::types::DbConnection,
+    unexpected_error,
+};
 
-use service_utils::result as superposition;
 use uuid::Uuid;
```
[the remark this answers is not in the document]
pratikmishra356: its getting imported with other imports from service_utils `use service_utils::{
    bad_argument, db_error, result as superposition, service::types::DbConnection,
    unexpected_error,
};`

crates/context_aware_config/src/api/config/mod.rs
pratikmishra356: @Datron its removed

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -98,24 +99,32 @@ fn is_not_modified(max_created_at: Option<NaiveDateTime>, req: &HttpRequest) ->
     max_created_at.is_some() && parsed_max <= last_modified
 }
 
-async fn generate_cac(
+pub fn generate_cac(
```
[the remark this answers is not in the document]
pratikmishra356: moved generate_cac in cac - helpers

crates/superposition/src/main.rs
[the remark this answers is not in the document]
pratikmishra356: @Datron arc is used to share the same mutex instance 
you can check the example for mutex implementation
` /// use std::sync::{Arc, Mutex};
    /// use std::thread;
    ///
    /// let mutex = Arc::new(Mutex::new(0));
    /// let c_mutex = Arc::clone(&mutex);
    ///
    /// thread::spawn(move || {
    ///     let mut lock = c_mutex.try_lock();
    ///     if let Ok(ref mut mutex) = lock {
    ///         **mutex = 10;
    ///     } else {
    ///         println!("try_lock failed");
    ///     }
    /// }).join().expect("thread::spawn failed");
    /// assert_eq!(*mutex.lock().unwrap(), 10);`

pratikmishra356:
> What happens when someone tries to update tags? Is that possible?

@Datron currently this is not supported , a different api will be needed
also we need to have list versions api with filter support 
these changes can be added later on , this pr already has lot of changes

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -98,52 +99,38 @@ fn is_not_modified(max_created_at: Option<NaiveDateTime>, req: &HttpRequest) ->
     max_created_at.is_some() && parsed_max <= last_modified
 }
 
-async fn generate_cac(
+pub fn generate_config_from_version(
+    version: Option<Value>,
```
ShubhranshuSanjeev, line 103:
```suggestion
    version: Option<i64>,
```

Why to parse from a `Value` here in this function, give a `i64` value

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -55,6 +60,8 @@ pub fn endpoints() -> Scope {
         .service(delete_context)
         .service(bulk_operations)
         .service(list_contexts)
+        .service(move_handler)
+        .service(delete_context)
```
ShubhranshuSanjeev, line 64: Why add these routes again? Remove these

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -5,11 +5,18 @@ use serde_json::{Map, Value};
 pub struct PutReq {
     pub context: Map<String, Value>,
     pub r#override: Map<String, Value>,
+    pub tags: Option<Vec<String>>,
```
ShubhranshuSanjeev, line 8: Can we accept tags as headers? These don't really belong here, and the name is also misleading should be `snapshot_tags`

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -147,25 +147,27 @@ async fn create(
...
-        }
-    }
+    conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+        let upsert = diesel::insert_into(default_configs)
+            .values(&default_config)
+            .on_conflict(db::schema::default_configs::key)
+            .do_update()
+            .set(&default_config)
+            .execute(transaction_conn);
+        add_config_version(&state, req.tags, transaction_conn)?;
```
ShubhranshuSanjeev, line 157: @pratikmishra356 , do we have to fail a request just because we weren't able to create a snapshot.
What I think, the user never intended to make a snapshot, he just wanted to create a default_config/context/....., creating a snapshot should be our separate task, I feel we really shouldn't fail the endpoints when we fail to create a snapshot

@knutties @Datron , what are you thoughts on this?

crates/context_aware_config/src/api/config/handlers.rs
pratikmishra356: @ShubhranshuSanjeev this function is getting called at three places , parsing query params there into i64 would be redundant, thats why parsing here at once

crates/context_aware_config/src/api/context/types.rs
pratikmishra356: headers can be used when its value is consistent across similar requests or let say it is a global level identifier for requests . 
although tags is consistent across all write requests , I think we should not add this in header because probably tags can be changed based on the different operations (c, u, d) on different entities (default_configs, contexts)

also we have other entities and request where this header is no need ( like functions, types, get apis ) so for me it probably seems more close to a create request than a global request identifier 
will change tags -> version_tags 

@knutties @Datron your thoughts ?

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: since currently we dont have any process/event based tracking mechanisms to do these operations in async , we might lose track of updates on configs thats why we followed this behavior , 
we assume every updates on default_config and contexts is crucial which results in changing the configs , and since now we are relying only on config_version table for serving configs ( i.e no more db fetch fetch from multiple tables ) it is important to have this both operation in sync

crates/context_aware_config/src/api/context/handlers.rs
pratikmishra356: it got added while merging conflicts will remove this

crates/context_aware_config/src/api/context/handlers.rs
ShubhranshuSanjeev: Can you put this back as it was? Since in future we might move away from doing snapshot in transaction and go ahead with some other method.

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/context/types.rs
ShubhranshuSanjeev: `tags` is a actually a metadata here, and not exactly the part of the request, are we using this to tag the actual resource we are touching in these request, the answer is no, these tags in fact are used to tag the versions created after the actual defined job for the request is completed.

ShubhranshuSanjeev: We are repeating everywhere too, all request types are changed, can't we do something to abstract this out.

sauraww (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -98,52 +118,32 @@ fn is_not_modified(max_created_at: Option<NaiveDateTime>, req: &HttpRequest) ->
...
-            .fold(Map::new(), |mut acc, item| {
-                acc.insert(item.0, item.1);
-                acc
-            });
-
-    Ok(Config {
-        contexts,
-        overrides,
-        default_configs,
+    let config = match version {
```
sauraww, line 125: Can we use if let some here ? 
Match case seems unnecessary here

sauraww (changes requested):

crates/context_aware_config/src/api/context/types.rs
pratikmishra356: added tags in headers

ShubhranshuSanjeev (changes requested):

crates/service_utils/src/service/types.rs
```diff
@@ -208,3 +209,24 @@ impl FromRequest for DbConnection {
         ready(result)
     }
 }
+
+pub struct CustomHeaders {
+    pub config_tags: Option<String>,
```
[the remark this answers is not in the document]
pratikmishra356: I have to perform split on string , so the logic is extract any custom headers as option<string> , and then in the next step we can add parsing and validation logic, here we should not add because these operation is individual based on headers

crates/context_aware_config/src/helpers.rs
```diff
@@ -287,6 +307,86 @@ pub fn calculate_context_priority(
     }
 }
 
+pub fn generate_cac(
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+) -> superposition::Result<Config> {
+    let contexts_vec = ctxt::contexts
```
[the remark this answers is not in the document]
pratikmishra356: this is when we are generating cac for adding in snapshot , this function is not used for serving apis

crates/context_aware_config/migrations/2024-04-22-122806_config_verions/up.sql
```diff
@@ -0,0 +1,14 @@
...
+--
+CREATE TABLE public.config_versions (
+    id bigint PRIMARY KEY,
+    config json NOT NULL,
+    config_hash TEXT NOT NULL,
+    tags varchar(100)[] check (array_position(tags, null) is null),
+    created_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS config_verions_tags_index ON public.config_versions USING gin(tags);
```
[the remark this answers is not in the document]
pratikmishra356: yes we can write db query , at the end it will generate sql statement only, index is added for fast retrieval since there can be multiple rows and with array of tags , performance wise it will be better

crates/superposition/src/main.rs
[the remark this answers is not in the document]
pratikmishra356: @Datron if we remove mutex, then other threads can generate same id at same time

crates/context_aware_config/migrations/2024-04-22-122806_config_verions/up.sql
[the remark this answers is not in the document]
Datron: No the places where this index benefits us is not supported 
https://www.postgresql.org/docs/current/arrays.html
it impacts us on the writes though. We should remove this indexing

crates/superposition/src/main.rs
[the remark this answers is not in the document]
Datron: @pratikmishra356 Why do we need both Arc and Mutex? Use either one

Datron (approved):

crates/superposition/src/main.rs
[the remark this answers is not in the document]
pratikmishra356: @Datron Reason for Using both is 
We are constructing Data(app_state) inside HttpServer::new so this is not being shared across multiple requests , and Mutex method does not have clone trait
so if dont use Arc then for each new request has their own mutex instance which results in same id generation while running tests , 
with Arc we are assuring that there is only one mutex instance stored on heap and shared across multiple requests

crates/service_utils/src/helpers.rs
```diff
@@ -369,3 +373,40 @@ where
...
+        log::error!("snowflake_id generation failed {}", e);
+        result::AppError::UnexpectedError(anyhow!("snowflake_id generation failed {}", e))
+    })?;
+    let id = snowflake_generator.real_time_generate();
+    // explicitly dropping snowflake_generator so that lock is released and it can be acquired in bulk-operations handler
+    drop(snowflake_generator);
+    Ok(id)
+}
+
+pub fn validate_config_tags(
```
ShubhranshuSanjeev, line 387:
```suggestion
pub fn parse_config_tags(
```

I think this function is parsing the header value, than just validating

crates/service_utils/src/service/types.rs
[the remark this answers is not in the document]
ShubhranshuSanjeev: :+1:

crates/service_utils/src/helpers.rs
pratikmishra356: No its also validating the regex match. Splitting string and then regex match on each elements

mahatoankitkumar (approved):

ShubhranshuSanjeev (approved):

sauraww (approved):

PR #11, opened 2024-04-25
fix: Do not lowercase dimension inputs while resolving

ayushjain17:
## Problem
Resolve screen converts all the values of dimensions to lower case

## Solution
Do not lowercase dimension inputs while resolving


mahatoankitkumar (approved):
Just add a before and after screenshot.

ShubhranshuSanjeev (approved):

PR #16, opened 2024-04-30
fix: fixed function template

mahatoankitkumar:
## Problem
fixed function template

## Solution
fixed function template
Screenshot 2024-05-02 at 12 04 42 PM

Screenshot 2024-05-02 at 12 01 04 PM

N.A.

## Pre-deployment activity
N.A.

## Post-deployment activity
N.A.

## API changes
N.A.

N.A.

### DIFF
--- a/crates/frontend/src/pages/function/function_create.rs
+++ b/crates/frontend/src/pages/function/function_create.rs
@@ -26,7 +26,8 @@ pub fn create_function_view() -> impl IntoView {
                     window.editor = monaco.editor.create(document.querySelector('.monaco'), {{
 
                         value: [
-                            'async function validate() {{',
+                            'async function validate(value, key) {{',
+                            '   return true;',
                             '}}'
                         ].join('\n'),
                         language: 'javascript',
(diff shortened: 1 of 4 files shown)


Datron (approved):

crates/frontend/src/pages/function/function_create.rs
```diff
@@ -26,7 +26,7 @@ pub fn create_function_view() -> impl IntoView {
                     window.editor = monaco.editor.create(document.querySelector('.monaco'), {{
 
                         value: [
-                            'async function validate() {{',
+                            'async function validate(value, key) {{',
```
knutties, line 29: Let us add a default return value to the template function as well
```suggestion
                            'async function validate(value, key) {{',
                            
                            return true;
                            }}'
```

knutties: The syntax of the suggestion might be off @mahatoankitkumar  - but you possibly got the point.  Basically want the template to convey that this function has to return a bool.

ShubhranshuSanjeev (approved):

PR #20, opened 2024-05-02
chore(ci): Build using Nix (Ubuntu & macOS)

srid:


.github/workflows/nix.yaml
```diff
@@ -0,0 +1,24 @@
+name: "CI Nix"
+on:
+  # Run only when pushing to main branch, and making PRs
+  push:
```
Datron, line 4: @srid can this be manually triggered instead of on every pull request and merge to main?

srid: @Datron CI should generally run on every PR and on commits to `main`. Why do you want to run it manually? The latter is risky that people may merge broken PRs into main.

Also, I think I should re-open the PR because someone seemed to have force-pushed to `main` branch.

Datron: @srid You can just rebase, should get rid of the conflicts. Sorry for the trouble, I force pushed to main.

> CI should generally run on every PR and on commits to main. Why do you want to run it manually? The latter is risky that people may merge broken PRs into main.
We already have a code check and build with cargo and tests, would `nixci build` do the same thing?

srid: It builds the Rust package (via cargo), runs tests (cargo tests) and effectively makes sure that `flake.nix` remains not-broken for the user.

Other checks in the existing like the one below can also be eventually moved to Nix, so that all of our checks run in CI via Nix. Just like it is in nammayatri. If there's consensus for that, I can help with doing it.

https://github.com/juspay/superposition/blob/01110e2f12dfc9d1af6b48b8f3af1350c68cac9a/.github/workflows/ci_check_pr.yaml#L64-L66

srid:
#27

PR #26, opened 2024-05-03
feat: Add JS experiment client

ayushjain17:
## Problem
Experimentation client not present in JS

## Solution
Add Experimentation client in JS


clients/js/src/index.ts
```diff
@@ -39,3 +39,61 @@ export class CacReader {
...
+                    return variant;
+                }
+            }
+        }
+        const variant_count = applicable_variants.length;
+        const range = traffic * variant_count;
+        if (toss >= range) {
+            return undefined;
+        }
+        let buckets = Array.apply(null, Array(variant_count)).map(function (_, i) { return traffic * (i + 1); });
```
ShubhranshuSanjeev, line 95: This is fine, but too computations happening here, making an array of `NULL`s, mapping over it, and then in the finding the index of the bucket and finally using the index to get the variant/bucket.
I think this can be just written in one single `for` loop.

ShubhranshuSanjeev (changes requested):

clients/js/src/index.ts
```diff
@@ -39,3 +39,61 @@ export class CacReader {
         return deepMerge(targetConfig, ...requiredOverrides);
     }
 }
+
+export class ExperimentReader {
+    experiments: ExperimentStore;
```
Datron, line 44: `ExperimentStore` is not really required, you could just use `Experiments` here

clients/js/src/index.ts
```diff
@@ -39,3 +39,61 @@ export class CacReader {
...
+            }
+        }
+        return variants;
+    };
+
+    getSatisfiedExperiments(data: IObject): Experiments {
+        const experiments = [];
+        const iter = this.experiments.values();
+
+        while (true) {
```
Datron, line 66: Can we just use a for loop? Infinite loops are dangerous and there are a limited number of experiments

clients/js/src/index.ts
```diff
@@ -39,3 +39,61 @@ export class CacReader {
...
 }
+
+export class ExperimentReader {
+    experiments: ExperimentStore;
+
+    constructor(experiments: Experiments) {
+        this.experiments = new Map(experiments.map(exp => [exp.id, exp]));
+    }
+
+    public getApplicableVariant(data: IObject, toss: number): Array<String> {
```
Datron, line 50: the type `number` allows floating point values too, can we add an assertion that toss is always a positive integer?

clients/js/src/index.ts
```diff
@@ -39,3 +39,61 @@ export class CacReader {
...
+                experiments.push(result.value);
+            }
+        }
+
+        return experiments;
+    };
+
+    // decide which variant to return among all applicable experiments
+    decideVariant(
+        traffic: number,
```
Datron, line 79: the type `number` encompasses decimal, float, hex, int, etc in JS. Can we check that all numbers provided are a positive integer?

clients/js/src/types.ts
```diff
@@ -13,4 +13,36 @@ export type DimensionConfig = {
...
+    INPROGRESS = "INPROGRESS",
+    CONCLUDED = "CONCLUDED",
+}
+
+export type Experiment = {
+    variants: Variants,
+    name: String,
+    id: String,
+    traffic_percentage: number,
+    context: Object,
```
Datron, line 42: Can context be a concrete type as well? It does have a definition in rust that can be used here.

Datron (changes requested):

clients/js/src/index.ts
ayushjain17: we cannot add restriction of positive integers only, to test EXPERIMENT variant we need to send -1

Datron: Sure, you can the allowed range as `(-1, 100)`

clients/js/src/types.ts
ayushjain17: Its of raw jsonlogic context type

ShubhranshuSanjeev (approved):

clients/js/src/index.ts
```diff
@@ -39,3 +39,52 @@ export class CacReader {
...
+        applicable_variants: Variants,
+        toss: number,
+    ): Variant | undefined {
+        if (!Number.isInteger(toss)) {
+            return undefined;
+        }
+        if (toss < 0) {
```
Datron, line 75: can we check that toss is an integer too?

clients/js/src/index.ts
Datron: Can we add the toss isInteger check here?

Datron (changes requested):

clients/js/src/index.ts
ayushjain17: have added the check [here](https://github.com/juspay/superposition/pull/26/files/f0b8b8802e60d6325d131caf71de0107c37b8424#diff-8392fd7ebcc540bad7b0c0d66740da3daf807d32253418c975825c96fa881bb5R72)

Datron (approved):

PR #28, opened 2024-05-06
Picaf 26152 reduce context tool

Sauravcv98:
## Problem
Reducing duplicate Contexts, which won't be used

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


Datron:
@Sauravcv98 can we squash these commits into 1?

Datron:
@Sauravcv98 no PICAF in github commits

PR #34, opened 2024-05-07
feat: delete default config api

pratikmishra356:
## Problem
Currently we dont have delete default config api because of whicj we had to execute db commands

## Solution
Created Delete default config api, key will deleted only when none of context's overrides contains this key

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
| {{host}}//default-config/{{key}}      | DELETE | na | 204 No content, Given key present in these contexts [] |


crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -1,35 +1,45 @@
...
     result as superposition,
     service::types::{AppState, DbConnection},
 };
 
 pub fn endpoints() -> Scope {
-    Scope::new("").service(create).service(get)
+    Scope::new("")
+        .service(create)
+        .service(get)
+        .service(delete_default_config)
```
mahatoankitkumar, line 42:
```suggestion
        .service(delete)
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+                unexpected_error!(err)
+            })?
+            .get(key)
+            .map_or((), |_| context_ids.push(context.id.to_owned()))
+    }
+    Ok(context_ids)
+}
+
+#[delete("/{key}")]
+async fn delete_default_config(
```
mahatoankitkumar, line 221:
```suggestion
async fn delete(
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+    path: Path<String>,
+    db_conn: DbConnection,
+    user: User,
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+
+    let key = path.into_inner();
+    fetch_default_key(&key, &mut conn)?;
+    let context_ids = check_key_in_context_overrides(&key, &mut conn)?;
+    if context_ids.len() == 0 {
```
mahatoankitkumar, line 231:
```suggestion
    if !context_ids.len.is_empty {
        Err(bad_argument!(
                "Given key present in these contexts: {}, please first remove keys from there",
                context_ids.join(",")
            ))
     }
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+    key: &String,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+) -> superposition::Result<Vec<String>> {
+    let result: Vec<Context> = contexts.load(conn).map_err(|err| {
+        log::error!("failed to fetch contexts with error: {}", err);
+        db_error!(err)
+    })?;
+
+    let mut context_ids = vec![];
+    for context in result.iter() {
```
mahatoankitkumar, line 208: use filter_map.

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: **delete** name collides with the diesel::delete function , thats why kept this as delete_default_config

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: filter_map wont work here , because we need to raise error and return , in filter_map we probably cant I tried

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: done

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: delete name collides with the diesel::delete function , thats why kept this as delete_default_config,
we might use diesel::delete at multiple places , so its better to rename one time used declarations

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+                Ok(HttpResponse::NoContent().finish())
+            }
+            Err(e) => {
+                log::error!("default config delete query failed with error: {e}");
+                Err(unexpected_error!("Something went wrong."))
+            }
+        }
+    } else {
+        Err(bad_argument!(
+            "Given key present in these contexts: {}, please first remove keys from there",
```
ShubhranshuSanjeev, line 248:
```suggestion
            "Given key already in use in contexts: {}",
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+
+    let key = path.into_inner();
+    fetch_default_key(&key, &mut conn)?;
+    let context_ids = check_key_in_context_overrides(&key, &mut conn)?;
+    if context_ids.is_empty() {
+        let deleted_row =
+            delete(default_configs.filter(db::schema::default_configs::key.eq(&key)))
+                .execute(&mut conn);
+        match deleted_row {
+            Ok(0) => Err(not_found!("default config Key `{}` doesn't exists", key)),
```
ShubhranshuSanjeev, line 236:
```suggestion
            Ok(0) => Err(not_found!("default config key `{}` doesn't exists", key)),
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+async fn delete_default_config(
+    path: Path<String>,
+    db_conn: DbConnection,
+    user: User,
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+
+    let key = path.into_inner();
+    fetch_default_key(&key, &mut conn)?;
+    let context_ids = check_key_in_context_overrides(&key, &mut conn)?;
```
ShubhranshuSanjeev, line 230: `check_key_in_context_overrides` is an internal function, any error returned by should not be given back in response, rather a simple message as `Something went wrong` should be good enough.
we can use `map_err` here to achieve the same.

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
     let result: Vec<DefaultConfig> = default_configs.get_results(&mut conn)?;
     Ok(Json(result))
 }
+
+pub fn check_key_in_context_overrides(
+    key: &String,
```
ShubhranshuSanjeev, line 199:
```suggestion
    key: &str,
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
     let result: Vec<DefaultConfig> = default_configs.get_results(&mut conn)?;
     Ok(Json(result))
 }
+
+pub fn check_key_in_context_overrides(
```
ShubhranshuSanjeev, line 198: The function name suggest it just checks, but we are returning back list of context ids where the key is being used, either change the function name or the result value type.

ShubhranshuSanjeev (changes requested):
Can you also make the necessary UI changes also to delete default config, in this PR itself?

crates/context_aware_config/src/api/default_config/handlers.rs
mahatoankitkumar: @ShubhranshuSanjeev the point is to inform the user about the key being used in the following contexts.

crates/context_aware_config/src/api/default_config/handlers.rs
mahatoankitkumar: Maybe you can name it `getKeyUsageContextList `

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: @ShubhranshuSanjeev this function returns two types of error db_error and unexpected error, 
both types were already handled in macro, message were passed according to that only

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+        log::error!("failed to fetch contexts with error: {}", err);
+        db_error!(err)
+    })?;
+
+    let mut context_ids = vec![];
+    for context in result.iter() {
+        from_value::<Map<String, Value>>(context.override_.to_owned())
+            .map_err(|err| {
+                log::error!("failed decode override into object: {}", err);
+                unexpected_error!(err)
```
ShubhranshuSanjeev, line 212: Passing any parsing error to the user is not needed, that's the point I am trying to make.

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,59 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
     Ok(Json(result))
 }
+
+pub fn get_key_usage_context_ids(
+    key: &str,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+) -> superposition::Result<Vec<String>> {
+    let result: Vec<Context> = contexts.load(conn).map_err(|err| {
+        log::error!("failed to fetch contexts with error: {}", err);
+        db_error!(err)
```
ShubhranshuSanjeev, line 204: Nor the DB error, user doesn't have to know that we have failed to load the context

ShubhranshuSanjeev: We can return the exact error from here.
But in the handler, its not really required to return the same error, why the user has to know that a db load failed to happen for context, they are not even making a request for context, its related to default_config, why to even provide a error that, its just an internal operation.

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: yes, but unexpected_error anyway ignores messages in the implementation

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: yes agree on this

crates/context_aware_config/src/api/default_config/handlers.rs
ShubhranshuSanjeev: Ok, sure

crates/context_aware_config/src/api/default_config/handlers.rs
ShubhranshuSanjeev: Ok sure

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -184,3 +194,60 @@ async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultCon
...
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+
+    let key = path.into_inner();
+    fetch_default_key(&key, &mut conn)?;
+    let context_ids = get_key_usage_context_ids(&key, &mut conn)
+        .map_err(|e| unexpected_error!("Something went wrong"))?;
+    if context_ids.is_empty() {
+        let deleted_row =
+            delete(default_configs.filter(db::schema::default_configs::key.eq(&key)))
```
mahatoankitkumar, line 234:
```suggestion
            diesel::delete(default_configs.filter(db::schema::default_configs::key.eq(&key)))
```

Now we can rename the function to just `delete` instead of `delete_default_config`.

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: done

mahatoankitkumar (approved):

crates/context_aware_config/src/api/default_config/handlers.rs
ShubhranshuSanjeev: +1, @pratikmishra356 can you make the changes

ShubhranshuSanjeev: I guess this is resolved

pratikmishra356: its already done

ShubhranshuSanjeev (approved):

PR #35, opened 2024-05-07
feat: filter support in experimentation client

mahatoankitkumar:
## Problem
Filter prefix from experiments.

## Solution
Added Filter prefix support in experimentation client.

### DIFF
--- a/crates/experimentation_client/src/interface.rs
+++ b/crates/experimentation_client/src/interface.rs
@@ -182,16 +182,17 @@ pub extern "C" fn get_applicable_variant(
         },
         Err(err) => return error_block(err),
     };
-    // println!("Fetching variantIds");
     let local = task::LocalSet::new();
-    let variants = local.block_on(&Runtime::new().unwrap(), unsafe {
+    let variants_result = local.block_on(&Runtime::new().unwrap(), unsafe {
         (*client).get_applicable_variant(&context, toss as i8)
     });
-    // println!("variantIds: {:?}", variants);
-    match serde_json::to_string::<Vec<String>>(&variants) {
-        Ok(result) => rstring_to_cstring(result).into_raw(),
-        Err(err) => error_block(err.to_string()),
-    }
+    variants_result
+        .map(|result| {
+            serde_json::to_string(&result)
+                .map(|json| rstring_to_cstring(json).into_raw())
+                .unwrap_or_else(|err| error_block(err.to_string()))
+        })
+        .unwrap_or_else(|err| error_block(err.to_string()))
 }
 
 #[no_mangle]
@@ -209,16 +210,15 @@ pub extern "C" fn get_satisfied_experiments(
 
     let local = task::LocalSet::new();
     let experiments = local.block_on(&Runtime::new().unwrap(), unsafe {
-        (*client).get_satisfied_experiments(&context)
+        (*client).get_satisfied_experiments(&context, None)
     });
     let experiments = match serde_json::to_value(experiments) {
         Ok(value) => value,
         Err(err) => return error_block(err.to_string()),
     };
-    match serde_json::to_string(&experiments) {
-        Ok(result) => rstring_to_cstring(result).into_raw(),
-        Err(err) => error_block(err.to_string()),
-    }
+    serde_json::to_string(&experiments)
+        .map(|exp| rstring_to_cstring(exp).into_raw())
+        .unwrap_or_else(|err| error_block(err.to_string()))
 }
 
 #[no_mangle]
--- a/crates/experimentation_client/src/lib.rs
+++ b/crates/experimentation_client/src/lib.rs
@@ -1,16 +1,21 @@
 mod interface;
 mod types;
-use std::{collections::HashMap, sync::Arc};
+mod utils;
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use chrono::{DateTime, TimeZone, Utc};
 use derive_more::{Deref, DerefMut};
-use serde_json::Value;
+use serde_json::{Map, Value};
 use tokio::{
     sync::RwLock,
     time::{self, Duration},
 };
 pub use types::{Config, Experiment, Experiments, Variants};
 use types::{ExperimentStore, ListExperimentsResponse, Variant, VariantType};
+use utils::MapError;
 
 #[derive(Clone, Debug)]
 pub struct Client {
@@ -52,7 +57,7 @@ impl Client {
                     self.client_config.tenant.to_string(),
                 )
                 .await
-                .unwrap();
+                .unwrap_or(HashMap::new());
 
                 let mut exp_store = self.experiments.write().await;
                 for (exp_id, experiment) in experiments.into_iter() {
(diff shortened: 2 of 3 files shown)


crates/experimentation_client/src/lib.rs
```diff
@@ -24,15 +30,16 @@ pub struct Client {
 // DO NOT let panics show up in library
 
 impl Client {
-    pub fn new(config: Config) -> Self {
-        Client {
+    pub fn new(config: Config) -> Result<Self, String> {
```
pratikmishra356, line 33: can you use app error here also

crates/experimentation_client/src/lib.rs
```diff
@@ -1,16 +1,22 @@
 mod interface;
 mod types;
-use std::{collections::HashMap, sync::Arc};
+mod utils;
+use std::{
+    collections::{HashMap, HashSet},
+    convert::identity,
```
pratikmishra356, line 6: its not getting used can you remove this

crates/experimentation_client/src/lib.rs
```diff
@@ -69,34 +76,98 @@ impl Client {
...
+        let filtered_running_experiments = running_experiments
             .iter()
             .filter(|(_, exp)| {
                 jsonlogic::apply(&exp.context, context) == Ok(Value::Bool(true))
             })
             .map(|(_, exp)| exp.clone())
-            .collect::<Experiments>()
+            .collect::<Experiments>();
+
+        if let Some(prefix) = context.get("prefix") {
```
pratikmishra356, line 109: can we use different param for prefix, since its not related to the context instead with the filter,
we also might need to change this in cac_client

crates/experimentation_client/src/lib.rs
```diff
@@ -69,34 +76,98 @@ impl Client {
...
+            let prefix_list: HashSet<&str> = prefix
+                .as_str()
+                .ok_or_else(|| {
+                    log::error!("Prefix is not a valid string.");
+                    format!("Prefix is not a valid string.")
+                })
+                .map_err_to_string()?
+                .split(",")
```
pratikmishra356, line 117: can we expect prefix as array of strings instead of comma separated string(since this is not query params of any api),

crates/experimentation_client/src/lib.rs
```diff
@@ -69,34 +76,98 @@ impl Client {
...
+                .collect();
+
+            let prefix_filtered_running_experiments: Vec<Experiment> =
```
pratikmishra356, line 120: @mahatoankitkumar this implementation might be inefficient at some point of time, 
can you instead include override_keys in experiment struct and , then we can directly filter out based on that

mahatoankitkumar: Why inefficient?

mahatoankitkumar: Why inefficient?

pratikmishra356: its inefficient because you are traversing over all the variants then repeated traversal over overrides of all the variants, 
and unnecessary skip for variants if decodes from_Value to value fails
so adding override_keys in experiment struct would be clean and better

mahatoankitkumar: Sure will simplify this.

mahatoankitkumar: But you can't add override_keys in the experiment struct they will be different for different variants.

pratikmishra356: No for all variants inside the same experiment override_keys will be identical , we are enforcing that in api

mahatoankitkumar: values are different.

pratikmishra356: values can be different , its not relevant in this case since we are filtering on keys of overrides not the values

crates/experimentation_client/src/lib.rs
```diff
@@ -69,34 +76,98 @@ impl Client {
         }
     }
 
-    pub async fn get_applicable_variant(&self, context: &Value, toss: i8) -> Vec<String> {
-        let experiments: Experiments = self.get_satisfied_experiments(context).await;
+    pub async fn get_applicable_variant(
+        &self,
+        context: &Value,
+        toss: i8,
+    ) -> Result<Vec<String>, String> {
```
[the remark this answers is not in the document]
Datron: You can't use AppError here, this will be integrated in other codebases that may or may not understand the struct AppError

crates/experimentation_client/src/interface.rs
```diff
@@ -187,9 +187,11 @@ pub extern "C" fn get_applicable_variant(
     let variants = local.block_on(&Runtime::new().unwrap(), unsafe {
         (*client).get_applicable_variant(&context, toss as i8)
     });
-    // println!("variantIds: {:?}", variants);
-    match serde_json::to_string::<Vec<String>>(&variants) {
-        Ok(result) => rstring_to_cstring(result).into_raw(),
+    match variants {
```
Datron, line 190: Can we use map() on variants so that it avoids this match statement? map_err for the error and you can end with ?

Datron (approved):

crates/experimentation_client/src/lib.rs
Datron: It makes sense to filter on the override_keys already present in Experiments struct

Datron (changes requested):

crates/experimentation_client/src/lib.rs
mahatoankitkumar: We are removing the keys from the overrides if that doesn't match the prefix filter, so will have to traverse all the variants in the overrides to mutate it.

crates/experimentation_client/src/lib.rs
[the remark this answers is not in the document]
mahatoankitkumar: Wanted to maintain consistency everywhere. Maybe in future.

crates/experimentation_client/src/interface.rs
mahatoankitkumar: This function does not return result type.

Datron: How does the match statement work?

crates/experimentation_client/src/lib.rs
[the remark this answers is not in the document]
Datron: Can we not do it then? Lets add it when the error comes

Datron (changes requested):
Had some questions

crates/experimentation_client/src/interface.rs
```diff
@@ -187,9 +187,11 @@ pub extern "C" fn get_applicable_variant(
     let variants = local.block_on(&Runtime::new().unwrap(), unsafe {
```
sauraww, line 187: Can we remove this line ?
  // println!("Fetching variantIds");

crates/experimentation_client/src/interface.rs
```diff
@@ -187,9 +187,11 @@ pub extern "C" fn get_applicable_variant(
...
         (*client).get_applicable_variant(&context, toss as i8)
     });
-    // println!("variantIds: {:?}", variants);
-    match serde_json::to_string::<Vec<String>>(&variants) {
-        Ok(result) => rstring_to_cstring(result).into_raw(),
+    match variants {
+        Ok(result) => match serde_json::to_string::<Vec<String>>(&result) {
```
sauraww, line 191:
```
variants_result
    .map(|result| {
        serde_json::to_string(&result)
            .map(|json| rstring_to_cstring(json).into_raw())
            .unwrap_or_else(|err| error_block(err.to_string()))
    })
    .unwrap_or_else(|err| error_block(err.to_string()))
```

crates/experimentation_client/src/interface.rs
```diff
@@ -209,7 +211,7 @@ pub extern "C" fn get_satisfied_experiments(
 
     let local = task::LocalSet::new();
     let experiments = local.block_on(&Runtime::new().unwrap(), unsafe {
-        (*client).get_satisfied_experiments(&context)
+        (*client).get_satisfied_experiments(&context, None)
     });
     let experiments = match serde_json::to_value(experiments) {
         Ok(value) => value,
```
sauraww, line 217:
```
let experiments = serde_json::to_value(experiments)
    .unwrap_or_else(|err| return error_block(err.to_string()));
```

sauraww: Can we make this change at all the places ? 
Ideally match statements should be used only for the cases where we have more than 2 conditions to match . 
For these cases , we can directly leverage map .

crates/experimentation_client/src/lib.rs
```diff
@@ -69,34 +74,92 @@ impl Client {
...
+                                    .filter(|(key, _)| {
+                                        prefix_list
+                                            .iter()
+                                            .any(|prefix_str| key.starts_with(prefix_str))
+                                    })
+                                    .collect();
+                                variant.overrides =
+                                    serde_json::to_value(filtered_override.clone())
+                                        .ok()?;
+                                if filtered_override.is_empty() {
```
sauraww, line 134: I think this check should be before calculating variant.overrides , we can save one operation here: 
```
        if filtered_override.is_empty() {
            return None; // Skip this variant
        }
        
        variant.overrides = serde_json::to_value(filtered_override).ok()?;
        Some(variant)
    })
    .collect();

```
Also you wouldn't need the else statement if you change the control structure.

sauraww (changes requested):

crates/experimentation_client/src/interface.rs
mahatoankitkumar: It's already removed, you seeing the code on the opposite(removed) side.

crates/experimentation_client/src/utils/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod core;
```
Datron, line 1: You can just put the code of core.rs into this file

pratikmishra356 (approved):

Datron (approved):

sauraww (approved):

PR #37, opened 2024-05-07
fix: fixed condition parsing and display

ShubhranshuSanjeev:
## Problem
We had inconsistent parsing logic for context, which led to wrong data display of context.

## Solution
Fix the parsing logic

### DIFF
--- a/crates/frontend/src/components/condition_pills/types.rs
+++ b/crates/frontend/src/components/condition_pills/types.rs
@@ -1,6 +1,145 @@
+use std::fmt::Display;
+
+use serde_json::{Map, Value};
+
+use crate::types::Context;
+
+#[derive(Debug, Clone)]
+pub enum ConditionOperator {
+    Is,
+    In,
+    Has,
+    Between,
+    Other(String),
+}
+
+impl Display for ConditionOperator {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Has => f.write_str("has"),
+            Self::Is => f.write_str("is"),
+            Self::In => f.write_str("in"),
+            Self::Between => f.write_str("between"),
+            Self::Other(o) => f.write_str(o),
+        }
+    }
+}
+
+impl From<(String, &Vec<Value>)> for ConditionOperator {
+    fn from(value: (String, &Vec<Value>)) -> Self {
+        let (operator, operands) = value;
+        let operand_0 = operands.get(0);
+        let operand_1 = operands.get(1);
+        let operand_2 = operands.get(2);
+        match (operator.as_str(), operand_0, operand_1, operand_2) {
+            // assuming there will be only two operands, one with the dimension name and other with the value
+            ("==", _, _, None) => ConditionOperator::Is,
+            ("<=", Some(_), Some(Value::Object(a)), Some(_)) if a.contains_key("var") => {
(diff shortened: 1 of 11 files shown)


mahatoankitkumar:
Would be helpful if you could add before after screenshots.

ShubhranshuSanjeev:
> Would be helpful if you could add before after screenshots.

No UI changes

crates/frontend/src/components/condition_pills/types.rs
```diff
@@ -1,6 +1,122 @@
...
+        if let Some(operator) = source.keys().next() {
+            let emty_vec = vec![];
+            let operands = source[operator].as_array().unwrap_or(&emty_vec);
+
+            let operator = ConditionOperator::from((operator.to_string(), operands));
+
+            let dimension_name = operands
+                .iter()
+                .find_map(|item| match item.as_object() {
+                    Some(o) if o.contains_key("var") => {
```
pratikmishra356, line 70: can we have better name instead of "o"

Sauravcv98 (approved):

crates/frontend/src/components/condition_pills/types.rs
```diff
@@ -1,6 +1,122 @@
...
 }
+
+impl TryFrom<&Map<String, Value>> for Condition {
+    type Error = &'static str;
+    fn try_from(source: &Map<String, Value>) -> Result<Self, Self::Error> {
+        if let Some(operator) = source.keys().next() {
+            let emty_vec = vec![];
+            let operands = source[operator].as_array().unwrap_or(&emty_vec);
+
+            let operator = ConditionOperator::from((operator.to_string(), operands));
```
pratikmishra356, line 65:
```suggestion
            let operator = ConditionOperator::from((operator.to_owned(), operands));
```

crates/frontend/src/components/condition_pills/types.rs
```diff
@@ -1,6 +1,122 @@
...
+#[derive(Debug, Clone)]
+pub enum ConditionOperator {
+    Is,
+    In,
+    Has,
+    Between,
+    Other(String),
+}
+
+impl Display for ConditionOperator {
```
sauraww, line 14: Can we use strum here ?

ShubhranshuSanjeev: There was no way in strum to serialize this `Other(String)` into to just use the inner value, so went ahead with this approach.

sauraww (approved):

crates/frontend/src/components/condition_pills/types.rs
ShubhranshuSanjeev: hey, `o` is just a throw away variable not even being used anywhere instead of that `Some()` arm, if its affecting readability I will change, if not lets go with this.
In my opinion I don't really think we need a full word for variable name for these kind of scenarios.

Sauravcv98 (approved):

sauraww (approved):

crates/frontend/src/components/condition_pills/types.rs
pratikmishra356: you are using o in` o.contains_key` and in `o["var"]`, when you are operating on any variable it would be better if it is not just o,a,b,c,d single alphabet, 
atleast `val` would work

pratikmishra356 (approved):

Issue #33, opened 2024-05-07
When an experiment is created from UI (and no operator is picked in context) - the operator defaults to has


Datron:
When an experiment is created from UI (and no operator is picked in context) - the operator defaults to has - we might need to default to is - we should force them to select one here.  Additionally a context with has operator and is operator isn’t being merged

PR #38, opened 2024-05-07
fix: creating experiments for default-config (no context)

Sauravcv98:
## Problem
We were not able to create experiments and conclude them for default config, that is without giving any contexts

## Solution
Made contexts optional, if it's None then we accept it as default config

## Pre-deployment activity
DB schema has been changed for experiments table, made context as nullable before it was not null

## Post-deployment activity

## API changes

### DIFF
--- a/crates/experimentation_platform/src/api/experiments/handlers.rs
+++ b/crates/experimentation_platform/src/api/experiments/handlers.rs
@@ -315,13 +315,33 @@ pub async fn conclude(
         })?;
 
         if variant.id == winner_variant_id {
-            let context_move_req = ContextMoveReq {
-                context: experiment_context.clone(),
-            };
+            if !experiment_context.is_empty() {
+                let context_move_req = ContextMoveReq {
+                    context: experiment_context.clone(),
+                };
+                operations.push(ContextAction::MOVE((context_id, context_move_req)));
+            } else {
+                for (key, val) in variant.overrides {
+                    let mut create_req = Map::new();
+                    create_req.insert("value".to_string(), val);
+                    let http_client = reqwest::Client::new();
+                    let url = state.cac_host.clone()
+                        + format!("/default-config/{key}").as_str();
+                    let _ = http_client
+                        .put(&url)
+                        .header("x-tenant", tenant.as_str())
+                        .header(
+                            "Authorization",
+                            format!("{} {}", user.get_auth_type(), user.get_auth_token()),
+                        )
+                        .json(&create_req)
+                        .send()
+                        .await;
+                }
+                operations.push(ContextAction::DELETE(context_id));
+            }
 
             is_valid_winner_variant = true;
-
-            operations.push(ContextAction::MOVE((context_id, context_move_req)));
         } else {
             // delete this context
             operations.push(ContextAction::DELETE(context_id));
--- a/crates/experimentation_platform/src/api/experiments/handlers.rs
+++ b/crates/experimentation_platform/src/api/experiments/handlers.rs
@@ -5,6 +5,7 @@ use actix_web::{
     web::{self, Data, Json, Query},
     HttpRequest, HttpResponse, Scope,
 };
+use anyhow::anyhow;
 use chrono::{DateTime, Duration, NaiveDateTime, Utc};
 use diesel::{
     r2d2::{ConnectionManager, PooledConnection},
@@ -12,12 +13,16 @@ use diesel::{
 };
 
 use service_utils::{
-    bad_argument, response_error, result as superposition, unexpected_error,
+    bad_argument,
+    helpers::{construct_request_headers, request},
+    response_error,
+    result::{self as superposition, AppError},
+    unexpected_error,
 };
 
 use superposition_types::{SuperpositionUser, User};
 
-use reqwest::{Response, StatusCode};
+use reqwest::{Method, Response, StatusCode};
 use service_utils::service::types::{AppState, DbConnection, Tenant};
 
 use super::{
@@ -322,21 +327,27 @@ pub async fn conclude(
                 operations.push(ContextAction::MOVE((context_id, context_move_req)));
             } else {
                 for (key, val) in variant.overrides {
-                    let mut create_req = Map::new();
-                    create_req.insert("value".to_string(), val);
-                    let http_client = reqwest::Client::new();
-                    let url = state.cac_host.clone()
-                        + format!("/default-config/{key}").as_str();
-                    let _ = http_client
-                        .put(&url)
-                        .header("x-tenant", tenant.as_str())
-                        .header(
+                    let create_req = HashMap::from([("value", val)]);
+
+                    let url = format!("{}/default-config/{}", state.cac_host, key);
+
+                    let headers = construct_request_headers(&[
+                        ("x-tenant", tenant.as_str()),
+                        (
                             "Authorization",
-                            format!("{} {}", user.get_auth_type(), user.get_auth_token()),
-                        )
-                        .json(&create_req)
-                        .send()
-                        .await;
+                            &format!(
+                                "{} {}",
+                                user.get_auth_type(),
+                                user.get_auth_token()
+                            ),
+                        ),
+                    ])
+                    .map_err(|err| AppError::UnexpectedError(anyhow!(err)))?;
+
+                    let _ =
+                        request::<_, Value>(url, Method::PUT, Some(create_req), headers)
+                            .await
+                            .map_err(|err| AppError::UnexpectedError(anyhow!(err)))?;
                 }
                 operations.push(ContextAction::DELETE(context_id));
             }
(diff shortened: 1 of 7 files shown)


crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -316,13 +306,43 @@ pub async fn conclude(
...
+                };
+                operations.push(ContextAction::MOVE((context_id, context_move_req)));
+            } else {
+                for (key, val) in variant.overrides {
+                    let mut create_req = Map::new();
+                    create_req.insert("value".to_string(), val);
+                    let http_client = reqwest::Client::new();
+                    let url = state.cac_host.clone()
+                        + format!("/default-config/{key}").as_str();
+                    let _ = http_client
```
mahatoankitkumar, line 331: use `request` function for api calls.

mahatoankitkumar (changes requested):

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -189,39 +197,44 @@ pub fn validate_experiment(
...
-
-    let mut updated_ctx = Map::new();
-    updated_ctx.insert(String::from("and"), serde_json::Value::Array(conditions));
 
-    match serde_json::to_value(updated_ctx) {
-        Ok(value) => Ok(value),
-        Err(_) => Err(bad_argument!(
-            "Failed to convert context to a valid JSON object. Check the request sent for correctness"
-        )),
+    if let Some(context_json) = opt_context_json {
```
mahatoankitkumar, line 211: Lets do case match for `if let` conditions having else condition too.
Improves readability.

crates/experimentation_platform/tests/experimentation_tests.rs
```diff
@@ -11,8 +11,8 @@ enum Dimensions {
     CLIENT(String),
 }
 
-fn single_dimension_ctx_gen(value: Dimensions) -> serde_json::Value {
-    match value {
+fn single_dimension_ctx_gen(value: Dimensions) -> Option<serde_json::Value> {
```
mahatoankitkumar, line 14: why return optional, if you are explicitly adding Some at the return statement.

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -51,33 +51,41 @@ pub fn validate_override_keys(override_keys: &Vec<String>) -> superposition::Res
...
+                is_overlapping = is_overlapping && test;
+
+                if !test {
+                    break;
+                }
+            }
 
-        if !test {
-            break;
+            Ok(is_overlapping)
```
mahatoankitkumar, line 86: Let's refactor this piece of code
Below code should work, though check it once.

```suggestion
           let dimensions_a = extract_dimensions(context_a)?;
            let dimensions_b = extract_dimensions(context_b)?;

            let (ref_dimensions, other_dimensions) = if dimensions_a.len() > dimensions_b.len() {
                (dimensions_b, dimensions_a)
            } else {
                (dimensions_a, dimensions_b)
            };

            let is_overlapping = ref_dimensions.iter().all(|(key, value)| {
                other_dimensions.contains_key(key) && &other_dimensions[key] == value
            });

            Ok(is_overlapping)
```

mahatoankitkumar (changes requested):

crates/experimentation_platform/src/api/experiments/helpers.rs
Sauravcv98: @mahatoankitkumar I'm sorry, i didn't get it 
there is if let condition along with else for the opt_context_json

crates/experimentation_platform/tests/experimentation_tests.rs
Sauravcv98: Hey because the type required is Option<Value>

crates/experimentation_platform/tests/experimentation_tests.rs
```diff
@@ -25,21 +25,24 @@ fn single_dimension_ctx_gen(value: Dimensions) -> serde_json::Value {
...
+    };
+    Some(context)
 }
 
-fn multiple_dimension_ctx_gen(values: Vec<Dimensions>) -> serde_json::Value {
+fn multiple_dimension_ctx_gen(values: Vec<Dimensions>) -> Option<serde_json::Value> {
     let mut conditions: Vec<serde_json::Value> = vec![];
     for val in values {
-        conditions.push(single_dimension_ctx_gen(val));
+        if let Some(context) = single_dimension_ctx_gen(val) {
```
mahatoankitkumar, line 35: If you see here you are doing `if let`. But if you just return `serde_json::Value` from `single_dimension_ctx_gen` you don't need to do this.

mahatoankitkumar (changes requested):

crates/experimentation_platform/src/api/experiments/helpers.rs
Sauravcv98: Done

crates/experimentation_platform/src/api/experiments/handlers.rs
Sauravcv98: Hey we don't have that function in the experiment_platform crate
It's only present in frontend crate 
Should we add ?

crates/experimentation_platform/tests/experimentation_tests.rs
Sauravcv98: Hey actually at other places we want the Option Value only like at test_are_overlapping_contexts function
We are using the Option <Value>

crates/experimentation_platform/src/api/experiments/helpers.rs
mahatoankitkumar: use case match

Sauravcv98: Sure

Sauravcv98: Done

Datron:
Does this PR solve https://github.com/juspay/superposition/issues/33?

crates/experimentation_platform/src/api/experiments/handlers.rs
mahatoankitkumar: It would be better to add a similar function to backend too.

crates/frontend/src/components/context_form/utils.rs
```diff
@@ -58,20 +58,25 @@ pub fn construct_context(
...
+    if conditions.is_empty() {
+        json!({})
     } else {
-        json!({ "and": condition_schemas })
-    };
+        let condition_schemas = conditions
+            .iter()
+            .map(|(variable, operator, value)| {
+                get_condition_schema(variable, operator, value, dimensions.clone())
+                    .unwrap()
```
mahatoankitkumar, line 68: if this returns optional then use `unwrap_option_or_default_with_error`.

mahatoankitkumar (changes requested):

crates/frontend/src/components/context_form/utils.rs
Sauravcv98: Hey the change is the indentation here and also this returns Result<Value,String>

crates/experimentation_platform/src/api/experiments/handlers.rs
Sauravcv98: Yeah Sure

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -202,36 +202,41 @@ pub fn add_variant_dimension_to_ctx(
...
+        "Context not an object. Ensure the context provided obeys the rules of JSON logic"
+    ))?;
 
-    match serde_json::to_value(updated_ctx) {
-        Ok(value) => Ok(value),
-        Err(_) => Err(bad_argument!(
-            "Failed to convert context to a valid JSON object. Check the request sent for correctness"
-        )),
+    if context.is_empty() {
+        Ok(variant_condition)
```
pratikmishra356, line 217: @Sauravcv98 this should also be in the form of {"and": [ variant_condition ] }

pratikmishra356 (changes requested):

crates/experimentation_platform/src/api/experiments/helpers.rs
Sauravcv98: Hey, but why ?
We can have
"in" : [
            variant,
            { "var": "variantIds" }
        ]

This would work right ?

pratikmishra356: yes it will work, I was thinking that we might be following this convention
but its fine you are right, let it be as it is

pratikmishra356 (approved):

crates/experimentation_platform/src/api/experiments/helpers.rs
Sauravcv98: got it, sure 👍

mahatoankitkumar (approved):

PR #41, opened 2024-05-09
docs: update README/experimentation/cac docs

knutties:
## Problem
This PR updates the docs to simplify some wording, use the same example as the proposed demo-app and makes both experimentation/cac docs refer to the same example.  It also links to the Postman collection as a substitute for API documentation

## Solution
Same as above

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
NA

NA

### DIFF
--- a/docs/context-aware-config.md
+++ b/docs/context-aware-config.md
@@ -1,7 +1,7 @@
 # Context Aware Config
 ---
 
-Context Aware Config (abbreviated as CAC) is the foundational service of the Superposition Platform that is configuration management system that can override config values under certain domain contexts. CAC makes it easy to configure your applications and make sure that you can change these configurations without causing issues in your application
+Context Aware Config (abbreviated as CAC) is the foundational service of the Superposition Platform.  It is the configuration management system that can override configuration values under certain domain contexts.
 
 - [Context Aware Config](#context-aware-config)
   - [Concepts](#concepts)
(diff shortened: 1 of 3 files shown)


docs/experimentation.md
```diff
@@ -1,27 +1,71 @@
...
-<br/>
+An experiment, as the name suggests, enables you to test and evaluate the
+behaviour of the system for different values of the same configuration. An
+experiment can have exactly one **CONTROL** variant and `n` **EXPERIMENTAL**
+variants, with each variant overriding/changing same set of keys in
+configuration. An experiment's scope can be controlled by declaring the
+context, which chalks out the sample set for the experiment from the
+population.
 
-![context-example](experiment-context-example.png)
```
knutties, line 14: @ShubhranshuSanjeev  - pardon me for removing the diagrams - wanted to refer to the same example as the proposed demo app - hence took the liberty.

docs/context-aware-config.md
```diff
@@ -13,92 +13,90 @@ Context Aware Config (abbreviated as CAC) is the foundational service of the Sup
...
+Default Configs is the set of all possible configuration keys of your application along with their default values.  One may think of default configs as the best assumption we can make about our configs.
 
-Example:
-Using our car classification example, we can establish defaults like:
-1. Number of doors - 4 (most cars do have 4 doors)
-2. Engine - Internal Combustion
-3. Chassis - Hatchback
-4. Color - Silver 
-5. interiors - leather
+Let us use a simple cab ride-hailing application that operates in different cities and capture some attributes that we might need configurable at a per city level.
```
Datron, line 24: @knutties can we link to the example code so that this folks can run the example too while reading this? Should we attach images or video?

Datron (approved):

docs/context-aware-config.md
knutties: I am thinking I should move the demo-app as a crate in here.  Or are you saying point it to mine for now and when it is moved - update the docs to point to local crate?

mahatoankitkumar (approved):
We might need to update the postman collection as we might not have updated it with the new api contracts.

knutties:
> We might need to update the postman collection as we might not have updated it with the new api contracts.

The API contracts haven't really changed for existing ones (except for maybe the things like prefix - which are feature additions).  But the current ones work I believe as we run the tests.  The new APIs are not added to tests I believe.  Yes - definitely needs updating - but yeah - we should move to documenting APIs itself directly.

PR #42, opened 2024-05-10
fix: setup superposition without nix

Datron:
## Problem
The local setup of superposition depends on nix with may impede some users from trying it out

## Solution
Provide an alternate way to use superposition without setting up nix

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -2,45 +2,64 @@
 
 This document outlines the setup process for the `Superposition Platform`.
 
-## Prerequisites
+## Installation for linux/MacOS
 
-* Docker installed and running.
-* Docker Compose installed.
-* [Follow installation instructions at Zero To Nix ↗](https://zero-to-nix.com/start/install) 
+- Install Rust using rustup
+    `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
+- Install wasm-pack
+    `curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh`
+- Install docker
+- make sure you have dependencies like libpq, openssl, libiconv,etc installed
+- make sure `make` command is installed
+  - Linux
+    ```bash
+    sudo apt-get install build-essential
+    ```
+  - MacOS
+    ```bash
+    xcode-select --install
+    ```
+- Clone the repository 
+    ```bash
+    git clone https://github.com/juspay/superposition.git
+    cd superposition
+    ```
+- Start docker daemon
+    - Linux
+        ```bash
+        sudo systemctl start docker
+        ```
+    - MacOS
+        ```bash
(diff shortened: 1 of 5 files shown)


docs/setup.md
```diff
@@ -2,45 +2,50 @@
 
 This document outlines the setup process for the `Superposition Platform`.
 
-## Prerequisites
+## Installation
```
knutties, line 5: Since I assume you have specified MacOS here - I assumed you want this for Linux/MacOS - should we call that out explicitly here?

Maybe @ShubhranshuSanjeev can validate steps on Linux.

docs/setup.md
```diff
@@ -2,45 +2,50 @@
...
 
-* Docker installed and running.
-* Docker Compose installed.
-* [Follow installation instructions at Zero To Nix ↗](https://zero-to-nix.com/start/install) 
+- Install Rust using rustup
+    `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
+- Install wasm-pack
+    `curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh`
+- Install docker
+- make sure you have dependencies like libpq, openssl, libiconv,etc installed
```
knutties, line 12: How does one do this in MacOS?  Any specific instructions needed for this ?

docs/setup.md
```diff
@@ -2,45 +2,50 @@
...
+- Clone the repository 
+    ```bash
+    git clone https://github.com/juspay/superposition.git
+    cd superposition
+    ```
+- Start docker daemon
+    ```bash
+    open --background -a Docker
```
knutties, line 20: This is MacOS specific I believe (should we give the Linux alternative as well)

docs/setup.md
Datron: @knutties I have written the steps to be platform independent, if necessary I can target ubuntu and macOS install instructions for libpq, openssl and libiconv

docs/setup.md
Datron: We can mention linux/MacOS. Only the steps to install libpq will change between the two.

docs/setup.md
Datron: Yeah let me add that

docs/setup.md
```diff
@@ -2,45 +2,64 @@
...
+    ```bash
+    xcode-select --install
+    ```
+- Clone the repository 
+    ```bash
+    git clone https://github.com/juspay/superposition.git
+    cd superposition
+    ```
+- Start docker daemon
+    - MacOS
```
knutties, line 28: better to keep the order the same in all examples (Linux first and then MacOS for consistency)

knutties (approved):

ShubhranshuSanjeev (approved):

PR #44, opened 2024-05-10
feat: reducing context tool

Sauravcv98:
## Problem
Too many Context duplicates, where some contexts are a subContext of other one's with the same overrides, so the parent one is not even required 
## Solution
Removing contexts if we find any one subcontext with the same override key, then we remove the override key from the parent context, now if all of the override keys are removed from the parent context then we can remove the entire context, so we just delete it.

### DIFF
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -42,7 +42,7 @@ pub fn endpoints() -> Scope {
     Scope::new("")
         .service(get)
         .service(get_resolved_config)
-        .service(reduce_context)
+        .service(reduce_config)
         .service(get_filtered_config)
 }
 
@@ -328,7 +328,7 @@ fn construct_new_payload(req_payload: &Map<String, Value>) -> web::Json<PutReq> 
     })
 }
 
-async fn reduce_context_key(
+async fn reduce_config_key(
     user: User,
     conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
     mut og_contexts: Vec<Context>,
@@ -364,24 +364,27 @@ async fn reduce_context_key(
         }
     }
 
-    contexts_overrides_values.sort_by(|a, b| {
-        (validate_dimensions_and_calculate_priority(
-            "context",
-            &(b.0).condition,
-            dimension_schema_map,
-        )
-        .unwrap())
-        .cmp(
-            &(validate_dimensions_and_calculate_priority(
-                "context",
-                &(a.0).condition,
-                dimension_schema_map,
-            )
-            .unwrap()),
-        )
-    });
+    let mut priorities = Vec::new();
 
-    let resolved_dimensions = reduce(contexts_overrides_values, default_config_val)?;
+    for (index, ctx) in contexts_overrides_values.iter().enumerate() {
+        let priority = validate_dimensions_and_calculate_priority(
+            "context",
+            &(ctx.0).condition,
+            dimension_schema_map,
+        )?;
+        priorities.push((index, priority))
+    }
+
+    // Sort the collected results based on priority
+    priorities.sort_by(|a, b| b.1.cmp(&a.1));
+
+    // Use the sorted indices to reorder the original vector
+    let sorted_priority_contexts = priorities
+        .into_iter()
+        .map(|(index, _)| contexts_overrides_values[index].clone())
+        .collect();
+
+    let resolved_dimensions = reduce(sorted_priority_contexts, default_config_val)?;
     for rd in resolved_dimensions {
         match (
             rd.get("can_be_reduced"),
@@ -446,7 +449,7 @@ async fn reduce_context_key(
 }
 
 #[put("/reduce")]
-async fn reduce_context(
+async fn reduce_config(
     req: HttpRequest,
     user: User,
     db_conn: DbConnection,
@@ -465,7 +468,7 @@ async fn reduce_context(
         let contexts = config.contexts;
         let overrides = config.overrides;
         let default_config = config.default_configs;
-        config = reduce_context_key(
+        config = reduce_config_key(
             user.clone(),
             &mut conn,
             contexts.clone(),
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -35,7 +35,7 @@ use service_utils::{bad_argument, db_error, unexpected_error};
 use itertools::Itertools;
 use jsonschema::JSONSchema;
 use service_utils::helpers::extract_dimensions;
-use service_utils::result as superposition;
+use service_utils::result::{self as superposition, AppError};
 use superposition_types::User;
 use uuid::Uuid;
 pub fn endpoints() -> Scope {
@@ -198,11 +198,11 @@ fn generate_subsets_keys(keys: Vec<String>) -> Vec<Vec<String>> {
 
 fn resolve(
     contexts_overrides_values: Vec<(Context, Map<String, Value>, Value, String)>,
-    default_config_val: Option<&Value>,
-) -> Vec<Map<String, Value>> {
+    default_config_val: &Value,
+) -> superposition::Result<Vec<Map<String, Value>>> {
     let mut dimensions: Vec<Map<String, Value>> = Vec::new();
     for (context, overrides, key_val, override_id) in contexts_overrides_values {
-        let mut ct_dimensions = extract_dimensions(&context.condition).unwrap();
+        let mut ct_dimensions = extract_dimensions(&context.condition)?;
         ct_dimensions.insert("key_val".to_string(), key_val);
         let request_payload = json!({
             "override": overrides,
@@ -217,19 +217,14 @@ fn resolve(
 
     //adding default config value
     let mut default_config_map = Map::new();
-    if let Some(some_default_config_val) = default_config_val {
-        default_config_map
-            .insert("key_val".to_string(), some_default_config_val.to_owned());
-    };
+    default_config_map.insert("key_val".to_string(), default_config_val.to_owned());
     dimensions.push(default_config_map);
 
     for (index1, c1) in dimensions.clone().iter().enumerate() {
-        let mut temp_c1 = c1.clone();
         let mut nc1 = c1.clone();
         nc1.remove("req_payload");
         let c1_val = nc1.remove("key_val");
         let all_subsets_c1 = generate_subsets(&nc1);
-        let mut can_be_removed = false;
         for (index2, c2) in dimensions.iter().enumerate() {
             let mut temp_c2 = c2.clone();
             temp_c2.remove("req_payload");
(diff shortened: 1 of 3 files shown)


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    //adding default config value
+    let mut default_config_map = Map::new();
+    if let Some(some_default_config_val) = default_config_val {
+        default_config_map
+            .insert("key_val".to_string(), some_default_config_val.to_owned());
+    };
+    dimensions.push(default_config_map);
+
+    for (index1, c1) in dimensions.clone().iter().enumerate() {
+        let mut temp_c1 = c1.clone();
```
Datron, line 227: Need better variable names

Datron: For the whole block actually

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+                overrides.clone(),
+                key_val.clone(),
+                override_id.to_string(),
+            ));
+        }
+    }
+    res
+}
+
+fn construct_new_payload(req_payload: &Map<String, Value>) -> web::Json<PutReq> {
```
Datron, line 285: Why not take a mutable cloned reference as an argument and skip the first line?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    mut og_contexts: Vec<Context>,
+    mut og_overrides: Map<String, Value>,
+    check_key: &str,
+    dimension_schema_map: &HashMap<String, (JSONSchema, i32)>,
+    default_config: Map<String, Value>,
+    is_approve: bool,
+) -> Config {
+    let default_config_val = default_config.get(check_key);
+    let mut contexts_overrides_values: Vec<(Context, Map<String, Value>, Value, String)> =
```
Datron, line 315: Might be good to type alias the tuple inside vector

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+                                    }
+                                    let mut elem = og_contexts[ctx_index].clone();
+                                    elem.override_with_keys = [new_id];
+                                    og_contexts[ctx_index] = elem;
+
+                                    if delete_old_oid {
+                                        og_overrides.remove(oid);
+                                    }
+                                }
+                            }
```
Datron, line 414: Can we simplify this whole code block? The nesting is insane.

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    }
+
+    //adding default config value
+    let mut default_config_map = Map::new();
+    if let Some(some_default_config_val) = default_config_val {
+        default_config_map
+            .insert("key_val".to_string(), some_default_config_val.to_owned());
+    };
+    dimensions.push(default_config_map);
+
```
Datron, line 225: Can we add a comment here, quickly explaining the algorithm below?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    dimension_schema_map: &HashMap<String, (JSONSchema, i32)>,
+    default_config: Map<String, Value>,
+    is_approve: bool,
+) -> Config {
+    let default_config_val = default_config.get(check_key);
+    let mut contexts_overrides_values: Vec<(Context, Map<String, Value>, Value, String)> =
+        Vec::new();
+
+    for (key, val) in og_overrides.clone() {
+        match val {
```
Datron, line 319: can you just do:
```suggestion
       val.as_object().unwrap_or(())
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+            Value::Object(obj_val) => {
+                if let Some(ans_val) = obj_val.get(check_key) {
```
Datron, line 321: ans_val, obj_val need better names. The code is hard to understand

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+        }
+    }
+
+    contexts_overrides_values.sort_by(|a, b| {
+        (validate_dimensions_and_calculate_priority(
+            "context",
+            &(b.0).condition,
+            dimension_schema_map,
+        )
+        .unwrap())
```
Datron, line 343: Can we avoid the raw unwrap?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+        .cmp(
+            &(validate_dimensions_and_calculate_priority(
+                "context",
+                &(a.0).condition,
+                dimension_schema_map,
+            )
+            .unwrap()),
```
Datron, line 350: Can we avoid the raw unwrap()

Datron (changes requested):
Changes required. Mostly documentation and formatting is what I saw in the first pass

crates/context_aware_config/src/api/config/handlers.rs
mahatoankitkumar: If pass the value instead of fetching it from the default config, then no need if this `if let` block.

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    web::Json(PutReq {
+        context: Map::new(),
+        r#override: Map::new(),
+    })
+}
+
+async fn reduce_context_key(
+    user: User,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    mut og_contexts: Vec<Context>,
```
mahatoankitkumar, line 307: mut not needed, as you are returning a new Config

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    db_conn: DbConnection,
+) -> actix_web::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+    let is_approve = req
+        .headers()
+        .get("x-approve")
+        .and_then(|value| value.to_str().ok().and_then(|s| s.parse::<bool>().ok()))
+        .unwrap_or(false);
+
+    let dimensions_schema_map = get_all_dimension_schema_map(&mut conn).unwrap();
```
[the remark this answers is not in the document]
Sauravcv98: Done

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: This value if not coming from default config, it's coming from overrides, have updated the code with better names, please review

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Hey, I would have to give a default value for unwrap_or, so i went with this

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Yeah, thanks for the suggestion
Implemented it

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+    user: User,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    mut og_contexts: Vec<Context>,
+    mut og_overrides: Map<String, Value>,
+    check_key: &str,
+    dimension_schema_map: &HashMap<String, (JSONSchema, i32)>,
+    default_config: Map<String, Value>,
+    is_approve: bool,
+) -> Config {
+    let default_config_val = default_config.get(check_key);
```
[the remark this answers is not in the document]
Sauravcv98: Hey from default config, i just need the value of that key alone

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Hey actually, og_contexts is getting used later and it's modified
So removing mut was not working

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,310 @@ async fn generate_cac(
...
+            sub.push(element.clone());
+            res.push(sub);
+        }
+    }
+    res
+}
+
+fn resolve(
+    contexts_overrides_values: Vec<(Context, Map<String, Value>, Value, String)>,
+    default_config_val: Option<&Value>,
```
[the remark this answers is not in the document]
Sauravcv98: Added the check before and returning the value

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Yes, will add it
Will make this block also a little better

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Yes

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Hey, in both unwrap above
I'm not able to propagate the errors or return it because it's inside the closure 
What i can do is create a new array first, which has all the priorities and then sort it 
Should i go ahead with that ?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -147,6 +161,329 @@ async fn generate_cac(
...
+
+    Ok(Config {
+        contexts: og_contexts,
+        overrides: og_overrides,
+        default_configs: default_config,
+    })
+}
+
+#[put("/reduce")]
+async fn reduce_context(
```
pratikmishra356, line 449: @Sauravcv98 is there any specific reason we kept this api under config and not context ?

Sauravcv98: Hey this actually reduces the entire Config, even the response of this is of type Config

pratikmishra356: then probably its name should be reduce_config @knutties @Datron  ?

pratikmishra356 (approved):

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Yeah true, was thinking the same
Will rename this to reduce_config

crates/context_aware_config/src/api/config/handlers.rs
mahatoankitkumar: yeah that would be better.

Datron: Can you make the closure return a Result type? You can then propagate errors through that?

crates/context_aware_config/src/api/config/handlers.rs
Sauravcv98: Hey actually, i'll have to do the same where i'm calling also
Like i would have to clone it over there and send it.

mahatoankitkumar (approved):

PR #48, opened 2024-05-10
feat: added support for update , create and clone of override

sauraww:
fixed override caching issue

## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -50,6 +50,7 @@ use service_utils::{bad_argument, result as superposition};
 pub fn endpoints() -> Scope {
     Scope::new("")
         .service(put_handler)
+        .service(update_override_handler)
         .service(move_handler)
         .service(delete_context)
         .service(bulk_operations)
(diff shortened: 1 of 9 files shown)


Datron (approved):

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -295,6 +315,40 @@ async fn put_handler(
...
+        }
+        Err(e) => {
+            log::error!("failed to update context with db error: {:?}", e);
+            Err(db_error!(e))
+        }
+    }
+}
+
+#[put("/overrides")]
+async fn update_override_handler(
```
Datron, line 341: Why do we need all this code? I thought update was straight forward and could be done with PUT context API

crates/frontend/src/api.rs
```diff
@@ -48,6 +48,33 @@ pub async fn fetch_default_config(
...
+    let url = format!("{}/context/{}", host, context_id);
+
+    let response = client
+        .delete(&url) // Make sure to pass the URL by reference here
+        .header("x-tenant", tenant)
+        .send()
+        .await
+        .map_err(|e| ServerFnError::ServerError(e.to_string()))?;
+
+    if response.status().is_success() {
```
Datron, line 68: @sauraww can we return the object that has been deleted?

crates/frontend/src/api.rs
```diff
@@ -48,6 +48,33 @@ pub async fn fetch_default_config(
...
+        Ok(())
+    } else {
+        Err(ServerFnError::ServerError(format!(
```
Datron, line 71: Can we use the error macros here?

crates/frontend/src/api.rs
```diff
@@ -48,6 +48,33 @@ pub async fn fetch_default_config(
...
+
+    // Use the first element of the context_id array in the URL
+    let url = format!("{}/context/{}", host, context_id);
+
+    let response = client
+        .delete(&url) // Make sure to pass the URL by reference here
+        .header("x-tenant", tenant)
+        .send()
+        .await
+        .map_err(|e| ServerFnError::ServerError(e.to_string()))?;
```
Datron, line 66: Can we use error macros here?

Datron (commented):
Do we need all that code to update overrides?

crates/context_aware_config/src/api/context/handlers.rs
sauraww: The context PUT api will merge the overrides, it will not replace it. 
What we wanted was to replace the overrides through update api.

PR #49, opened 2024-05-10
fix: fixed override_flow

sauraww:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/frontend/src/components/context_form/context_form.rs
+++ b/crates/frontend/src/components/context_form/context_form.rs
@@ -196,7 +196,7 @@ where
                             }
                         />
 
-                        <Show when=move || context.get().len() != 0>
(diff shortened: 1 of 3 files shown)


Datron (approved):

ShubhranshuSanjeev (approved):

crates/frontend/src/components/context_form/context_form.rs
```diff
@@ -196,7 +196,7 @@ where
                             }
                         />
 
-                        <Show when=move || context.get().len() != 0>
+                        <Show when=move || { context.get().len() != 0 && !disabled}>
```
Datron, line 199: We don't need brackets here right?

PR #63, opened 2024-05-15
feat: delete support for default config in UI

ShubhranshuSanjeev:
## Problem
No delete support for default config from UI

## Solution
Add delete button for each row in default config table, which calls `DELETE /default-config/{key}` endpoint.

### Additional changes
Refactored request function in `frontend/utils.rs` to return raw response instead of parsing it to a type.
Reason:
- Responses can be empty, not necessarily have a content.
- Parsing no-content to a type will always fail even if we are parsing it to `()` (unit), because `no-content` is not a valid JSON.
- Separating parsing of the response from the actual request part, will be bit cleaner, as request function will get bloated with conditionals to handle the `no-content` or any other scenario that I am not currently aware of.

### DIFF
--- a/crates/frontend/src/api.rs
+++ b/crates/frontend/src/api.rs
@@ -205,5 +205,7 @@ pub async fn delete_default_config(key: String, tenant: String) -> Result<(), St
         None::<serde_json::Value>,
         construct_request_headers(&[("x-tenant", &tenant)])?,
     )
-    .await
+    .await?;
+
+    Ok(())
 }
--- a/crates/frontend/src/api.rs
+++ b/crates/frontend/src/api.rs
@@ -5,7 +5,7 @@ use crate::{
         Config, DefaultConfig, Dimension, Experiment, ExperimentsResponse,
         FunctionResponse, ListFilters,
     },
-    utils::use_host_server,
+    utils::{construct_request_headers, get_host, request, use_host_server},
 };
 
 // #[server(GetDimensions, "/fxn", "GetJson")]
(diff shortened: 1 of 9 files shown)


crates/frontend/src/api.rs
```diff
@@ -194,3 +194,16 @@ pub async fn fetch_experiment(
         Err(e) => Err(ServerFnError::ServerError(e.to_string())),
     }
 }
+
+pub async fn delete_default_config(key: String, tenant: String) -> Result<(), String> {
+    let host = get_host();
+    let url = format!("{host}/default-config/{key}");
+
+    request(
```
mahatoankitkumar, line 202: @ShubhranshuSanjeev You would need to fix the request function a bit as it tries to return response message, and as the delete function as nothing in response it triggers a snack bar with the following error.
Screenshot 2024-05-17 at 11 55 29 AM

mahatoankitkumar (changes requested):

crates/frontend/src/api.rs
ShubhranshuSanjeev: Sure, let me take a look, and thanks for pointing this, I missed it.

mahatoankitkumar (approved):

sauraww (changes requested):

crates/frontend/src/pages/default_config/default_config.rs
```diff
@@ -186,6 +188,48 @@ pub fn default_config() -> impl IntoView {
...
         };
 
+        let delete_col_formatter = move |_: &str, row: &Map<String, Value>| {
+            let row_key = row["key"].as_str().map(|v| v.to_owned());
+            let is_folder = row_key.as_ref().is_some_and(|k| k.contains("."));
+
+            let key_name = StoredValue::new(row_key);
+
+            let handle_delete = move |_| {
+                match key_name.get_value() {
```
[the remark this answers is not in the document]
mahatoankitkumar: @sauraww error snack bar also needs to be triggered, using case match looks cleaner i guess.

Datron (approved):

sauraww (approved):

PR #66, opened 2024-05-16
fix: Added dropdown for enum selection

mahatoankitkumar:
## Problem
Add dropdown for enum selection

## Solution
Added dropdown for enum selection

https://github.com/juspay/superposition/assets/79959102/b59e138b-96e3-4e25-8e60-300de01d8ffd

### DIFF
--- a/crates/frontend/src/components/default_config_form/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form/default_config_form.rs
@@ -1,6 +1,3 @@
-pub mod types;
-pub mod utils;
-
 use leptos::*;
 use serde_json::{json, Number, Value};
 use std::str::FromStr;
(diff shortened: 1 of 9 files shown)


Datron:
@mahatoankitkumar can you add a video for dimensions too?

knutties:
@mahatoankitkumar - in the default config key addition section, the choose enum section show ideally go below the section where you configure the list of enums right ?  That is the natural order I believe.

What do you think?

mahatoankitkumar:
> @mahatoankitkumar - in the default config key addition section, the choose enum section show ideally go below the section where you configure the list of enums right ? That is the natural order I believe.
> 
> What do you think?

makes sense.

Datron:
@mahatoankitkumar can you include the video in overrides screen? An field thats an enum should just be a dropdown

PR #68, opened 2024-05-20
add cache control headers for all routes

pratikmishra356:
## Problem
Api response are getting cached on the client side, which gives inconsistent results

## Solution
Add cache-control: no-store headers , so that response should not cached in private and shared cache

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes

NA

### DIFF
--- a/crates/superposition/src/main.rs
+++ b/crates/superposition/src/main.rs
@@ -167,6 +167,7 @@ async fn main() -> Result<()> {
             .wrap(
                 actix_web::middleware::DefaultHeaders::new()
                     .add(("X-SERVER-VERSION", cac_version.to_string()))
+                    .add(("Cache-Control", "no-store".to_string()))
             )
             .service(web::redirect("/", ui_redirect_path.to_string()))
             .service(web::redirect("/admin", ui_redirect_path.to_string()))


crates/superposition/src/main.rs
```diff
@@ -167,6 +167,7 @@ async fn main() -> Result<()> {
             .wrap(
                 actix_web::middleware::DefaultHeaders::new()
                     .add(("X-SERVER-VERSION", cac_version.to_string()))
+                    .add(("Cache-Control", "no-store".to_string()))
```
knutties, line 170: Is this restricted to all responses or only front-end ?

mahatoankitkumar: I think this will affect the backend routes as well.

mahatoankitkumar (changes requested):

crates/superposition/src/main.rs
pratikmishra356: @knutties its mainly for backend , 
basically client calls backend apis to fetch data , now data gets cached in clients and served from cache which leads to inconsistent results , so we should have cache control for backend apis first

knutties (approved):

Datron (approved):
Looks good. Thanks @pratikmishra356.

mahatoankitkumar (approved):

PR #69, opened 2024-05-20
feat: support type templates

Datron:
## Problem
Users want to create and use their own types in default configs and overrides. Some teams like NY want to overwrite the default types to support null values

## Solution
https://github.com/juspay/superposition/discussions/52

## API changes
| Name | Method | URL | request | response |
|--------|--------|--------|--------|--------|
| List Type | GET | /types | None | `[{"created_at":"2024-05-13T12:02:02.030601","created_by":"user@superposition.io","display_name":"Number","id":"ed8c06e7-655b-4805-bd98-cd47f3a90cfc","last_modified":"2024-05-13T12:02:02.030601","type_name":"number","type_schema":{"type":"number"}},{"created_at":"2024-05-13T12:02:02.030601","created_by":"user@superposition.io","display_name":"Decimal","id":"c7b88aa0-dd0b-4b8d-9492-f2bb8701f8b9","last_modified":"2024-05-13T12:02:02.030601","type_name":"number","type_schema":{"type":"number"}},{"created_at":"2024-05-13T12:02:02.030601","created_by":"user@superposition.io","display_name":"Boolean","id":"16abae89-9fbd-488f-ac8b-d9ebdc8fae23","last_modified":"2024-05-13T12:02:02.030601","type_name":"boolean","type_schema":{"type":"boolean"}},{"created_at":"2024-05-13T12:02:02.030601","created_by":"user@superposition.io","display_name":"String (Enum)","id":"eed97c33-4a0d-4ab7-a7fe-8e1408f48e4f","last_modified":"2024-05-13T12:02:02.030601","type_name":"enum","type_schema":{"enum":"$replacement","type":"string"}},{"created_at":"2024-05-13T12:02:02.030601","created_by":"user@superposition.io","display_name":"String (Regex)","id":"2e5cd6a1-8b28-440a-83b1-3d0fce590340","last_modified":"2024-05-13T12:02:02.030601","type_name":"pattern","type_schema":{"pattern":"$replacement","type":"string"}}]` |
| Create Type | PUT | /types | `{ "type_schema": {"type": "integer"}, "display_name": "Integer", "type_name": "i16"}` | `{"message": "custom type created successfully"}` |
| Update Type | PUT | /types/{type_name} | `{"type": "integer"}` | `{"message": "custom type created successfully"}` |
| Delete Type | DELETE | /types/{id} | None | `{"message": "deleted custom type"}` | 

https://github.com/juspay/superposition/assets/15166178/5c4d9dba-3c19-4125-8518-d76622a96a9a

https://github.com/juspay/superposition/assets/15166178/9725b132-6601-4a7b-8471-415517f8c56b

https://github.com/juspay/superposition/assets/15166178/57ef9c43-c35c-4479-9978-6fad89e125ae


crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+use service_utils::service::types::DbConnection;
+use service_utils::{bad_argument, db_error, result as superposition};
+use superposition_types::User;
+
+use crate::api::custom_types::types::CreateTypeRequest;
+
+pub fn endpoints() -> Scope {
+    Scope::new("")
+        .service(list_types)
+        .service(create_or_update_type)
```
mahatoankitkumar, line 21: Lets follow having create and update as separate APIs.

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+async fn delete_type(
+    path: Path<String>,
+    db_conn: DbConnection,
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+    let path_id = path.into_inner();
+    let id = uuid::Uuid::from_str(path_id.as_str())
+        .map_err(|err| bad_argument!("Wrong custom type ID passed {}", err))?;
+    diesel::delete(dsl::jsonschema_types.filter(dsl::id.eq(id))).execute(&mut conn)?;
+    Ok(HttpResponse::Ok().json(json!({"message": "deleted custom type"})))
```
mahatoankitkumar, line 74: Lets change this to 204(No Content).

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+    let type_name = path.into_inner();
+    diesel::delete(dsl::jsonschema_types.filter(dsl::type_name.eq(type_name)))
+        .execute(&mut conn)?;
+    Ok(HttpResponse::Ok().json(json!({"message": "deleted custom type"})))
+}
+
+#[get("/{page}/{count}")]
+async fn list_types(
```
mahatoankitkumar, line 78:
```suggestion
async fn get(
```

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        .get_result(&mut conn)?;
+    let offset = (page - 1) * limit;
+    let custom_types: Vec<JsonSchemaTypes> = jsonschema_types::dsl::jsonschema_types
+        .order(jsonschema_types::dsl::created_at.desc())
+        .limit(limit)
+        .offset(offset)
+        .load(&mut conn)?;
+    let total_pages = (n_types as f64 / limit as f64).ceil() as u64;
+    Ok(HttpResponse::Ok().json(json!({
+        "total_pages": total_pages,
```
mahatoankitkumar, line 100: lets make a type for this.

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: @mahatoankitkumar In this case, there is a lot of code overlap. Should we still proceed with this? I found it redundant

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: This API is listing types, doesn't get a particular type. We have no ID and it maybe hard to search by the type_name so I went with list

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: Its a throwaway type, not used anywhere else hence I didn't type it. What would be the benefit of a type?

crates/context_aware_config/src/api/custom_types/mod.rs
mahatoankitkumar: Yeah in future code becomes more complex as we keep adding features or changes to this. We noticed this with default_config create/update. And we followed having separate APIs from functions onwards to prevent this.
And for common code you can make functions for them.

crates/context_aware_config/src/api/custom_types/mod.rs
mahatoankitkumar: But this is still a get all API right, if it was a get for a particular type then we could have had id in the params, which we don't have so this signifies GET ALL right?

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        .set((
+            jsonschema_types::type_schema.eq(request.type_schema.clone()),
+            jsonschema_types::created_by.eq(user.email),
+        ))
+        .execute(&mut conn)
+        .map_err(|err| {
+            log::error!("failed to insert custom type with error: {}", err);
+            db_error!(err)
+        })?;
+    Ok(HttpResponse::Ok().json(json!({"message": "custom type created successfully"})))
```
knutties, line 62: We should move to sending back the resource on successful creation instead of sending a message back.

crates/context_aware_config/src/api/custom_types/mod.rs
knutties: Agree - when we did `functions` we decided to have separate CREATE and UPDATE endpoints for clear behaviour separation.

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
```diff
@@ -0,0 +1,40 @@
...
+        NOW()
+    ),
+    (
+        'Boolean',
+        '{"type": "boolean"}',
+        'user@superposition.io',
+        NOW()
+    ),
+    (
+        'String (Enum)',
```
knutties, line 30: I think we should restrict the type name to have only permitted characters.  Something like function name in Cloudfront:

https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CreateFunction.html#API_CreateFunction_RequestSyntax

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: Right, which is why I went with `list_types` as the name. `get` doesn't really help identify what this function is really doing.

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
Datron: @knutties I will add validation for the type name here. I want to allow spaces since it doesn't seem to break anything, if someone adds too many spaces when trying to create a type they can quickly delete it too

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: Sure, will add this.

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
```diff
@@ -0,0 +1,40 @@
+-- Your SQL goes here
+CREATE TABLE jsonschema_types (
+    type_name TEXT PRIMARY KEY,
+    type_schema JSON NOT NULL,
+    created_by TEXT NOT NULL,
+    created_at TIMESTAMP NOT NULL,
+    last_modified TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+CREATE INDEX IF NOT EXISTS json_schema_types_index ON ONLY public.jsonschema_types(type_name, created_at);
```
ShubhranshuSanjeev, line 9: Why to create an index over `created_at`?

crates/context_aware_config/src/db/models.rs
```diff
@@ -74,3 +76,15 @@ pub struct EventLog {
     pub new_data: Option<Value>,
     pub query: String,
 }
+
+#[derive(Queryable, Selectable, Insertable, AsChangeset, Clone, Serialize, Debug)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+#[diesel(table_name = jsonschema_types)]
+#[diesel(primary_key(id))]
```
ShubhranshuSanjeev, line 83:
```suggestion
#[diesel(primary_key(type_name))]
```

there is no `id` column in `jsonschema_types` table

crates/frontend/src/components/type_template_form/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+            <div class="form-control grid w-full justify-end">
+                <Button
+                    class="pl-[70px] pr-[70px]".to_string()
+                    text="Submit".to_string()
+                    on_click=on_submit
+                />
+            </div>
+
+            {
```
ShubhranshuSanjeev, line 109: You can remove the `code block` here and `view!` too

crates/frontend/src/pages/custom_types/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+                            close_drawer(TYPE_DRAWER_ID);
+                        }/>
+
+                    </Drawer>
+                }
+            }
+        }}
+
+        <div class="p-8">
+            <Suspense fallback=move || view! { <p>"Loading..."</p> }>
```
ShubhranshuSanjeev, line 139: We have a `Skeleton` component now, that can be used here, instead of `Loading` text

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
knutties: @Datron - it will start breaking when we use function/types names as files (unless people take care).  For e.g. people might maintain some of the types/functions in their git repository to control edits to the functions (just like how we have a repo for AWS CFFs) - think infra as code.  In such cases these types might exist as files with the name of the type being the file name.  In such cases, scripts written have to be careful and handle spaces in file names correctly.

Datron: @knutties this makes sense, I hadn't considered this. I will allow [a-zA-Z_]+ as a pattern?

crates/frontend/src/pages/custom_types/mod.rs
Datron: Got it, will add it here

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
Datron: Answered in discussion

crates/context_aware_config/src/db/models.rs
Datron: Nice, I missed this

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
ShubhranshuSanjeev: https://github.com/juspay/superposition/discussions/52#discussioncomment-9516900

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
knutties: AWS uses Pattern: `^[a-zA-Z0-9-_]{1,64}$` for functions - which sounds like a good one to use.  We can use the same for functions as well.

Datron: Should we allow numbers @knutties? I think text is enough

knutties: Don't want to be too restrictive - particularly when people are playing around - numbers are sometimes handy to disambiguate stuff.

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
ShubhranshuSanjeev: https://github.com/juspay/superposition/discussions/52#discussioncomment-9523007

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/down.sql
```diff
@@ -0,0 +1,6 @@
+-- This file should undo anything in `up.sql`
+DELETE FROM TABLE jsonschema_types;
+
+DROP TABLE IF EXISTS jsonschema_types;
+
+DROP INDEX IF EXISTS json_schema_types_index;
```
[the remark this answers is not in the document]
srid: We check this in CI via Nix. nammayatri does it, through https://github.com/cachix/git-hooks.nix

The check can also (optionally) happen in pre-commit hook.

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+pub fn endpoints() -> Scope {
+    Scope::new("")
+        .service(list_types)
+        .service(create_type)
+        .service(update_type)
+        .service(delete_type)
+}
+
+#[put("")]
```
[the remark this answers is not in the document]
Datron: @knutties all of our creates/updates are PUTs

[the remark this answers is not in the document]
ShubhranshuSanjeev: @Datron , not really the one's with PUT for create are the old endpoints, you can check the functions and the experiment handlers, we have POST for create and PUT for update.
Can you update this?

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
```diff
@@ -0,0 +1,49 @@
+-- Your SQL goes here
+CREATE TABLE jsonschema_types (
+    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY,
+    type_name TEXT NOT NULL,
+    display_name TEXT NOT NULL,
+    type_schema JSON NOT NULL,
+    created_by TEXT NOT NULL,
+    created_at TIMESTAMP NOT NULL,
```
ShubhranshuSanjeev: @Datron , sorry for the late comment on this, but can we make this timestamp without timezone

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        ))
+        .get_result::<JsonSchemaTypes>(&mut conn)
+        .map_err(|err| {
+            log::error!("failed to insert custom type with error: {}", err);
+            db_error!(err)
+        })?;
+    Ok(HttpResponse::Ok().json(type_template))
+}
+
+#[post("")]
```
ShubhranshuSanjeev, line 72:
```suggestion
#[put("/{type_name}")]
```

crates/context_aware_config/src/api/custom_types/mod.rs
ShubhranshuSanjeev, line 29:
```suggestion
#[post("")]
```

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+async fn update_type(
+    request: Json<TypeTemplateRequest>,
```
ShubhranshuSanjeev, line 74: Can we have separate request type for this, the type name is the identifier here, which can be passed in the URL as a param?

The request body can only contain the schema, we should not really allow to change the type name, as this is a unique identifier, if the user wants he can delete the type and create a new one

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+    let type_name = path.into_inner();
+    let deleted_type =
+        diesel::delete(dsl::jsonschema_types.filter(dsl::type_name.eq(type_name)))
+            .get_result::<JsonSchemaTypes>(&mut conn)?;
+    Ok(HttpResponse::Ok().json(deleted_type))
+}
+
+#[get("/{page}/{count}")]
```
ShubhranshuSanjeev, line 130: Having the page and count in the URL makes it mandatory for the user to opt into pagination, i think this is not needed it can be in `query params`. 
For the cases like where we want to list all the types, for example the dimension_form or the default_config_form, we won't have a paginated view, we need all the types in one go, so that they can be displayed in the dropdown

crates/frontend/src/components/type_template_form/mod.rs
ShubhranshuSanjeev: @Datron , can you address this comment?

ShubhranshuSanjeev (changes requested):

ShubhranshuSanjeev:
> Reviewing this big PR is a pain, if possible can we divide the changes into two PRs, backend changes and the frontend changes.

We will review the backend changes and merge it first and then the frontend can go in.

crates/frontend/src/components/type_template_form/mod.rs
ShubhranshuSanjeev: @Datron , can you address this comment?

crates/context_aware_config/src/api/custom_types/mod.rs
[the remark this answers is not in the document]
Datron: Sure @ShubhranshuSanjeev , will do this

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: You cannot change the type name anyway, since the to-be-updated row is searched on the type name.

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: If they are query params, you have to assume a default anyway. This is a better way of organizing list APIs, they should all be paginated. The changes need to propagate to dimensions and default config as well so that the frontend doesn't have to guess the page length and count

Datron:
@ShubhranshuSanjeev you can review this PR commit-wise, I would recommend it going as a whole

crates/context_aware_config/src/api/custom_types/mod.rs
mahatoankitkumar: i too think these should be query_params instead of having it in URL.
https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28
https://medium.com/@premdattan/pagination-and-filter-url-strategy-b67d690120ae

crates/frontend/src/api.rs
```diff
@@ -194,3 +194,24 @@ pub async fn fetch_experiment(
...
+
+pub async fn fetch_types(
+    tenant: String,
+    page: i64,
+    count: i64,
+) -> Result<FetchCustomTypeResponse, ServerFnError> {
+    let client = reqwest::Client::new();
+    let host = use_host_server();
+
+    let url = format!("{host}/types/{page}/{count}");
```
mahatoankitkumar, line 206: Won't it be better to use `request` function for this?

Datron: Good point, I'll add this

crates/frontend/src/components/type_template_form/mod.rs
Datron: Done

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        bad_argument!(
+            "Invalid jsonschema sent in the request, validation error is: {}",
+            err.to_string()
+        )
+    })?;
+    let type_name = path.into_inner();
+    let regex = Regex::new(TYPE_NAME_REGEX).map_err(|err| {
+        unexpected_error!("could not parse regex due to: {}", err.to_string())
+    })?;
+    if !regex.is_match(type_name.as_str()) {
```
pratikmishra356, line 96: do we need to validate the key type_name ? as we are getting update requests , if it does not exist in db it will give badrequest

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,159 @@
+mod types;
+
```
pratikmishra356, line 2: can we rename this file and put outside this folder

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: If it does not exist in DB for update, it is a bad request. Lets validate the key name regardless.

crates/context_aware_config/src/api/custom_types/mod.rs
Datron: Leaving it as is, we can sync up on list API formats later

crates/context_aware_config/src/api/custom_types/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        .get_result(&mut conn)?;
+    let offset = (page - 1) * limit;
+    let custom_types: Vec<JsonSchemaTypes> = jsonschema_types::dsl::jsonschema_types
+        .order(jsonschema_types::dsl::created_at.desc())
+        .limit(limit)
+        .offset(offset)
+        .load(&mut conn)?;
+    let total_pages = (n_types as f64 / limit as f64).ceil() as u64;
+    Ok(HttpResponse::Ok().json(json!({
+        "total_pages": total_pages,
```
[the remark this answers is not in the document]
Datron: @pratikmishra356 any paginated list API should have these two fields along so that the UI can paginate easily rather than guessing how many pages are left. extra DB call is fine, count is fast

postman/cac/custom types/Get Types/event.prerequest.js
```diff
@@ -0,0 +1 @@
+// any prerequest js code goes here
```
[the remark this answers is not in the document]
Datron: Yeah we can, I'll delete it

mahatoankitkumar:
@Datron Can you improve this like showing the schema field on choosing, Show `Choose Schema` for the dropdown?
Screenshot 2024-06-06 at 1 05 03 PM

mahatoankitkumar:
@Datron Can you put a placeholder while creating a type template?

crates/frontend/src/pages/custom_types.rs
```diff
@@ -0,0 +1,200 @@
...
+    );
+    let selected_type = create_rw_signal::<Option<CustomTypeRow>>(None);
+    let table_columns = create_memo(move |_| {
+        vec![
+            Column::default("type_name".to_string()),
+            Column::default("type_schema".to_string()),
+            Column::default("created_by".to_string()),
+            Column::default("created_at".to_string()),
+            Column::default("last_modified".to_string()),
+            Column::new(
```
mahatoankitkumar, line 52: Can we separate delete and edit columns?

mahatoankitkumar (dismissed):

crates/frontend/src/pages/custom_types.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        <div class="p-8">
+            <Suspense fallback=move || view! { <Skeleton/> }>
+                <div class="pb-4">
+                    {move || {
+                        let types = types_resource.get().unwrap_or(vec![]);
+                        let data = types
+                            .iter()
+                            .map(|ele| {
+                                let mut ele_map = json!(ele).as_object().unwrap().to_owned();
```
mahatoankitkumar, line 149: use `unwrap_option_or_default_with_error` instead of unwrap.

crates/frontend/src/pages/custom_types.rs
Datron: There are actions you can perform on a row, why separate? Also, is there a benefit to two columns?

crates/frontend/src/pages/custom_types.rs
Datron: Nice, will use this

crates/frontend/src/components/type_template_form.rs
```diff
@@ -0,0 +1,124 @@
...
+        let type_schema = type_schema_rs.get();
+
+        let handle_submit_clone = handle_submit.clone();
+        spawn_local({
+            let handle_submit = handle_submit_clone;
+            async move {
+                let result = if edit {
+                    update_type(tenant_rs.get(), type_name, type_schema).await
+                } else {
+                    let payload = json!({
```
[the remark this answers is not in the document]
Datron: Don't see the value right now in making a type, we can add it later if required

crates/frontend/src/pages/custom_types.rs
mahatoankitkumar: This is a call to be made, if we want it to be this way, will have to change it in other places(where it is two different columns). If not then change it here.

flake.nix
```diff
@@ -30,6 +30,9 @@
 
       perSystem = { pkgs, self', config, ... }: {
         devShells.default = pkgs.mkShell {
+          shellHook = ''
+            export CARGO_HOME=$PWD/.cargo
```
srid, line 34: The devShell may not always start from project root. But you can use https://github.com/srid/flake-root to determine project root.

We use this trick in nammayatri for setting `CABAL_DIR`: https://github.com/nammayatri/common/blob/main/nix/haskell/no-global-cache.nix

Datron: Actually I added this for myself, it's not needed. Will remove this

crates/frontend/src/pages/custom_types.rs
Datron: We're doing this in overrides screen, where clone, edit and delete override are grouped for the card

mahatoankitkumar: Then better to do it for the default_config page too.

mahatoankitkumar: Screenshot 2024-06-07 at 12 37 02 PM
Here one can create key without choosing any schema also, is this as expected? You can show the schema field after choosing the schema.

Datron: Yeah this is expected, the type template is optional. You could use jsonschema that hasn't been defined before.

mahatoankitkumar: Maybe we can display that in any way there.

crates/context_aware_config/src/api/custom_types/handlers.rs
```diff
@@ -0,0 +1,160 @@
...
+        .service(update_type)
+        .service(delete_type)
+}
+
+#[post("")]
+async fn create_type(
+    request: Json<TypeTemplateRequest>,
+    db_conn: DbConnection,
+    user: User,
+) -> superposition::Result<HttpResponse> {
```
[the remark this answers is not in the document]
Datron: @pratikmishra356 I don't think a convention can be enforced here. Different handlers will have different responses. We can discuss it, but I have followed the convention used today with default_config, dimension, config and functions

crates/context_aware_config/src/api/custom_types/handlers.rs
```diff
@@ -0,0 +1,160 @@
...
+    let DbConnection(mut conn) = db_conn;
+    let _ = JSONSchema::compile(&request.type_schema).map_err(|err| {
```
mahatoankitkumar, line 37:
```suggestion
   JSONSchema::compile(&request.type_schema).map_err(|err| {
```

crates/context_aware_config/src/api/custom_types/handlers.rs
```diff
@@ -0,0 +1,160 @@
...
+    let DbConnection(mut conn) = db_conn;
+    let type_name = path.into_inner();
+    let deleted_type =
+        diesel::delete(dsl::jsonschema_types.filter(dsl::type_name.eq(type_name)))
+            .get_result::<JsonSchemaTypes>(&mut conn)?;
+    Ok(HttpResponse::Ok().json(deleted_type))
+}
+
+#[get("")]
+async fn list_types(
```
pratikmishra356, line 133: @Datron should we accept offset_value and count in query, 
as in the current case , if I am at page 5 and I change the count from 10 to 20 , then probably I will have to start from page 1 again

so in response we will send the offset value , and they will just send the number of  rows they want after an offset.
Page is probably from UI perspective  is what I think

crates/context_aware_config/src/api/custom_types/handlers.rs
```diff
@@ -0,0 +1,160 @@
...
+            "Invalid jsonschema sent in the request, schema: {:?} error: {}",
+            request.type_schema,
+            err
+        );
+        bad_argument!(
+            "Invalid jsonschema sent in the request, validation error is: {}",
+            err.to_string()
+        )
+    })?;
+    let type_name = request.type_name.clone();
```
mahatoankitkumar, line 48:
```suggestion
    let type_name = &request.type_name;
```

crates/context_aware_config/src/api/custom_types/handlers.rs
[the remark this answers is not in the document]
pratikmishra356: in Half of the places it is Json, and other half its HttpResponse, 
anyway in my headers pr I have added CustomRes for adding headers also.
Yes I think anyway is fine , just that we have to look inside the handler what response type it is returning

crates/context_aware_config/src/api/custom_types/handlers.rs
```diff
@@ -0,0 +1,160 @@
...
+async fn delete_type(
+    path: Path<String>,
+    db_conn: DbConnection,
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
+    let type_name = path.into_inner();
+    let deleted_type =
+        diesel::delete(dsl::jsonschema_types.filter(dsl::type_name.eq(type_name)))
+            .get_result::<JsonSchemaTypes>(&mut conn)?;
+    Ok(HttpResponse::Ok().json(deleted_type))
```
mahatoankitkumar, line 129: 204

crates/context_aware_config/src/api/custom_types/handlers.rs
Datron: @pratikmishra356 if you increase the count keeping the page the same, lets say from 10 to 20, you are shown the same 10 elements along with an additional 10 elements. You don't need to go back to page 1. That is why the offset is computed with the requested count in mind

pratikmishra356: No, lets say the earlier count was 10 and at page 4, then I will get data from 31 to 40 , now 
on that page only I changed the count to 20 , now offset value will be (4-1)*20 , so I will see data from 61 to 80
in the current page and I missed data from 31 to 60

crates/context_aware_config/src/api/custom_types/handlers.rs
Datron: 204 is not appropriate here, there is a response body

crates/context_aware_config/src/api/custom_types/handlers.rs
Datron: This is the avoid a warning with the compiler telling us that we are not using the result returned.
https://users.rust-lang.org/t/what-is-the-best-way-to-ignore-a-result/55187

crates/context_aware_config/src/api/custom_types/handlers.rs
mahatoankitkumar: I don't think there is a need for the response body?

Datron: The API caller gets information on what they deleted - frontends could reuse that to revert the action or log

crates/context_aware_config/src/api/custom_types/types.rs
```diff
@@ -0,0 +1,22 @@
...
+pub struct CustomTypeResponse {
+    pub type_schema: Value,
+    pub type_name: String,
+    pub created_at: String,
+    pub last_modified: String,
+    pub created_by: String,
+}
+#[derive(Debug, Clone, Deserialize)]
+pub struct QueryFilters {
+    pub count: Option<i64>,
```
pratikmishra356, line 20: @Datron can we make both i32 , as in code this allowing multiplying two 64 bit number which is resulting in thread panic

crates/superposition/src/main.rs
```diff
@@ -216,6 +216,11 @@ async fn main() -> Result<()> {
                             .wrap(AppExecutionScopeMiddlewareFactory::new(AppScope::CAC))
                             .service(functions::endpoints()),
                     )
+                    .service(
+                        scope("/types")
```
pratikmishra356, line 220: should we make it "/type"?

crates/frontend/src/components/type_template_form.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::components::type_template_form::utils::create_type;
+use crate::components::{button::Button, type_template_form::utils::update_type};
+use leptos::*;
+use serde_json::{from_str, json, Value};
+use web_sys::MouseEvent;
+
+#[component]
+pub fn type_template_form<NF>(
+    #[prop(default = false)] edit: bool,
+    #[prop(default = String::new())] type_name: String,
```
pratikmishra356, line 12: should we have default type_name also if we have default schema

crates/context_aware_config/src/api/custom_types/types.rs
Datron: @pratikmishra356 I didn't get you, why would i64 cause thread panic?

crates/frontend/src/components/type_template_form.rs
Datron: Hmm risky, that shouldn't be the case

crates/context_aware_config/src/api/custom_types/types.rs
pratikmishra356: in handler you are multiplying count and page , now if we send those two number greater than 2^32 it is causing thread panic, you can this

Datron: Okay, what is the worst case here? The thread panics, request is not completed right? Which is okay. This isn't an attack vector since actix will bring back the thread worker.

Datron: Also diesel's limit and offset functions expect i64

pratikmishra356 (approved):

ShubhranshuSanjeev (approved):

crates/frontend/src/components/type_template_form.rs
ShubhranshuSanjeev, line 1: @Datron , in handlers we are calling `type_template` as `custom_types`, db table is named `jsonschema_types` but migration file name mentions `custom_types`, can we use a single name throughout the service?

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
```diff
@@ -0,0 +1,42 @@
...
+    ),
+    (
+        'Boolean',
+        '{"type": "boolean"}',
+        'user@superposition.io',
+        NOW()
+    ),
+    (
+        'Enum',
+        '{"type": "string", "enum": "$replacement"}',
```
ShubhranshuSanjeev, line 33: Should we put an example array of options here, which can help user understand how to exactly use enum type?

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -179,160 +146,101 @@ where
...
                                 on:change=move |ev| {
-                                    logging::log!("{:?}", event_target_value(& ev));
-                                    set_config_value.set(event_target_value(&ev));
+                                    config_schema_ws
+                                        .set(string_to_value_closure(event_target_value(&ev)))
                                 }
                             >
 
-                                {config_value.get()}
+                                {format!("{}", config_schema_rs.get())}
```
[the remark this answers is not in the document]
Datron: Done

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
Datron: Any suggestions?

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
```diff
@@ -0,0 +1,42 @@
...
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    last_modified TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+CREATE INDEX IF NOT EXISTS json_schema_types_index ON public.jsonschema_types(type_name);
+CREATE INDEX IF NOT EXISTS json_schema_types_created_at_index ON public.jsonschema_types(created_at);
+CREATE INDEX IF NOT EXISTS json_schema_types_last_modifed_index ON public.jsonschema_types(last_modified);
+INSERT INTO jsonschema_types(type_name, type_schema, created_by, created_at)
+VALUES (
+        'Number',
+        '{"type": "number"}',
```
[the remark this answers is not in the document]
Datron: Yes we should

crates/context_aware_config/src/api/custom_types/handlers.rs
```diff
@@ -0,0 +1,160 @@
...
+            "Invalid jsonschema sent in the request, schema: {:?} error: {}",
+            request,
+            err
+        );
+        bad_argument!(
+            "Invalid jsonschema sent in the request, validation error is: {}",
+            err.to_string()
+        )
+    })?;
+    let type_name = path.into_inner();
```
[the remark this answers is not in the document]
Datron: It's required since I can reject a query without going to the database if the format is not correct.

crates/frontend/src/components/type_template_form.rs
Datron: Agreed, this is a mistake. Will take care of it.

crates/frontend/src/components/default_config_form.rs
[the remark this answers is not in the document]
Datron: @ShubhranshuSanjeev yes since it's only a type template, there is no way to make the template back to the type. We don't even associate the type template the config or dimension so hard to map this

crates/context_aware_config/migrations/2024-05-06-133756_custom_types/up.sql
Datron: I've added [android, ios]

ShubhranshuSanjeev (approved):

PR #72, opened 2024-05-22
fix: row numbering in paginated table

ShubhranshuSanjeev:
## Problem
In paginated tables, the rows number was always from 1 upto the row count for the particular page. For example, if row count is 10 for each page, rows on 2nd page of the table too had sr number `1,2,3,..,10`, which should have been `11,12,..20`.

## Solution
Move the pagination state inside the table, so that the row number can be calculated based on `current_page` and `count` of rows per page

### DIFF
--- a/crates/frontend/src/pages/function/function_list.rs
+++ b/crates/frontend/src/pages/function/function_list.rs
@@ -6,9 +6,8 @@ use leptos_router::A;
 use serde::{Deserialize, Serialize};
 
 use crate::components::skeleton::Skeleton;
-use crate::components::{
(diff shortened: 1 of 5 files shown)


crates/frontend/src/pages/function/function_list.rs
```diff
@@ -45,6 +44,26 @@ pub fn function_list() -> impl IntoView {
         },
     );
 
+    let handle_next_click = Callback::new(move |total_pages: i64| {
```
Datron, line 47: Can this be in utils somewhere or part of the table component?

Datron (approved):

mahatoankitkumar (approved):

PR #73, opened 2024-05-23
feat: Add JP version and regex operations in jsonlogic

ayushjain17:
## Problem
Semantic versioning and regex related operations not present in jsonlogic

## Solution
Add semver and regex operations in jsonlogic


Cargo.toml
```diff
@@ -41,7 +41,7 @@ chrono = { version = "0.4.26", features = ["serde"] }
 uuid = {version = "1.3.4", features = ["v4", "serde"]}
 reqwest = { version = "0.11.18", features = ["json"]}
 jsonschema = "~0.17"
-jsonlogic = "0.5.1"
+jsonlogic = { git = "https://github.com/juspay/jsonlogic_rs.git", rev = "4f1a8bcbb423c599eacf6c95ae17f5c0974bc071" }
```
Datron, line 44: Why did we make this change? All other changes are in JS

clients/js/src/index.ts
```diff
@@ -15,7 +15,12 @@ export class CacReader {
     defaultConfig: IObject;
 
     static {
-        jsonLogic.add_operation(">>", compareSemanticIsGreater);
+        jsonLogic.add_operation("match", matchRegex);
+        jsonLogic.add_operation("semver_eq", compareSemanticIsEqual);
+        jsonLogic.add_operation("semver_gt", compareSemanticIsGreater);
+        jsonLogic.add_operation("semver_gteq", compareSemanticIsGreaterEqual);
```
Datron, line 21:
```suggestion
        jsonLogic.add_operation("semver_ge", compareSemanticIsGreaterEqual);
```

clients/js/src/index.ts
```diff
@@ -15,7 +15,12 @@ export class CacReader {
...
+        jsonLogic.add_operation("semver_lt", compareSemanticIsLesser);
+        jsonLogic.add_operation("semver_lteq", compareSemanticIsLesserEqual);
```
Datron, line 23:
```suggestion
        jsonLogic.add_operation("semver_le", compareSemanticIsLesserEqual);
```

clients/js/src/utils/operations.ts
```diff
@@ -1,9 +1,33 @@
-const compareSemanticIsGreaterImp = function (version_a: string, version_b: string) {
-    return version_a.localeCompare(version_b, undefined, {numeric:true, sensitivity:'base'}) > 0;
+const compareSemantic = function (version_a: string, version_b: string): number {
+    return version_a.localeCompare(version_b, undefined, { numeric: true, sensitivity: 'base' });
```
Datron, line 2: the `locales` parameter and `options` parameter [are not supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#browser_compatibility) on android webview
Screenshot 2024-05-24 at 12 21 32 PM

Datron (changes requested):
Lack of support for android webviews means this won't work with SDK. Is that fine?

clients/js/src/index.ts
pratikmishra356, line 21: @ayushjain17 should we add one for between also ?

ayushjain17: @pratikmishra356 less than supports between operation as well, the behaviour of these version comparison operators are same as that of the normal comparison operators - wrt the number of operators

Cargo.toml
ayushjain17: to add the same operators in the rust client

Datron: Are those added in this PR? If not, can we make this change when we add the operations

ayushjain17: https://github.com/juspay/jsonlogic_rs/pull/1

ShubhranshuSanjeev (approved):

Cargo.toml
```diff
@@ -41,7 +41,7 @@ chrono = { version = "0.4.26", features = ["serde"] }
 uuid = {version = "1.3.4", features = ["v4", "serde"]}
 reqwest = { version = "0.11.18", features = ["json"]}
 jsonschema = "~0.17"
-jsonlogic = "0.5.1"
+jsonlogic = { git = "https://github.com/juspay/jsonlogic_rs.git", version = "0.5.2" }
```
Datron, line 44: Why add this without rust changes?

Datron (approved):

PR #78, opened 2024-05-27
feat: Add local storage support

mahatoankitkumar:
## Problem
Enable grouping toggle in default configs does not persist

## Solution
Added local storage support and grouping toggle in default configs persists now

### DIFF
--- a/crates/frontend/Cargo.toml
+++ b/crates/frontend/Cargo.toml
@@ -31,6 +31,7 @@ strum = { workspace = true }
 js-sys = "0.3.65"
 url = "2.5.0"
 once_cell = { workspace = true }
+leptos-use = { workspace = true}
 
 
 [features]
--- a/crates/frontend/src/pages/default_config/default_config.rs
+++ b/crates/frontend/src/pages/default_config/default_config.rs
@@ -7,7 +7,9 @@ use crate::components::stat::stat::Stat;
 use crate::components::table::{table::Table, types::Column};
 use crate::providers::alert_provider::enqueue_alert;
 use crate::types::BreadCrums;
-use crate::utils::unwrap_option_or_default_with_error;
+use crate::utils::{
+    get_local_storage, set_local_storage, unwrap_option_or_default_with_error,
+};
 use leptos::*;
 use leptos_router::{use_navigate, use_query_map};
 use serde_json::{json, Map, Value};
@@ -37,7 +39,8 @@ pub fn default_config() -> impl IntoView {
 
     let selected_config = create_rw_signal::<Option<RowData>>(None);
     let key_prefix = create_rw_signal::<Option<String>>(None);
-    let enable_grouping = create_rw_signal(false);
+    let enable_grouping =
+        create_rw_signal(get_local_storage::<bool>("enable_grouping").unwrap_or(false));
     let query_params = use_query_map();
     let bread_crums = Signal::derive(move || get_bread_crums(key_prefix.get()));
 
@@ -342,6 +345,10 @@ pub fn default_config() -> impl IntoView {
                                             on:click=move |_| {
                                                 folder_click_handler(None);
                                                 enable_grouping.set(!enable_grouping.get());
+                                                set_local_storage(
+                                                    "enable_grouping",
+                                                    &enable_grouping.get().to_string(),
+                                                );
                                             }
 
                                             class="cursor-pointer label mr-10"
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -5,11 +5,9 @@ use crate::{
     providers::alert_provider::enqueue_alert,
     types::{DefaultConfig, Dimension, Envs, ErrorResponse},
 };
+use cfg_if::cfg_if;
 use leptos::*;
-use reqwest::{
-    header::{HeaderMap, HeaderName, HeaderValue},
-    StatusCode,
-};
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use serde::de::DeserializeOwned;
 use serde_json::{Number, Value};
 use std::str::FromStr;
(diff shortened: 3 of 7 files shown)


crates/frontend/src/pages/default_config/default_config.rs
```diff
@@ -5,7 +5,9 @@ use crate::components::skeleton::Skeleton;
 use crate::components::stat::stat::Stat;
 use crate::components::table::{table::Table, types::Column};
 use crate::types::BreadCrums;
-use crate::utils::unwrap_option_or_default_with_error;
+use crate::utils::{
+    get_local_storage, set_local_storage, unwrap_option_or_default_with_error,
```
knutties, line 9: `unwrap_option_or_default_with_error` is unused ?

mahatoankitkumar: No it's being used.

crates/frontend/src/pages/default_config/default_config.rs
```diff
@@ -35,7 +37,11 @@ pub fn default_config() -> impl IntoView {
 
     let selected_config = create_rw_signal::<Option<RowData>>(None);
     let key_prefix = create_rw_signal::<Option<String>>(None);
-    let enable_grouping = create_rw_signal(false);
+    let enable_grouping = create_rw_signal(
```
ShubhranshuSanjeev, line 40:
```suggestion
    let enable_grouping =
        create_rw_signal(get_local_storage::<bool>("enable_grouping").unwrap_or(false));
```

crates/frontend/src/utils.rs
```diff
@@ -424,3 +425,29 @@ pub fn unwrap_option_or_default_with_error<T>(option: Option<T>, default: T) ->
         default
     })
 }
+
+pub fn get_local_storage(_key: &str) -> Option<String> {
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: @mahatoankitkumar , it would be better if we make this function generic, which will deserialize the value to a given type and then return it.

What do you think?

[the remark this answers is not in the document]
Datron: @ShubhranshuSanjeev you can't store objects in localstorage, only strings. 
https://stackoverflow.com/questions/2010892/how-to-store-objects-in-html5-localstorage-sessionstorage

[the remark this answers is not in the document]
Datron: lets leave it up to the calling function to decide the type after parsing it from strings

crates/frontend/src/pages/default_config/default_config.rs
Datron: I think cargo-fmt does this

Datron (approved):

crates/frontend/src/utils.rs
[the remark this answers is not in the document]
mahatoankitkumar: @Datron @ShubhranshuSanjeev Lets decide here first that who will decide the type after parsing it from string (the calling function or the called function) 🥲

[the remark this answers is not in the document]
ShubhranshuSanjeev: > @ShubhranshuSanjeev you can't store objects in localstorage, only strings.

Yes @Datron , I am aware of that, we set the value to a string only, but serde_json can parse those string to a type.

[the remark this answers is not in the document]
ShubhranshuSanjeev: >@Datron @ShubhranshuSanjeev Lets decide here first that who will decide the type after parsing it from string (the calling function or the called function)

@mahatoankitkumar , didn't really understood the question, if the function is generic then the caller decides the return type, otherwise the called function will dictate the return type.

While fetching from localstorage, we will always get a string, and depending on the requirement we will have to parse value to a certain type, doing this inside get_local_storage will reduce the repetition.

Debate here is for sure whether we want couple the parsing to the `get`, and as I can see parsing is going to be a subsequent step of `get_local_storage`, as everything is a `String`.

crates/frontend/src/pages/default_config/default_config.rs
ShubhranshuSanjeev: @Datron , I was suggesting after making return type of get_local_storage generic, we can re-write that piece of code like this, you can see I added added `::<bool>`, which is suggesting the type we want.

crates/frontend/Cargo.toml
```diff
@@ -31,6 +31,7 @@ strum = { workspace = true }
 js-sys = "0.3.65"
 url = "2.5.0"
 once_cell = { workspace = true }
+leptos-use = { workspace = true}
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: @mahatoankitkumar , can you check this?

crates/frontend/src/pages/default_config/default_config.rs
```diff
@@ -297,6 +300,8 @@ pub fn default_config() -> impl IntoView {
                                             on:click=move |_| {
                                                 folder_click_handler(None);
                                                 enable_grouping.set(!enable_grouping.get());
+                                                set_local_storage("enable_grouping", &enable_grouping.get().to_string());
+                                                logging::log!("<<>> set this {:?}", &enable_grouping.get().to_string());
```
ayushjain17, line 304: if we need to log this can we change it to something better, or lets remove it if the log is not intended

crates/frontend/src/utils.rs
```diff
@@ -424,3 +425,39 @@ pub fn unwrap_option_or_default_with_error<T>(option: Option<T>, default: T) ->
...
 }
+
+pub fn get_local_storage<T>(_key: &str) -> Option<T>
+where
+    T: serde::de::DeserializeOwned,
+{
+    cfg_if! {
+        if #[cfg(target_arch = "wasm32")] {
+            web_sys::window()
+                .and_then(|win| win.local_storage().ok())
```
ayushjain17, line 436:
```suggestion
                .and_then(|win| win.local_storage().ok().flatten())
```
you can use this to avoid adding the next line `.and_then(|storage| storage)`

crates/frontend/src/pages/default_config/default_config.rs
ayushjain17, line 304: if we need to log this can we change it to something better, or lets remove it if the log is not intended

crates/frontend/src/utils.rs
```diff
@@ -424,3 +425,39 @@ pub fn unwrap_option_or_default_with_error<T>(option: Option<T>, default: T) ->
...
+                .and_then(|storage| storage)
+                .and_then(|storage| {
+                    storage
+                        .get_item(_key)
+                        .ok()
+                        .flatten()
+                        .map(|value: String| serde_json::from_str::<T>(value.as_str()).ok())
```
ayushjain17, line 443:
```suggestion
                        .and_then(|value: String| serde_json::from_str::<T>(value.as_str()).ok())
```
you can use this to avoid adding the next line `.flatten()`

crates/frontend/src/utils.rs
```diff
@@ -424,3 +425,39 @@ pub fn unwrap_option_or_default_with_error<T>(option: Option<T>, default: T) ->
...
+            None
+        }
+    }
+}
+
+pub fn set_local_storage(_key: &str, _value: &str) -> Option<()> {
+    cfg_if! {
+        if #[cfg(target_arch = "wasm32")] {
+            web_sys::window()
+                .and_then(|win| win.local_storage().ok())
```
ayushjain17, line 456:
```suggestion
                .and_then(|win| win.local_storage().ok().flatten())
```
you can use this to avoid adding the next line `.and_then(|storage| storage)`

ayushjain17 (changes requested):

ShubhranshuSanjeev (approved):

ayushjain17 (approved):

PR #79, opened 2024-05-27
fix: stricter validation for key names in default config

knutties:
## Problem
Currently we allow trailing spaces and special characters in key names which is typically not needed.  This PR restricts the characters in the key name to the following regex: `^[a-zA-Z0-9-_.]{1,64}$`

## Solution
This PR uses the above regex to validate the key in the APIs - if found invalid - return a bad request response

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
| API      | PUT /config/key | same as existing with additional validations | same as existing |

NA


crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -50,11 +53,23 @@ async fn create(
...
+
+    if !regex.is_match(&key) {
+        return Err(bad_argument!(
+            "The key name {} is invalid, it should obey the regex {}",
+            key,
+            KEY_NAME_REGEX
+        ));
+    }
+
+    if key.starts_with(".") || key.ends_with(".") {
```
mahatoankitkumar, line 68: @knutties Can we try including this in the regex itself?

knutties: Good catch.  The regex gets a bit complicated to read - but I guess we can add a comment to explain that.  Will do this.

knutties: @mahatoankitkumar - addressed with a regex.  Please take a look.

mahatoankitkumar: @knutties Just a query, are we okay with key's name starting or ending with `-` or `_`?
Keys such as `_`, `_test_`, `----`, `____---`, `_____` would be valid for the above regex.

mahatoankitkumar (approved):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -28,12 +28,18 @@ use diesel::{
...
 use serde_json::{from_value, json, Map, Value};
 use service_utils::{
     result as superposition,
     service::types::{AppState, DbConnection},
 };
 
+// the first part of the regex covers for key names more than 1 character in length and
+// the second part of the regex covers for key names that is 1 character in length
+const KEY_NAME_REGEX: &str =
+    "^([a-zA-Z0-9-_][a-zA-Z0-9-_.]{0,62}[a-zA-Z0-9-_]|[a-zA-Z0-9-_])$";
```
ayushjain17, line 41: `^[a-zA-Z0-9-_]([a-zA-Z0-9-_\.]{0,62}[a-zA-Z0-9-_])?$`
can we try something like this, with this we would not need to have to break the regex into multiple parts without using any look arounds

ayushjain17: also we would need to do an escape sequence for `.` to `\.` as `.` is the accept any character regex symbol

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -50,11 +56,17 @@ async fn create(
...
+    let regex = Regex::new(KEY_NAME_REGEX).map_err(|err| {
+        unexpected_error!("could not parse regex due to: {}", err.to_string())
+    })?;
+
+    if !regex.is_match(&key) {
         return Err(bad_argument!(
-            "Configuration key cannot end with a '.' character. \
-            Please remove the trailing '.' in the key name."
+            "The key name {} is invalid, it should obey the regex {}. \
+            It can contain the following characters only [a-zA-Z0-9-_.] \
```
ayushjain17, line 66:
```suggestion
            It can contain the following characters only [a-zA-Z0-9-_\.] \
```

ayushjain17 (changes requested):
change `.` to `\.` in the regex

crates/context_aware_config/src/api/default_config/handlers.rs
knutties: Inside the character class the `.` need not be escaped I believe - it is matched literally.

ayushjain17: yeah I was not aware of it

ayushjain17 (approved):

PR #84, opened 2024-05-29
fix: Add prefix to client's methods

mahatoankitkumar:
## Problem
When running client in jenkins it shows `multiple definition of `several client's functions as exp_client and cac_client' s functions had similar name.

## Solution
Added prefix to the extern function for both cac and exp clients as `cac_` and `exp_` respectively.

### DIFF
--- a/headers/libexperimentation_client.h
+++ b/headers/libexperimentation_client.h
@@ -5,22 +5,22 @@
 
 typedef struct Arc_Client Arc_Client;
 
-int last_error_length(void);
+int expt_last_error_length(void);
 
-const char *last_error_message(void);
+const char *expt_last_error_message(void);
 
-void free_string(char *s);
+void expt_free_string(char *s);
 
-int new_client(const char *tenant, unsigned long update_frequency, const char *hostname);
+int expt_new_client(const char *tenant, unsigned long update_frequency, const char *hostname);
 
-void start_polling_update(const char *tenant);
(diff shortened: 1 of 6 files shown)


ayushjain17 (changes requested):
update haskell client as well

ayushjain17 (approved):

sauraww (approved):

headers/libexperimentation_client.h
```diff
@@ -5,22 +5,22 @@
 
 typedef struct Arc_Client Arc_Client;
 
-int last_error_length(void);
+int exp_last_error_length(void);
```
knutties, line 8: @mahatoankitkumar - can we use `expt` as the prefix for experimentation - that seems a little more standard and is also a little less ambiguous.

https://www.merriam-webster.com/dictionary/expt

knutties (changes requested):
Please consider changing the experimentation prefix to `expt`.

headers/libexperimentation_client.h
mahatoankitkumar: Yeah it's fine, but https://www.merriam-webster.com/dictionary/exp.

knutties: The ambiguity is explicit in that page 😄

Datron (approved):

knutties (approved):

PR #85, opened 2024-06-01
chore: address some clippy lints

SanchithHegde:
## Problem

The PR addresses some of the lints thrown by `cargo clippy --all-features --all-targets`.

## Solution

The PR can be easily reviewed one commit at a time so that the reviewers can gain context. I started addressing lints by applying the suggestions from clippy, and reviewed them before addressing the remaining set of warnings. The further changes were all done manually, and affect different parts of the code, specified in the commit message.

Note that this PR does not address all lints, but only some of them. I'll be addressing the remaining ones in different PRs.

One more thing I need help with is this function, which is indeed unused:
https://github.com/juspay/superposition/blob/d2ab501ec18fbd15499e8760aa8092e533acf8d6/crates/frontend/src/utils.rs#L17-L35

What should be done about it: should it be kept, or can it be removed?

What ENVs need to be added or changed: N/A

## Pre-deployment activity
Things needed to be done before deploying this change (if any): N/A

## Post-deployment activity
Things needed to be done after deploying this change (if any): N/A

## API changes

N/A

Describe any possible issues that could occur because of this change: N/A


crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -33,7 +33,7 @@ async fn create(
 ) -> superposition::Result<HttpResponse> {
     let DbConnection(mut conn) = db_conn;
 
-    if req.priority <= 0 {
+    if req.priority == 0 {
```
knutties, line 36: Why did clippy fix this - is it because priority is `uint` ?

SanchithHegde: This was not something clippy automatically fixed, clippy threw an error instead:

```text
error: this comparison involving the minimum or maximum element for this type contains a case that is always true or always false
  --> crates/context_aware_config/src/api/dimension/handlers.rs:36:8
   |
36 |     if req.priority <= 0 {
   |        ^^^^^^^^^^^^^^^^^
   |
   = help: because `0` is the minimum value for this type, the case where the two sides are not equal never occurs, consider using `req.priority == 0` instead
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons
   = note: `#[deny(clippy::absurd_extreme_comparisons)]` on by default
```

Considering that it's a `u16`, I changed the comparison to `==`.

crates/cac_client/src/interface.rs
```diff
@@ -111,22 +110,22 @@ pub extern "C" fn cac_new_client(
             .create_client(tenant.clone(), duration, hostname)
             .await
         {
-            Ok(_) => return 0,
+            Ok(_) => 0,
```
Datron, line 113: This doesn't terminate the function right? That should happen here. Early returns are required

crates/cac_client/src/lib.rs
```diff
@@ -166,10 +165,10 @@ impl Client {
                     .as_str()
                     .ok_or_else(|| {
                         log::error!("Prefix is not a valid string.");
-                        format!("Prefix is not a valid string.")
```
Datron, line 169: Do we avoid the macro here to reduce compile time? Whats the benefit of changing format! to to_string?

crates/caclang/src/helpers.rs
```diff
@@ -60,8 +60,8 @@ pub(crate) fn parse_and_validate_ctx(
 ) -> anyhow::Result<(Node, u64)> {
     let mut expr_ctx: Vec<String> = Vec::new();
     let mut priority: u64 = 0;
-    for rule in ctx.split(AND_TOKEN).into_iter() {
-        let parts: Vec<&str> = rule.trim().split_whitespace().collect();
+    for rule in ctx.split(AND_TOKEN) {
+        let parts: Vec<&str> = rule.split_whitespace().collect();
```
Datron, line 64: trim() is required here. Consider the statement
```
platform is android and client is 1mg
```
splitting on AND gives me
```
platform is android<space>
<space>client is 1mg
```
annoying to parse

crates/frontend/src/components/default_config_form.rs
```diff
@@ -222,37 +222,37 @@ where
 
                     <option
                         value="number"
-                        selected=move || { config_type.get() == "number".to_string() }
+                        selected=move || { config_type.get() == *"number" }
```
Datron, line 225: How does this work?

crates/frontend/src/pages/function/function_list.rs
```diff
@@ -123,13 +123,13 @@ pub fn function_list() -> impl IntoView {
                                             })
                                             .collect::<Vec<Map<String, Value>>>()
                                             .to_owned();
-                                        let total_pages = (v.functions.len() as f64 / 10 as f64)
+                                        let total_pages = (v.functions.len() as f64 / 10_f64)
```
Datron, line 126: Why replace `as` here? I didn't know we could use `_`

crates/service_utils/src/db/pgschema_manager.rs
```diff
@@ -58,7 +58,9 @@ impl From<Vec<ConnectionConfig>> for PgSchemaManager {
                 Pool::builder()
                     .max_size(config.count)
                     .build(manager)
-                    .expect(format!("Invalid config provided, {}", config.name).as_str()),
+                    .unwrap_or_else(|_| {
```
Datron, line 61: expect does the same thing right?

Datron (changes requested):
Just a few questions, mostly seems good!

crates/service_utils/src/db/pgschema_manager.rs
SanchithHegde:
```text
warning: use of `expect` followed by a function call
  --> crates/service_utils/src/db/pgschema_manager.rs:61:22
   |
61 |                     .expect(format!("Invalid config provided, {}", config.name).as_str()),
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `unwrap_or_else(|_| panic!("Invalid config provided, {}", config.name))`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
   = note: `#[warn(clippy::expect_fun_call)]` on by default
```

`expect` does the same thing, yes. It's only that the `format!()` will always be called (and thus, an allocation always happens) with `expect()`, while `unwrap_or_else()` involves lazy evaluation, so the `format!()` will only be called in case of errors.

Datron (approved):

crates/frontend/src/components/default_config_form.rs
sauraww: In the first line, the literal string "number" is converted to a String object using to_string() before the comparison.
In the second line, the literal string "boolean" is dereferenced using * before the comparison.
The second line is more idiomatic and performant in Rust, as it avoids the unnecessary creation of a String object and relies on the compiler's ability to optimize string literal comparisons.

sauraww (approved):

PR #86, opened 2024-06-02
chore: simplify module structure and follow Rust 2018 edition module naming convention

SanchithHegde:
## Problem

This PR addresses some more clippy lints in addition to those addressed in #85, specifically, those which concern module structure and module naming conventions.

## Solution

The PR includes the following changes:

1. Addresses the `clippy::module_inception` lint and removes unnecessary module nesting.
2. Updates all submodule file paths to follow the convention suggested since the Rust 2018 edition: using `submodule.rs` instead of `submodule/mod.rs`.
   - Refer to the relevant section in [The Rust Programming Language](https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths) book and [Rust Edition Guide](https://dev-doc.rust-lang.org/stable/edition-guide/rust-2018/path-changes.html#no-more-modrs) for more information.
   - In case the maintainers would still prefer using the old `mod.rs` convention, that can be done, it is recommended to use consistent naming regardless of what we end up with. Mixing both conventions is a bad idea.
3. Enables warnings for the [`clippy::mod_module_files`](https://rust-lang.github.io/rust-clippy/stable/index.html#mod_module_files) lint on a workspace level so that further usages of `mod.rs` files are avoided.
   - In case the maintainers prefer to use the old `mod.rs` convention, warnings on the [`clippy::self_named_module_files`](https://rust-lang.github.io/rust-clippy/stable/index.html#self_named_module_files) lint can be enabled instead.
4. Updates all modules with non-`snake_case` naming convention to use `snake_case` names instead.

Effectively, this PR boils down to some file renames and fixing imports for the affected modules. This PR, along with #85, addresses most of the clippy lints, very few of them still remain.

The changes in this PR can be verified by running the following command and confirming that clippy itself returns no warnings (ignoring those returned by the `cargo check`):

```shell
cargo clippy --all-features --all-targets -- -Aclippy::all -Wclippy::module_inception -Wclippy::mod_module_files
```

(The above command allows all lints, then explicitly enables warnings on the specified lints, thus helping ensure that they have been addressed.)

What ENVs need to be added or changed: N/A

## Pre-deployment activity
Things needed to be done before deploying this change (if any): N/A

## Post-deployment activity
Things needed to be done after deploying this change (if any): N/A

## API changes

N/A

I was not sure if the frontend would be affected because of the renames I'd done, but it turns out nothing is affected - I was able to run the server and use the UI without issues, attaching a screenshot of the same. So I'd expect no issues to pop up due to this change, apart from possible merge conflicts for other developers working on the codebase.

Screenshot of locally running server


Datron (approved):

pratikmishra356:
@SanchithHegde can you please rebase this PR

pratikmishra356 (approved):

Datron (approved):

PR #87, opened 2024-06-05
feat: add config version header in api response

pratikmishra356:
## Problem
With snapshot feature, we need to add **x-config-version** header in api response where any updates happening on context and default configs. we also need to add this header in config get apis that which version config is getting served

## Solution
Created CustomResp struct with responder trait to add custom headers in response in one go

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes

NA


crates/service_utils/Cargo.toml
```diff
@@ -35,6 +35,7 @@ reqwest = { workspace = true }
 thiserror = { workspace = true }
 once_cell = { workspace = true }
 regex = { workspace = true }
+mime = "0.3.17"
```
knutties, line 38: Can we inherit the version from the workspace - so that versions are same everywhere?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -289,7 +302,17 @@ async fn conclude_handler(
...
     .await?;
-    Ok(Json(ExperimentResponse::from(response)))
+    let headers = config_version_id.map_or(HashMap::new(), |v| {
+        HashMap::from([("x-config-version".to_string(), v)])
+    });
+    Ok(CustomResp {
+        body: ExperimentResponse::from(response),
+        headers: headers,
+    })
+    // Ok(resp_with_config_version_id(
```
mahatoankitkumar, line 312: ?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -506,8 +506,12 @@ async fn get(
                 .map_or_else(|_| json!(value), |int_val| json!(int_val)),
         );
     }
-
-    let config_version = validate_version_in_params(&mut query_params_map)?;
+    let mut resp = HttpResponse::Ok();
+    let config_version =
+        validate_version_in_params(&mut query_params_map)?.and_then(|v| {
+            resp.insert_header(("x-config-version", v.to_string().as_str()));
```
mahatoankitkumar, line 512: Better not to insert the header here.

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -527,9 +531,8 @@ async fn get(
         config = filter_config_by_dimensions(&config, &query_params_map)?
     }
 
-    let resp = HttpResponse::Ok().json(config);
+    let resp = resp.json(config);
     let audit_resp = add_audit_header(&mut conn, resp)?;
-
     add_last_modified_header(max_created_at, audit_resp)
```
mahatoankitkumar, line 536: Here at the bottom get all the headers in a map and use the type you made to construct the response.

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
pratikmishra356: yes will do it like that only

crates/context_aware_config/Cargo.toml
```diff
@@ -61,6 +61,7 @@ actix-files = { version = "0.6" }
 anyhow = { workspace = true }
 superposition_types = { path = "../superposition_types" }
 regex = { workspace = true }
+mime = "0.3.17"
```
mahatoankitkumar, line 64:
```suggestion
mime = { workspace = true }
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -62,43 +64,33 @@ fn validate_version_in_params(
         })
 }
 
-pub fn add_audit_header(
+pub fn add_audit_id(
```
Datron, line 67:
```suggestion
pub fn add_audit_id_to_header(
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -62,43 +64,33 @@ fn validate_version_in_params(
...
-        }
+        header_map.insert(AppHeader::XAuditId.to_string(), uuid.to_string());
     } else {
         log::error!("Failed to fetch contexts from event_log");
     }
-    Ok(res)
 }
 
-fn add_last_modified_header(
+fn add_last_modified(
```
Datron, line 83:
```suggestion
fn add_last_modified_to_header(
```

crates/service_utils/src/service/types.rs
```diff
@@ -29,6 +33,14 @@ pub enum AppEnv {
     DEV,
 }
 
+#[derive(Copy, Clone, Debug, strum_macros::Display)]
+#[strum(serialize_all = "kebab-case")]
+pub enum AppHeader {
```
Datron, line 38: nice

crates/service_utils/src/service/types.rs
```diff
@@ -228,3 +240,31 @@ impl FromRequest for CustomHeaders {
         ready(Ok(val))
     }
 }
+
+pub struct CustomResp<T> {
```
Datron, line 244:
```suggestion
pub struct AppResponse<T> {
```

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -463,17 +477,23 @@ async fn move_handler(
...
-        Ok(move_reponse)
+        let version_id = add_config_version(&state, tags, transaction_conn)?;
+        let headers = HashMap::from([(
+            AppHeader::XConfigVersion.to_string(),
+            version_id.to_string(),
+        )]);
+        Ok(CustomResp {
+            body: move_reponse,
+            headers: headers,
+        })
```
Datron, line 496: This whole part should be a function that takes the header enum and the response and wraps it appropriately

crates/service_utils/src/service/types.rs
```diff
@@ -228,3 +240,31 @@ impl FromRequest for CustomHeaders {
...
+    pub body: T,
+    pub headers: HashMap<String, String>,
+}
+impl<T: Serialize> Responder for CustomResp<T> {
+    type Body = EitherBody<String>;
```
Datron, line 249: This should be 
```suggestion
    type Body = EitherBody<T>;
```
right?

crates/service_utils/src/service/types.rs
```diff
@@ -228,3 +240,31 @@ impl FromRequest for CustomHeaders {
...
+
+    fn respond_to(self, _: &HttpRequest) -> HttpResponse<Self::Body> {
+        match (serde_json::to_string(&self.body), self.headers) {
```
Datron, line 252: Is there any benefit of doing this? We still end up writing the code to add headers in every function. Why can't I just add the headers as well? I don't need the responder for that.

Datron (changes requested):
Do we need the Responder Pattern here? In every function we have to still fetch and set the headers, I am wondering if this is a needless abstraction

crates/context_aware_config/src/api/context/handlers.rs
pratikmishra356: @Datron we have to construct the headerMap inside handler only , then only thing left is to construct CustomResp, do we need another function for this ?

Datron (approved):

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -265,7 +273,12 @@ async fn create(
     let inserted_experiment: Experiment = inserted_experiments.remove(0);
     let response = ExperimentCreateResponse::from(inserted_experiment);
 
-    Ok(Json(response))
+    let mut http_resp = HttpResponse::Ok();
+    config_version_id.and_then(|v| {
+        http_resp.insert_header((AppHeader::XConfigVersion.to_string(), v.clone()));
+        Some(v)
+    });
+    Ok(http_resp.json(response))
```
mahatoankitkumar, line 281: This block is repeatedly used everywhere. Can be made it into a function.

mahatoankitkumar (changes requested):

crates/experimentation_platform/src/api/experiments/handlers.rs
pratikmishra356: it  is dynamic with respect to handler , so it wouldn't reduce repetition , also this gives flexibility to add multiple headers and tweak httpresponse , so its fine

mahatoankitkumar: @pratikmishra356 You are inserting `AppHeader::XConfigVersion` in the handlers response. You can do that through a fucntion also by passing the `version_id` to it. Correct me if i am missing something.

pratikmishra356: No you are right , we can move config_version insertion in a function, 
but `http_resp.json(response)` should be kept inside the handler only

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -258,12 +266,17 @@ async fn delete(
...
                 Ok(0) => Err(not_found!("default config key `{}` doesn't exists", key)),
                 Ok(_) => {
-                    add_config_version(&state, tags, transaction_conn)?;
+                    let version_id = add_config_version(&state, tags, transaction_conn)?;
                     log::info!(
                         "default config key: {key} deleted by {}",
                         user.get_email()
                     );
-                    Ok(HttpResponse::NoContent().finish())
+                    Ok(HttpResponse::NoContent()
```
mahatoankitkumar, line 274: `add_config_version_to_header` can be added here also.
And you can add `add_config_version_to_header` in service_utils so that you don't have to repeat that in the both the crates.

mahatoankitkumar (dismissed):

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: both crates have different data types for config_version_header.  one has string , the other has i64

mahatoankitkumar: And why is it?
Can't both be made of same type?

pratikmishra356: first , that function should not be on service_utils as it is more of handler level function not global level
second , in experiment we are fetching from http headers of (context api response) , so there is not point of first mapping string to i64 , then calling the same function, it does not reduce code in anyway

pratikmishra356: @mahatoankitkumar adding add_config_version_to_header  function is unecessary here , because we aren't doing any checks or typ conversion , its pretty straight forward , just insert it in http_resp

mahatoankitkumar (approved):

PR #88, opened 2024-06-06
Rework async library in CAC client

Datron:
## Problem
Clippy while analyzing CAC client shows a error for calling await on a MutexGuard. This happens when we use the standard library RwLock. 

## Solution
Use tokio RwLock that handles this and is better at concurrent and parallel processing. Also add an integration example for cac client that lets implementers test out features of the client

Screenshot 2024-06-06 at 12 23 39 PM

### DIFF
--- a/crates/cac_client/src/interface.rs
+++ b/crates/cac_client/src/interface.rs
@@ -12,13 +12,13 @@ use std::{
     ffi::{c_int, CString},
     time::Duration,
 };
-use tokio::{runtime::Runtime, task};
+use tokio::runtime::Runtime;
 
 thread_local! {
     static LAST_ERROR: RefCell<Option<String>> = const { RefCell::new(None) };
 }
 
-pub static CAC_RUNTIME: Lazy<Runtime> =
+static CAC_RUNTIME: Lazy<Runtime> =
     Lazy::new(|| Runtime::new().expect("The runtime was not intialized"));
 
 macro_rules! null_check {
@@ -108,8 +108,7 @@ pub extern "C" fn cac_new_client(
     let hostname = unwrap_safe!(cstring_to_rstring(hostname), return 1);
 
     // println!("Creating cac client thread for tenant {tenant}");
-    let local = task::LocalSet::new();
-    local.block_on(&CAC_RUNTIME, async move {
+    CAC_RUNTIME.block_on(async move {
         match CLIENT_FACTORY
             .create_client(tenant.clone(), duration, hostname)
             .await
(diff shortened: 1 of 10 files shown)


crates/cac_client/src/interface.rs
```diff
@@ -105,7 +109,7 @@ pub extern "C" fn cac_new_client(
 
     // println!("Creating cac client thread for tenant {tenant}");
     let local = task::LocalSet::new();
-    local.block_on(&Runtime::new().unwrap(), async move {
+    local.block_on(&CAC_RUNTIME, async move {
```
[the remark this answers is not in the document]
Datron: We can do this for all functions expect `cac_start_polling_update` since I am expecting the FFI language to spawn a thread to run this function in the background

crates/cac_client/src/interface.rs
```diff
@@ -17,6 +18,9 @@ thread_local! {
     static LAST_ERROR: RefCell<Option<String>> = const { RefCell::new(None) };
 }
 
+pub static CAC_RUNTIME: Lazy<Runtime> =
```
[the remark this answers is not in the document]
Datron: Since its used only in this file, I'll keep it private. No point exposing it

crates/cac_client/src/interface.rs
```diff
@@ -127,10 +130,7 @@ pub extern "C" fn cac_start_polling_update(tenant: *const c_char) {
         null_check!(client, "CAC client for tenant not found", return);
         let local = task::LocalSet::new();
         // println!("in FFI polling");
-        local.block_on(
-            &Runtime::new().unwrap(),
-            (*client).clone().run_polling_updates(),
-        );
+        local.block_on(&CAC_RUNTIME, (*client).clone().run_polling_updates());
```
SanchithHegde, line 133: Can't we use `spawn()`?

```suggestion
        let _handle = CAC_RUNTIME.spawn((*client).clone().run_polling_updates());
```

I think it should be possible, since `run_polling_updates()` does not involve anything that "blocks" the current thread (neither the `tick()` on `tokio`'s `Interval`, nor does the config fetch). So I think it can be moved across threads by the `tokio` scheduler without issues.

Datron: The function update_cac() is blocking, it tries to acquire a write lock on its config data structure which won't resolve instantly.
https://github.com/juspay/superposition/pull/88/files/2d74390f9b6bd987715434a7ef164a389037316f#diff-0b91e6bae95567932f131f6998310c46af0fb7e3692e85cefeb6aff1a28d6689R140

SanchithHegde: Since we're now using [`tokio::sync::RwLock::write()`](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#method.write), it would not block the thread until the lock is acquired, but instead would allow other tasks to be run on the runtime:

> Locks this RwLock with exclusive write access, causing the current task to yield until the lock has been acquired.
>
> The calling task will yield while other writers or readers currently have access to the lock.
>
> Returns an RAII guard which will drop the write access of this RwLock when dropped.
>
> [...]

Datron: The expectation here is that this function should be run within a thread spawned by the implementing language. Would be better to block the thread so that it does not join back into the main thread

ShubhranshuSanjeev (approved):

SanchithHegde (approved):

mahatoankitkumar (approved):

PR #91, opened 2024-06-06
feat: Support relative path for frontend

ayushjain17:
All APIs from the front-end should be done to paths relative to the domain used to host the frontend application.  The API endpoints used by frontend - should not be hard-coded as the application can be hosted on any domain name.

Change log:
- fixed usage of __APP_ENV
- if the APP_ENV's host value is set as "" or "/", the paths for api calls are taken relative to frontend's host origin

### DIFF
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -77,7 +77,14 @@ pub fn get_host() -> String {
                 None
             }
         })
-        .unwrap_or(String::from("http://localhost:xxxx"));
+        .map_or(String::from("http://localhost:xxxx"), |host_from_env| {
+            let opt_host_origin =
+                web_sys::window().and_then(|window| window.location().origin().ok());
+            match (host_from_env.as_str(), opt_host_origin) {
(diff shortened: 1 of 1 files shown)


crates/frontend/src/utils.rs
```diff
@@ -78,6 +78,16 @@ pub fn get_host() -> String {
             }
         })
         .unwrap_or(String::from("http://localhost:xxxx"));
+    match host.as_str() {
```
Datron, line 81: This can be done inside match js_sys::eval() block, if not can we use if condition instead of match?

Datron: You can also avoid mutating host variable

Datron (changes requested):
A better description for the PR is needed to understand this change

pratikmishra356 (approved):

Datron (approved):

Datron (approved):

ShubhranshuSanjeev (approved):

PR #99, opened 2024-06-11
fix: parse type when submitting overrides, not on input

Datron:
## Problem
When you edit an override, the textarea sets the 'type' of values as String(value). On input in the field, the value is parsed to the correct type using the function `get_config_value`. Hence, if you are editing an input but leave the others as is, the submit request will fail because all types are sent as strings, even if they are not.

## Solution
Parse the type at the time of submitting the form, not on ever input


Datron:
https://github.com/juspay/superposition/assets/15166178/c7a98f7c-5e91-4463-8346-4f247c2a60b7

ShubhranshuSanjeev (approved):

Issue #117, opened 2024-06-13
Filter config by dimensions logic uses `equality check` on values mentioned in `context's` json-logic rather than json-logic apply (will need special partial apply) 


ShubhranshuSanjeev:
(From PR description)

Problem

    Filter currently works only for in operator with array and for all other types it checks direct equality
    Experiment filtering was not doing actual filter, rather it was applying the entire context
    Config prefix filtering was sent in context
    Prefix filter param of expt filtering not exposed to hs-client

Solution

    Use jsonlogic's partial_apply for config and experiment filtering
    Add separate function for Experiment filtering just like the filter function for config - also exposed to hs-client.
    Introduce separate param for prefix in config filtering.
    Exposed prefix filter param of expt filtering to hs-client.

PR #127, opened 2024-06-18
Use jsonlogic's partial_apply for config and experiment filtering

ayushjain17:
## Issue
#117 

## Problem
1. Filter currently works only for in operator with array and for all other types it checks direct equality
2. Experiment filtering was not doing actual filter, rather it was applying the entire context
3. Config prefix filtering was sent in context
4. Prefix filter param of expt filtering not exposed to hs-client

## Solution
1. Use jsonlogic's partial_apply for config and experiment filtering
2. Add separate function for Experiment filtering just like the filter function for config - also exposed to hs-client.
3. Introduce separate param for prefix in config filtering.
4. Exposed prefix filter param of expt filtering to hs-client.

### DIFF
--- a/crates/cac_client/src/interface.rs
+++ b/crates/cac_client/src/interface.rs
@@ -176,38 +176,43 @@ pub extern "C" fn cac_get_last_modified(client: *mut Arc<Client>) -> *const c_ch
 #[no_mangle]
 pub extern "C" fn cac_get_config(
     client: *mut Arc<Client>,
-    query: *const c_char,
+    filter_query: *const c_char,
+    filter_prefix: *const c_char,
 ) -> *const c_char {
-    let filter = if query.is_null() {
-        None
-    } else {
-        let filter_string = match cstring_to_rstring(query) {
-            Ok(s) => s,
-            Err(err) => {
-                update_last_error(err);
-                return std::ptr::null();
-            }
-        };
-        let filters: Map<String, Value> =
-            match serde_json::from_str::<Map<String, Value>>(filter_string.as_str()) {
-                Ok(json) => json,
-                Err(err) => {
-                    update_last_error(err.to_string());
-                    return std::ptr::null();
-                }
-            };
-        Some(filters)
-    };
-
     null_check!(
         client,
         "an invalid null pointer client is being used, please call get_client()",
         return std::ptr::null()
     );
+
+    let filters = if filter_query.is_null() {
+        None
+    } else {
+        let filter_string =
+            unwrap_safe!(cstring_to_rstring(filter_query), return std::ptr::null());
+        let filters: Map<String, Value> = unwrap_safe!(
+            serde_json::from_str::<Map<String, Value>>(filter_string.as_str()),
+            return std::ptr::null()
+        );
+
+        Some(filters).filter(|filters| !filters.is_empty())
+    };
+
+    let prefix_list = if filter_prefix.is_null() {
+        None
+    } else {
+        let filter_string =
+            unwrap_safe!(cstring_to_rstring(filter_prefix), return std::ptr::null());
+        let prefix_list: Vec<String> =
+            filter_string.split(',').map(String::from).collect();
+
+        Some(prefix_list).filter(|list| !list.is_empty())
+    };
+
     unwrap_safe!(
         unsafe {
             (*client)
-                .get_full_config_state_with_filter(filter)
+                .get_full_config_state_with_filter(filters, prefix_list)
                 .map(|config| {
                     rstring_to_cstring(serde_json::to_value(config).unwrap().to_string())
                         .into_raw()
--- a/crates/cac_client/src/lib.rs
+++ b/crates/cac_client/src/lib.rs
@@ -7,7 +7,7 @@ use chrono::{DateTime, Utc};
 use derive_more::{Deref, DerefMut};
 use reqwest::{RequestBuilder, Response, StatusCode};
 use serde::{Deserialize, Serialize};
-use serde_json::{Map, Value};
+use serde_json::{json, Map, Value};
 use std::{
     collections::{HashMap, HashSet},
     convert::identity,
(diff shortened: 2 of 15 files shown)


crates/experimentation_client/src/lib.rs
```diff
@@ -166,6 +169,49 @@ impl Client {
         Ok(experiments)
     }
 
+    fn filter_experiments_by_prefix(
```
ayushjain17, line 172: extracted as a common function

crates/cac_client/src/interface.rs
```diff
@@ -176,12 +176,19 @@ pub extern "C" fn cac_get_last_modified(client: *mut Arc<Client>) -> *const c_ch
 #[no_mangle]
 pub extern "C" fn cac_get_config(
     client: *mut Arc<Client>,
-    query: *const c_char,
+    filter_query: *const c_char,
+    filter_prefix: *const c_char,
 ) -> *const c_char {
-    let filter = if query.is_null() {
+    null_check!(
```
ayushjain17, line 182: INFO: moved [this check](https://github.com/juspay/superposition/pull/127/files#diff-1cc6f99f841b0d796e99630ff9386809a7735358480a55c7888c492a13c59581L202) here

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -637,7 +633,7 @@ async fn get_filtered_config(
         })?;
     let mut query_params_map: serde_json::Map<String, Value> = Map::new();
```
mahatoankitkumar, line 634: @ayushjain17 Can we remove this handler? I don't think this is being used anywhere, getConfig can be used for the same purpose.

crates/cac_client/src/lib.rs
```diff
@@ -352,14 +339,12 @@ pub fn filter_config_by_dimensions(
      -> superposition::Result<Vec<Context>> {
```
pratikmishra356, line 339: closure is not needed here

crates/context_aware_config/src/api/config/handlers.rs
ayushjain17: Sure, redirecting both the endpoints to the same url
or should I deprecate the filter endpoint ?

crates/cac_client/src/lib.rs
```diff
@@ -157,29 +155,20 @@ impl Client {
...
-
-            query_map.remove("prefix");
+        if let Some(prefix_list) = prefix {
+            config = filter_config_by_prefix(&config, &HashSet::from_iter(prefix_list))
+                .map_err_to_string()?;
+        }
 
-            if !query_map.is_empty() {
+        match query_data {
+            Some(query_map) if !query_map.is_empty() => {
```
Datron, line 167: Can we avoid the match here since the catch all case just throws `Unit`. If seems better, I would recommend `and_then`

crates/context_aware_config/src/api/config/handlers.rs
mahatoankitkumar: I think we should deprecate this endpoint as it has the same logic as the above.

crates/cac_client/src/lib.rs
```diff
@@ -157,29 +155,20 @@ impl Client {
...
-                    .map_err_to_string()?
-                    .split(',')
-                    .collect();
-                config =
-                    filter_config_by_prefix(&config, &prefix_list).map_err_to_string()?;
-            }
-
-            query_map.remove("prefix");
+        if let Some(prefix_list) = prefix {
+            config = filter_config_by_prefix(&config, &HashSet::from_iter(prefix_list))
```
Datron, line 162: `nitpick`
Can we take HashSet as function args instead of building it here from Vector?

Datron (changes requested):
You will need to pull my client changes for async locking and build on top of that too.

crates/cac_client/src/lib.rs
```diff
@@ -345,27 +332,20 @@ pub fn filter_config_by_prefix(
...
-                query_params_map.get(key).map_or(true, |val| {
-                    val == value || val.as_array().unwrap_or(&vec![]).contains(value)
-                })
-            });
-            if should_add_ctx {
-                filtered_context.push(context.clone());
+    let filtered_context = config
+        .contexts
+        .iter()
+        .filter_map(|context| {
```
pratikmishra356, line 340: can we use only filter

crates/context_aware_config/src/api/config/helpers.rs
```diff
@@ -2,39 +2,28 @@ use std::collections::HashSet;
...
 use super::types::{Config, Context};
 
-use serde_json::{Map, Value};
-use service_utils::{
-    helpers::extract_dimensions, result as superposition, unexpected_error,
-};
+use serde_json::{json, Map, Value};
+use service_utils::{result as superposition, unexpected_error};
 
 pub fn filter_context(
```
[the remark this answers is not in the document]
ayushjain17: filter operations are all based on partial_apply itself, I dont think we dont need to mention it separately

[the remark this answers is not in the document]
pratikmishra356: still its filtering based on a particular thing ,so its better to mention

crates/experimentation_client/src/interface.rs
```diff
@@ -208,9 +221,62 @@ pub extern "C" fn expt_get_satisfied_experiments(
...
+            }
+        };
+        Some(prefix_list)
+    };
+
+    let local = task::LocalSet::new();
+    let experiments = local.block_on(&Runtime::new().unwrap(), unsafe {
+        (*client).get_satisfied_experiments(&context, prefix_list)
+    });
+    let experiments = match serde_json::to_value(experiments) {
```
sauraww, line 241: `let experiments = serde_json::to_value(experiments).map_err(|err| error_block(err.to_string()))?;`

sauraww: Can't do this since it is not result type.

sauraww (approved):

Datron (approved):

PR #129, opened 2024-06-19
fix: Improve form inputs

mahatoankitkumar:
## Problem
Better form fields
- enum → Toggle
- boolean → dropdown with true and false options

### DIFF
--- a/crates/frontend/src/components/context_form.rs
+++ b/crates/frontend/src/components/context_form.rs
@@ -1,34 +1,37 @@
 pub mod utils;
 
 use crate::components::dropdown::{Dropdown, DropdownDirection};
-use crate::types::Dimension;
+use crate::components::input_components::{BooleanToggle, EnumDropdown};
+use crate::types::{Dimension, ContextCondition};
+use crate::utils::get_key_type;
 use leptos::*;
-use std::collections::HashSet;
+use serde_json::{Map, Value};
+use std::collections::{HashMap, HashSet};
 use web_sys::MouseEvent;
 
 #[component]
-pub fn context_form<NF>(
-    handle_change: NF,
+pub fn context_form(
     dimensions: Vec<Dimension>,
     #[prop(default = false)] is_standalone: bool,
-    context: Vec<(String, String, String)>,
+    context: ReadSignal<Vec<ContextCondition>>,
+    set_context: WriteSignal<Vec<ContextCondition>>,
     #[prop(default = String::new())] heading_sub_text: String,
     #[prop(default = false)] disabled: bool,
     #[prop(default = DropdownDirection::Right)] dropdown_direction: DropdownDirection,
     #[prop(default = false)] resolve_mode: bool,
 ) -> impl IntoView
-where
-    NF: Fn(Vec<(String, String, String)>) + 'static,
+// where
+//     NF: Fn(Vec<(String, String, String)>) + 'static,
 {
     let _has_dimensions = !dimensions.is_empty();
 
     let (used_dimensions, set_used_dimensions) = create_signal(
-        context
+        context.get()
             .iter()
-            .map(|(d, _, _)| d.clone())
+            .map(|ele| ele.dimension.get().clone())
             .collect::<HashSet<String>>(),
     );
-    let (context, set_context) = create_signal(context.clone());
+    // let (context, set_context) = create_signal(context.clone());
 
     let dimensions = StoredValue::new(dimensions);
 
@@ -40,17 +43,17 @@ where
         //TODO: submit logic for this
     };
 
-    create_effect(move |_| {
-        let f_context = context.get();
-        logging::log!("effect {:?}", f_context);
-        handle_change(f_context.clone());
-    });
+    // create_effect(move |_| {
+    //     let f_context = context.get();
+    //     logging::log!("effect {:?}", f_context);
+    //     handle_change(f_context.clone());
+    // });
 
     let handle_select_dropdown_option =
         Callback::new(move |selected_dimension: Dimension| {
             let dimension_name = selected_dimension.dimension;
             set_context.update(|value| {
-                value.push((dimension_name.clone(), "".to_string(), "".to_string()))
+                value.push(ContextCondition::new(dimension_name.clone(), "".to_string(), "".to_string()))
             });
             set_used_dimensions.update(|value: &mut HashSet<String>| {
                 value.insert(dimension_name);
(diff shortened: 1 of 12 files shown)


crates/frontend/src/components/dimension_form.rs
```diff
@@ -165,7 +165,7 @@ where
                             <textarea
                                 type="text"
                                 placeholder="JSON schema"
-                                class="input input-bordered mt-5 rounded-md resize-y w-full max-w-md"
+                                class="input input-bordered mt-5 rounded-md resize-y w-full max-w-md pt-3"
```
Datron, line 168: `nitpick`
mt-3 would be better

crates/frontend/src/components/default_config_form.rs
```diff
@@ -197,52 +197,117 @@ where
...
-                                        class="input input-bordered w-full max-w-md"
-                                        on:change=move |ev| {
-                                            logging::log!("{:?}", event_target_value(& ev));
-                                            set_config_value.set(event_target_value(&ev));
-                                        }
-                                    >
-
-                                        {config_value.get()}
-                                    </textarea>
+                let schema: Map<String, Value> = serde_json::from_value(config_schema_rs.get())
```
Datron, line 200: Should deciding input field based on schema type be a component of it's own? It is used across 3 files and should behave the same for all of them

Datron (changes requested):
I am thinking the input field should be made a component called DynamicInput that can change based on the JSON schema of an item. If that can be done in a follow up PR, I can go ahead and approve.

crates/frontend/src/components/dimension_form.rs
mahatoankitkumar: pt-3 is for the inner padding of the `textarea`, So that it looks cleaner, there is already a `mt` applied to it.

crates/frontend/src/components/default_config_form.rs
mahatoankitkumar: Sure.

crates/frontend/src/components/dimension_form.rs
pratikmishra356, line 1: #[prop(default = 0)] priority: u16,
can you change default to something else  , since we do not support priority 0 
better is to change it to option<number> 
and if someone dont provide number then you should not let it submit

crates/frontend/src/components/override_form.rs
pratikmishra356, line 1: Screenshot 2024-06-20 at 15 59 08
@mahatoankitkumar can you fix this override button it should be aligned with context button

crates/frontend/src/components/default_config_form.rs
ShubhranshuSanjeev, line 1: @mahatoankitkumar , we have similar fields for boolean and enum in both default_config and overrides form, can you make these a component and use them at both places, and also I think you should make similar changes for the context_form.

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/override_form.rs
mahatoankitkumar: Didn't get you

Datron (approved):

crates/frontend/src/components/input_components.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[component]
+pub fn enum_dropdown(
+    schema: Map<String, Value>,
+    config_value: String,
+    handle_change: Callback<String, ()>,
+    #[prop(default = String::new())] class: String,
+    #[prop(default = false)] disabled: bool,
+) -> impl IntoView {
+    let mut enum_array: Vec<String> = vec![];
```
pratikmishra356, line 14: @mahatoankitkumar lets not assume enum array , as array of string only , it can be integer or decimal also
can we have that support , because if I am declaring enum of number in UI form dropdown shows nothing

crates/frontend/src/components/context_form.rs
```diff
@@ -81,127 +84,193 @@ where
...
+                                .get()
+                                .into_iter()
+                                .enumerate()
+                                .map(|(idx, (dimension, mut operator, value))| {
+                                    let dimension_label = dimension.to_string();
+                                    let dimension_name = StoredValue::new(dimension.to_string());
+                                    let schema: Map<String, Value> = serde_json::from_value(
+                                            dimensions_map.get(&dimension_label).unwrap().schema.clone(),
+                                        )
+                                        .unwrap();
```
sauraww, line 104: Can we avoid doing plain unwrap here ?

sauraww (changes requested):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -197,52 +199,82 @@ where
...
-                                </label>
-                                {input_format}
-                            </div>
-                            <div class="divider"></div>
+                            <textarea
+                                type="text"
+                                placeholder="Value"
+                                class="input input-bordered w-full max-w-md pt-3"
+                                on:change=move |ev| {
+                                    logging::log!("{:?}", event_target_value(& ev));
```
sauraww, line 258: Here as well

crates/frontend/src/components/override_form.rs
```diff
@@ -40,6 +48,32 @@ where
...
+        get_config_value(
+            &config_key_value,
+            &value,
+            &default_config
+                .get_value()
+                .into_iter()
+                .map(ConfigType::DefaultConfig)
+                .collect::<Vec<_>>(),
+        )
+        .expect("can't parse default config key")
```
sauraww, line 61: Can we use unwrap_or instead and pass a json value having config_key_value : 
`unwrap_or(json!(config_key_value)`

sauraww (changes requested):

crates/frontend/src/utils.rs
```diff
@@ -449,3 +449,14 @@ pub fn set_local_storage(_key: &str, _value: &str) -> Option<()> {
         }
     }
 }
+
+pub fn get_key_type(schema: &Map<String, Value>) -> String {
+    if schema.contains_key("enum") {
+        String::from("ENUM")
+    } else {
+        match schema.get("type").unwrap_or(&Value::Null) {
+            Value::String(str_) => str_.to_ascii_uppercase(),
```
sauraww, line 458: `Value::String(str) => str.to_ascii_uppercase(),`

sauraww (changes requested):

Issue #111, opened 2024-06-13
Change default value for type template form


ShubhranshuSanjeev:
![image](https://github.com/juspay/superposition/assets/20094367/68d94048-c689-4b21-813b-3e9b40814cb3)

PR #131, opened 2024-06-20
Fix type template dropdown and form issues

### LINKED ISSUE #112: Default state `Choose a type template` shown in `Set Schema` dropdown, even if there is template selected
image

Datron:
## Problem
This PR closes #112 and #111 

## Solution
Screenshot 2024-06-20 at 4 50 23 PM
Screenshot 2024-06-20 at 4 50 35 PM

### DIFF
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -150,7 +150,9 @@ where
             <Suspense>
                 {move || {
                     let options = type_template_resource.get().unwrap_or(vec![]);
-                    let config_t = if config_type_rs.get().is_empty() {
+                    let config_t = if config_type_rs.get().is_empty() && edit {
+                        "change current type template".into()
+                    } else if config_type_rs.get().is_empty() && !edit {
                         "Choose a type template".into()
                     } else {
                         config_type_rs.get()
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -155,7 +155,11 @@ where
                     } else {
                         config_type_rs.get()
                     };
-                    let config_textarea = format!("{}", config_schema_rs.get());
+                    let config_textarea = if config_schema_rs.get().is_null() {
+                        String::from("")
+                    } else {
+                        format!("{}", config_schema_rs.get())
+                    };
                     view! {
                         <div class="form-control">
                             <label class="label">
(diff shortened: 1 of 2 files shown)


crates/frontend/src/components/default_config_form.rs
```diff
@@ -150,12 +150,18 @@ where
             <Suspense>
                 {move || {
                     let options = type_template_resource.get().unwrap_or(vec![]);
-                    let config_t = if config_type_rs.get().is_empty() {
+                    let config_t = if config_type_rs.get().is_empty() && edit {
+                        "change current type template".into()
```
sauraww, line 154: Why are we allowing the default config type to change ?
Wouldn't that be an issue for existing overrides , just wanted to understand the use case where we would need it.

sauraww (changes requested):

crates/frontend/src/components/override_form.rs
```diff
@@ -117,7 +117,7 @@ where
                                                             .map(ConfigType::DefaultConfig)
                                                             .collect::<Vec<_>>(),
                                                     )
-                                                    .expect("can't parse default config key");
+                                                    .unwrap_or(json!(input_value));
```
sauraww, line 120: @Datron we would also need to make the similar changes at line number 35 and 51 in `crates/frontend/src/components/context_form/utils.rs `

sauraww (changes requested):

crates/frontend/src/components/default_config_form.rs
Datron: @sauraww adding an enum, changing the pattern - these are few examples where we would want to update the type. I replicated what was the behaviour today. If we want to stop this, we discuss and do it in another PR

crates/frontend/src/components/override_form.rs
Datron: @sauraww why? There is no type templates in context form

Datron: If this is an issue, it seems out of scope of this PR

crates/frontend/src/components/default_config_form.rs
mahatoankitkumar: @sauraww this functionality of changing the schema of a default_config is already there. And the change would effect any new overrides i think, it wouldn't effect the existing ones i guess.

mahatoankitkumar (approved):

PR #133, opened 2024-06-25
Fix/tweak context parsing

mahatoankitkumar:
## Problem
For number in context, it would consider it as an integer as whole, even if it was an array

## Solution
Fixed the parsing for number

### DIFF
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -319,17 +319,22 @@ pub fn get_config_value(
         Some(ConfigValueType::Boolean) => bool::from_str(val)
             .map(Value::Bool)
             .map_err(|_| "Invalid boolean".to_string()),
-        Some(ConfigValueType::Number) => val
-            .parse::<i64>()
-            .map(|number| Value::Number(number.into()))
-            .or_else(|_| {
-                f64::from_str(val)
-                    .ok()
-                    .and_then(|num| Number::from_f64(num).map(Value::Number))
-                    .ok_or_else(|| {
-                        "Invalid decimal format or precision issue".to_string()
-                    })
-            }),
+        Some(ConfigValueType::Number) => {
+            let parsed_value: Value = serde_json::from_str(val)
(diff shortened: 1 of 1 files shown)


crates/frontend/src/utils.rs
```diff
@@ -319,17 +319,22 @@ pub fn get_config_value(
...
-                f64::from_str(val)
-                    .ok()
-                    .and_then(|num| Number::from_f64(num).map(Value::Number))
-                    .ok_or_else(|| {
-                        "Invalid decimal format or precision issue".to_string()
-                    })
-            }),
+        Some(ConfigValueType::Number) => {
+            let parsed_value: Value = serde_json::from_str(val)
+                .map_err(|_| "Invalid number or number array format".to_string())?;
```
sauraww, line 324: why do we need a number array format inside Number ?
Can you give an example or a screenshot ?

sauraww (changes requested):

crates/frontend/src/utils.rs
mahatoankitkumar: for `in` case, if you give [1,2,3], it will try to convert it as number as a whole.

sauraww: Can we separate this logic based on the operator? We need similar validation for booleans and strings as well. Also, the code looks messy. We should keep the existing logic, but move the array logic to a different function for better organization and handling.

sauraww: @mahatoankitkumar

sauraww (changes requested):

crates/frontend/src/utils.rs
mahatoankitkumar: @sauraww Why would we need similar validation for booleans and strings?

sauraww: There can be strings in the array right like list of clientIds, we can have a use case like that

sauraww (changes requested):

PR #134, opened 2024-06-26
fix: handled parsing for multiple possible types

ShubhranshuSanjeev:
## Problem
Parsing failed for values when type mentioned in schema had multiple value such as `{ type: ["null", "string"] }`.

## Solution
Try parsing for all possible types mentioned in the `type` property, return the parsed value for whichever type it succeeds

### DIFF
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -278,48 +278,71 @@ pub enum ConfigValueType {
     Boolean,
     Number,
     String,
+    Null,
+    Integer,
     Other,
+}
+
+impl FromStr for ConfigValueType {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "boolean" => Ok(ConfigValueType::Boolean),
+            "number" => Ok(ConfigValueType::Number),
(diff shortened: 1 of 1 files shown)


Datron (approved):

crates/frontend/src/utils.rs
```diff
@@ -278,48 +278,71 @@ pub enum ConfigValueType {
...
+impl FromStr for ConfigValueType {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "boolean" => Ok(ConfigValueType::Boolean),
+            "number" => Ok(ConfigValueType::Number),
+            "string" => Ok(ConfigValueType::String),
+            "null" => Ok(ConfigValueType::Null),
+            "integer" => Ok(ConfigValueType::Integer),
+            _ => Ok(ConfigValueType::Other),
```
pratikmishra356, line 295: We can implement for array and object also

pratikmishra356 (approved):

crates/frontend/src/utils.rs
pratikmishra356: its fine we can merge , anyway its working fine

PR #135, opened 2024-06-27
fix: allow positive i32 number for dimension priority in dimension create

ayushjain17:
## Problem
DB supports i32 values for priority but endpoints were restricted to u16

## Solution
allow positive i32 number for dimension priority in dimension create

### DIFF
--- a/crates/context_aware_config/src/api/dimension/handlers.rs
+++ b/crates/context_aware_config/src/api/dimension/handlers.rs
@@ -33,7 +33,7 @@ async fn create(
 ) -> superposition::Result<HttpResponse> {
     let DbConnection(mut conn) = db_conn;
 
-    if req.priority == 0 {
+    if req.priority <= 0 {
         return Err(bad_argument!("Priority should be greater than 0"));
     }
 
@@ -66,7 +66,7 @@ async fn create(
 
     let new_dimension = Dimension {
         dimension: create_req.dimension,
-        priority: i32::from(create_req.priority),
+        priority: create_req.priority,
         schema: schema_value,
         created_by: user.get_email(),
         created_at: Utc::now(),
--- a/crates/context_aware_config/src/api/dimension/types.rs
+++ b/crates/context_aware_config/src/api/dimension/types.rs
@@ -4,7 +4,7 @@ use serde_json::Value;
 #[derive(Debug, Deserialize)]
 pub struct CreateReq {
     pub dimension: String,
-    pub priority: u16,
+    pub priority: i32,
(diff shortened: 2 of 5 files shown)


crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -4,7 +4,7 @@ use serde_json::Value;
 #[derive(Debug, Deserialize)]
 pub struct CreateReq {
     pub dimension: String,
-    pub priority: u16,
+    pub priority: i32,
```
Datron, line 7: If we are expecting only positive numbers, maybe use u32?
```suggestion
    pub priority: u32,
```

crates/frontend/src/components/dimension_form.rs
```diff
@@ -195,8 +195,8 @@ where
                             class="input input-bordered w-full max-w-md"
                             value=priority.get()
                             on:change=move |ev| {
-                                logging::log!("{:?}", event_target_value(& ev).parse::< u16 > ());
-                                match event_target_value(&ev).parse::<u16>() {
+                                logging::log!("{:?}", event_target_value(& ev).parse::< u32 > ());
```
Datron, line 198:
```suggestion
                                logging::log!("{:?}", event_target_value(& ev).parse::<u32>());
```

Datron (approved):

crates/context_aware_config/src/api/dimension/types.rs
ayushjain17: avoided this, as we needed i32 in the next ops,
and then conversion from u32 to i32 is more likely to panic

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -37,6 +37,10 @@ async fn create(
         return Err(bad_argument!("Priority should be greater than 0"));
     }
 
+    if req.priority < 0 {
+        return Err(bad_argument!("Expected positive value for priority",));
+    }
+
```
Datron, line 43: Can we merge both of these checks?

crates/context_aware_config/src/api/dimension/types.rs
ShubhranshuSanjeev: Where exactly @ayushjain17 ?

ShubhranshuSanjeev (changes requested):

pratikmishra356 (approved):

ShubhranshuSanjeev (approved):

PR #145, opened 2024-07-02
fix: Fix context parsing

mahatoankitkumar:
## Problem
For number in context, it would consider it as an integer as whole, even if it was an array

## Solution
Fixed the parsing for number

### DIFF
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -9,7 +9,7 @@ use cfg_if::cfg_if;
 use leptos::*;
 use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use serde::de::DeserializeOwned;
-use serde_json::{Number, Value};
+use serde_json::Value;
 use std::str::FromStr;
 use url::Url;
 use wasm_bindgen::JsCast;
@@ -274,6 +274,7 @@ pub enum ConfigType {
     Dimension(Dimension),
 }
 
+#[derive(Clone, Debug)]
 pub enum ConfigValueType {
     Boolean,
     Number,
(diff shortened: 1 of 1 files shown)


sauraww (approved):

crates/frontend/src/utils.rs
ShubhranshuSanjeev, line 1: This could be the case for other types also, like string.

ShubhranshuSanjeev (changes requested):

crates/frontend/src/utils.rs
```diff
@@ -338,21 +338,22 @@ pub fn parse_value(val: &str, config_type: ConfigValueType) -> Result<Value, Str
...
+                Value::Number(num) => Ok(Value::Number(num)),
+                Value::Array(arr) => {
+                    for item in &arr {
+                        if !item.is_number() {
+                            return Err("Array contains non-number value".to_string());
+                        }
+                    }
+                    Ok(Value::Array(arr))
+                }
+                _ => Err("Invalid number or number array format".to_string()),
```
Datron, line 354: Can we also include the value that failed?

crates/frontend/src/utils.rs
```diff
@@ -338,21 +338,22 @@ pub fn parse_value(val: &str, config_type: ConfigValueType) -> Result<Value, Str
...
-            .map(|number| Value::Number(number.into()))
-            .or_else(|_| {
-                f64::from_str(val)
-                    .ok()
-                    .and_then(|num| Number::from_f64(num).map(Value::Number))
-                    .ok_or_else(|| {
-                        "Invalid decimal format or precision issue".to_string()
-                    })
-            }),
+        ConfigValueType::Number | ConfigValueType::Integer => {
```
Datron, line 341: We should treat these two differently, and parse them into their corresponding types. 
Integer - i32
Number - f32

Datron: Open to opinions on this

crates/frontend/src/utils.rs
```diff
@@ -338,21 +338,22 @@ pub fn parse_value(val: &str, config_type: ConfigValueType) -> Result<Value, Str
...
+                        if !item.is_number() {
+                            return Err("Array contains non-number value".to_string());
+                        }
+                    }
+                    Ok(Value::Array(arr))
+                }
+                _ => Err("Invalid number or number array format".to_string()),
+            }
+        }
         ConfigValueType::String => Ok(Value::String(val.to_string())),
```
pratikmishra356, line 357: array check should be for string also

ShubhranshuSanjeev (approved):

pratikmishra356 (approved):

PR #146, opened 2024-07-02
feat: added newtype for condition and override in context apis

pratikmishra356:
## Problem
Request Data Validation was not present in context apis

## Solution
Created newtype for condition and override and added validation there

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes

NA


Datron:
@pratikmishra356 Can we add tests to check these validations?

crates/service_utils/src/result.rs
```diff
@@ -50,6 +50,28 @@ impl AppError {
...
             .json(response)
     }
+    pub fn message(&self) -> String {
+        use diesel::result::DatabaseErrorKind as diesel_error_kind;
+        use diesel::result::Error as diesel_error;
+        match self {
+            AppError::ValidationError(msg)
+            | AppError::BadArgument(msg)
+            | AppError::NotFound(msg) => msg.to_owned(),
+            AppError::UnexpectedError(_) => String::from("Something went wrong"),
```
Datron, line 60: Lets stop doing "Something went wrong". Need a better error message here

Datron: Also log and throw errors if appropriate

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -1,11 +1,61 @@
...
+        Condition::new(s).map_err(|err| err.message())
+    }
+}
+
+#[derive(Deserialize, Serialize, Clone, AsRef, Deref, Debug, Eq, PartialEq)]
+#[serde(try_from = "Map<String,Value>")]
+pub struct Override(Map<String, Value>);
+
+impl Override {
+    pub fn new(override_map: Map<String, Value>) -> superposition::Result<Self> {
```
Datron, line 37: Should we validate the schema of the override value as well using json schema?

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -1,11 +1,61 @@
...
+use service_utils::{bad_argument, result as superposition};
+
+#[derive(Deserialize, Clone, AsRef, Deref, Debug, Eq, PartialEq)]
+#[serde(try_from = "Map<String,Value>")]
+pub struct Condition(Map<String, Value>);
+
+impl Condition {
+    pub fn new(condition_map: Map<String, Value>) -> superposition::Result<Self> {
+        if !condition_map.is_empty() {
+            jsonlogic::apply(&json!(condition_map), &json!({})).map_err(|msg| {
```
Datron, line 13: We can use jsonlogic::compile here if we want to check validity instead of applying it

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -17,16 +17,16 @@ use diesel::{
...
 type DBConnection = PooledConnection<ConnectionManager<PgConnection>>;
 
 pub fn validate_condition_with_functions(
     conn: &mut DBConnection,
-    context: &Value,
+    condition: &Condition,
 ) -> superposition::Result<()> {
     use dimensions::dsl;
-    let context = extract_dimensions(context)?;
+    let context = extract_dimensions(&json!(**condition))?;
```
Datron, line 29: derive_more has an auto deref trait. Can we use that to avoid our code looking like C++?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -190,8 +190,9 @@ fn create_ctx_from_put_req(
     conn: &mut DBConnection,
     user: &User,
 ) -> superposition::Result<Context> {
-    let ctx_condition = Value::Object(req.context.to_owned());
-    let ctx_override: Value = req.r#override.to_owned().into();
+    let ctx_condition = req.context.to_owned();
+    let condition_val = json!(*ctx_condition);
```
Datron, line 194: Use derive_more Deref to avoid this.

crates/context_aware_config/src/api/context/helpers.rs
Datron: Can we also add a to_json() method on newtypes? Skip one more step here

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -297,23 +298,27 @@ fn get_contextids_from_overrideid(
...
 
-fn construct_new_payload(req_payload: &Map<String, Value>) -> web::Json<PutReq> {
+fn construct_new_payload(
+    req_payload: &Map<String, Value>,
+) -> superposition::Result<web::Json<PutReq>> {
     let mut res = req_payload.clone();
     res.remove("to_be_deleted");
     res.remove("override_id");
     res.remove("id");
     if let Some(Value::Object(res_context)) = res.get("context") {
```
Datron, line 308: Can we do negative checks first to avoid this nesting?
```suggestion
    if res.get("context").is_none() {
```
Throw error if true, process if false

Datron (changes requested):
Are all validations for context and overrides included here or more expected in another PR? Can we add tests to check the negative cases for validations so we can guarantee that they work in the future too?

crates/service_utils/src/result.rs
pratikmishra356: @Datron The string returned from message function should be same as what string returned using error macro.
For logging , wherever we pass error in these macro like bad_argument!(err) -> we are logging error there itself. Like you can check new function.
This function just helps in getting the message which Apperror sends , so that we can pass the same message into other error types like here serde deserialisation error

crates/superposition_types/src/lib.rs
```diff
@@ -74,3 +77,276 @@ impl FromRequest for User {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq)]
```
ayushjain17, line 81: can we use something like this:
```suggestion
#[derive(Debug, Clone, PartialEq)]
pub enum ValidationType {
    DEFAULT,
    EXPERIMENTAL,
    #[cfg(feature = "db_type_validation_disabled")]
    DB,
}

pub fn get_db_validation_type() -> ValidationType {
    #[cfg(feature = "db_type_validation_disabled")]
    return ValidationType::DB;
    #[cfg(not(feature = "db_type_validation_disabled"))]
    return ValidationType::DEFAULT;
}
```

we can utilise rust features here, and anyone who has old version of code can go ahead with `db_type_validation_disabled` feature (or whatever you want to call it) as newer versions will not have discrepancy in DB data

wherever you want to use ValidationType::DB, just use this function `get_db_validation_type` directly, which will resolve the value based on the feature which the user of superposition requires

later on, when we are sure that data is not longer corrupt in DB, we can get the feature removed altogether

ayushjain17: places having a switch case can use something like:
```
            #[cfg(feature = "db_type_validation_disabled")]
            ValidationType::DB => (),
```

crates/context_aware_config/src/api/config/types.rs
```diff
@@ -1,17 +1,20 @@
...
     pub default_configs: Map<String, Value>,
 }
 
 #[derive(Serialize, Clone, Deserialize)]
 pub struct Context {
     pub id: String,
-    pub condition: Value,
+    pub condition: Condition,
     pub priority: i32,
     pub override_with_keys: [String; 1],
```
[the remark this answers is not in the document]
pratikmishra356: @mahatoankitkumar `[String; 1]` seems enough for this validation and scope of this pr is only for context and overrides

crates/context_aware_config/src/helpers.rs
```diff
@@ -323,21 +327,27 @@ pub fn generate_cac(
...
-            };
-            ctxts.push(ctxt);
-            overrides.insert(override_id, override_);
-            (ctxts, overrides)
-        },
-    );
+    let mut contexts = Vec::new();
+    let mut overrides: HashMap<String, Overrides> = HashMap::new();
+
+    for (id, condition, priority_, override_id, override_) in contexts_vec.iter() {
```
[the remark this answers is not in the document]
pratikmishra356: can be used , but it gives no advantage over for loop and readability reduces as multiple variables are being used here

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -399,7 +401,7 @@ fn r#move(
 ) -> superposition::Result<PutResp> {
     use contexts::dsl;
     let req = req.into_inner();
-    let ctx_condition = Value::Object(req.context);
+    let ctx_condition = json!(req.context.into_inner());
```
ayushjain17, line 404: can we let it be Value::Object itself
```suggestion
    let ctx_condition = Value::Object(req.context.into_inner());
```

ayushjain17 (commented):
Can we avoid json! macro and specify Value::Object directly and keep parameter or variables name same as condition for Condition types to avoid inconsistency

crates/superposition_types/src/lib.rs
```diff
@@ -74,3 +77,307 @@ impl FromRequest for User {
         }
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Copy, Serialize)]
+pub struct Cac<T>(T);
```
Datron, line 82: Can we implement deref trait instead of `into_inner`?

Datron (approved):
few nitpicks but overall looks good

crates/superposition_types/src/lib.rs
pratikmishra356: @Datron into_inner gives the type wrapped inside i.e T, 
Deref trait mainly dereference the the value , 
and also Deref wont work in this case as Deref on CAC requires copy trait for Condition , and we cant have Copy trait for condition as it wraps serde Map which does not have the Copy trait either 

```
cannot move out of dereference of `Cac<superposition_types::Condition>`
move occurs because value has type `superposition_types::Condition`, which does not implement the `Copy` trait
```

ayushjain17 (approved):

mahatoankitkumar (approved):

PR #157, opened 2024-07-12
fix: Improve form inputs

mahatoankitkumar:
## Problem
Better form fields
- enum → Toggle
- boolean → dropdown with true and false options

https://github.com/user-attachments/assets/5152c2fb-316d-4617-86aa-09d90460ec8c

### DIFF
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -2,7 +2,7 @@ pub mod types;
 pub mod utils;
 
 use leptos::*;
-use serde_json::{json, Value};
+use serde_json::{json, Map, Value};
 use std::str::FromStr;
 use web_sys::MouseEvent;
 
@@ -11,8 +11,10 @@ use crate::{
     components::{
         button::Button,
         dropdown::{Dropdown, DropdownBtnType, DropdownDirection},
+        input_components::{BooleanToggle, EnumDropdown},
     },
     types::{FunctionsName, TypeTemplate},
+    utils::get_key_type,
 };
 
 use self::{types::DefaultConfigCreateReq, utils::create_default_config};
@@ -183,8 +185,8 @@ where
 
                             <textarea
                                 type="text"
-                                placeholder="Enter a JSON schema"
-                                class="input input-bordered mt-5 rounded-md resize-y w-full max-w-md"
+                                placeholder="JSON schema"
+                                class="input input-bordered mt-5 rounded-md resize-y w-full max-w-md pt-3"
                                 rows=8
                                 on:change=move |ev| {
                                     config_schema_ws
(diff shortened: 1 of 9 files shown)


Datron (approved):
Lets get this in

pratikmishra356 (approved):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -203,52 +205,82 @@ where
...
+                    }
+                    "NUMBER" | "INTEGER" => {
+                        view! {
+                            <input
+                                type="number"
+                                placeholder="Value"
+                                class="input input-bordered w-full max-w-md"
+                                value=config_value.get()
+                                on:change=move |ev| {
+                                    logging::log!("{:?}", event_target_value(& ev));
```
sauraww, line 250: Can you add the desciption for the log here ?

crates/frontend/src/components/default_config_form.rs
```diff
@@ -203,52 +205,82 @@ where
...
-                                </label>
-                                {input_format}
-                            </div>
-                            <div class="divider"></div>
+                            <textarea
+                                type="text"
+                                placeholder="Value"
+                                class="input input-bordered w-full max-w-md pt-3"
+                                on:change=move |ev| {
+                                    logging::log!("{:?}", event_target_value(& ev));
```
sauraww, line 264: Here as well

sauraww (changes requested):

crates/frontend/src/components/input_components.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[component]
+pub fn enum_dropdown(
+    schema: Map<String, Value>,
+    config_value: String,
+    handle_change: Callback<String, ()>,
+    #[prop(default = String::new())] class: String,
+    #[prop(default = false)] disabled: bool,
+) -> impl IntoView {
+    let (value, set_value) = create_signal(config_value.replace("\\", ""));
```
sauraww, line 14: Can we avoid replacing it ? 
What if someone actually wants to send this in the input ?

sauraww (changes requested):

crates/frontend/src/components/dimension_form.rs
```diff
@@ -209,8 +209,9 @@ where
                                     ev.prevent_default();
                                 }
                             }
+
                             on:change=move |ev| {
-                                logging::log!("{:?}", event_target_value(&ev).parse::<u32>());
+                                logging::log!("{:?}", event_target_value(& ev).parse::< u32 > ());
```
sauraww, line 214: Can you add the log description here as well ?

sauraww (changes requested):

crates/frontend/src/utils.rs
```diff
@@ -516,3 +516,14 @@ pub fn set_local_storage(_key: &str, _value: &str) -> Option<()> {
         }
     }
 }
+
+pub fn get_key_type(schema: &Map<String, Value>) -> String {
+    if schema.contains_key("enum") {
+        String::from("ENUM")
+    } else {
+        match schema.get("type").unwrap_or(&Value::Null) {
```
sauraww, line 524: Why are we using unwrap and match both here ? 
Possible to change it to : 
`match schema.get("type") {
    Some(Value::String(str_)) => str_.to_ascii_uppercase(),
    _ => String::from("STRING"),
}
`

sauraww (dismissed):

PR #173, opened 2024-07-16
feat: Mandatory dimensions feature

mahatoankitkumar:
## Problem
Mandatory Dimensions Feature

## Env
`MANDATORY_DIMENSIONS="dev:clientId,fare;test:fare"`

https://github.com/user-attachments/assets/44cbe739-ebaa-466f-b6b4-e34a1b0805c2

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,3 +24,5 @@ TENANTS=dev,test
 TENANT_MIDDLEWARE_EXCLUSION_LIST="/health,/assets/favicon.ico,/pkg/frontend.js,/pkg,/pkg/frontend_bg.wasm,/pkg/tailwind.css,/pkg/style.css,/assets,/admin,/"
 SERVICE_PREFIX=""
 SERVICE_NAME="CAC"
+MANDATORY_DIMENSIONS=""
+# MANDATORY_DIMENSIONS="dev:clientId,fare;test:fare"
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -22,7 +22,7 @@ use crate::{
     },
 };
 use actix_web::web::Data;
-use service_utils::service::types::{AppHeader, AppState, CustomHeaders};
+use service_utils::service::types::{AppHeader, AppState, CustomHeaders, Tenant};
 
 use actix_web::{
     delete, get, post, put,
@@ -45,7 +45,8 @@ use std::collections::HashMap;
 use superposition_types::{SuperpositionUser, User};
 
 use super::helpers::{
-    validate_condition_with_functions, validate_override_with_functions,
+    validate_condition_with_functions, validate_condition_with_mandatory_dimensions,
+    validate_override_with_functions,
 };
 
 use superposition_macros::{
@@ -192,11 +193,18 @@ fn create_ctx_from_put_req(
     req: Json<PutReq>,
     conn: &mut DBConnection,
     user: &User,
+    state: &Data<AppState>,
+    tenant: &Tenant,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
     let condition_val = json!(ctx_condition);
     let r_override = req.r#override.clone().into_inner();
     let ctx_override = json!(r_override.to_owned());
+    validate_condition_with_mandatory_dimensions(
+        &ctx_condition,
+        &state.mandatory_dimensions,
+        tenant,
+    )?;
     validate_override_with_default_configs(conn, &r_override)?;
     validate_condition_with_functions(conn, &ctx_condition)?;
     validate_override_with_functions(conn, &r_override)?;
@@ -299,9 +307,11 @@ pub fn put(
     conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
     already_under_txn: bool,
     user: &User,
+    state: &Data<AppState>,
+    tenant: &Tenant,
 ) -> superposition::Result<PutResp> {
     use contexts::dsl::contexts;
-    let new_ctx = create_ctx_from_put_req(req, conn, user)?;
+    let new_ctx = create_ctx_from_put_req(req, conn, user, &state, tenant)?;
 
     if already_under_txn {
         diesel::sql_query("SAVEPOINT put_ctx_savepoint").execute(conn)?;
@@ -330,15 +340,15 @@ async fn put_handler(
     req: Json<PutReq>,
     mut db_conn: DbConnection,
     user: User,
+    tenant: Tenant,
 ) -> superposition::Result<HttpResponse> {
     let tags = parse_config_tags(custom_headers.config_tags)?;
     db_conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
-        let put_response = put(req, transaction_conn, true, &user).map_err(
-            |err: superposition::AppError| {
+        let put_response = put(req, transaction_conn, true, &user, &state, &tenant)
+            .map_err(|err: superposition::AppError| {
                 log::info!("context put failed with error: {:?}", err);
                 err
-            },
-        )?;
+            })?;
         let version_id = add_config_version(&state, tags, transaction_conn)?;
         let mut http_resp = HttpResponse::Ok();
 
@@ -355,9 +365,11 @@ fn override_helper(
     conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
     already_under_txn: bool,
     user: &User,
+    state: &Data<AppState>,
+    tenant: Tenant,
 ) -> superposition::Result<PutResp> {
     use contexts::dsl::contexts;
-    let new_ctx = create_ctx_from_put_req(req, conn, user)?;
+    let new_ctx = create_ctx_from_put_req(req, conn, user, state, &tenant)?;
     if already_under_txn {
         diesel::sql_query("SAVEPOINT insert_ctx_savepoint").execute(conn)?;
     }
@@ -385,15 +397,17 @@ async fn update_override_handler(
     req: Json<PutReq>,
     mut db_conn: DbConnection,
     user: User,
+    tenant: Tenant,
 ) -> superposition::Result<HttpResponse> {
     let tags = parse_config_tags(custom_headers.config_tags)?;
     db_conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
-        let override_resp = override_helper(req, transaction_conn, true, &user).map_err(
-            |err: superposition::AppError| {
-                log::info!("context put failed with error: {:?}", err);
-                err
-            },
-        )?;
+        let override_resp =
+            override_helper(req, transaction_conn, true, &user, &state, tenant).map_err(
+                |err: superposition::AppError| {
+                    log::info!("context put failed with error: {:?}", err);
+                    err
+                },
+            )?;
         let version_id = add_config_version(&state, tags, transaction_conn)?;
         let mut http_resp = HttpResponse::Ok();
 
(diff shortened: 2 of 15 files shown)


crates/superposition/src/main.rs
```diff
@@ -83,6 +84,25 @@ async fn main() -> Result<()> {
...
+    let mandatory_dimensions: Map<String, Value> =
+        get_from_env_unsafe::<String>("MANDATORY_DIMENSIONS")
+            .expect("MANDATORY_DIMENSIONS is not set")
+            .split(';')
+            .filter_map(|ele| {
+                let arr: Vec<&str> = ele.split(':').collect();
+                if arr.len() == 2 {
+                    let key = arr[0].to_string();
+                    let values =
+                        arr[1].split(',').map(String::from).collect::<Vec<String>>();
```
ayushjain17, line 96: since you need to map again on the values after split, you don't need to collect here, you can directly map it here as well
```suggestion
                        arr[1].split(',').map(String::from).map(Value::String).collect();
```

crates/superposition/src/main.rs
```diff
@@ -83,6 +84,25 @@ async fn main() -> Result<()> {
...
+                    Some((
+                        key,
+                        Value::Array(values.into_iter().map(Value::String).collect()),
```
ayushjain17, line 99: this would then simply become
```suggestion
                        Value::Array(values.collect()),
```

crates/frontend/src/pages/context_override.rs
```diff
@@ -151,7 +151,25 @@ pub fn context_override() -> impl IntoView {
 
     let handle_context_create = Callback::new(move |_| {
         set_form_mode.set(Some(FormMode::Create));
-        set_selected_data.set(None);
+        let PageResource {
```
ayushjain17, line 154: you can write it like this as well:
```suggestion
        let PageResource { dimensions, .. } = page_resource.get().unwrap_or_default();
```

you wont need to write the other fields and then ignore them explicitly by using underscore

crates/frontend/src/pages/context_override.rs
```diff
@@ -151,7 +151,25 @@ pub fn context_override() -> impl IntoView {
...
+            config: _,
+            dimensions,
+            default_config: _,
+        } = page_resource.get().unwrap_or_default();
+        let context_with_mandatory_dimensions = dimensions
+            .into_iter()
+            .filter_map(|dim| {
+                if dim.mandatory {
+                    Some((dim.dimension, String::from(""), String::from("")))
```
ayushjain17, line 163: if the 2nd and 3rd argument of the tuple are always empty string, do we at all need these 2 params of the tuple ?

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -8,6 +8,7 @@ pub struct CreateReq {
     pub schema: Value,
     #[serde(default, deserialize_with = "deserialize_option")]
     pub function_name: Option<Value>,
+    pub mandatory: Option<bool>,
```
ayushjain17, line 11: this should not be there right ?
users wont be able to change if a dimension is mandatory or not, right ?

as the data is coming from ENV at the moment

crates/frontend/src/pages/context_override.rs
ayushjain17, line 154: you can write it like this as well:
```suggestion
        let PageResource { dimensions, .. } = page_resource.get().unwrap_or_default();
```

you wont need to write the other fields and then ignore them explicitly by using underscore

crates/context_aware_config/src/api/dimension/helpers.rs
```diff
@@ -0,0 +1,29 @@
...
+use service_utils::service::types::Tenant;
+
+pub fn get_mandatory_dimensions(
+    mandatory_dimensions: &Map<String, Value>,
+    tenant: Tenant,
+) -> Vec<String> {
+    let result: Vec<String> = mandatory_dimensions
+        .get(&tenant.0.to_string())
+        .and_then(|v| v.as_array())
+        .unwrap_or(&vec![])
```
ayushjain17, line 11: shouldn't this be better
```suggestion
        .map_or_else(|| vec![], |arr| {
            arr.iter().filter_map(|value| value.as_str()).map(String::from).collect()
        |)
```

ayushjain17 (changes requested):
Not a complete review, still reviewing

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -101,9 +122,31 @@ async fn create(
...
+                .and_then(|value| value.as_object().cloned())
+        })
+        .collect();
+
+    let mandatory_dimensions =
+        get_mandatory_dimensions(&state.mandatory_dimensions, tenant);
+
+    let new_map: Vec<Map<String, Value>> = dimensions_vec
+        .into_iter()
+        .map(|ele| insert_mandatory_dimension(ele, mandatory_dimensions.clone()))
```
ayushjain17, line 148: instead of doing all these conversions (in the newly code) to insert the field `mandatory`
it be better if we create a separate type for Dimension which includes the field mandatory as well, that way the existence of the field `mandatory` gets documented as well along with being able to make this code cleaner by implementing a From trait for the newly added type from the already existing Dimension type

something like:
```suggestion
        let mandatory_dimensions =
               get_mandatory_dimensions(&state.mandatory_dimensions, tenant);
        result.into_iter().map(DimensionWithMandatory::from).map(|d| {
              DimensionWithMandatory { mandatory: mandatory_dimensions.contains(d.dimension), ..d}
        })
        .collect();
```

the from trait can be something like:
```
impl From<Dimension> for DimensionWithMandatory {
    fn from(value: Dimension) -> Self {
        DimensionWithMandatory {
            mandatory: false,
            ..value.into()
        }
    }
}
```

this setup seems cleaner than what we have right now, with all the conversions and getter around Value type

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -80,14 +83,32 @@ async fn create(
         .get_result::<Dimension>(&mut conn);
 
     match upsert {
-        Ok(upserted_dimension) => Ok(HttpResponse::Created().json(upserted_dimension)),
+        Ok(upserted_dimension) => {
+            let dimension_map = serde_json::to_value(&upserted_dimension)
```
ayushjain17, line 87: here as well
(same as below comment)

crates/context_aware_config/src/api/dimension/helpers.rs
```diff
@@ -0,0 +1,29 @@
+use serde_json::{Map, Value};
+use service_utils::service::types::Tenant;
+
+pub fn get_mandatory_dimensions(
```
ayushjain17, line 4: can we rather move this functionality to the Tenant enum itself

```
impl Tenant {
     pub get_mandatory_dimensions (&self, mandatory_dimensions: &Map<String, Value>) -> Vec<String> {
          mandatory_dimensions
               .get(self.0.to_string())
               .and_then(|v| v.as_array())
               .map_or_else(|| vec![], |arr| {
                     arr.iter().filter_map(|value| value.as_str()).map(String::from).collect()
               |)
     }
}
```

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -18,10 +19,35 @@ use diesel::{
...
 
+pub fn validate_condition_with_mandatory_dimensions(
+    context: &Condition,
+    mandatory_dimensions_map: &Map<String, Value>,
+    tenant: Tenant,
+) -> superposition::Result<()> {
+    let context_map = extract_dimensions(context)?;
+    let dimensions_list: Vec<String> = context_map.keys().cloned().collect();
+    let binding = json!([]);
+    let mandatory_dimensions = mandatory_dimensions_map
```
ayushjain17, line 34: can we not reuse the function `get_mandatory_dimensions ` here?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -646,16 +661,18 @@ async fn bulk_operations(
...
         for action in reqs.into_inner().into_iter() {
+            let tenant = tenant.clone();
             match action {
                 ContextAction::Put(put_req) => {
-                    let put_resp = put(Json(put_req), transaction_conn, true, &user)
-                        .map_err(|err| {
-                            log::error!(
-                                "Failed at insert into contexts due to {:?}",
+                    let put_resp =
+                        put(Json(put_req), transaction_conn, true, &user, &state, tenant)
```
ayushjain17, line 668: we can rather accept a reference of the tenant in the `put` function, then you can avoid cloning tenant

ayushjain17 (changes requested):
Completed review

crates/frontend/src/pages/context_override.rs
mahatoankitkumar: Yes because its vec<(String,String, String)>, dimension, operator and value.

crates/context_aware_config/src/api/dimension/types.rs
mahatoankitkumar: forgot to remove

pratikmishra356:
@mahatoankitkumar we could have changed the Tenant structure itself 
```
pub struct Tenant {
   name: 
   mandatory_dimensions : 
}
```
and create this object on every request

mahatoankitkumar:
> @mahatoankitkumar we could have changed the Tenant structure itself
> 
> ```
> pub struct Tenant {
>    name: 
>    mandatory_dimensions : 
> }
> ```
> 
> and create this object on every request
@pratikmishra356 Lets not change the structure of Tenant right now, as this implementation will be discarded afterwards.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -192,11 +193,18 @@ fn create_ctx_from_put_req(
...
     conn: &mut DBConnection,
     user: &User,
+    state: &Data<AppState>,
+    tenant: Tenant,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
     let condition_val = json!(ctx_condition);
     let r_override = req.r#override.clone().into_inner();
     let ctx_override = json!(r_override.to_owned());
+    validate_condition_with_mandatory_dimensions(
```
pratikmishra356, line 203: reviewing this logic

pratikmishra356 (changes requested):

pratikmishra356:
@mahatoankitkumar I see no checks for move api , we should have validation on move api also right ?

mahatoankitkumar:
> @mahatoankitkumar I see no checks for move api , we should have validation on move api also right ?

yeah right.

crates/superposition/src/main.rs
```diff
@@ -83,6 +84,25 @@ async fn main() -> Result<()> {
             .split(',')
             .map(String::from)
             .collect::<HashSet<String>>();
+    let mandatory_dimensions: Map<String, Value> =
+        get_from_env_unsafe::<String>("MANDATORY_DIMENSIONS")
+            .expect("MANDATORY_DIMENSIONS is not set")
+            .split(';')
```
pratikmishra356, line 90: better to create this part a function and write tests for it since we are doing 3 times split with using array index

mahatoankitkumar: This solution is not permanent, will get removed afterwards when tenant config gets in.
I think this can go right now.

Datron (approved):

.env.example
```diff
@@ -23,4 +23,5 @@ ENABLE_TENANT_AND_SCOPE=true
 TENANTS=dev,test
 TENANT_MIDDLEWARE_EXCLUSION_LIST="/health,/assets/favicon.ico,/pkg/frontend.js,/pkg,/pkg/frontend_bg.wasm,/pkg/tailwind.css,/pkg/style.css,/assets,/admin,/"
 SERVICE_PREFIX=""
-SERVICE_NAME="CAC"
\ No newline at end of file
+SERVICE_NAME="CAC"
+MANDATORY_DIMENSIONS=""
```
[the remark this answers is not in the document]
mahatoankitkumar: `MANDATORY_DIMENSIONS="dev:clientId,fare;test:fare"`

crates/superposition/src/main.rs
```diff
@@ -83,6 +84,25 @@ async fn main() -> Result<()> {
...
+            .filter_map(|ele| {
+                let arr: Vec<&str> = ele.split(':').collect();
+                if arr.len() == 2 {
+                    let key = arr[0].to_string();
+                    let values =
+                        arr[1].split(',').map(String::from).collect::<Vec<String>>();
+                    Some((
+                        key,
```
pratikmishra356, line 98: can we trim the key value here ?

pratikmishra356 (approved):

ayushjain17 (approved):

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -646,16 +661,21 @@ async fn bulk_operations(
     let mut response = Vec::<ContextBulkResponse>::new();
     conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
         for action in reqs.into_inner().into_iter() {
+            let tenant = tenant.clone();
```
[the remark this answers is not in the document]
mahatoankitkumar: Its already removed you published after the update.

PR #174, opened 2024-07-16
fix: remove process, filter env variables

pratikmishra356:
## Problem
First process was allowed in js_function sandbox environment
we need to enable env variables that is required for js_functions only

## Solution
Remove process from vm , delete all env variable keys except keys mentioned under `FUNCTION_ENV_VARIABLES`

NA

## Pre-deployment activity
Add FUNCTION_ENV_VARIABLES and mention all env that is required for js_function

## Post-deployment activity
NA

## API changes

NA

### DIFF
--- a/crates/context_aware_config/src/validation_functions.rs
+++ b/crates/context_aware_config/src/validation_functions.rs
@@ -4,6 +4,9 @@ use std::str;
 use superposition_macros::{unexpected_error, validation_error};
 use superposition_types::result as superposition;
 
+static FUNCTION_ENV_VARIABLES: &str =
+    "HTTP_PROXY,HTTPS_PROXY,HTTP_PROXY_HOST,HTTP_PROXY_PORT,NO_PROXY";
+
 fn type_check_validate(code_str: &str) -> String {
     format!(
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -23,4 +23,5 @@ ENABLE_TENANT_AND_SCOPE=true
 TENANTS=dev,test
 TENANT_MIDDLEWARE_EXCLUSION_LIST="/health,/assets/favicon.ico,/pkg/frontend.js,/pkg,/pkg/frontend_bg.wasm,/pkg/tailwind.css,/pkg/style.css,/assets,/admin,/"
 SERVICE_PREFIX=""
-SERVICE_NAME="CAC"
\ No newline at end of file
+SERVICE_NAME="CAC"
+FUNCTION_ENV_VARIABLES="SERVICE_NAME"
```
Datron, line 27: This should not be an ENV. Let us hard code it in file, we have too many ENVs today

Datron (changes requested):

Datron (approved):

mahatoankitkumar (approved):

PR #180, opened 2024-07-23
fix: resolve page issue

mahatoankitkumar:
## Problem

- Resolve API didn't work for boolean values because of parsing query params
- Resolve page enum and toggle didn't work
- Resolve page un-strike of the resolved config didn't work

## Solution
Solved the above issues

https://github.com/user-attachments/assets/34a0f36c-3112-48d5-925c-f9fdfced28e3

### DIFF
--- a/crates/frontend/src/pages/home.rs
+++ b/crates/frontend/src/pages/home.rs
@@ -15,9 +15,7 @@ use serde_json::{Map, Value};
 use strum::EnumProperty;
 use strum_macros::Display;
 use wasm_bindgen::JsCast;
-use web_sys::{
-    HtmlButtonElement, HtmlInputElement, HtmlSelectElement, HtmlSpanElement, MouseEvent,
-};
+use web_sys::{HtmlButtonElement, HtmlSpanElement, MouseEvent};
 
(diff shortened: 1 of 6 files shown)


crates/frontend/src/pages/home.rs
```diff
@@ -66,6 +66,7 @@ pub fn home() -> impl IntoView {
         },
     );
 
+    let (context, set_context) = create_signal::<Vec<(String, String, String)>>(vec![]);
```
Datron, line 69:
```suggestion
    let (context_rs, context_ws) = create_signal::<Vec<(String, String, String)>>(vec![]);
```

Datron (approved):

crates/frontend/src/pages/home.rs
ShubhranshuSanjeev: @Datron , leptos follows context and set_context convention in their docs, lets stick with that.

Datron: @ShubhranshuSanjeev that nomenclature doesn't help identify the type of instrument you are working with. We should self document and differentiate between signals, resources, etc.

pratikmishra356 (approved):

PR #184, opened 2024-07-29
feat: monaco as a component

Datron:
## Problem
General text area inputs are not good enough for long JSON objects like the type schemas supported by superposition. We need a better input box, something like monaco

## Solution
Add monaco editor as a component to superposition, making it reusable and configurable. Also monaco-rust uses wasm bindgen and the component is written in rust. This PR also adds brings in better JS syntax help and  completion to the functions UI.

TODO: Support for JSON schema checks will be added in a separate PR. This is the foundation for that

Screenshot 2024-07-29 at 3 33 55 PM

https://github.com/user-attachments/assets/f6db1943-4467-4a7a-ad71-c4527c9d2794

https://github.com/user-attachments/assets/6ae7ff97-e174-4bd5-9d3f-c37334eb0a50


crates/frontend/src/components/monaco_editor.rs
```diff
@@ -0,0 +1,58 @@
...
+use monaco::api::CodeEditor;
+use monaco::sys::editor::{IEditorMinimapOptions, IStandaloneEditorConstructionOptions};
+#[derive(Debug, Clone, strum_macros::Display)]
+#[strum(serialize_all = "lowercase")]
+pub enum Languages {
+    Javascript,
+    Json,
+}
+
+pub type EditorModelCell = Rc<RefCell<Option<CodeEditor>>>;
```
pratikmishra356, line 13: @Datron any specific reason to use Rc<RefCell>. , we can use just Option<CodeEditor> also

pratikmishra356 (approved):

ShubhranshuSanjeev (approved):

PR #188, opened 2024-07-30
fix(cac_client): Run `fixDarwinDylibNames` during `postInstall`

shivaraj-bh:
## Problem

`.dylib` files contain information about their location. When the rust library is built using Nix, the build happens in a temporary build directory, which is removed after the build and once a corresponding `/nix/store/…` path is created. In the metadata of the `.dylib` files is the path to the temporary build directory, which needs to be patched.

## Solution

This PR ensures the `.dylib` files are always searched for in the respective `/nix/store/...` path, instead of the temporary build directory.

### DIFF
--- a/nix/rust.nix
+++ b/nix/rust.nix
@@ -20,12 +20,19 @@
             crane = {
               args = {
                 buildInputs = lib.optionals isDarwin
-                  (with apple_sdk.frameworks; [
-                    Security
+                  ([
+                    apple_sdk.frameworks.Security
+                    pkgs.fixDarwinDylibNames
                   ]) ++ [
                   pkgs.postgresql_12
                   pkgs.openssl
                 ];
+              };
+              extraBuildArgs = {
+                # https://discourse.nixos.org/t/how-to-use-install-name-tool-on-darwin/9931/2
+                postInstall = ''
+                  ${if isDarwin then "fixDarwinDylibNames" else ""}
+                '';
               };
             };
           };


srid:
Related: does `cabal repl` work? Is this up to date?

https://github.com/juspay/superposition/blob/73b5dbcb4b77c9c6234d1ac28be47798e84be1f1/clients/haskell/default.nix#L14-L16

srid (commented):
> `.dylib` search path is set to temporary build directory

Could you describe the problem in detail? `nix build` works in CI,[^1] so what problem does it fix for the developer?

[^1]: https://github.com/juspay/superposition/actions/runs/10092003297/job/27904741482

shivaraj-bh:
> Could you describe the problem in detail?

Updated the description

> nix build works in CI,[1](https://github.com/juspay/superposition/pull/188#user-content-fn-1-620ae21cbadad1bba087656328e5ac5d) so what problem does it fix for the developer?

For the Haskell clients in this repository, the rust library’s `.dylib` output is a runtime dependency, whereas for the downstream projects that use these Haskell clients, it is a build time dependency.

If you try to `nix run .#hs-cac-client` on the current `main`, it should fail with:
```sh
dyld[29980]: Library not loaded: /private/tmp/nix-build-cac_client-0.14.0.drv-0/source/target/release/deps/libcac_client.dylib
  Referenced from: <5BE2F4DC-80E3-3116-9B16-728C33666BAF> /nix/store/an93yxkmpmlc3g60gg9x259gxikfxybk-hs-cac-client-0.1.0.0/bin/hs-exe
  Reason: tried: '/private/tmp/nix-build-cac_client-0.14.0.drv-0/source/target/release/deps/libcac_client.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/private/tmp/nix-build-cac_client-0.14.0.drv-0/source/target/release/deps/libcac_client.dylib' (no such file), '/private/tmp/nix-build-cac_client-0.14.0.drv-0/source/target/release/deps/libcac_client.dylib' (no such file), '/usr/local/lib/libcac_client.dylib' (no such file), '/usr/lib/libcac_client.dylib' (no such file, not in dyld cache)
```

shivaraj-bh:
> Related: does `cabal repl` work? Is this up to date?
> 
> https://github.com/juspay/superposition/blob/73b5dbcb4b77c9c6234d1ac28be47798e84be1f1/clients/haskell/default.nix#L14-L16

Yes, works. On Linux and macOS

srid:
> If you try to `nix run .#hs-cac-client`

Let's add this to CI (in this PR) so as to prevent this from happening again.

image

Datron (approved):

ShubhranshuSanjeev (approved):

PR #192, opened 2024-07-31
Add last modified for audit log

pratikmishra356:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/migrations/2024-04-22-122806_config_verions/down.sql
+++ b/crates/context_aware_config/migrations/2024-04-22-122806_config_verions/down.sql
@@ -1,1 +1,3 @@
 -- This file should undo anything in `up.sql`
+DELETE FROM public.config_versions;
+DROP TABLE IF EXISTS public.config_versions;
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -223,6 +223,8 @@ fn create_ctx_from_put_req(
         override_: ctx_override.to_owned(),
         created_at: Utc::now(),
         created_by: user.get_email(),
+        last_modified_at: Utc::now().naive_utc(),
+        last_modified_by: user.get_email(),
     })
 }
 
@@ -234,6 +236,7 @@ pub fn hash(val: &Value) -> String {
 fn update_override_of_existing_ctx(
     conn: &mut PgConnection,
     ctx: Context,
+    user: &User,
 ) -> superposition::Result<PutResp> {
     use contexts::dsl;
     let mut new_override: Value = dsl::contexts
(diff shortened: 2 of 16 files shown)


crates/context_aware_config/migrations/2024-04-22-122806_config_verions/down.sql
```diff
@@ -1 +1,3 @@
 -- This file should undo anything in `up.sql`
+DELETE FROM public.config_versions;
+DROP TABLE public.config_versions;
```
Datron, line 3:
```suggestion
DROP TABLE IF EXISTS public.config_versions;
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -117,8 +131,10 @@ async fn create(
         value,
         schema,
         function_name,
-        created_by: user.get_email(),
-        created_at: Utc::now(),
+        created_by: created_by_val,
+        created_at: created_at_val,
+        last_modified_at: Utc::now().naive_utc(),
```
Datron, line 136:
```suggestion
        last_modified_at: created_at_val,
```
These two should be the same right? Lets not recompute the time here

Datron (changes requested):
Few small comments, can we quickly resolve them?

crates/context_aware_config/src/api/default_config/handlers.rs
pratikmishra356: no @Datron , created_at_val can be from already existing row, thats why recomputed the time

Datron: @pratikmishra356 this is inside create function, how can created_at_val previously exist? Either way, the mark is for this row right? created_at should be recomputed and the same should be used for last_modified_at

pratikmishra356: @Datron the handler name is create , but we have only one api for create or update/ the handler name is misguiding 
we create default_config struct from combination of req data and existing data , 
`let result = fetch_default_key(&key, &mut conn);`

Datron (approved):

pratikmishra356:
do not merge this without letting the team know , as it contains db changes it might block cac deployment

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -264,11 +273,16 @@ fn replace_override_of_existing_ctx(
         override_id: new_override_id,
         ..ctx
     };
-    diesel::update(dsl::contexts)
+    let update_resp = diesel::update(dsl::contexts)
```
mahatoankitkumar, line 276: Can we remove this code repetition?
same as line 251

mahatoankitkumar (changes requested):

mahatoankitkumar (approved):

PR #193, opened 2024-08-01
fix: Snapshot version parsing

ayushjain17:
## Problem
when someone passes `?version=7224713864877502464` as the snapshot version the code currently fails to parse this is as a valid i64 version value and only accepts `?version="7224713864877502464"` as a correct input

## Solution
Replace as_str conversion of version with as_i64

### DIFF
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -46,16 +46,13 @@ fn validate_version_in_params(
     query_params_map
         .remove("version")
         .map_or(Ok(None), |version| {
-            version
-                .as_str()
-                .and_then(|val| val.to_owned().parse::<i64>().ok())
-                .map_or_else(
-                    || {
-                        log::error!("failed to decode version as integer: {}", version);
-                        Err(bad_argument!("version is not of type integer"))
-                    },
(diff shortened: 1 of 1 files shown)


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -49,6 +49,7 @@ fn validate_version_in_params(
             version
                 .as_str()
```
ShubhranshuSanjeev, line 50: The reason for failure can be the use of `as_str` function returns `Some` value only when the JSON value is a `String`.

This can be resolved by doing `version.as_i64`

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
ayushjain17: yeah we want to support the user to send the data in both the formats, thats why didn't replace the previous set of lines with this

ShubhranshuSanjeev: "" are unnecessary in the query param value

ShubhranshuSanjeev (approved):

pratikmishra356 (approved):

PR #194, opened 2024-08-01
fix: haskell-client-doc

mahatoankitkumar:
## Problem
haskell client doc

## Solution
fixed haskell client doc

### DIFF
--- a/docs/client-context-aware-configuration.md
+++ b/docs/client-context-aware-configuration.md
@@ -295,8 +295,14 @@ Get the full config definition of your tenants configuration from superposition.
 ##### Funtion Definition
 
 ```
-getFullConfigStateWithFilter :: ForeignPtr CacClient -> Maybe String -> IO (Either Error Value)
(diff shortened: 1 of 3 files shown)


ayushjain17:
can we also, update the getConfig function in haskell, it is supposed to take Maybe [String] and concat them with "," and send it to cac client, but currently it is taking Maybe String (already assuming that the filters are joined with "," before passing it to the function)

pratikmishra356 (approved):

docs/client-context-aware-configuration.md
```diff
@@ -295,7 +295,7 @@ Get the full config definition of your tenants configuration from superposition.
 ##### Funtion Definition
 
 ```
-getFullConfigStateWithFilter :: ForeignPtr CacClient -> Maybe String -> IO (Either Error Value)
+getFullConfigStateWithFilter :: ForeignPtr CacClient -> Maybe String -> Maybe String -> IO (Either Error Value)
```
Datron, line 298: @mahatoankitkumar can we add the table explaining what each parameter is? I missed this the last time

Datron (approved):

PR #195, opened 2024-08-01
feat: adding go cac and experimentation client wrapper

namitgoel:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/clients/go/cacclient/main.go
+++ b/clients/go/cacclient/main.go
@@ -1,0 +1,186 @@
+package cacclient
+
+/*
+#include "libcac_client.h"
+*/
+import "C"
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"unsafe"
+)
+
+type MergeStrategy string
+
+const (
+	MERGE   MergeStrategy = "MERGE"
+	REPLACE MergeStrategy = "REPLACE"
+)
+
+// CacClient struct
+type CacClient struct {
+	tenant           string
+	pollingFrequency int
+	cacHostName      string
+	delimiter        string
+}
+
+// NewCacClient creates a new CacClient
+func NewCacClient(tenantName string, pollingFrequency int, cacHostName string) (*CacClient, error) {
+	if tenantName == "" {
+		return nil, errors.New("tenantName cannot be Empty")
+	}
+	if cacHostName == "" {
+		return nil, errors.New("cacHostName cannot be Empty")
+	}
+	client := &CacClient{tenant: tenantName, pollingFrequency: pollingFrequency, cacHostName: cacHostName, delimiter: ","}
+	resp := C.cac_new_client(C.CString(tenantName), C.ulong(pollingFrequency), C.CString(cacHostName))
+	if resp == 1 {
+		errorMessage := client.GetLastErrorMessage()
+		fmt.Printf("Some Error Occur while creating new client: %s\n", errorMessage)
+	}
+	return client, nil
+}
+
+// GetLastErrorMessage gets the last error message from the C library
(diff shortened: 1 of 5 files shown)


clients/go/cacclient/main.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// StartCacPollingUpdate starts polling updates in a separate goroutine
+func (c *CacClient) StartCacPollingUpdate() {
+	tenant := C.CString(c.tenant)
+	go C.cac_start_polling_update(tenant)
+}
+
+// GetCacConfig gets the configuration from the C library
+func (c *CacClient) GetCacConfig(filterQuery string, filterPrefix string) string {
```
ayushjain17, line 56: where ever we are accepting filters, they have to go equivalent of `Option<Vec<String>>` and then we get Some value then we join the array based on the join character for that particular function, else we send null pointer

ayushjain17, line 56: params like filter query should take Maybe Json object and context should take Json object type of data, and then the function should stringify the json value and pass it to the FFI functions

in case of no json, simply pass null pointer, 

it needs to be the same across other clients as well, please make the changes there as well
and if possible raise a fix PR for python client as well, adding the same changes

ayushjain17 (changes requested):
Fix type in the functions

clients/go/cacclient/main.go
namitgoel: @ayushjain17 updated

namitgoel: done

Datron:
@namitgoel Add go version in nix flake too

clients/go/main.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func main() {
+	// Example usage
+	tenant := "dev"
+	pollingFrequency := 1
+	cachostName := "http://localhost:8080"
+
+	client := cacclient.NewCacClient(tenant, pollingFrequency, cachostName)
+	client.CreateNewCacClient()
```
Datron, line 16: Why do we create a client twice? Can we do these operations in one call?

clients/go/main.go
```diff
@@ -0,0 +1,22 @@
...
+	fmt.Println("Default Configs", client.GetCacConfig(nil, nil))
+
+	expClient := expclient.NewExperimentationClient(tenant, pollingFrequency, cachostName)
+	expClient.CreateNewExperimentationClient()
```
Datron, line 20: Same here

clients/go/main.go
```diff
@@ -0,0 +1,22 @@
...
+	fmt.Println("Running experiments:", expClient.GetRunningExperiments())
```
Datron, line 21: @namitgoel please add examples for fetching a config, resolving, etc. Basically one for each function in the client

Datron (changes requested):

clients/go/main.go
namitgoel: done

clients/go/main.go
namitgoel: done

clients/go/main.go
namitgoel: done

Datron (approved):

mahatoankitkumar (approved):

pratikmishra356 (changes requested):
@namitgoel @Datron function's output type is string , ideally it should match whatever type is there in rust client

Datron:
@pratikmishra356 its fine, the implementor can add a parse to convert to json or his type

PR #198, opened 2024-08-05
[Client] Java client

mahatoankitkumar:
## Problem
Java CAC Client


clients/java/src/main/java/CAC/ExperimentationClient.java
ayushjain17, line 1: we keep the two clients separately in different projects

clients/java/exp-client/gradle/libs.versions.toml
```diff
@@ -0,0 +1,10 @@
+# This file was generated by the Gradle 'init' task.
+# https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format
+
+[versions]
+guava = "33.1.0-jre"
```
Datron, line 5: Why do we import guava? Do we need this?

clients/java/cac-client/src/main/java/CAC/CacClient.java
```diff
@@ -0,0 +1,127 @@
...
+    public String getStringAndFree(Pointer ptr) {
+        if (ptr == null) {
+            return null;
+        }
+        try {
+            return ptr.getString(0);
+        } finally {
+            freeString(ptr);
+        }
+    }
```
Datron, line 126: Can we write this better?

clients/java/cac-client/src/main/java/CAC/Client.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.concurrent.CountDownLatch;
+
+import jnr.ffi.Pointer;
+
+public class Client {
+
+    private static void callCacClient() {
+        String dylib = "cac_client";
+        File currentDir = new File(System.getProperty("user.dir"));
+        String libraryPath = currentDir.getParentFile().getParentFile().getParentFile() + "/target/debug";
```
Datron, line 14: No hardcoding the path please, use env SUPERPOSITION_LIB_PATH

clients/java/cac-client/src/main/java/CAC/Client.java
```diff
@@ -0,0 +1,109 @@
...
+
+        System.out.println("CAC Client");
+
+        System.out.println("---------------------");
+
+        CacClient wrapper = new CacClient(libraryPath, dylib);
+
+        int newClient;
+        try {
+            newClient = wrapper.cacNewClient(tenant, 1, "http://localhost:8080");
```
Datron, line 27: The URL should not be hardcoded, is this an example file?

clients/java/cac-client/src/main/java/CAC/Client.java
```diff
@@ -0,0 +1,109 @@
+package CAC;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
+
+import jnr.ffi.Pointer;
+
+public class Client {
```
Datron, line 9: Can we name this class better? It seems like a Client implementation

clients/java/exp-client/src/main/java/CAC/Client.java
```diff
@@ -0,0 +1,116 @@
+package CAC;
```
Datron, line 1: Package should be different

Datron (changes requested):

clients/java/exp-client/gradle/libs.versions.toml
mahatoankitkumar: Not sure, let me check.

clients/java/cac-client/gradle/libs.versions.toml
```diff
@@ -0,0 +1,10 @@
+# This file was generated by the Gradle 'init' task.
+# https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format
+
+[versions]
+guava = "33.1.0-jre"
+junit-jupiter = "5.10.2"
+
+[libraries]
+guava = { module = "com.google.guava:guava", version.ref = "guava" }
+junit-jupiter = { module = "org.junit.jupiter:junit-jupiter", version.ref = "junit-jupiter" }
```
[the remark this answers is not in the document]
mahatoankitkumar: will check

clients/java/cac-client/src/main/java/CAC/Client.java
mahatoankitkumar: Yes this is an example file.

Datron:
Can you add java and gradle versions in flakes.nix as well?

Datron (approved):

ShubhranshuSanjeev (approved):

PR #199, opened 2024-08-07
fix: newtypes for entity fiels

pratikmishra356:
## Problem
We need strong type validation for reduce repetitive checks

## Solution
Added newtypes and validation for name field of default_config, functions, type templates , dimension, also added priority type for dimension
for path type value , when validation fails it was giving 404 notfound error , that's why overriden pathconfig error in app_data

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes

NA

### DIFF
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -10,12 +10,15 @@ use superposition_macros::{
 };
 use superposition_types::{result as superposition, SuperpositionUser, User};
 
-use crate::{api::context::helpers::validate_value_with_function, db::models};
 use crate::{
-    api::functions::helpers::get_published_function_code,
+    api::{
+        context::helpers::validate_value_with_function,
+        default_config::types::DefaultConfigKey,
+        functions::helpers::get_published_function_code,
+    },
     db::{
         self,
-        models::{Context, DefaultConfig},
+        models::{self, Context, DefaultConfig},
         schema::{contexts::dsl::contexts, default_configs::dsl},
     },
     helpers::add_config_version,
@@ -32,10 +35,7 @@ use diesel::{
 };
 use diesel::{Connection, SelectableHelper};
 use jsonschema::{Draft, JSONSchema, ValidationError};
-use regex::Regex;
 use serde_json::{from_value, json, Map, Value};
-
-const KEY_NAME_REGEX: &str = "^[a-zA-Z0-9-_]([a-zA-Z0-9-_.]{0,254}[a-zA-Z0-9-_])?$";
 
 pub fn endpoints() -> Scope {
     Scope::new("").service(create).service(get).service(delete)
@@ -44,7 +44,7 @@ pub fn endpoints() -> Scope {
 #[put("/{key}")]
 async fn create(
     state: Data<AppState>,
-    key: web::Path<String>,
+    key: web::Path<DefaultConfigKey>,
     custom_headers: CustomHeaders,
     request: web::Json<CreateReq>,
     db_conn: DbConnection,
@@ -52,22 +52,8 @@ async fn create(
 ) -> superposition::Result<HttpResponse> {
     let DbConnection(mut conn) = db_conn;
     let req = request.into_inner();
-    let key = key.into_inner();
+    let key = key.into_inner().into();
     let tags = parse_config_tags(custom_headers.config_tags)?;
-
-    let regex = Regex::new(KEY_NAME_REGEX).map_err(|err| {
-        unexpected_error!("could not parse regex due to: {}", err.to_string())
-    })?;
-
-    if !regex.is_match(&key) {
-        return Err(bad_argument!(
-            "The key name {} is invalid, it should obey the regex {}. \
-            It can contain the following characters only [a-zA-Z0-9-_.] \
-            and it should not start or end with a '.' character.",
-            key,
-            KEY_NAME_REGEX
-        ));
-    }
 
     if req.value.is_none() && req.schema.is_none() && req.function_name.is_none() {
         log::error!("No data provided in the request body for {key}");
@@ -252,7 +238,7 @@ pub fn get_key_usage_context_ids(
 #[delete("/{key}")]
 async fn delete(
     state: Data<AppState>,
-    path: Path<String>,
+    path: Path<DefaultConfigKey>,
     custom_headers: CustomHeaders,
     db_conn: DbConnection,
     user: User,
@@ -260,7 +246,7 @@ async fn delete(
     let DbConnection(mut conn) = db_conn;
     let tags = parse_config_tags(custom_headers.config_tags)?;
 
-    let key = path.into_inner();
+    let key: String = path.into_inner().into();
     fetch_default_key(&key, &mut conn)?;
     let context_ids = get_key_usage_context_ids(&key, &mut conn)
         .map_err(|_| unexpected_error!("Something went wrong"))?;
--- a/crates/context_aware_config/src/api/default_config/types.rs
+++ b/crates/context_aware_config/src/api/default_config/types.rs
@@ -1,5 +1,7 @@
+use derive_more::{AsRef, Deref, DerefMut, Into};
 use serde::{Deserialize, Deserializer};
 use serde_json::{Map, Value};
+use superposition_types::RegexEnum;
 
 #[derive(Debug, Deserialize)]
 pub struct CreateReq {
@@ -8,6 +10,25 @@ pub struct CreateReq {
     pub schema: Option<Map<String, Value>>,
     #[serde(default, deserialize_with = "deserialize_option")]
     pub function_name: Option<Value>,
+}
+
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "String")]
+pub struct DefaultConfigKey(String);
+impl DefaultConfigKey {
+    pub fn validate_data(name: String) -> Result<Self, String> {
+        let name = name.trim();
+        RegexEnum::DefaultConfigKey
+            .match_regex(name)
+            .map(|_| Self(name.to_string()))
+    }
+}
+
+impl TryFrom<String> for DefaultConfigKey {
+    type Error = String;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Ok(Self::validate_data(value)?)
+    }
 }
 
 fn deserialize_option<'de, D>(deserializer: D) -> Result<Option<Value>, D::Error>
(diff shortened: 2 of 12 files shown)


crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -20,6 +19,13 @@ use crate::{
     },
     helpers::add_config_version,
 };
+use crate::{
+    api::{
+        context::helpers::validate_value_with_function,
+        default_config::types::DefaultConfigKey,
+    },
+    db::models,
```
ayushjain17, line 27: can we instead update the [above imports](https://github.com/juspay/superposition/pull/199/files#diff-debfcffd0fcd387cbeb80d1f272aab5fab878afcee71711199a19035dafe8e9cL18) instead

we can change
```models::{Context, DefaultConfig}```

to 
```models::{self, Context, DefaultConfig}```

File: crates/context_aware_config/src/api/default_config/types.rs [hunk 2]
```diff
@@ -10,6 +12,38 @@ pub struct CreateReq {
     pub function_name: Option<Value>,
 }
 
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "String")]
+pub struct DefaultConfigKey(String);
+impl DefaultConfigKey {
+    pub fn validate_data(name: String) -> Result<Self, String> {
+        let regex_str: &str = "^[a-zA-Z0-9-_]([a-zA-Z0-9-_.]{0,254}[a-zA-Z0-9-_])?$";
```
ayushjain17, line 20: wont it be better to keep it as a const, outside the function ?

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -10,6 +12,38 @@ pub struct CreateReq {
...
+
+        let regex = Regex::new(regex_str).map_err(|err| {
+            log::error!("error while validating defaultConfigKey : {err}");
+            "Something went wrong".to_string()
+        })?;
+
+        if !regex.is_match(&name) {
+            return Err(format!(
+                "The key name {} is invalid, it should obey the regex {}. \
```
ayushjain17, line 29: Not mandatory, but should we start considering this syntax for simple variable values ?
```suggestion
                "The key name {name} is invalid, it should obey the regex {regex_str}. \
```

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -10,6 +12,38 @@ pub struct CreateReq {
...
+        } else {
+            Ok(Self(name))
+        }
+    }
+}
+
+impl TryFrom<String> for DefaultConfigKey {
+    type Error = String;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Ok(DefaultConfigKey::validate_data(value)?)
```
ayushjain17, line 43:
```suggestion
        Ok(Self::validate_data(value)?)
```

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -10,6 +12,38 @@ pub struct CreateReq {
     pub function_name: Option<Value>,
 }
 
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "String")]
+pub struct DefaultConfigKey(String);
```
ayushjain17, line 17: shouldn't this be simply `ConfigKey`, seems a bit confusing naming it specifically `DefaultConfigKey`, or do we have some specific reason to name it this way ?

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -1,18 +1,72 @@
...
+        } else {
+            Ok(Self(priority_val))
+        }
+    }
+}
+
+impl TryFrom<i32> for Priority {
+    type Error = String;
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        Ok(Priority::validate_data(value)?)
```
ayushjain17, line 34:
```suggestion
        Ok(Self::validate_data(value)?)
```

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -1,18 +1,72 @@
...
+        } else {
+            Ok(Self(name))
+        }
+    }
+}
+
+impl TryFrom<String> for DimensionName {
+    type Error = String;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Ok(DimensionName::validate_data(value)?)
```
ayushjain17, line 66:
```suggestion
        Ok(Self::validate_data(value)?)
```

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -1,18 +1,72 @@
...
+        Ok(Priority::validate_data(value)?)
+    }
+}
+
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "String")]
+pub struct DimensionName(String);
+impl DimensionName {
+    pub fn validate_data(name: String) -> Result<Self, String> {
+        let regex_str: &str = "^[a-zA-Z0-9-_]([a-zA-Z0-9-_.]{0,254}[a-zA-Z0-9-_])?$";
```
ayushjain17, line 43: lets reuse the previous string itself, once its made as a const

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -1,18 +1,72 @@
...
+
+        let regex = Regex::new(regex_str).map_err(|err| {
+            log::error!("error while validating dimensionName : {err}");
+            "Something went wrong".to_string()
+        })?;
+
+        if !regex.is_match(&name) {
+            return Err(format!(
+                "The key name {} is invalid, it should obey the regex {}. \
+                It can contain the following characters only [a-zA-Z0-9-_.] \
```
ayushjain17, line 53: we can have this also extracted as a const `[a-zA-Z0-9-_.]`, being used twice, can be easier to maintain later
we can probably have some trait implemented here
something like, we can have an enum for all regexs

```
const KEY_NAME_REGEX: &str = "^[a-zA-Z0-9-_]([a-zA-Z0-9-_.]{0,254}[a-zA-Z0-9-_])?$";
const KEY_NAME_REGEX_WORDS: &str = "It can contain the following characters only [a-zA-Z0-9-_.] \
                                                                          and it should not start or end with a '.' character.";

const FUNCTION_NAME_REGEX: &str = "^[a-zA-Z0-9-_]{1,64}$";
const FUNCTION_NAME_REGEX_WORDS: &str = "It can contain the following characters only [a-zA-Z0-9-_]";

pub enum Regex {
   KeyName,
   FunctionName
}

impl Regex {
   fn get_error_message(&self) -> String {
      match self {
         Self::KeyName => KEY_NAME_REGEX_WORDS,
         Self::FunctionName => FUNCTION_NAME_REGEX_WORDS,
      }.to_string()
   }
}

impl Display for Regex {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let regex = match self {
            Self::KeyName => KEY_NAME_REGEX,
            Self::FunctionName => FUNCTION_NAME_REGEX,
        }
        .to_string();
        write!(f, "{regex}")
    }
}
```

we can keep extending it further on addition of new other variants of the enum, it will help keep all the necessary things tight and mandatory

`TypeName` would also be a variant of it, not adding it in the code snippet, you can add this and others also

crates/context_aware_config/src/api/functions/types.rs
```diff
@@ -10,12 +12,43 @@ pub struct UpdateFunctionRequest {
...
+        } else {
+            Ok(Self(name))
+        }
+    }
+}
+
+impl TryFrom<String> for FunctionName {
+    type Error = String;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Ok(FunctionName::validate_data(value)?)
```
ayushjain17, line 48:
```suggestion
        Ok(Self::validate_data(value)?)
```

crates/context_aware_config/src/api/type_templates/types.rs
```diff
@@ -20,3 +22,33 @@ pub struct QueryFilters {
...
+        } else {
+            Ok(Self(name))
+        }
+    }
+}
+
+impl TryFrom<String> for TypeTemplateName {
+    type Error = String;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Ok(TypeTemplateName::validate_data(value)?)
```
ayushjain17, line 52:
```suggestion
        Ok(Self::validate_data(value)?)
```

crates/context_aware_config/src/api/type_templates/types.rs
```diff
@@ -20,3 +22,33 @@ pub struct QueryFilters {
...
+        let regex_str: &str = "^[a-zA-Z0-9-_]{1,64}$";
+
+        let regex = Regex::new(regex_str).map_err(|err| {
+            log::error!("error while validating typeTemplateName : {err}");
+            "Something went wrong".to_string()
+        })?;
+
+        if !regex.is_match(&name) {
+            return Err(format!(
+                "The type name {} is invalid, it should obey the regex {}",
```
ayushjain17, line 40: lets add a to word extension of the regex here as well, like in other places

ayushjain17 (changes requested):
added comments for code refactoring/cleanup

crates/context_aware_config/src/api/default_config/types.rs
pratikmishra356: Generally for all types I followed entityname<>fieldname

crates/context_aware_config/src/api/dimension/types.rs
pratikmishra356: Lets keep different regex for different entity so that the regex logic doesn't overlap

crates/context_aware_config/src/api/type_templates/types.rs
```diff
@@ -20,3 +22,33 @@ pub struct QueryFilters {
     pub count: Option<i64>,
     pub page: Option<i64>,
 }
+
+#[derive(Debug, Deserialize, Serialize, AsRef, Deref, DerefMut, Into, Clone)]
```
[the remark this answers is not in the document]
pratikmishra356: These derive macro should be used as per needed and we should not have all the macros for multiple types , as this will increase compilation time , also adding or removing wouldn't be straight forward in that case

ayushjain17 (approved):

mahatoankitkumar (approved):

PR #203, opened 2024-08-14
fix: changing aws kms library to aws-sdk-kms to rusto

namitgoel:
## Problem
we are using aws kms to store secrets like db password and when we are deploying superposition service to different region aws kms needs which region to call to decrypt those keys for that the library we were using was rusto which has a map for the regions but when we tried to deployed in sp-south-2 region it failed because the map does not contain that region.

## Solution
solution is from rusto we moved to different library aws-sdk-kms that handles it.

No env change
## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,8 +52,6 @@ reqwest = { version = "0.11.18", features = ["json"] }
 jsonschema = "~0.17"
 jsonlogic = { git = "https://github.com/juspay/jsonlogic_rs.git", version = "0.5.3" }
 rs-snowflake = "0.6.0"
-rusoto_kms = "0.48.0"
-rusoto_signature = "0.48.0"
 bytes = "1.4.0"
 rusoto_core = "0.48.0"
 rand = "0.8.5"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -69,6 +67,8 @@ leptos_actix = { version = "0.6.11" }
 thiserror = { version = "1.0.57" }
 leptos-use = "0.10.3"
 mime = "0.3.17"
+aws-sdk-kms = {version = "1.38.0"}
+aws-config = { version = "1.1.1", features = ["behavior-version-latest"] }
```
ayushjain17, line 71: this is an un-supported version combination with respect to aws-sdk-kms, use:
```suggestion
aws-config = { version = "1.1.7", features = ["behavior-version-latest"] }
```

ayushjain17 (changes requested):
Version conflict

Cargo.toml
namitgoel: @ayushjain17 updated

ayushjain17 (approved):

pratikmishra356 (approved):

PR #206, opened 2024-08-19
chore(nix): Switch to omnix

srid:
Use https://omnix.page/om/ci.html to build Nix outputs.

`omnix` is now installed in self-hosted runners.


Datron (approved):
@srid what is the advantage of using omnix, why are we switching to it?

mahatoankitkumar (approved):

srid:
@Datron `nixci` is being superceded by `omnix`. All new development happens on Omnix, rather than nixci. See `#omnix` channel in our Slack.

Issue #189, opened 2024-07-31
All APIs should return the entity it acted on


Datron:
APIs should return the entity they acted on in update, create and delete APIs. This makes it easier for scripts to operate on. Requires changes in API endpoints and the leptos frontend

PR #208, opened 2024-08-20
fix: return default config object when created

Datron:
## Problem
Most APIs return the entity object created, but this is not done for default config

## Solution
return the row that was created/updated in the database

Solves #189

### DIFF
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -35,7 +35,7 @@ use diesel::{
 };
 use diesel::{Connection, SelectableHelper};
 use jsonschema::{Draft, JSONSchema, ValidationError};
-use serde_json::{from_value, json, Map, Value};
+use serde_json::{from_value, Map, Value};
 
 pub fn endpoints() -> Scope {
     Scope::new("").service(create).service(get).service(delete)
@@ -189,9 +189,7 @@ async fn create(
         AppHeader::XConfigVersion.to_string(),
         version_id.to_string(),
     ));
-    Ok(http_resp.json(json!({
-        "message": "DefaultConfig created/updated successfully."
-    })))
+    Ok(http_resp.json(default_config))
 }
 
 fn fetch_default_key(


ShubhranshuSanjeev (approved):

crates/context_aware_config/src/api/default_config/handlers.rs
ShubhranshuSanjeev, line 1:
```
async fn create(
    state: Data<AppState>,
    key: web::Path<DefaultConfigKey>,
    custom_headers: CustomHeaders,
    request: web::Json<CreateReq>,
    db_conn: DbConnection,
    user: User,
) -> superposition::Result<HttpResponse>
```

A suggestion, for this handler, I think we can mention the type of response `DefaultConfig` rather than just `HttpResponse`

ayushjain17: yeah, we have been using HttpResponse everywhere as the return type, 
instead we should make it more specific
```actix_web::Result<Json<DefaultConfig>>```

**Review by mahatoankitkumar — APPROVED** [review]

File: crates/context_aware_config/src/api/default_config/handlers.rs [hunk 1]
```diff
@@ -189,9 +189,7 @@ async fn create(
         AppHeader::XConfigVersion.to_string(),
         version_id.to_string(),
     ));
-    Ok(http_resp.json(json!({
-        "message": "DefaultConfig created/updated successfully."
-    })))
+    Ok(http_resp.json(json!(default_config)))
```
ayushjain17, line 192:
```suggestion
    Ok(http_resp.json(default_config))
```

json! macro is not needed here, http_resp.json function as well as `Json` type of actix_web by default serializes its data

ayushjain17 (dismissed):
json! macro is not needed here, http_resp.json function as well as Json type of actix_web by default serializes its data

crates/context_aware_config/src/api/default_config/handlers.rs
Datron: Nice! I didn't know that

ayushjain17 (approved):

PR #209, opened 2024-08-22
fix: Add pagination to list APIs

mahatoankitkumar:
## Problem
Add pagination to dimension and default_config list APIs

https://github.com/user-attachments/assets/3bb751c0-f6c8-4e2d-a760-b7c8d3692efa

### DIFF
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -34,19 +34,18 @@ use service_utils::{
 use superposition_macros::response_error;
 use superposition_macros::{bad_argument, db_error, unexpected_error};
 use superposition_types::{
-    custom_query::{self as superposition_query, CustomQuery, QueryMap},
+    custom_query::{
+        self as superposition_query, CustomQuery, PaginationParams, QueryMap,
+    },
     result as superposition, Cac, Condition, Config, Context, Overrides,
     PaginatedResponse, TenantConfig, User,
 };
 use uuid::Uuid;
 
-use crate::api::dimension::get_all_dimension_schema_map;
-use crate::api::{
-    context::{
-        delete_context_api, hash, put, validate_dimensions_and_calculate_priority, PutReq,
-    },
-    type_templates::types::QueryListFilters,
+use crate::api::context::{
+    delete_context_api, hash, put, validate_dimensions_and_calculate_priority, PutReq,
 };
+use crate::api::dimension::get_all_dimension_schema_map;
 use crate::db::models::ConfigVersion;
 use crate::{
     db::schema::{config_versions::dsl as config_versions, event_log::dsl as event_log},
@@ -754,24 +753,33 @@ async fn get_resolved_config(
 #[get("/versions")]
 async fn get_config_versions(
     db_conn: DbConnection,
-    filters: Query<QueryListFilters>,
+    filters: Query<PaginationParams>,
 ) -> superposition::Result<Json<PaginatedResponse<ConfigVersion>>> {
     let DbConnection(mut conn) = db_conn;
+
+    if let Some(true) = filters.all {
+        let config_versions: Vec<ConfigVersion> =
+            config_versions::config_versions.get_results(&mut conn)?;
+        return Ok(Json(PaginatedResponse {
+            total_pages: 1,
+            total_items: config_versions.len() as i64,
+            data: config_versions,
+        }));
+    }
 
     let n_version: i64 = config_versions::config_versions
         .count()
         .get_result(&mut conn)?;
+
+    let limit = filters.count.unwrap_or(10);
     let mut builder = config_versions::config_versions
         .into_boxed()
-        .order(config_versions::created_at.desc());
-    if let Some(limit) = filters.count {
-        builder = builder.limit(limit);
-    }
+        .order(config_versions::created_at.desc())
+        .limit(limit);
     if let Some(page) = filters.page {
-        let offset = (page - 1) * filters.count.unwrap_or(10);
+        let offset = (page - 1) * limit;
         builder = builder.offset(offset);
     }
-    let limit = filters.count.unwrap_or(10);
     let config_versions: Vec<ConfigVersion> = builder.load(&mut conn)?;
     let total_pages = (n_version as f64 / limit as f64).ceil() as i64;
     Ok(Json(PaginatedResponse {
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -34,16 +34,19 @@ use service_utils::{
 use superposition_macros::response_error;
 use superposition_macros::{bad_argument, db_error, unexpected_error};
 use superposition_types::{
-    custom_query::{self as superposition_query, CustomQuery, QueryFilters, QueryMap},
+    custom_query::{self as superposition_query, CustomQuery, QueryMap},
     result as superposition, Cac, Condition, Config, Context, Overrides,
     PaginatedResponse, TenantConfig, User,
 };
 use uuid::Uuid;
 
-use crate::api::context::{
-    delete_context_api, hash, put, validate_dimensions_and_calculate_priority, PutReq,
-};
 use crate::api::dimension::get_all_dimension_schema_map;
+use crate::api::{
+    context::{
+        delete_context_api, hash, put, validate_dimensions_and_calculate_priority, PutReq,
+    },
+    type_templates::types::QueryListFilters,
+};
 use crate::db::models::ConfigVersion;
 use crate::{
     db::schema::{config_versions::dsl as config_versions, event_log::dsl as event_log},
@@ -751,7 +754,7 @@ async fn get_resolved_config(
 #[get("/versions")]
 async fn get_config_versions(
     db_conn: DbConnection,
-    filters: Query<QueryFilters>,
+    filters: Query<QueryListFilters>,
 ) -> superposition::Result<Json<PaginatedResponse<ConfigVersion>>> {
     let DbConnection(mut conn) = db_conn;
 
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -9,13 +9,14 @@ use service_utils::{
 use superposition_macros::{
     bad_argument, db_error, not_found, unexpected_error, validation_error,
 };
-use superposition_types::{result as superposition, User};
+use superposition_types::{result as superposition, PaginatedResponse, User};
 
 use crate::{
     api::{
         context::helpers::validate_value_with_function,
         default_config::types::DefaultConfigKey,
         functions::helpers::get_published_function_code,
+        type_templates::types::QueryListFilters,
     },
     db::{
         self,
@@ -30,7 +31,7 @@ use crate::helpers::put_config_in_redis;
 
 use actix_web::{
     delete, get, put,
-    web::{self, Data, Json, Path},
+    web::{self, Data, Json, Path, Query},
     HttpResponse, Scope,
 };
 use chrono::Utc;
@@ -211,11 +212,31 @@ fn fetch_default_key(
 }
 
 #[get("")]
-async fn get(db_conn: DbConnection) -> superposition::Result<Json<Vec<DefaultConfig>>> {
+async fn get(
+    db_conn: DbConnection,
+    filters: Query<QueryListFilters>,
+) -> superposition::Result<Json<PaginatedResponse<DefaultConfig>>> {
     let DbConnection(mut conn) = db_conn;
 
-    let result: Vec<DefaultConfig> = dsl::default_configs.get_results(&mut conn)?;
-    Ok(Json(result))
+    let n_default_configs: i64 = dsl::default_configs.count().get_result(&mut conn)?;
+    let mut builder = dsl::default_configs
+        .into_boxed()
+        .order(dsl::created_at.desc());
+    if let Some(limit) = filters.count {
+        builder = builder.limit(limit);
+    }
+    if let Some(page) = filters.page {
+        let offset = (page - 1) * filters.count.unwrap_or(10);
+        builder = builder.offset(offset);
+    }
+    let limit = filters.count.unwrap_or(10);
+    let result: Vec<DefaultConfig> = builder.load(&mut conn)?;
+    let total_pages = (n_default_configs as f64 / limit as f64).ceil() as i64;
+    Ok(Json(PaginatedResponse {
+        total_pages,
+        total_items: n_default_configs,
+        data: result,
+    }))
 }
 
 pub fn get_key_usage_context_ids(
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -9,14 +9,15 @@ use service_utils::{
 use superposition_macros::{
     bad_argument, db_error, not_found, unexpected_error, validation_error,
 };
-use superposition_types::{result as superposition, PaginatedResponse, User};
+use superposition_types::{
+    custom_query::PaginationParams, result as superposition, PaginatedResponse, User,
+};
 
 use crate::{
     api::{
         context::helpers::validate_value_with_function,
         default_config::types::DefaultConfigKey,
         functions::helpers::get_published_function_code,
-        type_templates::types::QueryListFilters,
     },
     db::{
         self,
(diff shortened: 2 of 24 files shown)


crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -15,6 +15,7 @@ use crate::{
         context::helpers::validate_value_with_function,
         default_config::types::DefaultConfigKey,
         functions::helpers::get_published_function_code,
+        type_templates::types::QueryFilters,
```
pratikmishra356, line 18: why we are using type_templates queryfilter , can we make this generic listquery type

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -113,10 +116,22 @@ async fn get(
     db_conn: DbConnection,
     state: Data<AppState>,
     tenant: Tenant,
+    filters: Query<QueryFilters>,
```
pratikmishra356, line 119: can we handler non positive cases , only allow > 0 for count and page

pratikmishra356 (changes requested):

Datron:
@mahatoankitkumar please add screen recordings for each page so it's easy to understand the changes

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -204,11 +205,31 @@ fn fetch_default_key(
...
+        .into_boxed()
+        .order(dsl::created_at.desc());
+    if let Some(limit) = filters.count {
+        builder = builder.limit(limit);
+    }
+    if let Some(page) = filters.page {
+        let offset = (page - 1) * filters.count.unwrap_or(10);
+        builder = builder.offset(offset);
+    }
+    let limit = filters.count.unwrap_or(10);
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: This will require changes in frontend too, as frontend relies on the fact that without params it will get the full list.

crates/frontend/src/pages/default_config.rs
```diff
@@ -26,19 +29,35 @@ pub struct RowData {
 #[component]
 pub fn default_config() -> impl IntoView {
     let tenant_rs = use_context::<ReadSignal<String>>().unwrap();
+    let enable_grouping =
+        create_rw_signal(get_local_storage::<bool>("enable_grouping").unwrap_or(false));
+    let (filters, set_filters) = create_signal(if enable_grouping.get() {
```
ShubhranshuSanjeev, line 34: This will cause hydration issues, as enable_grouping is only available in localstorage, and the list filters will update the number of list items.

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -204,11 +205,31 @@ fn fetch_default_key(
...
+    let result: Vec<DefaultConfig> = builder.load(&mut conn)?;
+    let total_pages = (n_default_configs as f64 / limit as f64).ceil() as u64;
+    Ok(HttpResponse::Ok().json(json!({
```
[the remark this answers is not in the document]
mahatoankitkumar: Yeah had created in my last PR, Will rebase and change

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -180,7 +180,7 @@ pub struct StatusTypes(
 );
 
 #[derive(Deserialize, Debug)]
-pub struct ListFilters {
+pub struct ExpListFilters {
```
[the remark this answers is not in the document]
mahatoankitkumar: Will make it consistent everywhere by putting it in superposition_types

crates/frontend/src/pages/default_config.rs
ayushjain17: @mahatoankitkumar this needs to be done in this way only: https://github.com/juspay/superposition/pull/209/files#diff-f14ee882e4ca49f81aa2293a656f92603255460bc68c7247b510d60ad6f7c4e6R64

or we start accepting cookies for this

mahatoankitkumar: Sure i leave it as it is, but cookies has the same problem, already tried it.

crates/context_aware_config/src/api/default_config/handlers.rs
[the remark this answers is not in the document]
mahatoankitkumar: @ShubhranshuSanjeev in frontend we need the full list in one case that is when the grouping is enabled, so i don't think this would be a good idea.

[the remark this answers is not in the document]
Datron: @mahatoankitkumar agree with @ShubhranshuSanjeev, the component in frontend should also paginate in a way. I'm guessing the override form is using this, if we scroll to the end it should load the next page

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -119,10 +120,22 @@ async fn create(
...
+    let mut builder = dimensions.into_boxed().order(created_at.desc());
+    if let Some(limit) = filters.count {
+        builder = builder.limit(limit);
+    }
+    if let Some(page) = filters.page {
+        let offset = (page - 1) * filters.count.unwrap_or(10);
+        builder = builder.offset(offset);
+    }
+    let limit = filters.count.unwrap_or(10);
+    let result: Vec<Dimension> = builder.load(&mut conn)?;
```
Datron, line 137: Could we apply the same API behaviour here, where if limit and page are not passed they are assumed?

crates/frontend/src/pages/default_config.rs
```diff
@@ -70,6 +100,26 @@ pub fn default_config() -> impl IntoView {
         navigate(redirect_url.as_str(), Default::default());
     };
 
+    let handle_next_click = Callback::new(move |total_pages: i64| {
+        set_filters.update(|f| {
+            f.page = match f.page {
+                Some(p) if p < total_pages => Some(p + 1),
+                Some(p) => Some(p),
+                None => None,
+            }
```
Datron, line 109: Can we abstract this out into frontend/utils?

crates/frontend/src/pages/function.rs
```diff
@@ -89,7 +89,9 @@ pub fn function_page() -> impl IntoView {
                         let (function_rs, _) = create_signal(function);
                         publish_error_ws.set("".to_string());
                         match function_rs.get().published_at {
-                            Some(val) => show_publish_ws.set(val < function_rs.get().draft_edited_at),
+                            Some(val) => {
+                                show_publish_ws.set(val < function_rs.get().draft_edited_at)
```
Datron, line 93: We don't need the brackets here, right?

Datron:
@mahatoankitkumar can we add pagination to overrides page as well? Should that be a different PR

crates/frontend/src/pages/default_config.rs
```diff
@@ -228,8 +278,15 @@ pub fn default_config() -> impl IntoView {
                     }
                 }}
                 {move || {
-                    let default_config = default_config_resource.get().unwrap_or(vec![]);
+                    let default_config = default_config_resource
+                        .get()
+                        .unwrap_or(PaginatedResponse {
+                            total_items: 1,
```
pratikmishra356, line 284: here default items and page is 1, in other places it is 0

pratikmishra356:
Screenshot 2024-10-11 at 16 48 00
@mahatoankitkumar shouldn't it should show total number of keys at top , instead of number of keys in that page ?

crates/context_aware_config/src/api/default_config/handlers.rs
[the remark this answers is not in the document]
mahatoankitkumar: @Datron the default-config page with the grouping enabled needs the whole default-config.

crates/context_aware_config/src/api/dimension/handlers.rs
mahatoankitkumar: @Datron the default-config page with the grouping enabled needs the whole default-config.

crates/frontend/src/pages/function.rs
mahatoankitkumar: i just formatted, sure will remove this.

crates/frontend/src/pages/default_config.rs
mahatoankitkumar: I think it should be 1 as default as 0 makes no sense here

mahatoankitkumar:
> @mahatoankitkumar shouldn't it should show total number of keys at top , instead of number of keys in that page ?

@pratikmishra356 Now thats a UX call we can take, as i have seen at multiple places its actually like
`showing 1-10 of 50` which gives a proper idea.

crates/context_aware_config/src/api/default_config/handlers.rs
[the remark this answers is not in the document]
mahatoankitkumar: @Datron `dsl::default_configs.count().get_result(&mut conn)?;`  which gives the total_items returns i64, converting it to u32 can be unsafe right, better we can convert it to u64 that u32?

crates/context_aware_config/src/api/default_config/handlers.rs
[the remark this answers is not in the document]
Datron: Hmm okay, this makes it a little tricky. You may need to reverse this then, the API is paginated by default and you can pass a query param option --all to get all configs for grouping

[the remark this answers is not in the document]
ayushjain17: @mahatoankitkumar you can have a look at the changes I made for the `context/list` endpoint, that is also paginated along with filters, similar to the grouping operation which are you doing

so based on the the grouping, you can decide to paginate the response differently, but the response would still remain paginated

Datron (approved):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -211,11 +212,32 @@ fn fetch_default_key(
 }
 
 #[get("")]
```
pratikmishra356, line 214: should we use the same route convention as contexts i.e adding /list and support single get ?

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -9,13 +9,14 @@ use service_utils::{
...
 };
-use superposition_types::{result as superposition, User};
+use superposition_types::{result as superposition, PaginatedResponse, User};
 
 use crate::{
     api::{
         context::helpers::validate_value_with_function,
         default_config::types::DefaultConfigKey,
         functions::helpers::get_published_function_code,
+        type_templates::types::QueryListFilters,
```
pratikmishra356, line 19: @mahatoankitkumar can you transfer this type to superposition_types

crates/context_aware_config/src/api/type_templates/types.rs
```diff
@@ -18,9 +18,10 @@ pub struct TypeTemplateResponse {
     pub created_by: String,
 }
 #[derive(Debug, Clone, Deserialize)]
-pub struct QueryFilters {
+pub struct QueryListFilters {
     pub count: Option<i64>,
```
pratikmishra356, line 22: non positive integers are not restricted , it will lead to error

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -165,7 +165,7 @@ pub struct StatusTypes(
...
```
pratikmishra356, line 168: why we dont have **all** filter here ?

crates/frontend/src/api.rs
```diff
@@ -13,12 +14,26 @@ use crate::{
...
+    tenant: String,
+) -> Result<PaginatedResponse<Dimension>, ServerFnError> {
     let client = reqwest::Client::new();
     let host = use_host_server();
 
-    let url = format!("{}/dimension", host);
-    let response: Vec<Dimension> = client
+    let mut query_params = vec![];
+    if let Some(page) = filters.page {
+        query_params.push(format!("page={}", page));
```
pratikmishra356, line 26: dont we have functionality to add query params in url ?

crates/context_aware_config/src/api/default_config/handlers.rs
mahatoankitkumar: Single get is not required as of now,
we can develop it once needed
And /list makes sense, will make the change.

crates/frontend/src/api.rs
mahatoankitkumar: Yes we are doing this
checking here as its optional.

crates/frontend/src/pages/custom_types.rs
```diff
@@ -29,11 +30,15 @@ const TYPE_DRAWER_ID: &str = "type_template_drawer";
...
+    let (filters_rs, _) = create_signal(TypeFilter {
+        page: 1,
+        count: 10,
+        all: false,
+    });
     let types_resource = create_blocking_resource(
         move || (tenant_rs.get(), filters_rs.get()),
         |(t, filter)| async move {
-            match fetch_types(t, filter.page, filter.count).await {
+            match fetch_types(t, filter.page, filter.count, false).await {
```
ShubhranshuSanjeev, line 41:
```suggestion
            match fetch_types(t, filter.page, filter.count, filter.all).await {
```

ShubhranshuSanjeev (changes requested):

crates/frontend/src/api.rs
```diff
@@ -56,11 +83,12 @@ pub async fn fetch_snapshots(
     tenant: String,
     page: i64,
     count: i64,
+    all: bool,
```
ShubhranshuSanjeev, line 86: @mahatoankitkumar , can we either pass `ListFilters` struct or its properties individually as parameters to all these api functions. I would prefer individual parameters but I am okay with anything as long as it is consistent.

Why I prefer individual parameters?
As we won't have to construct a dummy ListFilters struct instance to pass the values which is being done at some places, I can directly write `0` or `true`, etc, and also removes the baggage of cloning if we want to reuse the values, as these primitive types implement Copy and its also cheap (not really a deal breaker).

Removing the baggage of cloning a dummy struct instance just because we want to reuse the values is a deal breaker for me.
https://github.com/juspay/superposition/pull/209/files#diff-e4e1d9e001dbf707cd00c7d34f98e60a017d96fd2be2230a10e5e625307819b4R46-R54

ShubhranshuSanjeev (changes requested):

crates/frontend/src/api.rs
mahatoankitkumar: For this we can raise a different PR to make it consistent.
I'll raise the next PR on this, no worries.

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -751,24 +753,27 @@ async fn get_resolved_config(
...
+    filters: Query<PaginationParams>,
 ) -> superposition::Result<Json<PaginatedResponse<ConfigVersion>>> {
     let DbConnection(mut conn) = db_conn;
 
     let n_version: i64 = config_versions::config_versions
         .count()
         .get_result(&mut conn)?;
+
+    let limit = if filters.all.unwrap_or(false) {
+        filters.count.unwrap_or(n_version)
```
ShubhranshuSanjeev, line 765:
```suggestion
        n_version
```

ShubhranshuSanjeev: https://github.com/juspay/superposition/pull/209/files#r1829150741

ShubhranshuSanjeev (changes requested):

postman/cac/custom types/Get Types/event.test.js
```diff
@@ -7,42 +7,41 @@ pm.test('expect response be 200', function () {
     console.log("The API returned the response", modified_response);
     pm.expect(JSON.stringify(modified_response)).to.be.eq(JSON.stringify([
```
Datron, line 8: Why did this change?

Datron (commented):
@mahatoankitkumar can you resolve some of @ShubhranshuSanjeev and @pratikmishra356 's comments?

postman/cac/custom types/Get Types/event.test.js
mahatoankitkumar: Had to change the order to pass the test, don't know why was it failing though.

Datron (approved):

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -563,15 +565,18 @@ async fn list_experiments(
...
 
-    let limit = filters.count.unwrap_or(10);
-    let offset = (filters.page.unwrap_or(1) - 1) * limit;
+    let number_of_experiments = count_query.count().get_result(&mut conn)?;
+    let limit = if pagination_params.all.unwrap_or(false) {
+        pagination_params.count.unwrap_or(number_of_experiments)
+    } else {
+        pagination_params.count.unwrap_or(10)
+    };
+    let offset = (pagination_params.page.unwrap_or(1) - 1) * limit;
```
ShubhranshuSanjeev, line 574: These needed to be changed too.

ShubhranshuSanjeev (approved):

PR #210, opened 2024-08-23
feat: Add get_applicable_variants as expt endpoint

ayushjain17:
## Problem
get_applicable_variants only exist as a functionality inside client

## Solution
Add get_applicable_variants as  an endpoint in experimentation

## API changes
| /experiments/applicable_variants      | POST | {context: Value, toss: i8} | Vec<Variants> |

Code duplication

### DIFF
--- a/crates/experimentation_platform/src/api/experiments/handlers.rs
+++ b/crates/experimentation_platform/src/api/experiments/handlers.rs
@@ -27,8 +27,8 @@ use superposition_types::{
 use super::{
     helpers::{
         add_variant_dimension_to_ctx, check_variant_types,
-        check_variants_override_coverage, extract_override_keys, validate_experiment,
-        validate_override_keys,
+        check_variants_override_coverage, decide_variant, extract_override_keys,
+        validate_experiment, validate_override_keys,
     },
     types::{
         AuditQueryFilters, ConcludeExperimentRequest, ContextAction, ContextBulkResponse,
@@ -39,8 +39,11 @@ use super::{
 };
 
 use crate::{
-    db::models::{EventLog, Experiment, ExperimentStatusType},
-    db::schema::{event_log::dsl as event_log, experiments::dsl as experiments},
+    api::experiments::types::ApplicableVariantsQuery,
+    db::{
+        models::{EventLog, Experiment, ExperimentStatusType},
+        schema::{event_log::dsl as event_log, experiments::dsl as experiments},
+    },
 };
 
 use serde_json::{json, Map, Value};
@@ -51,6 +54,7 @@ pub fn endpoints(scope: Scope) -> Scope {
         .service(create)
         .service(conclude_handler)
         .service(list_experiments)
+        .service(get_applicable_variants)
         .service(get_experiment_handler)
         .service(ramp)
         .service(update_overrides)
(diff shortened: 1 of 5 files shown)


crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -442,6 +446,85 @@ pub async fn conclude(
...
+            unexpected_error!("Something went wrong.")
+        })? {
+            variants.push(v)
+        }
+    }
+
+    Ok(HttpResponse::Ok().json(variants))
+}
+
+fn decide_variant(
```
mahatoankitkumar, line 500: move to helper.rs

mahatoankitkumar (changes requested):

mahatoankitkumar (approved):

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -442,6 +446,59 @@ pub async fn conclude(
...
+
+    let experiments = experiment_list.filter_map(|exp| {
+        let is_empty = exp
+            .context
+            .as_object()
+            .map_or(false, |context| context.is_empty());
+        if is_empty {
+            Some(exp.clone())
+        } else {
+            match jsonlogic::partial_apply(
```
pratikmishra356, line 470: why partial apply ?

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -129,6 +135,68 @@ pub enum ContextBulkResponse {
...
+            .remove("toss")
+            .and_then(|toss| toss.as_i64())
+            .and_then(|toss| {
+                if -1 <= toss && toss <= 100 {
+                    Some(toss as i8)
+                } else {
+                    None
+                }
+            })
+            .ok_or_else(|| {
```
pratikmishra356, line 158: shouldn't toss be optional and default should be -1 ?

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -129,6 +135,68 @@ pub enum ContextBulkResponse {
...
+            })?;
+
+        Ok(Self {
+            toss,
+            context: value,
+        })
+    }
+}
+
+impl FromRequest for ApplicableVariantsQuery {
```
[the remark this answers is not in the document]
pratikmishra356: We can use try_from in the handler , and expect query param as HashMap<String,String>

crates/experimentation_platform/src/api/experiments/types.rs
ayushjain17: get_applicable_variant is based on toss itself, toss is mandatory here

ayushjain17: > shouldn't toss be optional and default should be -1 ?

get_applicable_variant is based on toss itself, toss is mandatory here

crates/experimentation_platform/src/api/experiments/types.rs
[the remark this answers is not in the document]
ayushjain17: doing it manually, it becomes quite flaky and something which is like the dev has to separately ensure the safety of the type and not the system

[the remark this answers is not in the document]
pratikmishra356: But then everyone will add from_request impl for all the types

pratikmishra356 (approved):

PR #213, opened 2024-08-26
feat: delete dimension api and ui

pratikmishra356:
## Problem
Implement delete dimension api

## Solution
Implemented delete dimension api and ui , restricted to delete variantIds and also dimensions which are used in any contexts

NA

## Pre-deployment activity
NA

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/dimension/handlers.rs
+++ b/crates/context_aware_config/src/api/dimension/handlers.rs
@@ -1,26 +1,34 @@
 use crate::{
-    api::dimension::types::CreateReq,
-    db::{models::Dimension, schema::dimensions::dsl::*},
+    api::dimension::{types::CreateReq, utils::get_dimension_usage_context_ids},
+    db::{
+        models::Dimension,
+        schema::{dimensions, dimensions::dsl::*},
+    },
     helpers::validate_jsonschema,
 };
 use actix_web::{
-    get, put,
-    web::{self, Data},
+    delete, get, put,
+    web::{self, Data, Path},
     HttpResponse, Scope,
 };
 use chrono::Utc;
-use diesel::RunQueryDsl;
+use diesel::{
+    delete, Connection, ExpressionMethods, QueryDsl, RunQueryDsl, SelectableHelper,
+};
 use jsonschema::{Draft, JSONSchema};
 use serde_json::Value;
-use superposition_macros::{bad_argument, unexpected_error};
+use superposition_macros::{bad_argument, not_found, unexpected_error};
 use superposition_types::{result as superposition, SuperpositionUser, User};
 
 use service_utils::service::types::{AppState, DbConnection, Tenant};
 
-use super::types::DimensionWithMandatory;
+use super::types::{DeleteReq, DimensionWithMandatory};
 
 pub fn endpoints() -> Scope {
-    Scope::new("").service(create).service(get)
+    Scope::new("")
+        .service(create)
+        .service(get)
+        .service(delete_dimension)
 }
 
 #[put("")]
@@ -130,4 +138,47 @@ async fn get(
         .collect();
 
     Ok(HttpResponse::Ok().json(dimensions_with_mandatory))
+}
+
+#[delete("/{name}")]
+async fn delete_dimension(
+    path: Path<DeleteReq>,
+    user: User,
+    db_conn: DbConnection,
+) -> superposition::Result<HttpResponse> {
+    let name: String = path.into_inner().into();
+    let DbConnection(mut conn) = db_conn;
+    dimensions::dsl::dimensions
+        .filter(dimensions::dimension.eq(&name))
+        .select(Dimension::as_select())
+        .get_result(&mut conn)?;
+    let context_ids = get_dimension_usage_context_ids(&name, &mut conn)
+        .map_err(|_| unexpected_error!("Something went wrong"))?;
+    if context_ids.is_empty() {
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            use dimensions::dsl;
+            diesel::update(dsl::dimensions)
+                .filter(dsl::dimension.eq(&name))
+                .set((
+                    dsl::last_modified_at.eq(Utc::now().naive_utc()),
+                    dsl::last_modified_by.eq(user.get_email()),
+                ))
+                .execute(transaction_conn)?;
+            let deleted_row = delete(dsl::dimensions.filter(dsl::dimension.eq(&name)))
+                .execute(transaction_conn);
+            match deleted_row {
+                Ok(0) => Err(not_found!("Dimension `{}` doesn't exists", name)),
+                Ok(_) => Ok(HttpResponse::NoContent().finish()),
+                Err(e) => {
+                    log::error!("dimension delete query failed with error: {e}");
+                    Err(unexpected_error!("Something went wrong."))
+                }
+            }
+        })
+    } else {
+        Err(bad_argument!(
+            "Given key already in use in contexts: {}",
+            context_ids.join(",")
+        ))
+    }
 }
--- a/crates/context_aware_config/src/api/dimension/types.rs
+++ b/crates/context_aware_config/src/api/dimension/types.rs
@@ -90,3 +90,19 @@ impl DimensionWithMandatory {
         }
     }
 }
+
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "String")]
+pub struct DeleteReq(String);
+
+impl TryFrom<String> for DeleteReq {
+    type Error = String;
+    fn try_from(name: String) -> Result<Self, Self::Error> {
+        let name = name.trim();
+        if name == "variantIds" {
+            Err("variantIds cannot be deleted".to_string())
+        } else {
+            Ok(Self(name.to_owned()))
+        }
+    }
+}
--- a/crates/context_aware_config/src/api/dimension/utils.rs
+++ b/crates/context_aware_config/src/api/dimension/utils.rs
@@ -1,13 +1,17 @@
-use std::collections::HashMap;
-
-use crate::db::{models::Dimension, schema::dimensions::dsl::*};
-use diesel::RunQueryDsl;
+use crate::db::{
+    models::{Context, Dimension},
+    schema::{contexts::dsl::contexts, dimensions::dsl::*},
+};
 use diesel::{
     r2d2::{ConnectionManager, PooledConnection},
-    PgConnection,
+    PgConnection, RunQueryDsl,
 };
 use jsonschema::{Draft, JSONSchema};
-use superposition_types::result as superposition;
+use serde_json::Map;
+use service_utils::helpers::extract_dimensions;
+use std::collections::HashMap;
+use superposition_macros::{db_error, unexpected_error};
+use superposition_types::{result as superposition, Cac, Condition};
 
 pub fn get_all_dimension_schema_map(
     conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
@@ -27,4 +31,31 @@ pub fn get_all_dimension_schema_map(
         .collect();
 
     Ok(dimension_schema_map)
+}
+
+pub fn get_dimension_usage_context_ids(
+    key: &str,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+) -> superposition::Result<Vec<String>> {
+    let result: Vec<Context> = contexts.load(conn).map_err(|err| {
+        log::error!("failed to fetch contexts with error: {}", err);
+        db_error!(err)
+    })?;
+
+    let mut context_ids = vec![];
+    for context in result.iter() {
+        let condition = Cac::<Condition>::try_from_db(
+            context.value.as_object().unwrap_or(&Map::new()).clone(),
+        )
+        .map_err(|err| {
+            log::error!("generate_cac : failed to decode context from db {}", err);
+            unexpected_error!(err)
+        })?
+        .into_inner();
+
+        extract_dimensions(&condition)?
+            .get(key)
+            .map(|_| context_ids.push(context.id.to_owned()));
+    }
+    Ok(context_ids)
 }
--- a/crates/frontend/src/pages/dimensions.rs
+++ b/crates/frontend/src/pages/dimensions.rs
@@ -2,13 +2,14 @@ use crate::components::dimension_form::DimensionForm;
 use crate::components::drawer::{close_drawer, open_drawer, Drawer, DrawerBtn};
 use crate::components::skeleton::Skeleton;
 use crate::components::{
+    delete_modal::DeleteModal,
     stat::Stat,
     table::{types::Column, Table},
 };
 use leptos::*;
 use serde_json::{json, Map, Value};
 
-use crate::api::fetch_dimensions;
+use crate::api::{delete_dimension, fetch_dimensions};
 
 #[derive(Clone, Debug, Default)]
 pub struct RowData {
@@ -22,6 +23,8 @@ pub struct RowData {
 #[component]
 pub fn dimensions() -> impl IntoView {
     let tenant_rs = use_context::<ReadSignal<String>>().unwrap();
+    let (delete_modal_visible_rs, delete_modal_visible_ws) = create_signal(false);
+    let (delete_id_rs, delete_id_ws) = create_signal::<Option<String>>(None);
     let dimensions_resource = create_blocking_resource(
         move || tenant_rs.get(),
         |current_tenant| async move {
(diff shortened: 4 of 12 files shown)


crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -131,3 +144,57 @@ async fn get(
...
+
+#[delete("/{name}")]
+async fn delete_dimension(
+    path: Path<String>,
+    user: User,
+    mut db_conn: DbConnection,
+) -> superposition::Result<HttpResponse> {
+    let name = path.into_inner();
+    fetch_dimension(&name, &mut db_conn)?;
+    if name == "variantIds" {
```
mahatoankitkumar, line 184: Can we make the path params a type and have this check in that type?

mahatoankitkumar (changes requested):

crates/frontend/src/pages/dimensions.rs
```diff
@@ -64,13 +65,26 @@ pub fn dimensions() -> impl IntoView {
...
+                    }
+                });
+            };
 
             view! {
-                <span class="cursor-pointer" on:click=edit_click_handler>
-                    {edit_icon}
-                </span>
+                <div class="join">
+                    <span class="cursor-pointer" on:click=edit_click_handler.clone()>
```
ayushjain17, line 81: why this extra cloning of the function?
this was not there earlier

pratikmishra356: removed , it got added by mistake

crates/frontend/src/pages/dimensions.rs
```diff
@@ -50,6 +50,7 @@ pub fn dimensions() -> impl IntoView {
                 _ => Some(json!(function_name.replace('"', ""))),
             };
             let mandatory = row["mandatory"].as_bool().unwrap_or(false);
+            let dimesnion_name: StoredValue<_> = StoredValue::new(row_dimension.clone());
```
mahatoankitkumar, line 53: Why do we need this? You can use `row_dimension` directly right?

mahatoankitkumar (changes requested):

crates/frontend/src/pages/dimensions.rs
pratikmishra356: No, you cant use row_dimension directly since the value gets moved inside the delete handler

mahatoankitkumar: Then why use `StoredValue ` can directly clone and use right?

mahatoankitkumar: And add delete modal for the delete confirmation, the component is already there.

pratikmishra356: Done

crates/frontend/src/pages/dimensions.rs
```diff
@@ -50,6 +73,7 @@ pub fn dimensions() -> impl IntoView {
                 _ => Some(json!(function_name.replace('"', ""))),
             };
             let mandatory = row["mandatory"].as_bool().unwrap_or(false);
+            let dimesnion_name = row_dimension.clone();
```
mahatoankitkumar, line 76:
```suggestion
            let dimension_name = row_dimension.clone();
```
Would be better if this can be done.

mahatoankitkumar (changes requested):
And it shows Something went wrong instead of the actual error when trying to delete variantIds dimension through the UI.

crates/frontend/src/pages/dimensions.rs
pratikmishra356: Yes , separated client_error logic and server_error logic

mahatoankitkumar (approved):

crates/frontend/src/pages/dimensions.rs
```diff
@@ -173,7 +204,13 @@ pub fn dimensions() -> impl IntoView {
                         </div>
                     }
                 }}
-
+            <DeleteModal
+                modal_visible=delete_modal_visible
```
ayushjain17, line 208:
```suggestion
                modal_visible=delete_id.is_some()
```

crates/frontend/src/pages/dimensions.rs
```diff
@@ -173,7 +204,13 @@ pub fn dimensions() -> impl IntoView {
...
+                confirm_delete=confirm_delete
+                set_modal_visible=set_delete_modal_visible
```
ayushjain17, line 210: ideally this should have been an on_close prop, because a modal can never set itself as visible, it can only close itself, and the prop should be called in all those places where the Modal wants to trigger a close action, and then the on_close prop passed from here should do whatever is needed with respect to on_close action.

The caller should not have to rely on what is being sent from the modal to this function

crates/frontend/src/pages/dimensions.rs
```diff
@@ -32,10 +35,30 @@ pub fn dimensions() -> impl IntoView {
         },
     );
 
+    let confirm_delete = Callback::new(move |_| {
+        if let Some(id) = delete_id.get().clone() {
+            spawn_local(async move {
```
ayushjain17, line 40: when an api call is going on, shouldn't the interactions be totally blocked here ?
if it fails or succeeds the user will totally lose context, they could have to some other section assuming that the operation (api call) happened instantaneously and that too it was a success

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -28,3 +32,42 @@ pub fn get_all_dimension_schema_map(
...
+        let condition = Cac::<Condition>::try_from_db(
+            context.value.as_object().unwrap_or(&Map::new()).clone(),
+        )
+        .map_err(|err| {
+            log::error!("generate_cac : failed to decode context from db {}", err);
+            unexpected_error!(err)
+        })?
+        .into_inner();
+
+        let dimension_map = extract_dimensions(&condition)?;
```
ayushjain17, line 56: This should be enough here, i guess
```suggestion
        extract_dimensions(&condition)?
            .get(key)
            .map(|_| context_ids.push(context.id.to_owned()));
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -131,3 +144,54 @@ async fn get(
...
+}
+
+#[delete("/{name}")]
+async fn delete_dimension(
+    path: Path<DeleteReq>,
+    user: User,
+    mut db_conn: DbConnection,
+) -> superposition::Result<HttpResponse> {
+    let name = path.into_inner().into();
+    fetch_dimension(&name, &mut db_conn)?;
```
ayushjain17, line 183: lets return 204 when its Not Found over here, for other errors, we can still return back the error from here

crates/frontend/src/pages/dimensions.rs
```diff
@@ -22,6 +23,8 @@ pub struct RowData {
 #[component]
 pub fn dimensions() -> impl IntoView {
     let tenant_rs = use_context::<ReadSignal<String>>().unwrap();
+    let (delete_modal_visible, set_delete_modal_visible) = create_signal(false);
```
ayushjain17, line 26: can be optimised to one single signal, but DeleteDimensions modal bad structure seem to be a blocker for these changes

ayushjain17 (dismissed):

crates/context_aware_config/src/api/dimension/handlers.rs
mahatoankitkumar: Why 204? Not found is fine i guess here?

pratikmishra356: @ayushjain17 its debatable , either way is fine 404 or 204
https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time

ayushjain17: @mahatoankitkumar 204 is No Content, which is the standard response on a successful delete,
so when someone is trying to delete an object which does not exist in the first place, services generally still return 204 (saying that the delete operation is done successfully) instead of throwing an error

though, we need to still return error response in case its not a Not Found error

mahatoankitkumar: @ayushjain17 As @pratikmishra356 said its debatable. There are more documentation in support of 404 in such cases
https://medium.com/@satokenta940/the-complete-guide-to-understanding-http-delete-requests-04a041822784#:~:text=404%20Not%20Found%3A%20The%20resource,or%20has%20already%20been%20deleted.
https://cloud.google.com/storage/docs/xml-api/delete-object#:~:text=If%20you%20make%20a%20DELETE,NoSuchKey%20in%20the%20Code%20element.

mahatoankitkumar: https://http.dev/delete

mahatoankitkumar: https://www.mscharhag.com/api-design/rest-deleting-resources

mahatoankitkumar: https://dev.to/cesareferrari/deleting-a-resource-via-http-delete-5hd9

crates/frontend/src/pages/dimensions.rs
ShubhranshuSanjeev: Agree

crates/frontend/src/pages/dimensions.rs
```diff
@@ -22,6 +23,8 @@ pub struct RowData {
...
+    let (delete_id, set_delete_id) = create_signal::<Option<String>>(None);
```
ShubhranshuSanjeev, line 27:
```suggestion
    let (delete_id_rs, delete_id_ws) = create_signal::<Option<String>>(None);
```

crates/frontend/src/utils.rs
```diff
@@ -459,14 +459,29 @@ where
         .map_err(|err| err.to_string())?;
 
     let status = response.status();
-    if status.is_client_error() || status.is_server_error() {
+    if status.is_client_error() {
```
ShubhranshuSanjeev, line 462: Why make this change?

crates/frontend/src/pages/dimensions.rs
```diff
@@ -64,13 +88,20 @@ pub fn dimensions() -> impl IntoView {
...
 
             view! {
-                <span class="cursor-pointer" on:click=edit_click_handler>
-                    {edit_icon}
-                </span>
+                <div class="join">
+                    <span class="cursor-pointer" on:click=edit_click_handler>
+                        <i class="ri-pencil-line ri-xl text-blue-500"></i>
+                    </span>
+                    <span class="cursor-pointer text-red-500" on:click=handle_dimension_delete>
```
ShubhranshuSanjeev, line 101: Should we hide the delete button and the edit button for `variantIds`? :thinking:

crates/context_aware_config/src/api/dimension/handlers.rs
ShubhranshuSanjeev: We should stick with 404 when the key is not found. We will keep this consistent across all apis.

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -131,3 +144,54 @@ async fn get(
...
+    let context_ids = get_dimension_usage_context_ids(&name, &mut db_conn)
+        .map_err(|_| unexpected_error!("Something went wrong"))?;
+    if context_ids.is_empty() {
+        db_conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
```
ShubhranshuSanjeev, line 187: Why a transaction, and why create helper function is it even an operation which is going to be used anywhere else?

ShubhranshuSanjeev (dismissed):

crates/frontend/src/utils.rs
pratikmishra356: https://github.com/juspay/superposition/pull/213#pullrequestreview-2265521287

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: It is in transaction because , we are doing first update and then delete
and making a different function is because , its better to have db query part as a separate function from the api handler

crates/frontend/src/pages/dimensions.rs
pratikmishra356: Hiden delete button , let this be editable , once can edit priority and add function

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: @pratikmishra356 why is it better to have the DB query as a separate function from the API handler?

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -131,3 +144,54 @@ async fn get(
 
     Ok(HttpResponse::Ok().json(dimensions_with_mandatory))
 }
+
+pub fn delete_dimension_helper(
```
Datron, line 148: Few things:
- why is this function public
- why is it required? This code can be in the handler itself

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -90,3 +90,24 @@ impl DimensionWithMandatory {
...
+impl DeleteReq {
+    pub fn validate_data(name: String) -> Result<Self, String> {
+        let name = name.trim();
+        if name == "variantIds" {
+            Err("variantIds cannot be deleted".to_string())
+        } else {
+            Ok(Self(name.to_owned()))
+        }
+    }
+}
```
Datron, line 106: Is this used only in try_from? Can this code be there itself? Do we need this impl?

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -28,3 +32,41 @@ pub fn get_all_dimension_schema_map(
...
+pub fn fetch_dimension(
+    key: &String,
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+) -> superposition::Result<Dimension> {
+    let res = dimensions::dsl::dimensions
+        .filter(dimensions::dimension.eq(key))
+        .select(Dimension::as_select())
+        .get_result(conn)?;
+    Ok(res)
+}
```
Datron, line 72: I think this is used in a single place, can we remove this function and just run the query directly in the handler?

postman/cac/Dimension/Delete Dimension/event.test.js
```diff
@@ -0,0 +1,3 @@
+pm.test("204 check", function () {
```
Datron, line 1: Can we test for 404 as well?

postman/cac/Dimension/Delete Dimension/response.json
```diff
@@ -0,0 +1 @@
+[]
```
Datron, line 1: remove this file

Datron (changes requested):
Frontend is fine, need to reduce the number of functions in backend

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: @Datron , this way it looks clean and better readability , easy debug, as handler function should have the business logic and complex queries can be under helper function,

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: this function need not to be public , i have removed

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -131,3 +141,58 @@ async fn get(
...
+        Ok(_) => {
+            log::info!("{name} dimension deleted by {}", user.get_email());
+            Ok(())
+        }
+        Err(e) => {
+            log::error!("dimension delete query failed with error: {e}");
+            Err(unexpected_error!("Something went wrong."))
+        }
+    }
+}
```
Datron, line 171: Can we get rid of this function as well? Merge it into the handler

crates/frontend/src/utils.rs
```diff
@@ -459,14 +459,29 @@ where
...
+            .await
+            .map_err(|_| String::from("Something went wrong"))?;
+        let error_msg = serde_json::from_slice::<ErrorResponse>(&resp_bytes).map_or(
+            String::from(
+                std::str::from_utf8(&resp_bytes.to_vec())
+                    .unwrap_or("Something went wrong"),
+            ),
+            |error| error.message,
+        );
+        logging::error!("{}", error_msg);
```
sauraww, line 474: Can you add description in the log here ?

crates/frontend/src/pages/dimensions.rs
```diff
@@ -32,10 +35,30 @@ pub fn dimensions() -> impl IntoView {
         },
     );
 
+    let confirm_delete = Callback::new(move |_| {
+        if let Some(id) = delete_id_rs.get().clone() {
```
sauraww, line 39: Why are we cloning here ?

sauraww (changes requested):

pratikmishra356:
Screenshot 2024-09-05 at 17 24 53
Screenshot 2024-09-05 at 17 25 08

sauraww (approved):

PR #214, opened 2024-08-27
chore: add example app

Datron:
## Problem
No example app for superposition

## Solution
bring in @knutties example app


srid:
@Datron CI should succeed now. But please first merge https://github.com/juspay/superposition/pull/206

examples/superposition-demo-app/README.md
```diff
@@ -0,0 +1,6 @@
+# Superposition Demo App setup instructions
+
+1. Compile and run the app
+```sh
+make superposition-example
+```
```
mahatoankitkumar, line 6: Shouldn't we have commands to run the sql file here and in the makefile?

mahatoankitkumar:
Just a small nitpick, if we can add an override with hour_of_the_day as condition.

knutties (approved):

mahatoankitkumar (approved):

examples/superposition-demo-app/src/templates/index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="/static/htmx.js"></script>
```
ayushjain17, line 4: why are we simply not using cdn provided by the packages, here ?
its anyways a network call over here, so why not fetch the original one, why make copies ?

knutties: You can work on the project without internet 😉 @ayushjain17

docker-compose/postgres/db_init.sql
ShubhranshuSanjeev, line 1: image
image

PR #216, opened 2024-08-28
fix: empty-override-prompt-text

pratikmishra356:
## Problem
In case of empty overrides list , page does not have any prompt text to hint user
 
## Solution
Added prompt text in case of empty page

## Pre-deployment activity

## Post-deployment activity

## API changes

Screenshot 2024-08-28 at 15 51 30

### DIFF
--- a/crates/frontend/src/pages/context_override.rs
+++ b/crates/frontend/src/pages/context_override.rs
@@ -335,20 +335,35 @@ pub fn context_override() -> impl IntoView {
                                 (context.clone(), overrides)
                             })
                             .collect::<Vec<(Context, Map<String, Value>)>>();
-                        ctx_n_overrides
-                            .into_iter()
-                            .map(|(context, overrides)| {
-                                view! {
-                                    <ContextCard
-                                        context=context
-                                        overrides=overrides
-                                        handle_edit=handle_context_edit
-                                        handle_clone=handle_context_clone
-                                        handle_delete=handle_context_delete
-                                    />
-                                }
-                            })
-                            .collect_view()
+                        if ctx_n_overrides.is_empty() {
+                            view! {
+                                <div class="flex-row" style="margin-top:20rem;">
+                                    <div class="flex justify-center text-gray-400">
+                                    <i class="ri-file-add-line ri-xl"></i>
+                                    </div>
+                                    <div class="flex mt-4 font-semibold items-center text-gray-400 text-xl justify-center">
+                                    "Start with creating an override"
+                                    </div>
+                                </div>
+
+                            }.into_view()
+                        } else {
+                            ctx_n_overrides
+                                .into_iter()
+                                .map(|(context, overrides)| {
+                                    view! {
+                                        <ContextCard
+                                            context=context
+                                            overrides=overrides
+                                            handle_edit=handle_context_edit
+                                            handle_clone=handle_context_clone
+                                            handle_delete=handle_context_delete
+                                        />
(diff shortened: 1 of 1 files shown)


crates/frontend/src/pages/context_override.rs
```diff
@@ -335,20 +335,30 @@ pub fn context_override() -> impl IntoView {
...
-                                        handle_edit=handle_context_edit
-                                        handle_clone=handle_context_clone
-                                        handle_delete=handle_context_delete
-                                    />
-                                }
-                            })
-                            .collect_view()
+                        if ctx_n_overrides.is_empty() {
+                            view! {
+                                <div class="flex font-semibold items-center justify-center">
```
Datron, line 340: @pratikmishra356 Can we check if this is required in any other screen?

Datron: Also can this be vertically centered?

pratikmishra356: I checked this is not required in other screen, 
I think it should not be vertically centre as one had to look down a bit, 
since overrides text and create button is on top it is better to be aligned towards top

Datron: You can make the text bigger, and then center it. May need an icon as well

crates/frontend/src/pages/context_override.rs
```diff
@@ -335,20 +335,30 @@ pub fn context_override() -> impl IntoView {
...
+                                "Overrides has not been created yet"
```
Datron, line 341:
```suggestion
                                "Overrides have not been created yet"
```

Datron (changes requested):
Small fixes

crates/frontend/src/pages/context_override.rs
pratikmishra356: sure checking

ShubhranshuSanjeev:
Can you change the commit message?

And we should the change message to `Start with creating an override`, the text color can be bit grayed out along with it an icon can be added to this.

Datron (approved):

crates/frontend/src/pages/context_override.rs
```diff
@@ -335,20 +335,35 @@ pub fn context_override() -> impl IntoView {
...
-                                        context=context
-                                        overrides=overrides
-                                        handle_edit=handle_context_edit
-                                        handle_clone=handle_context_clone
-                                        handle_delete=handle_context_delete
-                                    />
-                                }
-                            })
-                            .collect_view()
+                        if ctx_n_overrides.is_empty() {
```
ShubhranshuSanjeev, line 338: You can use `<Show>` component to achieve this.
And the else branch in not needed I guess, move the code out of the else branch.

ShubhranshuSanjeev (dismissed):
You can use <Show> component to achieve this.
And the else branch in not needed I guess, move the code out of the else branch.

mahatoankitkumar:
Should we also get this out in other pages like default_config, dimensions, functions. Instead of showing empty tables?

pratikmishra356:
@mahatoankitkumar other pages have different structure (tabular) so it is not needed there i guess

crates/frontend/src/pages/context_override.rs
Datron: @ShubhranshuSanjeev can we take this later? I think this is fine for now.

sauraww (approved):

PR #219, opened 2024-09-03
fix: python client

pratikmishra356:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/clients/python/expclient/client.py
+++ b/clients/python/expclient/client.py
@@ -1,6 +1,7 @@
 import ctypes
 import os
 import threading
+import ast
 
 platform = os.uname().sysname.lower()
 lib_path = os.environ.get("SUPERPOSITION_LIB_PATH")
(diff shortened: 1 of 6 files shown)


clients/python/main.py
```diff
@@ -1,22 +1,19 @@
 from expclient import ExperimentationClient
-from cacclient import CacClient 
+from cacclient import CacClient , MergeStrategy
 import http.server
 import socketserver
 
 try:
-    tenant_name = "dev"
+    tenant_name = "test"
```
mahatoankitkumar, line 7: Let it be `dev` for consistency around all clients

clients/python/main.py
```diff
@@ -1,22 +1,19 @@
...
     cac_client = CacClient(tenant_name, polling_frequency, cac_host_name)
 
-    exp_client.create_new_experimentation_client()
-    cac_client.create_new_cac_client()
-
     cac_client.start_cac_polling_update()
     exp_client.start_experimentation_polling_update()
 
-    PORT = 8002
+    PORT = 8003
```
mahatoankitkumar, line 16: Okay with this, but any reason to change this?

crates/cac_client/src/lib.rs
```diff
@@ -99,6 +99,9 @@ impl Client {
         let reqwc = clone_reqw(&reqw)?;
         let resp = reqwc.send().await.map_err_to_string()?;
         let last_modified_at = get_last_modified(&resp);
+        if resp.status().is_client_error() {
+            return Err("Invalid tenant given".to_string());
```
mahatoankitkumar, line 103:
```suggestion
            return Err("Invalid tenant".to_string());
```

mahatoankitkumar (dismissed):

clients/python/main.py
pratikmishra356: No, I will change

clients/python/expclient/client.py
```diff
@@ -15,6 +16,33 @@
...
+
+# x = AttrDict({'some': 'value'})
+# x['blah'] = 'hello world'
+# print(x.blah)  # hello world
+# print(x.some)  # value
+
+# # you can also assign value this way
+# x.foo = 'bar'
+# print(x['foo'])   # bar
+
```
Datron, line 45: remove comments

crates/experimentation_client/src/interface.rs
```diff
@@ -211,17 +211,20 @@ pub extern "C" fn expt_get_satisfied_experiments(
...
+                    .await
+                    .map(|exp| {
+                        rstring_to_cstring(serde_json::to_value(exp).unwrap().to_string())
+                            .into_raw()
+                    }),
+                std::ptr::null_mut()
+            )
+        }
+    })
 }
```
Datron, line 228: @pratikmishra356 we can't use EXP_RUNTIME for this, it needs localSet for this. Can we keep the same code, just unwrap safely and return that

crates/experimentation_client/src/interface.rs
```diff
@@ -250,30 +253,33 @@ pub extern "C" fn expt_get_filtered_satisfied_experiments(
...
```
Datron, line 270: Same as above

Datron (approved):

ShubhranshuSanjeev (approved):

PR #217, opened 2024-09-02
fix: Display unsupported operators in override UI and enhance Context Form

sauraww:
Enhance context form and display unsupported operators in override UI

## Problem
Describe the problem you are trying to solve here

## Solution
Provide a brief summary of your solution so that reviewers can understand your code

## Environment variable changes

What ENVs need to be added or changed

## Pre-deployment activity
Things needed to be done before deploying this change (if any)

## Post-deployment activity
Things needed to be done after deploying this change (if any)

## API changes
| Endpoint        | Method           | Request body  | Response Body |
| ------------- |:-------------:| -----:| ----------------:|
| API      | GET/POST, etc | request | response |

## Possible Issues in the future
Describe any possible issues that could occur because of this change

PR #230, opened 2024-09-07
fix: Revert "fix: display unsupported operators in overrides UI and e…

Datron:
…nhance context_form (#217)"

This reverts commit c615b77f0f9e67a7184974c2c5eddb7b6e8f6f45.


Datron:
This PR is not needed

PR #240, opened 2024-09-13
chore: Add tailwind extension settings to workspace settings for vscode

ayushjain17:
## Problem
Tailwind extension fails to show the value of the tailwind classes on hover

## Solution
Add tailwind extension settings to workspace settings for vscode

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,15 @@
+{
+  "tailwindCSS.classAttributes": [
+    "class",
+    "className",
+    "ngClass",
+    "[\\S]*[cC]lass[\\S]*"
+  ],
+  "files.associations": {
+    "*.rs": "rust"
+  },
+  "tailwindCSS.includeLanguages": {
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tailwindCSS.classAttributes": [
+    "class",
+    "className",
+    "ngClass",
+    "[\\S]*[cC]lass[\\S]*"
+  ],
```
Datron, line 7: These are settings for react and angular right? We don't need them I guess?

Datron (approved):

.vscode/settings.json
ayushjain17: `className` is used in React but is not restricted to it

Have added these patterns to allow tailwind to give details of class on hover when the name of the property follows either of the regex and not `class` only which is the default behavior

ShubhranshuSanjeev (approved):

PR #241, opened 2024-09-16
feat: Use Variant type in model

ayushjain17:
## Problem
using Value type in model.rs for Variants

## Solution
use proper type for Variants in models instead of serde_json::Value type

### DIFF
--- a/crates/experimentation_platform/src/db/models.rs
+++ b/crates/experimentation_platform/src/db/models.rs
@@ -2,10 +2,13 @@ use crate::db::schema::*;
 use chrono::{DateTime, NaiveDateTime, Utc};
 
 use diesel::{
-    query_builder::QueryId, Insertable, Queryable, QueryableByName, Selectable,
+    deserialize::FromSqlRow, expression::AsExpression, query_builder::QueryId,
+    sql_types::Json, Insertable, Queryable, QueryableByName, Selectable,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use superposition_derives::{JsonFromSql, JsonToSql};
+use superposition_types::{Exp, Overrides};
 
 #[derive(
     Debug,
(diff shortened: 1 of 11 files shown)


crates/superposition_proc_macros/src/lib.rs
```diff
@@ -0,0 +1,48 @@
+extern crate proc_macro;
```
Datron, line 1: We already have a macros crate right? Can we group all macros in one place?

Datron: @ayushjain17 should we use proc_macros here? Why not directly implement `ToSql` and `FromSql` traits on a struct?

If the argument is that we need this later, can we add it when we need it? I don't want to prematurely optimize

crates/experimentation_platform/src/db/models.rs
```diff
@@ -25,6 +28,39 @@ pub enum ExperimentStatusType {
...
+
+impl Variants {
+    pub fn new(data: Vec<Variant>) -> Self {
+        Self(data)
+    }
+
+    pub fn into_inner(self) -> Vec<Variant> {
+        self.0
+    }
+}
```
Datron, line 62: You can use strum::macros Deref to handle into_inner and you may not need the new function

Datron (changes requested):
Can we see if the proc_macros are required?

crates/superposition_proc_macros/src/lib.rs
ayushjain17: proc_macros can be in a crate which are used to write proc_macros only, you cannot export other macros from that crate, it stops compiling with that error itself

ayushjain17: > can we add it when we need it? I don't want to prematurely optimize

it is already added now, so it would basically be like, remove the optimisation which is already added and then add it later on again when we need it

and mostly, we forget to catch and reduce these redundancies later on when they occur and then we add more work to just add refactor to optimise these

when we can already solve it when we have already identified it as a future use-case

also, it mainly doesn't make sense to un-optimise something which has already been optimised

crates/experimentation_platform/src/db/models.rs
ayushjain17: using that adds more complexity as to add that you need to implement Copy trait, and then String does not implement Copy, so its not possible to add Deref for this

Datron: Can you check default_config API file? It has this as a struct
```
#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
#[serde(try_from = "String")]
pub struct DefaultConfigKey(String);
```

You can probably do this

crates/superposition_proc_macros/src/lib.rs
Datron: > it is already added now, so it would basically be like, remove the optimisation which is already added and then add it later on again when we need it

Its an unoptimization right now, given the added compile time for macros. It's not used anywhere else. Why do we need it as a proc_macro instead of a direct trait implementation?

crates/experimentation_platform/src/db/models.rs
Datron: Also for tenant struct
```
#[derive(Deref, DerefMut, Clone, Debug)]
pub struct Tenant(pub String);
```

crates/superposition_proc_macros/src/lib.rs
ShubhranshuSanjeev: @Datron , I think the intention here is to put up an easy way for other people who will need to do something similar in future. @ayushjain17 has already figured out the solution to this kind of problem and put it in place, this will helpful.

And for the compile time argument, yes this will increase, but our project already uses a lot of derive macros, the delta here will be minimal, we won't be anyway writing that many macros.

sauraww: image
These are the results of profiling . The above is for main which takes 81.7 seconds 
This one is from ayush's branch , it takes 80.4 seconds : 
image
Now , we can decide what to go forward with.

Datron (approved):

PR #242, opened 2024-09-17
feat: custom query-param extractor for platform queries

ShubhranshuSanjeev:
## Problem
In current implementation for filtering config/other resources using query params, variables such as   
- config version (version), merge strategy(merge_strategy), etc that are platform specific as passed along with the 
- filters for dimensions  (<dimension_name>=<dimension_value>)
Since dimension names are user-defined there are chances of conflict between dimension names and platform specific params.

## Solution
Namespacing platform specific params with prefixes such as `platform` can be one of the way to resolve the conflict between platform params and dimension names.
There are multiple ways of namespacing query params with `platform`:
- `platform_<param name>`
- `platform.<param name>`
- `platform[<param name>]`

Since we allow use of `_` and `.` in names of dimensions and default config there will be possibility of conflict, which leaves us with `platform[<param name>]`.

## Implementation
Wrote a custom query extractor trait which can be implemented with your choice of namespace for query params. 
The user of trait has to provide:
- `regex` to match a specific pattern, example `platform\[(?<query_name>.*)\]`, 
- and a `capture_group` name, which will be `query_name` for the above regex example.

`regex` is used to identify the namespaced queries and the `capture_group` is used to get the actual query name. 

## Related work
- Handlers have to updated to use namespaced queries
- Support non-namespaced params for backward compatibility
- Changes in clients, to use namespaced query params for making request

### DIFF
--- a/crates/superposition_types/src/custom_query.rs
+++ b/crates/superposition_types/src/custom_query.rs
@@ -1,0 +1,88 @@
+use std::collections::HashMap;
+
+use actix_web::web::Query;
+use regex::Regex;
+use serde::de::DeserializeOwned;
+
+pub trait CustomQuery: Sized {
+    type Inner: DeserializeOwned;
+
+    fn regex_pattern() -> &'static str;
+    fn capture_group() -> &'static str;
+
+    fn query_regex() -> Regex {
+        Regex::new(Self::regex_pattern()).unwrap()
+    }
+
+    fn extract_query(
+        query_string: &str,
+    ) -> Result<Self, actix_web::error::QueryPayloadError> {
+        let query_map = Query::<HashMap<String, String>>::from_query(query_string)?;
+        let filtered_query = Self::filter_and_transform_query(query_map.into_inner());
+        let inner = Query::<Self::Inner>::from_query(&filtered_query)?.into_inner();
+        Ok(Self::new(inner))
+    }
+
+    fn filter_and_transform_query(query_map: HashMap<String, String>) -> String {
+        let regex = Self::query_regex();
+        query_map
+            .into_iter()
+            .filter_map(|(k, v)| {
+                Self::extract_key(&regex, &k).map(|pk| format!("{pk}={v}"))
+            })
+            .collect::<Vec<String>>()
+            .join("&")
+    }
+
+    fn extract_key(regex: &Regex, key: &str) -> Option<String> {
+        regex
+            .captures(key)
+            .and_then(|captures| captures.name(Self::capture_group()))
+            .map(|m| m.as_str().to_owned())
+    }
+
+    fn new(inner: Self::Inner) -> Self;
+    fn into_inner(self) -> Self::Inner;
+}
+
+#[derive(Debug, Clone)]
+pub struct PlatformQuery<T: DeserializeOwned>(pub T);
+
+impl<T> CustomQuery for PlatformQuery<T>
+where
+    T: DeserializeOwned,
+{
+    type Inner = T;
+
+    fn regex_pattern() -> &'static str {
+        r"platform\[(?<query_name>.*)\]"
+    }
+
+    fn capture_group() -> &'static str {
+        "query_name"
+    }
+
+    fn into_inner(self) -> T {
+        self.0
+    }
+
+    fn new(inner: Self::Inner) -> Self {
+        Self(inner)
+    }
+}
+
+impl<T> actix_web::FromRequest for PlatformQuery<T>
+where
+    T: DeserializeOwned,
+{
+    type Error = actix_web::Error;
+    type Future = std::future::Ready<Result<Self, Self::Error>>;
+
+    fn from_request(
+        req: &actix_web::HttpRequest,
+        _: &mut actix_web::dev::Payload,
+    ) -> Self::Future {
+        use std::future::ready;
+        ready(Self::extract_query(req.query_string()).map_err(actix_web::Error::from))
+    }
+}
--- a/crates/superposition_types/src/lib.rs
+++ b/crates/superposition_types/src/lib.rs
@@ -1,4 +1,5 @@
 #![deny(unused_crate_dependencies)]
+pub mod custom_query;
 #[cfg(feature = "result")]
 pub mod result;
 


Cargo.toml
```diff
@@ -15,7 +15,8 @@ members = [
     "examples/experimentation_client_integration_example",
     "examples/cac_client_integration_example",
     "examples/superposition-demo-app",
-    "crates/superposition_macros"
+    "crates/superposition_macros",
+    "crates/superposition_proc_macros"
```
Datron, line 19: Can we call this crate superposition_derives? That seems to be the conventional way

crates/service_utils/src/service/types.rs
```diff
@@ -256,3 +259,7 @@ impl FromRequest for CustomHeaders {
         ready(Ok(val))
     }
 }
+
+#[derive(Debug, Clone, CustomQuery)]
+#[custom_query(regex = r"platform\[(?<query_name>.*)\]", capture_group = "query_name")]
+pub struct PlatformQuery<T: DeserializeOwned>(pub T);
```
Datron, line 265: Should this be in superposition_types?

crates/superposition_proc_macros/src/lib.rs
```diff
@@ -0,0 +1,138 @@
+use proc_macro::TokenStream;
```
Datron, line 1: Why a proc_macro? Do we expect PlatformQuery like functionality elsewhere? Why not add PlatformQuery to superposition types and just implement the trait CustomQuery on it? The struct PlatformQuery can store it's pattern, capture group, etc in the struct itself

Datron (changes requested):

crates/superposition_proc_macros/src/lib.rs
ShubhranshuSanjeev: The aim was to prevent repetition of logic related to  namespacing query params with some other prefix as we discussed in the meeting. 

CustomQuery does solve the repetition part but there are other boilerplate code like implementing FromRequest, etc, which I thought since we are trying to make this easily usable why not solve this completely.

Proc macros as you can see in code makes it way easier for use.

Cargo.toml
ShubhranshuSanjeev: Sure

crates/superposition_proc_macros/src/lib.rs
Datron: Can we do it at a later point in time, when we need it?

crates/superposition_derives/src/lib.rs
```diff
@@ -0,0 +1,138 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::punctuated::Punctuated;
+use syn::{parse::Parse, parse_macro_input, DeriveInput, Error, LitStr, Meta, Token};
+
+#[proc_macro_derive(CustomQuery, attributes(custom_query))]
+pub fn derive_custom_query(input: TokenStream) -> TokenStream {
```
Datron, line 7: If we can avoid this proc macro, please try

Datron (approved):

crates/superposition_derives/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = &input.ident;
+    let generics = &input.generics;
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+
+    let custom_query_attr = match extract_attributes(&input.attrs) {
+        Ok(Some(attr)) => attr,
+        Ok(None) => {
+            return Error::new_spanned(&input, "Missing #[custom_query] attribute")
```
ayushjain17, line 16:
```suggestion
            return Error::new_spanned(&input, "`custom_query ` attributes are needed for deriving `CustomQuery ` on a structure.")
```

crates/superposition_types/src/custom_query.rs
```diff
@@ -0,0 +1,50 @@
...
+        let inner = Query::<Self::Inner>::from_query(&filtered_query)?.into_inner();
+        Ok(Self::from_inner(inner))
+    }
+
+    fn filter_and_transform_query(query_map: HashMap<String, String>) -> String {
+        let regex = Self::query_regex();
+        query_map
+            .into_iter()
+            .filter_map(|(k, v)| {
+                Self::extract_key(&regex, &k).map(|pk| format!("{pk}={v}"))
```
ayushjain17, line 32: just to verify, this is converting `platform[version]=version1` to `version=version1`, or am I missing here something ?

crates/superposition_types/Cargo.toml
```diff
@@ -17,6 +17,7 @@ diesel = { workspace = true, optional = true }
 anyhow = { workspace = true, optional = true }
 jsonlogic = { workspace = true }
 regex = { workspace = true }
+superposition_derives = { path="../superposition_derives" }
```
ayushjain17, line 20:
```suggestion
superposition_derives = { path = "../superposition_derives" }
```

crates/superposition_derives/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        impl #impl_generics #name #ty_generics #where_clause {
+            pub fn into_inner(self) -> #inner_type {
+                self.0
+            }
+        }
+
+        impl #impl_generics actix_web::FromRequest for #name #ty_generics #where_clause {
```
ayushjain17, line 55: shouldn't this be left on individual types to implement?
here we will be able to return `HashMap<String, String>` and the dimensions sent needs to converted in `Map<String, Value>` in most of our use cases

crates/superposition_derives/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+                        ..
+                    }) = nv.value
+                    {
+                        capture_group = Some(lit_str);
+                    }
+                }
+                _ => {
+                    return Err(syn::Error::new_spanned(
+                        attr,
+                        "Expected `regex` or `capture_group`",
```
ayushjain17, line 104: can we check if the compile time errors are co-relatable with `CustomQuery`

ayushjain17 (changes requested):

crates/superposition_types/src/custom_query.rs
ShubhranshuSanjeev: Yup, that's what happening

ayushjain17 (approved):

PR #243, opened 2024-09-17
fix: Frontend related issues and warnings

ayushjain17:
## Problem

1. `dlmalloc` failing in FE after upgrading to rust 1.78.0
2. unhandled warnings use of signals in non-reactive sections
3. content mismatch issue while using key grouping
4. range for ramp is allowed till 100

## Solution

1. Revert back to rust 1.76.0
2. convert `get` to `get_untracked` and other similar fixes (might not have solved all)
3. move the check of local storage value to a `create_effect` to avoid initial value mismatch between client and SSR
4. limit the max value of ramp based on the allowed max value

### DIFF
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -141,7 +141,7 @@ where
                     type="text"
                     placeholder="Key"
                     class="input input-bordered w-full max-w-md"
-                    value=config_key.get()
+                    value=config_key.get_untracked()
                     on:change=move |ev| {
                         let value = event_target_value(&ev);
                         set_config_key.set(value);
--- a/crates/frontend/src/pages/experiment_list.rs
+++ b/crates/frontend/src/pages/experiment_list.rs
@@ -42,7 +42,7 @@ pub fn experiment_list() -> impl IntoView {
     });
 
     let (reset_exp_form, set_exp_form) = create_signal(0);
-    let table_columns = create_memo(move |_| experiment_table_columns());
+    let table_columns = store_value(experiment_table_columns());
 
     let combined_resource: Resource<(String, ListFilters), CombinedResource> =
         create_blocking_resource(
@@ -177,7 +177,7 @@ pub fn experiment_list() -> impl IntoView {
                                                     cell_class="min-w-48 font-mono".to_string()
                                                     rows=data
                                                     key_column="id".to_string()
-                                                    columns=table_columns.get()
+                                                    columns=table_columns.get_value()
                                                     pagination=pagination_props
                                                 />
                                             </ConditionCollapseProvider>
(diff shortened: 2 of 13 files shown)


crates/frontend/src/pages/experiment_list.rs
```diff
@@ -177,7 +177,7 @@ pub fn experiment_list() -> impl IntoView {
                                                     cell_class="min-w-48 font-mono".to_string()
                                                     rows=data
                                                     key_column="id".to_string()
-                                                    columns=table_columns.get()
+                                                    columns=table_columns.get_untracked()
```
ShubhranshuSanjeev, line 180: @ayushjain17 , thinking we can get rid of create_memo for table_columns?

crates/frontend/src/pages/home.rs
```diff
@@ -303,8 +303,10 @@ pub fn home() -> impl IntoView {
         }
         // resolve the context and get the config that would apply
         spawn_local(async move {
-            let context = gen_query_context(context_rs.get());
-            let mut config = match resolve_config(tenant_rs.get(), context).await.unwrap()
+            let context = gen_query_context(context_rs.get_untracked());
```
ShubhranshuSanjeev, line 306: Why replace with `get_untracked` here? If `get` is not reactive here then move the `get` call to outside of async block. Let's keep the `get`.

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -141,7 +141,7 @@ where
                     type="text"
                     placeholder="Key"
                     class="input input-bordered w-full max-w-md"
-                    value=config_key.get()
+                    value=config_key.get_untracked()
```
Datron, line 144: Are we sure get_untracked is right here? Do we not want reactivity?

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.78.0"
+channel = "1.76.0"
```
Datron, line 2: Why is `dlmalloc` failing? We can debug that, rather than downgrading the rust version

Datron (changes requested):

crates/frontend/src/components/default_config_form.rs
ayushjain17: it was anyways not reactive earlier at this point, this just makes it explicit and avoids warnings

rust-toolchain.toml
ayushjain17: to fix this you need to upgrade wasm-bindgen to `2.92.0` or higher, and once you do that web-sys starts failing to compile and then we need to fix that then the chain continues

Datron (approved):

ShubhranshuSanjeev (approved):

PR #244, opened 2024-09-22
fix: case-insensitive dropdown search

ShubhranshuSanjeev:
## Problem
Case sensitive dropdown search in UI

## Solution
transforming search term and option labels to lowercase before matching


ayushjain17 (approved):

ayushjain17 (approved):
Can we also add auto focus on dropdown opening, shouldn't be a big change

Datron (approved):

PR #245, opened 2024-09-22
fix: using HtmlDisplay trait to show json in html

ShubhranshuSanjeev:
### Description
- Fixes display of JSON values across the frontend.
- Uses `html_display` to render json values
- Removes `trim` from current frontend code


crates/frontend/src/components/context_card.rs
```diff
@@ -31,9 +31,10 @@ pub fn context_card(
         .clone()
         .into_iter()
         .map(|(k, v)| {
-            let k = Value::String(k.trim_matches('"').to_string());
-            let v = Value::String(format!("{}", v).trim_matches('"').to_string());
-            Map::from_iter(vec![(String::from("KEY"), k), (String::from("VALUE"), v)])
+            Map::from_iter(vec![
+                (String::from("KEY"), Value::String(k.clone())),
+                (String::from("VALUE"), v),
+            ])
```
Datron, line 37: was this tested? Or can we remove this because of @sauraww 's change earlier?

Datron (changes requested):
Write a PR description, attach images of the change

crates/frontend/src/components/table.rs
```diff
@@ -80,9 +65,8 @@ pub fn table(
                                         .filter(|column| !column.hidden)
                                         .map(|column| {
                                             let cname = &column.name;
-                                            let value: String = generate_table_row_str(
-                                                row.get(cname).unwrap_or(&Value::String("".to_string())),
-                                            );
+                                            let value: String =
+                                                row.get(cname).unwrap_or(&Value::String(String::new())).html_display();
```
ayushjain17, line 69: what difference is `html_display` creating here ?

crates/frontend/src/components/experiment.rs
```diff
@@ -162,62 +187,52 @@ where
...
                             }
                             view
                         }}
 
                     </div>
                 </div>
-            </div> <div class="card bg-base-100 max-w-screen shadow m-5">
+            </div>
+            <div class="card bg-base-100 max-w-screen shadow m-5">
                 <div class="card-body">
```
ayushjain17, line 226: meanwhile can we also remove this redundant `div`

crates/frontend/src/components/experiment.rs
```diff
@@ -162,62 +187,52 @@ where
...
                     <h2 class="card-title">Variants</h2>
                     <div class="overflow-x-auto overflow-y-auto">
```
ayushjain17, line 228: this as well seems redundant

crates/frontend/src/components/experiment.rs
```diff
@@ -162,62 +187,52 @@ where
                             for token in contexts.clone() {
                                 let (dimension, values) = (token.left_operand, token.right_operand);
                                 let mut value_views = Vec::new();
-
                                 for value in values.iter() {
                                     if value.is_object() && value.get("var").is_some() {
                                         continue;
                                     }
-
                                     let value_str = match value {
```
ayushjain17, line 194: can this not be replaced by `html_display` ?

ayushjain17 (changes requested):
not complete review

crates/frontend/src/components/table.rs
ShubhranshuSanjeev: It's similar to earlier implementation. Just using `html_display` as a standard across the frontend.

crates/frontend/src/components/experiment.rs
ShubhranshuSanjeev: Good point, will do it

crates/frontend/src/components/experiment.rs
ShubhranshuSanjeev: Sure

crates/frontend/src/components/context_card.rs
ShubhranshuSanjeev: This is tested, I am just getting rid of trim logic, as `html_display` will take care of showing these values without quotes in frontend.

crates/frontend/src/components/table.rs
ShubhranshuSanjeev: This is similar to `generate_table_row_str`, just using html_display everywhere for uniformity.

crates/frontend/src/components/dropdown.rs
```diff
@@ -66,7 +68,16 @@ where
             class=("dropdown-top", dropdown_direction == DropdownDirection::Top)
             class=("dropdown-down", dropdown_direction == DropdownDirection::Down)
         >
-            <label tabindex="0" class=format!("{} {}", class, btn_class)>
+            <label
+                tabindex="0"
+                class=format!("{} {}", class, btn_class)
+                on:click:undelegated=move |_| {
```
ayushjain17, line 74: also, how was this supposed to work ?

crates/frontend/src/components/dropdown.rs
```diff
@@ -86,7 +97,8 @@ where
                                     type="text"
                                     class="grow"
                                     placeholder="Search"
-                                    name=name
+                                    ref_=node_ref
```
ayushjain17, line 100: wont this suffice ?
```suggestion
                                    autofocus={true}
```

ayushjain17 (changes requested):

crates/frontend/src/components/dropdown.rs
ShubhranshuSanjeev: That wasn't working, so went ahead with this approach.

crates/frontend/src/components/dropdown.rs
ShubhranshuSanjeev: Didn't get your question

crates/frontend/src/components/experiment.rs
```diff
@@ -24,136 +55,131 @@ where
     HC: Fn() + 'static + Clone,
     HE: Fn() + 'static + Clone,
 {
-    let experiment_rc = Rc::new(experiment.clone());
-    let contexts = Rc::clone(&experiment_rc).context.clone();
+    let experiment = store_value(experiment);
+    let contexts = experiment.with_value(|v| v.context.clone());
+    let badge_class = match experiment.with_value(|v| v.status) {
+        ExperimentStatusType::CREATED => "badge text-white ml-3 mb-1 badge-xl badge-info",
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: Sure

crates/frontend/src/components/dropdown.rs
ayushjain17: i was asking how was this change helping in solving the autofocus issue, couldn't understand the logic by looking at the code

ShubhranshuSanjeev: To further add on this, the component library we are using keeps the dropdown mounted on the DOM, so autofocus do focus on the input box but since we make other interactions that focus is lost.

ShubhranshuSanjeev: Other way of doing this can be taking over the logic to display dropdown using signals, but I don't want to manually manage close for any click outside of dropdown.

ayushjain17: > keeps the dropdown mounted on the DOM

why is it always mounted ?
shouldn't we just unmount it when not in use
anyways, the behaviour which we simply wanted is the moment it becomes visible (gets mounted in ideal case), the input component should get focused

ShubhranshuSanjeev: > shouldn't we just unmount it when not in use

Ideally yes.

> why is it always mounted ?

That's how daisy UI does it.
It just changes the visibility to hidden and opacity to 0.

ayushjain17 (approved):

PR #246, opened 2024-09-23
feat: Tenant specific config support via .cac.toml

ayushjain17:
## Problem
Tenant specific configuration not present

## Solution
Get tenant-config via cac using `cac_toml` which reads `Superposition.cac.toml`


crates/cac_toml/Cargo.toml
```diff
@@ -17,4 +17,4 @@ path = "src/bin.rs"
 clap = { version = "4.3.0", features = ["derive"] }
 pest = "2.6"
 pest_derive = "2.6"
-toml = { version = "0.8.8", features = ["preserve_order"] }
+toml = { workspace = true }
```
knutties, line 20: Will this pick-up the `preserve_order` if someone just includes `cac_toml` outside of this workspace?  Could we validate that?

ayushjain17: yes this will set the features for the crate, in the same way it is specified in the workspace `Cargo.toml`
have verified this

crates/superposition_types/src/lib.rs
```diff
@@ -257,6 +257,46 @@ impl Display for RegexEnum {
     }
 }
 
+pub trait SuperpositionTenantConfig {
```
Datron, line 260: This could be a struct with pub members right? Why define tenant config as a trait behaviour? No one can implement this elsewhere and pass it around. Why are we making it generic? Note: If the answer summarizes to "We will need this in the future", my response is "lets add it in the future". This is unecessary code complexity right now

crates/superposition/src/main.rs
```diff
@@ -64,69 +48,16 @@ async fn main() -> Result<()> {
         prefix => "/".to_owned() + prefix,
     };
 
-    let cac_host: String = get_from_env_unsafe("CAC_HOST").expect("CAC host is not set");
     let cac_port: u16 = get_from_env_unsafe("PORT").unwrap_or(8080);
```
Datron, line 51: Nice, we needed this

Datron (changes requested):
Let me know if we can avoid the type ceremony regarding TenantConfig being a trait. Else this is fine

crates/superposition_types/src/lib.rs
Datron: I'm guessing this is because of internal asks for auth? The tenant config struct internal to Juspay will be different I'm assuming

Datron (approved):

Datron (approved):

crates/superposition/src/main.rs
```diff
@@ -153,51 +89,39 @@ async fn main() -> Result<()> {
...
         let leptos_options = &conf.leptos_options;
         let site_root = &leptos_options.site_root;
         let leptos_envs = ui_envs.clone();
-        let cac_host = cac_host.to_owned() + base.as_str();
         App::new()
             .wrap(Compress::default())
+            .app_data(app_state.clone())
             .wrap_fn(|req, srv| {
-                let user = User::default();
+                let state = req.app_data::<Data<AppState>>().unwrap();
```
ShubhranshuSanjeev, line 103: @ayushjain17 , can you explain why this change is needed?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -234,30 +244,48 @@ async fn create(
     // creating variants' context in CAC
     let http_client = reqwest::Client::new();
     let url = state.cac_host.clone() + "/context/bulk-operations";
-    let headers_map = construct_header_map(tenant.as_str(), custom_headers.config_tags)?;
+    let user_str = serde_json::to_string(&user).map_err(|err| {
```
ShubhranshuSanjeev, line 247: What are we trying to achieve with `x-user` header?

ShubhranshuSanjeev (changes requested):

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: Someone who has access for only making changes in experimentation and not in context-aware-config, for them all requests of experimenation would start failing as the when these api calls would be made to make changes in context-aware-config, the authorization would fail due to lack of access.

So Internal authorization has been added
Now since this Internal auth would not have the details of the current user, and on validating it would give the details of a super-user then the audit logs would show wrong info regarding the maker of the contexts in context-aware-config, so in this case since experimentation has already decoded the user, it just forwards the user data 

And this user data is only to be consumed if the Internal auth check is successful

ShubhranshuSanjeev (approved):

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -794,29 +841,47 @@ async fn update_overrides(
...
+    let headers_map = construct_header_map(tenant.as_str(), extra_headers)?;
 
     let response = http_client
         .put(&url)
         .headers(headers_map.into())
         .header(
-            "Authorization",
-            format!("{} {}", user.get_auth_type(), user.get_auth_token()),
+            header::AUTHORIZATION,
+            format!("Internal {}", state.superposition_token),
```
Datron, line 866: What is internal auth?

Datron (changes requested):
want to understand Authorization internal before approving

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: This is not a supported authentication scheme
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes

ShubhranshuSanjeev (changes requested):
Can we restrict the internal token access to be only used by experimentation platform and also that this token should not have the full access, just access to CRUD of contexts/overrides will work?

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -234,30 +244,48 @@ async fn create(
...
+        .into_iter()
+        .filter_map(|item| match item {
+            ContextBulkResponse::PUT(context) => Some(context),
+            ContextBulkResponse::DELETE(_) => None,
+            _ => {
+                log::error!("Unexpected response item: {:?}", item);
+                None
+            }
+        })
+        .collect::<Vec<_>>();
```
Datron, line 288: nitpick: specify types here, makes it easy to read on a glance

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -371,6 +399,16 @@ pub async fn conclude(
                 };
                 operations.push(ContextAction::MOVE((context_id, context_move_req)));
             } else {
+                let user_str = serde_json::to_string(&user).map_err(|err| {
```
Datron, line 402: Do we want to send the whole user object in the headers or only the email address?

Datron (approved):
Minor nitpicks and optimizations, not a PR blocker

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: Whole user, as what all data would be needed later on the context api or which api being called would not be known

ShubhranshuSanjeev (approved):

PR #247, opened 2024-09-24
Dependency pruning

ayushjain17:
## Problem

1. `make run` compiles code twice
2. Redundant dependencies across all crates
3. Single use dependency specified at workspace `Cargo.toml`
4. `superposition_types` uses `diesel` as dependency, which causes dependency issues in 3rd party repos using this crate
5. dependencies in random order
6. `cac_client` returning data of types `superposition_types::result::Result<T>` in cases where the return should neither be a `Result` type or can simply be a `Result<T, String>`

## Solution

1. Remove redundant compile step
2. Remove unused dependencies, mandate `#![deny(unused_crate_dependencies)]` for all crates
3. Localise single use dependency to the crates instead of workspace `Cargo.toml`
4. make `result` type as a feature for `superposition_types` (make `diesel` optional)
5. sorted dependencies in alphabetical order
6. Update return type of `cac_client` functions to direct `T` or `Result<T, String>`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,30 +3,6 @@
 version = 3
 
 [[package]]
-name = "actix"
-version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f728064aca1c318585bf4bb04ffcfac9e75e508ab4e8b1bd9ba5dfe04e2cbed5"
-dependencies = [
- "actix-rt",
- "actix_derive",
- "bitflags 1.3.2",
- "bytes",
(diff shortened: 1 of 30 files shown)


Datron:
@ayushjain17 can we separate out dependency pruning and tenant config? That would make this PR way easier to review

ayushjain17:
> review

@Datron this PR is only for dependency pruning only, tenant config is just a pre-commit

If needed, you can review this PR, i will get the commit separated

Datron (approved):

PR #248, opened 2024-09-25
feat: get config-versions api

mahatoankitkumar:
## Problem
get config-versions api

## API changes
| {{host}}/config-versions     | GET | None | `{
    "total_pages": total_pages,
    "total_items": n_version,
    "data": config_versions
    }`


ayushjain17:
Can you please add the query param info as well in the description

crates/context_aware_config/src/api/config_versions/handlers.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub fn endpoints() -> Scope {
+    Scope::new("").service(get)
+}
+
+#[get("")]
+async fn get(
+    db_conn: DbConnection,
+    filters: Query<QueryFilters>,
+) -> superposition::Result<HttpResponse> {
```
sauraww, line 23: Instead of HttpResponse , Can you change it to a type as SnapshotResponse or something?

sauraww (changes requested):

crates/context_aware_config/src/api/config_versions/handlers.rs
mahatoankitkumar: Yeah can be done but the response now is 
`"total_pages": total_pages,
    "total_items": n_version,
    "data": config_versions`
    but as pagination is there for so many APIs
    Maybe something like this can be done
    `enum DataType<A, B, C> {
    TypeA(Vec<A>),
    TypeB(Vec<B>),
    TypeC(Vec<C>),
}

struct PaginatedResponse<A, B, C> {
    total_pages: i64,
    total_items: i64,
    data: DataType<A, B, C>,
}`

Opinions?

sauraww: Yeah , that looks a better generic solution to me , we can do that.

sauraww (changes requested):

crates/context_aware_config/src/api/config_versions/handlers.rs
ayushjain17: @mahatoankitkumar you don't need an enum for this
the PaginationResponse should be a generic type capable of returning any paginated response for any data type, an enum should be needed for it, as in case of Enum it would lead to unnecessary boilerplate code as well

simply doing this should be enough:
```
pub struct PaginatedResponse<T> {
      pub total_pages: i64,
      pub total_items: i64,
      data: T,
}
```

crates/context_aware_config/src/api/config_versions.rs
JamesGeorg, line 1: Can we sync one single name for this, frontend we are referring it as Snapshot here Config Version?

JamesGeorg (dismissed):

crates/context_aware_config/src/api/config_versions/handlers.rs
ShubhranshuSanjeev: One more addition to this, T should be deserializable, which can be done by `struct PaginatedResponse<T: DeserializeOwned>`.

crates/context_aware_config/src/api/config_versions.rs
mahatoankitkumar: Yeah @JamesGeorg we have decided to go with config-versions, as this is not a complete representation of what snapshot is.

crates/context_aware_config/src/types.rs
```diff
@@ -0,0 +1,14 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct QueryFilters {
+    pub count: Option<i64>,
+    pub page: Option<i64>,
+}
+
+#[derive(Serialize, Debug, Clone, Deserialize)]
+pub struct PaginatedResponse<T> {
```
ayushjain17, line 10: should we move this to `superposition_types` in some new file for such common utility types as it might be useful for experimentation as well for list experiments

crates/context_aware_config/src/api/config_versions/handlers.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub fn endpoints() -> Scope {
+    Scope::new("").service(get)
+}
+
+#[get("")]
+async fn get(
+    db_conn: DbConnection,
+    filters: Query<QueryFilters>,
+) -> superposition::Result<Json<PaginatedResponse<ConfigVersion>>> {
```
Datron, line 26: Do we need the Json wrapper struct? Can we implement something on PaginatedResponse<T> to avoid this?

crates/context_aware_config/src/types.rs
Datron: Agreed, this template can be common throughout the repo

Datron (approved):
Small comments, nothing that is a PR blocker

crates/superposition/src/main.rs
```diff
@@ -245,6 +245,11 @@ async fn main() -> Result<()> {
                             .wrap(AppExecutionScopeMiddlewareFactory::new(AppScope::CAC))
                             .service(type_templates::endpoints()),
                     )
+                    .service(
+                        scope("/config-versions")
```
Datron, line 249: we already have `/config`, should this be `/config/versions`?

Datron (changes requested):
Should we change the endpoint? which works better?

crates/superposition/src/main.rs
mahatoankitkumar: @Datron thats a call to be made, if we want that then better move this handler to config handlers.

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -656,3 +661,33 @@ async fn get_resolved_config(
...
+        .count()
+        .get_result(&mut conn)?;
+    let mut builder = config_versions::config_versions
+        .into_boxed()
+        .order(config_versions::created_at.desc());
+    if let Some(limit) = filters.count {
+        builder = builder.limit(limit);
+    }
+    if let Some(page) = filters.page {
+        let offset = (page - 1) * filters.count.unwrap_or(10);
```
pratikmishra356, line 682: @mahatoankitkumar can we handle the negative value for offset as it is not allowed in offset and giving 5xx
`OFFSET must not be negative`

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -656,3 +661,33 @@ async fn get_resolved_config(
...
+        builder = builder.offset(offset);
+    }
+    let limit = filters.count.unwrap_or(10);
+    let config_versions: Vec<ConfigVersion> = builder.load(&mut conn)?;
+    let total_pages = (n_version as f64 / limit as f64).ceil() as i64;
```
pratikmishra356, line 687: can we handle case for limit value as 0 also since we are dividing with it

mahatoankitkumar: Yeah sure will add a check that these should be > 0.

Datron (approved):

pratikmishra356 (approved):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -654,3 +657,33 @@ async fn get_resolved_config(
...
+        .into_boxed()
+        .order(config_versions::created_at.desc());
+    if let Some(limit) = filters.count {
+        builder = builder.limit(limit);
+    }
+    if let Some(page) = filters.page {
+        let offset = (page - 1) * filters.count.unwrap_or(10);
+        builder = builder.offset(offset);
+    }
+    let limit = filters.count.unwrap_or(10);
```
ShubhranshuSanjeev, line 681: This is wrong, why use a default count as 10 and calculating the number of pages after loading the full list of the config_versions.

PR #257, opened 2024-09-30
fix: JSONSchema generated forms for contexts

ShubhranshuSanjeev:
## Problem
Context forms were generated using lot of broken, clumsy pattern matching logic, parsing of values and generating final context json's logic was spread across utility functions.  

## Solution
Replaced pattern matching over string to guess the type of value with algebraic representation of JsonLogic.
Single Input component, which works seemlessly with the schema type of the dimensions and render appropriate fields along with handling parsing and input validation.


sauraww (changes requested):

crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,285 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Operand {
+    Value(serde_json::Value),
+    Dimension(serde_json::Value),
+}
+
+impl Operand {
+    pub fn from_operand_json(value: Value) -> Self {
+        match value {
+            Value::Object(ref o) if o.contains_key("var") => Operand::Dimension(value),
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: I want a reference.

crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,278 @@
...
+
+#[derive(Debug, Clone, Deref, DerefMut, Serialize, Deserialize)]
+pub struct Operands(pub Vec<Operand>);
+
+impl FromIterator<Value> for Operands {
+    fn from_iter<T: IntoIterator<Item = Value>>(iter: T) -> Self {
+        Operands(
+            iter.into_iter()
+                .map(Operand::from_operand_json)
+                .collect::<Vec<Operand>>(),
```
sauraww, line 33: Can we make a type for Vec<Operand> as Operands as we made for Conditions ?

crates/frontend/src/pages/context_override.rs
```diff
@@ -172,71 +167,78 @@ pub fn context_override() -> impl IntoView {
...
     });
 
-    let handle_context_edit =
-        Callback::new(move |data: (Context, Map<String, Value>)| {
-            let (context, overrides) = data;
-            let conditions = extract_conditions(&Value::Object(context.condition.into()));
+    let on_context_edit = Callback::new(move |data: (Context, Map<String, Value>)| {
+        let (context, overrides) = data;
+        let conditions =
+            Conditions::from_context_json(&context.condition.into()).unwrap();
```
sauraww, line 211: Can we avoid using plain unwrap here ?

crates/frontend/src/pages/context_override.rs
```diff
@@ -172,71 +167,78 @@ pub fn context_override() -> impl IntoView {
...
+    });
 
-    let handle_context_clone =
-        Callback::new(move |data: (Context, Map<String, Value>)| {
-            let (context, overrides) = data;
-            let conditions = extract_conditions(&Value::Object(context.condition.into()));
+    let on_context_clone = Callback::new(move |data: (Context, Map<String, Value>)| {
+        let (context, overrides) = data;
+        let conditions =
+            Conditions::from_context_json(&context.condition.into()).unwrap();
```
sauraww, line 225: Same here

crates/frontend/src/pages/context_override.rs
```diff
@@ -172,71 +167,78 @@ pub fn context_override() -> impl IntoView {
...
-            set_form_mode.set(Some(FormMode::Create));
+        selected_context_ws.set(Some(Data {
+            context: conditions,
+            overrides: overrides.into_iter().collect::<Vec<(String, Value)>>(),
+        }));
+        set_form_mode.set(Some(FormMode::Create));
 
-            open_drawer("context_and_override_drawer");
-        });
+        open_drawer("context_and_override_drawer");
```
sauraww, line 233: Can we create ENUMs for these strings , it's a little dangerous to keep it as strings

sauraww (changes requested):

crates/frontend/src/pages/context_override.rs
ShubhranshuSanjeev: Drawers are going away in my next PR, so this will be redundant work.

crates/frontend/src/logic.rs
ShubhranshuSanjeev: Didn't really understand the point.

crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+impl TryFrom<(&Operator, &str, &SchemaType)> for Operands {
+    type Error = String;
+    fn try_from(
+        (operator, d_name, r#type): (&Operator, &str, &SchemaType),
+    ) -> Result<Self, Self::Error> {
+        match operator {
+            Operator::Is => Ok(Operands(vec![
+                Operand::Dimension(json!({ "var": d_name })),
```
ShreyBana, line 45: Hmm, a little odd that dimension is also an operand. :thinking:

crates/frontend/src/pages/context_override.rs
```diff
@@ -172,71 +167,88 @@ pub fn context_override() -> impl IntoView {
...
+        match Conditions::from_context_json(&context.condition.into()) {
+            Ok(conditions) => {
+                selected_context_ws.set(Some(Data {
+                    context: conditions,
+                    overrides: overrides.into_iter().collect::<Vec<(String, Value)>>(),
+                }));
+                set_form_mode.set(Some(FormMode::Edit));
+                open_drawer("context_and_override_drawer");
+            }
+            Err(e) => {
```
ShreyBana, line 219: At this point in the program, if this call errs out, there's no useful error we can report to the user to un-block them. So maybe, it makes sense to use some kind of **front-end panic** which let's us write code as if there was no chance of an error to begin with. What do you think?

ShreyBana (approved):

crates/frontend/src/components/condition_pills.rs
```diff
@@ -164,11 +133,17 @@ pub fn condition(
                 .clone()>
                 {conditions
```
sauraww, line 134: Can we avoid multiple get_value calls and precompute it before hand : 
`let conditions_list = conditions.get_value(); `
We can call it afterwards like : 
```
conditions_list
                    .iter()
                    .enumerate()
                    .map(|(idx, condition)| {
                        let item_id = format!("{}-{}", id, idx);
                        view! {
                            <ConditionExpression
                                condition=condition.clone()
                                id=item_id
                                list_id=id.clone()
                            />
                        }
```

sauraww (changes requested):

crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,357 @@
...
+pub enum Expression {
+    Is(serde_json::Value),
+    In(serde_json::Value),
+    Has(serde_json::Value),
+    Between(serde_json::Value, serde_json::Value),
+    Other(String, Vec<serde_json::Value>),
+}
+
+pub fn is_variable(v: &serde_json::Value) -> bool {
+    v.is_object() && v.as_object().unwrap().contains_key("var")
```
Datron, line 19: risky unwrap()

crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,357 @@
...
+    fn try_from_expression_map(
+        source: &Map<String, serde_json::Value>,
+    ) -> Result<Self, &'static str> {
+        if let Some(operator) = source.keys().next() {
+            let operands = source[operator]
+                .as_array()
+                .cloned()
+                .ok_or("Invalid operands list for context")?;
+
+            let operand_0 = operands.first();
```
Datron, line 36:
```suggestion
            let operand_0 = operands.get(0);
```

crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,357 @@
...
+}
+
+impl Operator {
+    pub fn from_operator_input(source: String) -> Self {
+        match source.as_str() {
+            "==" => Operator::Is,
+            "<=" => Operator::Between,
+            "in" => Operator::In,
+            "has" => Operator::Has,
+            other => Operator::Other(other.to_string()),
```
Datron, line 213: Should this be _?

Datron (changes requested):
Few concerns

sauraww (changes requested):

ShreyBana (approved):

sauraww (approved):

crates/frontend/src/logic.rs
ShubhranshuSanjeev: done

PR #259, opened 2024-10-07
fix: add prefix filter in resolve endpoint

mahatoankitkumar:
## Problem
add prefix filter in resolve endpoint

### DIFF
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, str::FromStr};
 
 use super::helpers::{
-    filter_config_by_dimensions, filter_config_by_prefix, get_query_params_map,
+    apply_prefix_filter_to_config, filter_config_by_dimensions, get_query_params_map,
 };
 use super::types::{Config, Context};
 use crate::api::context::{
@@ -559,17 +559,7 @@ async fn get(
     let mut config_version = validate_version_in_params(&mut query_params_map)?;
     let mut config = generate_config_from_version(&mut config_version, &mut conn)?;
 
-    if let Some(prefix) = query_params_map
-        .get("prefix")
-        .and_then(|prefix| prefix.as_str())
-    {
-        config = filter_config_by_prefix(
-            &config,
-            &prefix.split(',').map(String::from).collect(),
-        )?
-    }
-
-    query_params_map.remove("prefix");
+    config = apply_prefix_filter_to_config(&mut query_params_map, config)?;
 
     if !query_params_map.is_empty() {
         config = filter_config_by_dimensions(&config, &query_params_map)?
@@ -601,7 +591,9 @@ async fn get_resolved_config(
     }
 
     let mut config_version = validate_version_in_params(&mut query_params_map)?;
-    let config = generate_config_from_version(&mut config_version, &mut conn)?;
+    let mut config = generate_config_from_version(&mut config_version, &mut conn)?;
+
+    config = apply_prefix_filter_to_config(&mut query_params_map, config)?;
 
     let cac_client_contexts = config
         .contexts
--- a/crates/context_aware_config/src/api/config/helpers.rs
+++ b/crates/context_aware_config/src/api/config/helpers.rs
@@ -24,6 +24,24 @@ pub fn filter_context(
         .collect()
 }
 
+pub fn apply_prefix_filter_to_config(
+    query_params_map: &mut Map<String, Value>,
+    mut config: Config,
+) -> superposition::Result<Config> {
+    if let Some(prefix) = query_params_map
+        .get("prefix")
+        .and_then(|prefix| prefix.as_str())
+    {
+        config = filter_config_by_prefix(
+            &config,
+            &prefix.split(',').map(String::from).collect(),
+        )?
+    }
+
+    query_params_map.remove("prefix");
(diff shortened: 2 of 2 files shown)


crates/context_aware_config/src/api/config/handlers.rs
ShubhranshuSanjeev, line 1: Why not get the filtered config and then directly apply the eval over it, rather than evaluating and then filtering out the keys?

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
mahatoankitkumar: filtering out the resolved config would be much faster i guess.

ShubhranshuSanjeev: You mean performance wise?

mahatoankitkumar: yes

ShubhranshuSanjeev: Can you explain, how you came to that conclusion?

mahatoankitkumar: I guess it was intuitive, parsing the whole config and filtering out the default_config, context and overrides would take a significant time in case of a large config and just filtering the resolved config would take much lesser time i guess. Correct me if i am missing something.

ShubhranshuSanjeev: Why not, in the eval helper function for all the evaluated contexts, remove the non-required keys and then merge, are we not operating on extra keys while merging if we don't filter out.

ayushjain17: as far as I remember, eval was our slower operation here (way slow as compared to filtering out keys from overrides)
so doing filter would help us here, as the number of contexts on which I have to perform eval has high chances of getting reduced, so overall this is likely to improve perf

mahatoankitkumar: > eval was our slower operation here (way slow as compared to filtering out keys from overrides)
I think this is debatable.

Datron: eval is the slower operation, given how json logic is structured. Its probably O(nlogn), filter is O(n) here. 

The performance gain in doing filter first and then eval might not justify the code complexity. I'm okay not optimizing right now, and doing it when performance becomes a problem. Filtering after eval on key value pairs is the simpler way, (done in the client as well). If it's fine, lets just go with that.

mahatoankitkumar: If that is the case then we can do the filter first now it self, that isn't a big change.

Datron (approved):

crates/context_aware_config/src/api/config/handlers.rs
ShubhranshuSanjeev: > eval is the slower operation, given how json logic is structured. Its probably O(nlogn), filter is O(n) here.
> 
> The performance gain in doing filter first and then eval might not justify the code complexity. I'm okay not optimizing right now, and doing it when performance becomes a problem. Filtering after eval on key value pairs is the simpler way, (done in the client as well). If it's fine, lets just go with that.

@mahatoankitkumar made the changes, but I don't support this, no point reworking same piece of code again and again.

crates/context_aware_config/src/api/config/helpers.rs
```diff
@@ -24,6 +24,24 @@ pub fn filter_context(
...
+        config = filter_config_by_prefix(
+            &config,
+            &prefix.split(',').map(String::from).collect(),
+        )?
+    }
+
+    query_params_map.remove("prefix");
+    Ok(config)
+}
+
```
ShubhranshuSanjeev, line 44:
```suggestion
pub fn apply_prefix_filter_to_config(
    query_params_map: &mut Map<String, Value>,
    config: Config,
) -> superposition::Result<Config> {
    if let Some(prefix) = query_params_map
        .get("prefix")
        .and_then(|prefix| prefix.as_str())
    {
        return filter_config_by_prefix(
            &config,
            &prefix.split(',').map(String::from).collect(),
        );
    }

    query_params_map.remove("prefix");
    Ok(config)
}

```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -559,17 +559,7 @@ async fn get(
...
-        .and_then(|prefix| prefix.as_str())
-    {
-        config = filter_config_by_prefix(
-            &config,
-            &prefix.split(',').map(String::from).collect(),
-        )?
-    }
-
-    query_params_map.remove("prefix");
+    config = apply_prefix_filter_to_config(&mut query_params_map, config)?;
```
ShubhranshuSanjeev, line 562: Do we even need this function?

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/config/helpers.rs
mahatoankitkumar: This way `query_params_map.remove("prefix");` won't be performed after applying filter.

ShubhranshuSanjeev: Ok, I overlooked that.

ShubhranshuSanjeev (approved):

PR #265, opened 2024-10-16
feat: Webhook trigger for experiments

mahatoankitkumar:
## Problem
Webhook trigger for experiments

## Solution
 Webhook trigger for experiments (Create, Conclude, ramp, update_overrides)

## Tenant Config changes

`experiments_webhook_config = { "value" = {}, "schema" = { "type" = "object", properties = {"url" = { "type" = "string" }, "method" = { "enum" = ["Post","Put","Get"], "type" = "string" },"headers" = { "type" = "array", "items" = { "type" = "string", "enum" = ["ConfigVersion"] }}, authorization = {"type" = "object", properties = { "key" = { "type" = "string" }, "value" = { "type" = "string" }}}, "required"= ["url", "method", "headers", "authorization"], "additionalProperties" = false}}}
`

## Post-deployment activity

Tenant Config Changes

### DIFF
--- a/crates/experimentation_platform/src/api/experiments/handlers.rs
+++ b/crates/experimentation_platform/src/api/experiments/handlers.rs
@@ -1,9 +1,6 @@
-use std::{
-    collections::{HashMap, HashSet},
-    str::FromStr,
-};
+use std::collections::{HashMap, HashSet};
 
-use actix_http::header::{self, HeaderMap, HeaderName, HeaderValue};
+use actix_http::header::{self};
 use actix_web::{
     get, patch, post, put,
     web::{self, Data, Json, Query},
@@ -17,18 +14,24 @@ use diesel::{
 };
 use reqwest::{Method, Response, StatusCode};
 use serde_json::{json, Map, Value};
-use service_utils::helpers::{construct_request_headers, generate_snowflake_id, request};
+use service_utils::helpers::{
+    construct_request_headers, execute_webhook_call, generate_snowflake_id, request,
+};
 use service_utils::service::types::{
     AppHeader, AppState, CustomHeaders, DbConnection, Tenant,
 };
 use superposition_macros::{bad_argument, response_error, unexpected_error};
-use superposition_types::{result as superposition, Condition, Exp, Overrides, User};
+use superposition_types::{
+    result::{self as superposition},
+    webhook::{WebhookConfig, WebhookEvent},
+    Condition, Exp, Overrides, TenantConfig, User,
+};
 
 use super::{
     helpers::{
         add_variant_dimension_to_ctx, check_variant_types,
         check_variants_override_coverage, decide_variant, extract_override_keys,
-        validate_experiment, validate_override_keys,
+        fetch_cac_config, validate_experiment, validate_override_keys,
     },
     types::{
         ApplicableVariantsQuery, AuditQueryFilters, ConcludeExperimentRequest,
@@ -38,9 +41,12 @@ use super::{
     },
 };
 
-use crate::db::{
-    models::{EventLog, Experiment, ExperimentStatusType, Variant, Variants},
-    schema::{event_log::dsl as event_log, experiments::dsl as experiments},
+use crate::{
+    api::experiments::helpers::construct_header_map,
+    db::{
+        models::{EventLog, Experiment, ExperimentStatusType, Variant, Variants},
+        schema::{event_log::dsl as event_log, experiments::dsl as experiments},
+    },
 };
 
 pub fn endpoints(scope: Scope) -> Scope {
@@ -53,33 +59,6 @@ pub fn endpoints(scope: Scope) -> Scope {
         .service(get_experiment_handler)
         .service(ramp)
         .service(update_overrides)
-}
-
-fn construct_header_map(
-    tenant: &str,
-    other_headers: Vec<(&str, String)>,
-) -> superposition::Result<HeaderMap> {
-    let mut headers = HeaderMap::new();
-    let tenant_val = HeaderValue::from_str(tenant).map_err(|err| {
-        log::error!("failed to set header: {}", err);
-        unexpected_error!("Something went wrong")
-    })?;
-    headers.insert(HeaderName::from_static("x-tenant"), tenant_val);
-    for (header, value) in other_headers {
-        let header_name = HeaderName::from_str(header).map_err(|err| {
-            log::error!("failed to set header: {}", err);
-            unexpected_error!("Something went wrong")
-        })?;
-
-        HeaderValue::from_str(value.as_str())
-            .map(|header_val| headers.insert(header_name, header_val))
-            .map_err(|err| {
-                log::error!("failed to set header: {}", err);
-                unexpected_error!("Something went wrong")
-            })?;
-    }
-
-    Ok(headers)
 }
 
 fn add_config_version_to_header(
@@ -151,6 +130,7 @@ async fn create(
     db_conn: DbConnection,
     tenant: Tenant,
     user: User,
+    tenant_config: TenantConfig,
 ) -> superposition::Result<HttpResponse> {
     use crate::db::schema::experiments::dsl::experiments;
     let mut variants = req.variants.to_vec();
@@ -305,8 +285,19 @@ async fn create(
         .get_results(&mut conn)?;
 
     let inserted_experiment: Experiment = inserted_experiments.remove(0);
-    let response = ExperimentCreateResponse::from(inserted_experiment);
-
+    let response = ExperimentCreateResponse::from(inserted_experiment.clone());
+    if let WebhookConfig::Enabled(experiments_webhook_config) =
+        tenant_config.experiments_webhook_config
+    {
+        execute_webhook_call(
+            &experiments_webhook_config,
+            &inserted_experiment,
+            &config_version_id,
+            WebhookEvent::ExperimentCreated,
+            &state.http_client,
+        )
+        .await?;
+    }
     let mut http_resp = HttpResponse::Ok();
     add_config_version_to_header(&config_version_id, &mut http_resp);
     Ok(http_resp.json(response))
@@ -320,11 +311,12 @@ async fn conclude_handler(
     req: web::Json<ConcludeExperimentRequest>,
     db_conn: DbConnection,
     tenant: Tenant,
+    tenant_config: TenantConfig,
     user: User,
 ) -> superposition::Result<HttpResponse> {
     let DbConnection(conn) = db_conn;
     let (response, config_version_id) = conclude(
-        state,
+        &state,
         path.into_inner(),
         custom_headers.config_tags,
         req.into_inner(),
@@ -333,13 +325,27 @@ async fn conclude_handler(
         user,
     )
     .await?;
+
+    if let WebhookConfig::Enabled(experiments_webhook_config) =
+        tenant_config.experiments_webhook_config
+    {
+        execute_webhook_call(
+            &experiments_webhook_config,
+            &response,
+            &config_version_id,
+            WebhookEvent::ExperimentConcluded,
+            &state.http_client,
+        )
+        .await?;
+    }
+
     let mut http_resp = HttpResponse::Ok();
     add_config_version_to_header(&config_version_id, &mut http_resp);
     Ok(http_resp.json(ExperimentResponse::from(response)))
 }
 
 pub async fn conclude(
-    state: Data<AppState>,
+    state: &Data<AppState>,
     experiment_id: i64,
     config_tags: Option<String>,
     req: ConcludeExperimentRequest,
@@ -604,10 +610,13 @@ pub fn get_experiment(
 
 #[patch("/{id}/ramp")]
 async fn ramp(
+    data: Data<AppState>,
     params: web::Path<i64>,
     req: web::Json<RampRequest>,
     db_conn: DbConnection,
     user: User,
+    tenant: Tenant,
+    tenant_config: TenantConfig,
 ) -> superposition::Result<Json<ExperimentResponse>> {
     let DbConnection(mut conn) = db_conn;
     let exp_id = params.into_inner();
@@ -644,6 +653,28 @@ async fn ramp(
         ))
         .get_result(&mut conn)?;
 
+    let (_, config_version_id) = fetch_cac_config(&tenant, &data).await?;
+
+    let webhook_event = if new_traffic_percentage == 0
+        && matches!(experiment.status, ExperimentStatusType::CREATED)
+    {
+        WebhookEvent::ExperimentStarted
+    } else {
+        WebhookEvent::ExperimentInprogress
+    };
+    if let WebhookConfig::Enabled(experiments_webhook_config) =
+        tenant_config.experiments_webhook_config
+    {
+        execute_webhook_call(
+            &experiments_webhook_config,
+            &updated_experiment,
+            &config_version_id,
+            webhook_event,
+            &data.http_client,
+        )
+        .await?;
+    }
+
     Ok(Json(ExperimentResponse::from(updated_experiment)))
 }
 
@@ -655,6 +686,7 @@ async fn update_overrides(
     db_conn: DbConnection,
     req: web::Json<OverrideKeysUpdateRequest>,
     tenant: Tenant,
+    tenant_config: TenantConfig,
     user: User,
 ) -> superposition::Result<HttpResponse> {
     let DbConnection(mut conn) = db_conn;
@@ -863,6 +895,19 @@ async fn update_overrides(
             experiments::last_modified_by.eq(user.get_email()),
         ))
         .get_result::<Experiment>(&mut conn)?;
+
+    if let WebhookConfig::Enabled(experiments_webhook_config) =
+        tenant_config.experiments_webhook_config
+    {
+        execute_webhook_call(
+            &experiments_webhook_config,
+            &updated_experiment,
+            &config_version_id,
+            WebhookEvent::ExperimentUpdated,
+            &state.http_client,
+        )
+        .await?;
+    }
 
     let mut http_resp = HttpResponse::Ok();
     add_config_version_to_header(&config_version_id, &mut http_resp);
--- a/crates/experimentation_platform/src/api/experiments/helpers.rs
+++ b/crates/experimentation_platform/src/api/experiments/helpers.rs
@@ -1,11 +1,14 @@
+use actix_http::header::{self, HeaderMap, HeaderName, HeaderValue};
+use actix_web::web::Data;
 use diesel::pg::PgConnection;
 use diesel::{BoolExpressionMethods, ExpressionMethods, QueryDsl, RunQueryDsl};
 use serde_json::{Map, Value};
 use service_utils::helpers::extract_dimensions;
-use service_utils::service::types::ExperimentationFlags;
+use service_utils::service::types::{AppState, ExperimentationFlags, Tenant};
 use std::collections::HashSet;
+use std::str::FromStr;
 use superposition_macros::{bad_argument, unexpected_error};
-use superposition_types::{result as superposition, Condition, Exp, Overrides};
+use superposition_types::{result as superposition, Condition, Config, Exp, Overrides};
 
 use crate::db::models::{Experiment, ExperimentStatusType, Variant, VariantType};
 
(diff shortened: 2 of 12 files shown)


crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -268,3 +272,67 @@ pub fn decide_variant(
...
+        HeaderValue::from_str(value.as_str())
+            .map(|header_val| headers.insert(header_name, header_val))
+            .map_err(|err| {
+                log::error!("failed to set header: {}", err);
+                unexpected_error!("Something went wrong")
+            })?;
+    }
+
+    Ok(headers)
+}
```
mahatoankitkumar, line 301: Just moved here from handlers.

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -333,9 +327,24 @@ async fn conclude_handler(
...
+        WebhookEvent::ExperimentConcluded,
+    )
+    .await
+    {
+        Ok(_) => {
+            let mut http_resp = HttpResponse::Ok();
+            add_config_version_to_header(&config_version_id, &mut http_resp);
+            Ok(http_resp.json(ExperimentResponse::from(response)))
+        }
+        Err(webhook_error) => Err(superposition::AppError::WebhookError(anyhow!(
```
Datron, line 344: @mahatoankitkumar don't use anyhow here

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -872,9 +897,21 @@ async fn update_overrides(
...
         .get_result::<Experiment>(&mut conn)?;
 
-    let mut http_resp = HttpResponse::Ok();
-    add_config_version_to_header(&config_version_id, &mut http_resp);
-    Ok(http_resp.json(ExperimentResponse::from(updated_experiment)))
+    match execute_webhook_call(
+        tenant_config.experiments_webhook_config,
+        updated_experiment.clone(),
+        &config_version_id,
+        WebhookEvent::ExperimentInprogress,
```
Datron, line 904: I you can't update_overrides while the experiment is inptogress, so you can propogate created here

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+pub async fn execute_webhook_call<T>(
+    webhook_object: Value,
+    payload: T,
+    config_version_opt: &Option<String>,
+    event: WebhookEvent,
+) -> Result<(), AppError>
+where
+    T: Serialize,
+{
+    match serde_json::from_value::<Webhook>(webhook_object) {
```
Datron, line 424: @mahatoankitkumar can we use `?` to unwrap the object earlier, and avoid this nesting?

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+        Ok(webhook) => {
+            let http_client = reqwest::Client::new();
```
Datron, line 426: We can add this to AppState, creating a new HttpClient is expensive operation. This can then be used everywhere else in Superposition

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+                .collect::<Vec<(String, String)>>();
+
+            if let Ok(auth) =
+                serde_json::from_value::<Authorization>(webhook.authorization)
+            {
+                let auth_token_value: String =
+                    get_from_env_unsafe(&auth.value).map_err(|err| {
+                        log::error!("Failed to retrieve authentication token for the webhook with error: {}", err);
+                        AppError::WebhookError(
+                            anyhow!("Failed to retrieve authentication token for the webhook. Please verify the credentials in TenantConfig.")
```
Datron, line 451: Avoid using anyhow

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+            let response = request_builder
+                .headers(headers.into())
+                .json(&WebhookResponse { event, payload })
+                .send()
+                .await;
+
+            match response {
+                Ok(res) => match res.status() {
+                    StatusCode::OK => {
+                        log::info!("webhook call succeeded: {:?}", res);
```
Datron, line 482:
```suggestion
                        log::info!("webhook call succeeded: {}", res.to_string());
```

Datron: We maybe shouldn't log the response, the service we called could return a huge file or response and break us. We can instead log some details - like status code, request headers, response headers. We don't care about the body of the response

crates/superposition_macros/src/lib.rs
```diff
@@ -57,3 +57,13 @@ macro_rules! response_error {
...
     };
 }
+
+#[macro_export]
+macro_rules! webhook_error {
+    ($msg: literal, $($args: tt)*) => {
+        superposition_types::result::AppError::WebhookError(anyhow::anyhow!(format!($msg, $($args)*)))
+    };
+    ($err: tt) => {
+        superposition_types::result::AppError::WebhookError(anyhow::anyhow!($err.to_string()))
```
Datron, line 67: Skip using anyhow

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,57 @@ impl Display for RegexEnum {
...
+pub struct Webhook {
+    pub url: String,
+    pub method: HttpMethod,
+    pub headers: Vec<HeadersEnum>,
+    pub authorization: Value,
+}
+
+#[derive(Serialize, Deserialize)]
+pub enum WebhookEvent {
+    ExperimentCreated,
```
Datron, line 283: We should have an experiment started event, that is fired when ramp 0 is called for the first time

Datron (changes requested):

crates/superposition/Superposition.cac.toml
```diff
@@ -1,9 +1,12 @@
 [default-config]
 mandatory_dimensions = { "value" = [
 ], "schema" = { "type" = "array", "items" = { "type" = "number" } } }
+experiments_webhook_config = { "value" = {}, "schema" = { "type" = "object", properties = {"url" = { "type" = "string" }, "method" = { "enum" = ["Post","Put","Get"], "type" = "string" },"headers" = { "type" = "array", "items" = { "type" = "string", "enum" = ["ConfigVersion"] }}, authorization = {"type" = "object", properties = { "key" = { "type" = "string" }, "value" = { "type" = "string" }}}, "required"= ["url", "method", "headers", "authorization"], "additionalProperties" = false}}}
```
ayushjain17, line 4:
```suggestion
experiments_webhook_config = { "value" = {}, "schema" = { "type" = "object", properties = {"url" = { "type" = "string" }, "method" = { "enum" = ["Post", "Put", "Get"], "type" = "string" }, "headers" = { "type" = "array", "items" = { "type" = "string", "enum" = ["ConfigVersion"] }}, authorization = {"type" = "object", properties = { "key" = { "type" = "string" }, "value" = { "type" = "string" }}}, "required"= ["url", "method", "headers", "authorization"], "additionalProperties" = false}}}
```

ayushjain17, line 4: you have added this `"required"= ["url", "method", "headers", "authorization"]`, but the default value is set as `"value" = {}`

even though the validation is not enabled, lets add stuff as per proper validations only, 
because I am not sure with all the keys marked as required and then with proper default value, how are we going to disable the webhook call for those who dont need it

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
         }
     }
 }
+
+pub async fn execute_webhook_call<T>(
```
ayushjain17, line 415: review pending for this function

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,59 @@ impl Display for RegexEnum {
...
+#[derive(Serialize, Deserialize)]
+pub struct WebhookResponse<T> {
+    pub event: WebhookEvent,
+    pub payload: T,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
 pub struct TenantConfig {
     pub mandatory_dimensions: Vec<String>,
+    pub experiments_webhook_config: Value,
```
ayushjain17, line 299: we know the type of data here, so lets not use Value here and lets use a proper concrete type, which should be `Webhook` type, right ?
```suggestion
    pub webhook_config: Webhook,
```

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,59 @@ impl Display for RegexEnum {
     }
 }
 
-#[derive(Clone, Deserialize)]
+#[derive(Clone, Debug, Deserialize)]
+pub enum HeadersEnum {
+    ConfigVersion,
```
ayushjain17, line 249: since, we are adding this, can we also replace all the usages of "x-config-version" with this Enum

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,59 @@ impl Display for RegexEnum {
...
+            HeadersEnum::ConfigVersion => write!(f, "Config-Version"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub enum HttpMethod {
+    Get,
+    Post,
+    Put,
```
ayushjain17, line 264: is there any particular reason to restrict to these 3 only ?

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,59 @@ impl Display for RegexEnum {
...
+    pub key: String,
+    pub value: String,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Webhook {
+    pub url: String,
+    pub method: HttpMethod,
+    pub headers: Vec<HeadersEnum>,
+    pub authorization: Value,
```
ayushjain17, line 278: shouldn't the type here be `Authorization` ?
```suggestion
    pub authorization: Authorization,
```

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+        Ok(webhook) => {
+            let http_client = reqwest::Client::new();
+            let mut header_array = webhook
+                .headers
+                .into_iter()
+                .filter_map(|key| match key {
+                    HeadersEnum::ConfigVersion => {
+                        if let Some(config_version) = config_version_opt {
+                            Some((
+                                "x-config-version".to_string(),
```
ayushjain17, line 434:
```suggestion
                                String::from(key),
```

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+where
+    T: Serialize,
+{
+    match serde_json::from_value::<Webhook>(webhook_object) {
+        Ok(webhook) => {
+            let http_client = reqwest::Client::new();
+            let mut header_array = webhook
+                .headers
+                .into_iter()
+                .filter_map(|key| match key {
```
ayushjain17, line 430: why do we need to do a filter operation over here ?
this is counter intuitive, `webhook.headers`, is of type `Vec<HeadersEnum>`, so it does not make sense to filter here

to avoid repetition of keys in the headers (which headers anyways disregards and keeps only one entry as its a map), you can simply add if checks for each value present in the `HeadersEnum` which makes the handling of each header explicit, as it is supposed to be different behaviour for each

or this could have simply been a `map`

crates/service_utils/src/helpers.rs
ayushjain17, line 424: this should not be needed, the type should handle it by default

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+                            ))
+                        } else {
+                            None
+                        }
+                    }
+                })
+                .collect::<Vec<(String, String)>>();
+
+            if let Ok(auth) =
+                serde_json::from_value::<Authorization>(webhook.authorization)
```
ayushjain17, line 445: same, this should not be needed here, the type should handle it by default

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+                        log::error!("Failed to retrieve authentication token for the webhook with error: {}", err);
+                        AppError::WebhookError(
+                            format!("Failed to retrieve authentication token for the webhook. Please verify the credentials in TenantConfig.")
+                        )
+                    })?;
+                header_array.push((auth.key, auth_token_value));
+            }
+
+            let mut headers = HeaderMap::new();
+            header_array.iter().for_each(|(name, value)| {
```
ayushjain17, line 458: I think, we had a function for doing this, can you please verify once
most probably `construct_request_headers` from `service_utils` should be the one

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +411,91 @@ pub fn parse_config_tags(
...
+                .await;
+
+            match response {
+                Ok(res) => match res.status() {
+                    StatusCode::OK => {
+                        log::info!("webhook call succeeded: {:?}", res.status());
+                        Ok(())
+                    }
+                    _ => {
+                        log::error!("Webhook call failed with status code: {:?}, response headers: {:?}", res.status(), res.headers());
```
ayushjain17, line 486: shouldn't we log here the response body, instead of the response headers ?

ayushjain17 (changes requested):
pending comments plus these as all

crates/service_utils/src/helpers.rs
ayushjain17: this is done

crates/service_utils/src/helpers.rs
mahatoankitkumar: cc: @Datron

crates/service_utils/src/helpers.rs
mahatoankitkumar: Already had tried this, but construct_request_headers takes &[&str, &str]

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -305,11 +286,23 @@ async fn create(
...
+    let response = ExperimentCreateResponse::from(inserted_experiment.clone());
 
-    let mut http_resp = HttpResponse::Ok();
-    add_config_version_to_header(&config_version_id, &mut http_resp);
-    Ok(http_resp.json(response))
+    match execute_webhook_call(
+        tenant_config.experiments_webhook_config,
+        inserted_experiment,
+        &config_version_id,
+        WebhookEvent::ExperimentCreated,
```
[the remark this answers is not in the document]
mahatoankitkumar: I guess this is fine, it would be an overkill i guess

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,57 @@ impl Display for RegexEnum {
     }
 }
 
-#[derive(Clone, Deserialize)]
+#[derive(Clone, Debug, Deserialize)]
```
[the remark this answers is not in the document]
mahatoankitkumar: but its related to this only right

crates/service_utils/src/helpers.rs
mahatoankitkumar: But config_version_id is optional, so filter_map

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,59 @@ impl Display for RegexEnum {
...
+pub struct Webhook {
+    pub url: String,
+    pub method: HttpMethod,
+    pub headers: Vec<HeadersEnum>,
+    pub authorization: Value,
+}
+
+#[derive(Serialize, Deserialize)]
+pub enum WebhookEvent {
```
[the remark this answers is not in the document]
mahatoankitkumar: Why is this needed?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -333,13 +327,28 @@ async fn conclude_handler(
         user,
     )
     .await?;
+
+    if let (true, Some(experiments_webhook_config)) = (
+        tenant_config.experiments_webhook_config.enabled,
+        tenant_config.experiments_webhook_config.configuration,
```
Datron, line 333: I thought we added enabled to avoid parsing configuration?

Datron (approved):
Looks good, few questions

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -648,6 +660,29 @@ async fn ramp(
         ))
         .get_result(&mut conn)?;
 
+    let (_, config_version_id) = fetch_cac_config(&tenant, &data).await?;
+
+    let webhook_event = if new_traffic_percentage == 0
```
ayushjain17, line 665: if the new ramp is 0 and previous ramp is not 0, but something else like 10 or 20, in that case it should not be `ExperimentStarted`

ayushjain17 (changes requested):
mostly pending comments

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -305,8 +286,20 @@ async fn create(
         .get_results(&mut conn)?;
 
     let inserted_experiment: Experiment = inserted_experiments.remove(0);
-    let response = ExperimentCreateResponse::from(inserted_experiment);
-
+    let response = ExperimentCreateResponse::from(inserted_experiment.clone());
+    if let (true, Some(experiments_webhook_config)) = (
```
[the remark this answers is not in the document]
mahatoankitkumar: cc: @Datron

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -872,6 +908,20 @@ async fn update_overrides(
...
 
+    if let (true, Some(experiments_webhook_config)) = (
+        tenant_config.experiments_webhook_config.enabled,
+        tenant_config.experiments_webhook_config.configuration,
+    ) {
+        execute_webhook_call(
+            experiments_webhook_config,
+            updated_experiment.clone(),
+            &config_version_id,
+            WebhookEvent::ExperimentUpdated,
```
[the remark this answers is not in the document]
mahatoankitkumar: I guess this is fine, it would be an overkill i guess

crates/experimentation_platform/src/api/experiments/handlers.rs
mahatoankitkumar: But if the previous ramp is not 0, then the previous experimentStatus won't be CREATED, it would be INPROGRESS

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,92 @@ impl Display for RegexEnum {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Webhook {
+    pub url: String,
+    pub method: HttpMethod,
+    pub headers: Vec<HeadersEnum>,
+    pub authorization: Authorization,
+}
+
+#[derive(Serialize, Deserialize)]
+pub enum WebhookEvent {
```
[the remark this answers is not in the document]
mahatoankitkumar: Why is this not usable for CAC, the webhook object will be similar, it can be reused. If it can't be reused then it can be renamed then. But now i think this is fine

crates/service_utils/src/helpers.rs
```diff
@@ -401,3 +410,78 @@ pub fn parse_config_tags(
...
+where
+    T: Serialize,
+{
+    let mut header_array = webhook_config
+        .headers
+        .into_iter()
+        .filter_map(|key| match key {
+            HeadersEnum::ConfigVersion => {
+                if let Some(config_version) = config_version_opt {
+                    Some(("x-config-version".to_string(), config_version.to_string()))
```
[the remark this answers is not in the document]
mahatoankitkumar: please explain with code, i didn't get you here.

crates/superposition_types/src/lib.rs
```diff
@@ -243,9 +244,92 @@ impl Display for RegexEnum {
...
+    pub configuration: Option<Webhook>,
+}
+
+impl<'de> Deserialize<'de> for WebhookConfig {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct WebhookConfigHelper {
```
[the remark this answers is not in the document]
mahatoankitkumar: Isn't it better to throw proper error than unwraping

crates/experimentation_platform/src/api/experiments/handlers.rs
mahatoankitkumar: But this check would be needed right?

crates/service_utils/src/helpers.rs
mahatoankitkumar: How please explain

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -268,3 +272,67 @@ pub fn decide_variant(
...
+        )
+        .send()
+        .await;
+
+    match response {
+        Ok(res) => {
+            let config_version = res
+                .headers()
+                .get("x-config-version")
+                .and_then(|val| val.to_str().map_or(None, |v| Some(v.to_string())));
```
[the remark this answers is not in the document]
mahatoankitkumar: > to_str gives result, if we just do map it will return result not option

crates/superposition_types/src/webhook.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        let helper = WebhookConfigHelper::deserialize(deserializer)?;
+        match (helper.enabled, helper.configuration) {
+            (true, None) => Err(serde::de::Error::custom(
+                "Configuration must be provided when enabled is true",
+            )),
+            (false, Some(_)) => Err(serde::de::Error::custom(
+                "Configuration must be None when enabled is false",
+            )),
```
ayushjain17, line 76: this should not be failing a case, someone can have their data configured but they can simply opt to have it switched off
resulting in
```suggestion
            (false, Some(_)) => Ok(Self::Disbled),
```

crates/superposition_types/src/webhook.rs
```diff
@@ -0,0 +1,81 @@
...
+            (false, None) => Ok(Self::Disbled),
+            (true, Some(webhook)) => Ok(Self::Enabled(webhook)),
```
ayushjain17, line 78: rather this can be simplified to
```suggestion
            (true, None) => Err(serde::de::Error::custom(
                "Configuration must be provided when enabled is true",
            )),
            (true, Some(webhook)) => Ok(Self::Enabled(webhook)),
            (false, _) => Ok(Self::Disbled),
```

crates/superposition/Superposition.cac.toml
```diff
@@ -1,9 +1,28 @@
...
+], "additionalProperties" = false } } }, "required" = [
+    "enabled",
+] } }
 
 [dimensions]
 tenant = { schema = { "type" = "string", "enum" = ["test", "dev"] } }
 
 [context."$tenant == 'dev'"]
 mandatory_dimensions = []
+# experiments_webhook_config = { "enabled" = true, "configuration" = { "url" = "http://localhost:8080/config/test", "method" = "Get", "headers" = [
```
ayushjain17, line 26: lets have it added as with "enabled" = false and the config added, will ensure things are getting validated in local

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -15,20 +12,28 @@ use diesel::{
     r2d2::{ConnectionManager, PooledConnection},
     ExpressionMethods, PgConnection, QueryDsl, RunQueryDsl,
 };
+
```
ayushjain17, line 15: can we remove this line, causing formatting issue

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -15,20 +12,28 @@ use diesel::{
...
+use service_utils::helpers::{
+    construct_request_headers, execute_webhook_call, generate_snowflake_id, request,
+};
+
```
ayushjain17, line 19: can we remove this line, causing formatting issue

ayushjain17 (changes requested):

ayushjain17 (approved):

PR #266, opened 2024-10-16
fix: compile-time configurable NY-delimiter

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/frontend/src/pages/default_config.rs
+++ b/crates/frontend/src/pages/default_config.rs
@@ -6,14 +6,18 @@ use crate::components::skeleton::Skeleton;
 use crate::components::stat::Stat;
 use crate::components::table::{types::Column, Table};
 
+use crate::schema::HtmlDisplay;
 use crate::types::BreadCrums;
-use crate::utils::{
-    get_local_storage, set_local_storage, unwrap_option_or_default_with_error,
-};
+use crate::utils::{get_local_storage, set_local_storage};
 use leptos::*;
 use leptos_router::{use_navigate, use_query_map};
 use serde_json::{json, Map, Value};
-use std::collections::HashSet;
+
+#[cfg(feature = "multi-delimiter")]
+const GROUPING_DELIMITER: &[char] = &['.', ':'];
+
+#[cfg(not(feature = "multi-delimiter"))]
+const GROUPING_DELIMITER: &[char] = &['.'];
 
 #[derive(Clone, Debug, Default)]
 pub struct RowData {
@@ -40,7 +44,7 @@ pub fn default_config() -> impl IntoView {
     let key_prefix = create_rw_signal::<Option<String>>(None);
     let enable_grouping = create_rw_signal(false);
     let query_params = use_query_map();
-    let bread_crums = Signal::derive(move || get_bread_crums(key_prefix.get()));
+    let bread_crums = Signal::derive(move || utils::get_bread_crums(key_prefix.get()));
 
     create_effect(move |_| {
         let enable_grouping_val =
@@ -59,7 +63,7 @@ pub fn default_config() -> impl IntoView {
         }
     });
 
-    let folder_click_handler = move |key_name: Option<String>| {
+    let on_folder_click = move |key_name: Option<String>| {
         let tenant = tenant_rs.get();
         let redirect_url = match key_name {
             Some(prefix) => format!("admin/{tenant}/default-config?prefix={prefix}"),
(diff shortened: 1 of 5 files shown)


crates/superposition_types/src/lib.rs
```diff
@@ -222,7 +234,7 @@ impl RegexEnum {
     fn get_error_message(&self) -> String {
         match self {
             Self::DefaultConfigKey => ALPHANUMERIC_WITH_DOT_WORDS,
-            Self::DimensionName => ALPHANUMERIC_WITH_DOT_WORDS,
+            Self::DimensionName => ALPHANUMERIC_WITHOUT_DOT,
```
ayushjain17, line 237: This is wrong

ayushjain17 (changes requested):

crates/superposition_types/src/lib.rs
ShubhranshuSanjeev: Can you please elaborate, what's wrong?

ayushjain17: If you will check [here](https://github.com/juspay/superposition/blob/fix%2Fdelimiter/crates%2Fsuperposition_types%2Fsrc%2Flib.rs#L249)
You can see for DimensionName, the regex is for with dot and you have changed the error message to without dot

ayushjain17:
Lets not label anything with `ny` and lets keep it a generic feature (no matter for what short time period we would be keeping this feature)

crates/superposition_types/src/lib.rs
ShubhranshuSanjeev: I meant to change the regex. My bad didn't confirm before pushing this change.

mahatoankitkumar (changes requested):

crates/frontend/src/pages/default_config.rs
```diff
@@ -308,37 +284,32 @@ pub fn default_config() -> impl IntoView {
...
 #[component]
-pub fn bread_crums<NF>(
+pub fn bread_crums(
     bread_crums: Vec<BreadCrums>,
-    folder_click_handler: NF,
-) -> impl IntoView
-where
-    NF: Fn(Option<String>) + 'static + Clone,
-{
+    #[prop(into)] on_folder_click: Callback<Option<String>, ()>,
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: Saves us from writing `clone()` everywhere, handled internally by Leptos when we actually use the value.
Callback internally uses `StoredValue` which is just an id to memory arena (something similar) in leptos. Makes it easy to use in `move` closures.

crates/frontend/src/pages/default_config.rs
```diff
@@ -6,14 +6,18 @@ use crate::components::skeleton::Skeleton;
...
 use leptos::*;
 use leptos_router::{use_navigate, use_query_map};
 use serde_json::{json, Map, Value};
-use std::collections::HashSet;
+
+#[cfg(feature = "ny-delimiter")]
+const GROUPING_DELIMITER: &[char] = &['.', ':'];
+
+#[cfg(not(feature = "ny-delimiter"))]
+const GROUPING_DELIMITER: &[char] = &['.'];
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: This didn't work, this is the best I was able to do.
```rust
cfg_if::cfg_if! {
    if #[cfg(feature = "ny-delimiter")] {
        const GROUPING_DELIMITER: &[char] = &['.', ':'];
    } else {
        const GROUPING_DELIMITER: &[char] = &['.'];
    }
}
```
Which I think is similar to what I wrote. I will leave it as it is.

crates/superposition_types/src/lib.rs
```diff
@@ -234,7 +246,7 @@ impl Display for RegexEnum {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let regex = match self {
             Self::DefaultConfigKey => ALPHANUMERIC_WITH_DOT,
-            Self::DimensionName => ALPHANUMERIC_WITH_DOT,
+            Self::DimensionName => ALPHANUMERIC_WITHOUT_DOT,
```
ayushjain17, line 249: why are we changing this ?
we do intend to allow dimension names like `{key1}.{key2}`

ayushjain17 (changes requested):

Datron:
Write a proper PR description

ayushjain17:
we would need changes in Makefile as well, right ?

Datron:
Closing PR since this change does not need to be merged to main

PR #268, opened 2024-10-21
feat: add cache in cac client

pratikmishra356:
## Problem
Add cache functionality in client eval

## Solution
Used mini-moka , and added new optional parameters while creating new client

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/cac_client/src/interface.rs
```diff
@@ -102,15 +102,40 @@ pub extern "C" fn cac_new_client(
...
+    cache_max_capacity: *const c_char,
+    cache_ttl: *const c_ulong,
+    cache_tti: *const c_ulong,
 ) -> c_int {
     let duration = Duration::new(update_frequency, 0);
     let tenant = unwrap_safe!(cstring_to_rstring(tenant), return 1);
     let hostname = unwrap_safe!(cstring_to_rstring(hostname), return 1);
 
+    let cache_max_capacity = if !cache_max_capacity.is_null() {
+        Some(unsafe { *cache_max_capacity as u64 })
```
ayushjain17, line 114: this should be in `unwrap_safe!`

crates/cac_client/src/interface.rs
```diff
@@ -102,15 +102,40 @@ pub extern "C" fn cac_new_client(
...
+    } else {
+        None
+    };
+    let cache_ttl = if !cache_ttl.is_null() {
+        Some(unsafe { *cache_ttl as u64 })
```
ayushjain17, line 119: this should be in `unwrap_safe!`

crates/cac_client/src/interface.rs
```diff
@@ -102,15 +102,40 @@ pub extern "C" fn cac_new_client(
...
+    } else {
+        None
+    };
+    let cache_tti = if !cache_tti.is_null() {
+        Some(unsafe { *cache_tti as u64 })
```
ayushjain17, line 124: this should be in `unwrap_safe!`

crates/cac_client/Cargo.toml
```diff
@@ -19,6 +19,8 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 tokio = { version = "1.29.1", features = ["full"] }
+mini-moka = { version = "0.10.3" }
+itertools = "0.10.5"
```
ayushjain17, line 23: `context-aware-config` crate is also using this, can you please mark both as workspace and move this version definition to the common one

crates/cac_client/src/lib.rs
```diff
@@ -204,11 +238,33 @@ impl Client {
...
-        }
-        Ok(cac)
+        let filter_string = if let Some(mut vec) = filter_keys.clone() {
+            vec.sort();
+            vec.join(",")
+        } else {
+            "null".to_string()
+        };
+        let hash_key = json_to_sorted_string(&json!(query_data))
+            + "$"
```
ayushjain17, line 248: can we use different delimiter here and the next one, it can cause collision as `json_to_sorted_string` also adds ["$"](https://github.com/juspay/superposition/pull/268/files#diff-0cce34f446e173a9f372d1cc5b7a96287a40a48b0871aab96c3818d11b7fc4fbR14)

crates/cac_client/src/utils.rs
```diff
@@ -1 +1,29 @@
+use itertools::{self, Itertools};
+use serde_json::Value;
 pub mod core;
+
+pub fn json_to_sorted_string(v: &Value) -> String {
```
ayushjain17, line 5: let's expose this function from lib.rs of this crate and use it `context-aware-config` crate as well, delete the copy present in `context-aware-config`

lets make a conscious effort to avoid code repetition

crates/cac_client/src/lib.rs
```diff
@@ -102,7 +112,24 @@ impl Client {
...
+            .weigher(|_key, value: &Map<String, Value>| -> u32 {
+                json!(value.to_owned())
+                    .to_string()
+                    .len()
+                    .try_into()
+                    .unwrap_or(u32::MAX)
+            })
+            // max size of cache in mb
+            .max_capacity(
+                cache_max_capacity.map_or(10 * 1024 * 1024, |v| v * 1024 * 1024),
```
ayushjain17, line 125: lets define the default value as a separate const value ?

crates/cac_client/src/lib.rs
```diff
@@ -102,7 +112,24 @@ impl Client {
...
+            )
+            // Time to live (TTL): in minutes
+            .time_to_live(Duration::from_secs(cache_ttl.map_or(180 * 60, |v| v * 60)))
```
ayushjain17, line 128: lets define the default value as a separate const value ?

crates/cac_client/src/lib.rs
```diff
@@ -102,7 +112,24 @@ impl Client {
...
+            // Time to idle (TTI):  in minutes
+            .time_to_idle(Duration::from_secs(cache_tti.map_or(30 * 60, |v| v * 60)))
```
ayushjain17, line 130: lets define the default value as a separate const value ?

crates/cac_client/src/lib.rs
```diff
@@ -84,10 +87,17 @@ fn get_last_modified(resp: &Response) -> Option<DateTime<Utc>> {
 }
 
 impl Client {
+    /** cache_max_capacity: Max size of cache in mb, default 10 mb
+     *  cache_ttl : Time to live value in minutes, default 180 minutes
```
ayushjain17, line 91:
```suggestion
     *  cache_ttl: Time to live value in minutes, default 180 minutes
```

crates/cac_client/src/lib.rs
```diff
@@ -102,7 +112,24 @@ impl Client {
             return Err("Invalid tenant".to_string());
         }
         let config = resp.json::<Config>().await.map_err_to_string()?;
-
+        let config_cache = Cache::builder()
+            .weigher(|_key, value: &Map<String, Value>| -> u32 {
+                json!(value.to_owned())
```
ayushjain17, line 117: lets avoid macros when it is known that the data is already of `serde_json::Value` compatible type
```suggestion
                Value::Object(value.to_owned())
```

crates/cac_client/src/lib.rs
```diff
@@ -204,11 +238,33 @@ impl Client {
...
+            vec.join(",")
+        } else {
+            "null".to_string()
+        };
+        let hash_key = json_to_sorted_string(&json!(query_data))
+            + "$"
+            + &merge_strategy.clone().to_string()
+            + "$"
+            + &filter_string;
+        let value = self.config_cache.get(&hash_key).unwrap_or({
```
ayushjain17, line 252:
```suggestion
        let value = self.config_cache.get(&hash_key).unwrap_or_else(|| {
```

crates/cac_client/src/lib.rs
```diff
@@ -204,11 +238,33 @@ impl Client {
...
+            let cac = self.config.read().await;
+            let mut evaled_cac = eval::eval_cac(
+                cac.default_configs.to_owned(),
+                &cac.contexts,
+                &cac.overrides,
+                &query_data,
+                merge_strategy,
+            )?;
```
ayushjain17, line 260: shouldn't we cache the error  as well to avoid evaluation when we knows it going to fail anyways ?

ayushjain17, line 260: `unwrap_or_else` would not work if you want to use early return from here
in that case you would have to convert this to a `match` block

crates/cac_client/src/lib.rs
```diff
@@ -188,14 +216,20 @@ impl Client {
...
-        let cac = self.config.read().await;
-        eval::eval_cac(
-            cac.default_configs.to_owned(),
-            &cac.contexts,
-            &cac.overrides,
-            &query_data,
-            merge_strategy,
-        )
+        let context_key = json_to_sorted_string(&json!(query_data));
+        let value = self.config_cache.get(&context_key).unwrap_or({
```
ayushjain17, line 220:
```suggestion
        let value = self.config_cache.get(&context_key).unwrap_or_else(|| {
```

crates/cac_client/src/lib.rs
```diff
@@ -188,14 +216,20 @@ impl Client {
...
+            let cac = self.config.read().await;
+            let evaled_val = eval::eval_cac(
+                cac.default_configs.to_owned(),
+                &cac.contexts,
+                &cac.overrides,
+                &query_data,
+                merge_strategy,
+            )?;
```
ayushjain17, line 228: shouldn't we cache the error as well to avoid evaluation when we knows it going to fail anyways ?

ayushjain17, line 228: `unwrap_or_else` would not work if you want to use early return from here
in that case you would have to convert this to a `match` block

crates/cac_client/src/lib.rs
```diff
@@ -188,14 +216,20 @@ impl Client {
...
     ) -> Result<Map<String, Value>, String> {
-        let cac = self.config.read().await;
-        eval::eval_cac(
-            cac.default_configs.to_owned(),
-            &cac.contexts,
-            &cac.overrides,
-            &query_data,
-            merge_strategy,
-        )
+        let context_key = json_to_sorted_string(&json!(query_data));
```
ayushjain17, line 219:
```suggestion
        let context_key = json_to_sorted_string(&Value::Object(query_data));
```

crates/cac_client/src/lib.rs
```diff
@@ -204,11 +238,33 @@ impl Client {
...
-            cac = filter_keys_by_prefix(cac, keys);
-        }
-        Ok(cac)
+        let filter_string = if let Some(mut vec) = filter_keys.clone() {
+            vec.sort();
+            vec.join(",")
+        } else {
+            "null".to_string()
+        };
+        let hash_key = json_to_sorted_string(&json!(query_data))
```
ayushjain17, line 247:
```suggestion
        let hash_key = json_to_sorted_string(&Value::Object(query_data))
```

ayushjain17 (changes requested):
FFI client side changes review pending

crates/cac_client/src/interface.rs
ayushjain17, line 114: also, lets add a function maybe `clong_to_ru64` just like `cstring_to_rstring`

clients/python/cac_client/cac_client/client.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, config_dict):
 class CacClient:
     rust_lib = ctypes.CDLL(lib_path)
 
-    rust_lib.cac_new_client.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p]
+    rust_lib.cac_new_client.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p, ctypes.POINTER(ctypes.c_int), ctypes.POINTER(ctypes.c_int), ctypes.POINTER(ctypes.c_int)]
```
ayushjain17, line 38: should the types be `c_int` or something like `c_long`, because the headers has type `long`, right ?

clients/python/cac_client/cac_client/client.py
```diff
@@ -66,16 +66,24 @@ class CacClient:
...
             raise ValueError("tenantName cannot be null or empty")
 
         self.tenant = tenant_name
         self.polling_frequency = polling_frequency
         self.cac_host_name = cac_host_name
+        self.cache_max_capacity = cache_max_capacity
+        self.cache_ttl = cache_ttl
+        self.cache_tti = cache_tti
+
+        max_capacity_ptr = ctypes.pointer(ctypes.c_int(cache_max_capacity)) if cache_max_capacity is not None else ctypes.POINTER(ctypes.c_int)()
```
ayushjain17, line 80: can this simply not be written as:
```suggestion
        max_capacity_ptr = ctypes.pointer(ctypes.c_int(cache_max_capacity))
```
sending `None` as param should be same as not sending any param, right ?

clients/python/cac_client/cac_client/client.py
```diff
@@ -66,16 +66,24 @@ class CacClient:
...
+        ttl_ptr = ctypes.pointer(ctypes.c_int(cache_ttl)) if cache_ttl is not None else ctypes.POINTER(ctypes.c_int)()
```
ayushjain17, line 81: same here

clients/python/cac_client/cac_client/client.py
```diff
@@ -66,16 +66,24 @@ class CacClient:
...
+        tti_ptr = ctypes.pointer(ctypes.c_int(cache_tti)) if cache_tti is not None else ctypes.POINTER(ctypes.c_int)()
```
ayushjain17, line 82: same here

clients/haskell/hs-cac-client/src/Client.hs
```diff
@@ -80,12 +81,22 @@ getError = c_last_error_message
 cleanup :: [Ptr a] -> IO ()
 cleanup items = mapM free items $> ()
 
-createCacClient:: Tenant -> Integer -> String -> IO (Either Error ())
-createCacClient tenant frequency hostname = do
+allocateAndPoke :: Maybe Integer -> IO (Ptr CULong)
```
ayushjain17, line 84: should we name it something like
```suggestion
allocateCLongPtr :: Maybe Integer -> IO (Ptr CULong)
```
although its using `poke` internally, having `poke` in name, made it really confusing to understand what was going on here

ayushjain17 (changes requested):
FFI client side changes review

clients/go/cacclient/main.go
```diff
@@ -28,15 +28,30 @@ type CacClient struct {
...
+		cacheTimeToLive = nil
+	}
+	if cacheTTI != nil {
+		cacheTimeToIdle = &(C.ulong(*cacheTTI))
+	} else {
+		cacheTimeToIdle = nil
+	}
 	client := &CacClient{tenant: tenantName, pollingFrequency: pollingFrequency, cacHostName: cacHostName, delimiter: ","}
-	resp := C.cac_new_client(C.CString(tenantName), C.ulong(pollingFrequency), C.CString(cacHostName))
+	resp := C.cac_new_client(C.CString(tenantName), C.ulong(pollingFrequency), C.CString(cacHostName), cacheCapacity, cacheTimeToLive, cacheTimeToIdle)
```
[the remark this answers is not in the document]
pratikmishra356: yes

crates/cac_client/src/interface.rs
pratikmishra356: @ayushjain17 unwrap_safe! is for unwraping and handling error ,
but dereferencing a pointer is unsafe function , so I need unsafe block

crates/cac_client/src/lib.rs
pratikmishra356: we can use `?`

clients/python/cac_client/cac_client/client.py
pratikmishra356: No , it should be pointer only

clients/python/cac_client/cac_client/client.py
pratikmishra356: @ayushjain17 we cannot write in that way as cache_max_capacity can be None , we cannot do ctypes.pointer(ctypes.c_int(cache_max_capacity))

crates/cac_client/Cargo.toml
```diff
@@ -19,6 +19,8 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 tokio = { version = "1.29.1", features = ["full"] }
+mini-moka = { workspace = true }
+itertools = { workspace = true }
```
ayushjain17, line 23: there seems to be a misunderstanding, I was asking to move only `itertools` to workspace as that is the only only common one, and let `mini-moka` be defined here only

also, update the `Cargo.toml` in `context-aware-config` to make the `itertools` over there also pick from workspace

also, can you please make sure that the entries in all the 3 modified toml files are alphabetically sorted

clients/go/cacclient/main.go
```diff
@@ -28,15 +28,30 @@ type CacClient struct {
 }
 
 // NewCacClient creates a new CacClient
-func NewCacClient(tenantName string, pollingFrequency int, cacHostName string) (*CacClient, error) {
+func NewCacClient(tenantName string, pollingFrequency int, cacHostName string, cacheMaxCapacity *int, cacheTTL *int, cacheTTI *int) (*CacClient, error) {
```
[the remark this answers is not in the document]
ayushjain17: this is pending

ayushjain17 (changes requested):

clients/python/cac_client/cac_client/client.py
ayushjain17: no, my question was, 
is `ctypes.pointer(ctypes.c_int(None))` and `ctypes.pointer(ctypes.c_int())` different ?
aren't these two the same ?

pratikmishra356: no these two are not same, c_int() expects an integer value
also in None case it is `ctypes.POINTER(ctypes.c_int)()` not `ctypes.pointer(ctypes.c_int())`

pratikmishra356:
@ayushjain17 I have used go formatter , it does not formats that function part

clients/go/expclient/main.go
```diff
@@ -1,7 +1,7 @@
 package expclient
 
 /*
-#include "libexperimentation_client.h"
+#include "../../../headers/libexperimentation_client.h"
```
ayushjain17, line 4: @Datron can you please verify the change in the path of headers file

clients/python/cac_client/cac_client/client.py
ayushjain17: oh right, missed that subtle difference

crates/cac_client/src/lib.rs
```diff
@@ -169,24 +201,77 @@ impl Client {
         self.last_modified.read().await.clone()
     }
 
+    pub async fn eval(
```
ayushjain17, line 204: we have removed this function
`eval` is no longer needed

crates/cac_client/src/lib.rs
```diff
@@ -169,24 +201,77 @@ impl Client {
...
         &self,
         query_data: Map<String, Value>,
         filter_keys: Option<Vec<String>>,
         merge_strategy: MergeStrategy,
     ) -> Result<Map<String, Value>, String> {
-        let cac = self.config.read().await;
-        let mut config = cac.to_owned();
-        if let Some(keys) = filter_keys {
-            config = config.filter_by_prefix(&HashSet::from_iter(keys));
+        // let cac = self.config.read().await;
```
ayushjain17, line 232: do we need to keep this commented section of code ?

clients/go/expclient/main.go
ayushjain17, line 4: @Datron can you please verify the path change for the header file

clients/go/cacclient/main.go
```diff
@@ -1,7 +1,7 @@
 package cacclient
 
 /*
-#include "libcac_client.h"
+#include "../../../headers/libcac_client.h"
```
ayushjain17, line 4: @Datron can you please verify the path change for the header file

clients/haskell/hs-cac-client/src/Client.hs
ayushjain17, line 1: dont we need to update the path for the header files for Haskell client ?

clients/java/cac-client/src/main/java/cac_client/CacClient.java
ayushjain17, line 1: dont we need to update the path for the header files for Java client

ayushjain17 (changes requested):
Few points of verification and earlier missed out errors

clients/python/cac_client/cac_client/client.py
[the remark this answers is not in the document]
pratikmishra356: no need to update python client , as we not need to explicitly pass None value

clients/java/cac-client/src/main/java/cac_client/CacClient.java
pratikmishra356: I dont think so

clients/haskell/hs-cac-client/src/Client.hs
pratikmishra356: No

clients/go/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+
```
ayushjain17, line 5: can we remove this newline between imports

ayushjain17 (approved):
Verification of path changes for header files is pending across the different FFI clients, otherwise its fine

Datron (changes requested):

crates/cac_client/src/interface.rs
```diff
@@ -102,15 +110,27 @@ pub extern "C" fn cac_new_client(
     tenant: *const c_char,
     update_frequency: c_ulong,
     hostname: *const c_char,
+    cache_max_capacity: *const c_ulong,
+    cache_ttl: *const c_ulong,
+    cache_tti: *const c_ulong,
```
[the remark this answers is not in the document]
ShubhranshuSanjeev: I would also prefer not using null pointer, but won't this make the contract inconsistent, if someone is using a Rust client they can give None or Some but for other client they have to give an int or long everytime.

ShubhranshuSanjeev (changes requested):

clients/haskell/hs-cac-client/src/Client.hs
```diff
@@ -108,7 +119,7 @@ getFullConfigStateWithFilter client mbFilters mbPrefix = do
     cPrefix <- case mbPrefix of
         Just prefix -> newCAString (intercalate "," prefix)
         Nothing     -> return nullPtr
-    config <- withForeignPtr client $ \client -> c_get_config client cFilters cPrefix
+    config <- withForeignPtr client $ \val -> c_get_config val cFilters cPrefix
```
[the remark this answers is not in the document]
pratikmishra356: @Datron it was giving error , as we are using client name in functional parameter also

clients/go/main.go
```diff
@@ -12,18 +12,18 @@ func main() {
...
 	cachostName := "http://localhost:8080"
 
-	client, error := cacclient.NewCacClient(tenant, pollingFrequency, cachostName)
+	client, error := cacclient.NewCacClient(tenant, pollingFrequency, cachostName, nil, nil, nil)
 	if error != nil {
 		fmt.Println(error)
 	}
 
-	fmt.Println("\n------------Configs----------------------------\n")
+	fmt.Println("\n------------Configs----------------------------")
```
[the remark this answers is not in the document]
pratikmishra356: as it is already using println , that's why

Datron (commented):
Looks good, can you update the haskell client variable?

Datron (approved):

PR #276, opened 2024-10-30
feat: use new input type for json schema inputs as well

Datron:
## Problem
The input field to enter a type schema for a config was set to textarea

## Solution
Use the smarter input component that we already use for overrides

https://github.com/user-attachments/assets/e187f0b2-7073-4e07-8932-5a60edfea1c2

### DIFF
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -28,7 +28,7 @@ pub fn default_config_form<NF>(
     #[prop(default = String::new())] config_key: String,
     #[prop(default = String::new())] config_type: String,
     #[prop(default = Value::Null)] type_schema: Value,
-    #[prop(default = String::new())] config_value: String,
+    #[prop(default = Value::Null)] config_value: Value,
     #[prop(default = None)] function_name: Option<Value>,
     #[prop(default = None)] prefix: Option<String>,
     handle_submit: NF,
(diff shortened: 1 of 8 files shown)


crates/frontend/src/components/default_config_form.rs
```diff
@@ -226,49 +213,84 @@ where
...
+                    "OBJECT" => {
+                        let config_type_schema = SchemaType::Single(JsonSchemaType::from(&config_schema_rs.get()));
+                        view! {
+                            <Input
+                                id="default-config-value"
+                                class="mt-5 rounded-md resize-y w-full max-w-md pt-3"
+                                schema_type=config_type_schema
+                                value=config_value_rs.get()
+                                on_change=Callback::new(move |new_default_config: Value| {
+                                    logging::log!("{:?}", new_default_config);
```
sauraww, line 269: Can we add some description here ?

sauraww (approved):

pratikmishra356 (approved):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -226,49 +213,84 @@ where
...
+                        if value.is_null() {
+                            config_value_ws.set(Value::Bool(false));
                         }
                         view! {
                             <BooleanToggle
-                                value={value.parse::<bool>().unwrap_or(false)}
-                                on_change=Callback::new(move |flag: bool| {
-                                    set_config_value.set(flag.to_string());
-                                })
+                                value=value.as_bool().unwrap_or(false)
```
ShubhranshuSanjeev, line 222: @Datron , the pattern matching can be removed and in place the Input component can be used.
```rust
let schema_type = SchemaType::try_from(schema);
let enum_variants = EnumVariants::try_from(schema);
let input_type = InputType::from((schema_type, variants));
```
These snippets can be used to determine the required type and enum options, and then can be directly passed to the `Input` component.

```rust
<Input
    id=id
    class=""
    r#type=input_type.unwrap()
    value=value
    schema_type=schema_type.unwrap()
    on_change=move |value| {}
/>
```

PR #279, opened 2024-11-07
feat: priority recompute restructure

pratikmishra356:
## Problem
User has to declare this exponential value, as value is exponential it has limitation on the total number of dimensions that can be defined

## Solution
Postgresql's Numeric Type Approach

For dimension we can have another column named order which will store order of dimension's specificity ,
the higher order is means more specific
For context , we will have another column named weightage , where we will calculate the value just like we do now
it will summation of (2^ dimension's order value)

since context's weightage value can be too large , so we will use postgresql Numeric type

CREATE TABLE large_numbers (
    id SERIAL PRIMARY KEY,
    big_value NUMERIC
);
Rust/ Diesel supports BigDecimal type which is compatible with Numeric type

## Pre-deployment activity

## Post-deployment activity
link for migration/backward compatibility: https://github.com/juspay/superposition/discussions/277#discussioncomment-11318312

## API changes


crates/context_aware_config/migrations/2024-11-05-083200_add_priority_restructure/up.sql
```diff
@@ -0,0 +1,8 @@
+-- Your SQL goes here
+ALTER TABLE public.dimensions
+add column position integer DEFAULT 1 NOT NULL;
+
+ALTER TABLE public.contexts
```
Datron, line 5: Can we write a query to copy priority to weightage

crates/context_aware_config/migrations/2024-11-05-083200_add_priority_restructure/up.sql
```diff
@@ -0,0 +1,8 @@
+-- Your SQL goes here
+ALTER TABLE public.dimensions
```
Datron, line 2: Can you backfill this based on priority? Can be computed from priority

crates/context_aware_config/Cargo.toml
```diff
@@ -11,6 +11,7 @@ actix-http = "3.3.1"
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 base64 = { workspace = true }
+bigdecimal = { version = "0.3.1" , features= ["serde"]}
```
Datron, line 14: Do we need bigdecimal?

crates/context_aware_config/src/helpers.rs
```diff
@@ -38,6 +42,12 @@ use crate::db::{
     },
 };
 
+pub struct DimensionData {
```
Datron, line 45: derive debug and Clone

crates/context_aware_config/src/helpers.rs
```diff
@@ -38,6 +42,12 @@ use crate::db::{
...
+    pub schema: JSONSchema,
+    pub priority: i32,
```
Datron, line 47:
```suggestion
    pub weight: i32,
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -167,15 +177,15 @@ pub fn validate_jsonschema(
...
+    dimension_schema_map: &HashMap<String, DimensionData>,
 ) -> Result<i32, String> {
     let get_priority = |key: &str, val: &Value| -> Result<i32, String> {
         if key == "var" {
             let dimension_name =
                 val.as_str().ok_or("failed to decode dimension as str")?;
             dimension_schema_map
                 .get(dimension_name)
-                .map(|(_, priority)| priority)
+                .map(|dimension_val| &dimension_val.priority)
```
Datron, line 188:
```suggestion
                .map(|dimension_val| &dimension_val.weight)
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -197,6 +207,43 @@ pub fn calculate_context_priority(
     }
 }
 
+fn calculate_2_exponenet_to_bigdecimal(exponent: u32) -> Result<BigDecimal, String> {
```
Datron, line 210: Why use bigdecimal instead of big int?

Datron, line 210:
```suggestion
fn calculate_weight_from_index(exponent: u32) -> Result<BigDecimal, String> {
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -197,6 +207,43 @@ pub fn calculate_context_priority(
...
+        msg
+    })?;
+
+    let mut weightage = BigDecimal::from(0);
+    for dimension in dimensions {
+        let position = dimension_position_map
+            .get(dimension.as_str())
+            .map(|x| x.position)
+            .ok_or_else(|| {
+                let msg = String::from("Dimension not found in Dimension schema map");
```
Datron, line 239: Can we also log the name of the dimension we couldn't parse?

crates/context_aware_config/migrations/2024-11-05-083200_add_priority_restructure/up.sql
```diff
@@ -0,0 +1,8 @@
...
+add column position integer DEFAULT 1 NOT NULL;
+
+ALTER TABLE public.contexts
+add column weightage numeric(1000,0) DEFAULT 1 NOT NULL;
```
Datron, line 6: Should we call this weight?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -527,7 +529,8 @@ async fn reduce_config(
         .and_then(|value| value.to_str().ok().and_then(|s| s.parse::<bool>().ok()))
         .unwrap_or(false);
 
-    let dimensions_schema_map = get_all_dimension_schema_map(&mut conn)?;
+    let dimensions_vec = get_dimension_data(&mut conn)?;
```
Datron, line 532: Should `get_dimension_data` return a map?

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -81,6 +83,14 @@ pub struct PriorityRecomputeResponse {
     pub new_priority: i32,
 }
 
+#[derive(Serialize)]
+pub struct WeightageRecomputeResponse {
+    pub id: String,
+    pub condition: Condition,
+    pub old_weightage: BigDecimal,
+    pub new_weightage: BigDecimal,
```
Datron, line 91:
```suggestion
    pub old_weight: BigDecimal,
    pub new_weight: BigDecimal,
```

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -35,6 +36,36 @@ impl TryFrom<i32> for Priority {
     }
 }
 
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "Option<i32>")]
+pub struct Position(i32);
+impl Position {
+    fn validate_data(position_val: Option<i32>) -> Result<Self, String> {
+        if let Some(val) = position_val {
+            if val < 0 {
```
Datron, line 45: We should allow 0 right? 2^0 is 1

Datron (changes requested):

crates/context_aware_config/migrations/2024-11-05-083200_add_priority_restructure/up.sql
pratikmishra356: @Datron we can write , but we will anyway have to re populate dimension position and then recompute

migration approach: https://github.com/juspay/superposition/discussions/277#discussioncomment-11318312

crates/context_aware_config/migrations/2024-11-05-083200_add_priority_restructure/up.sql
pratikmishra356: we can but , serial order is being not followed in all tenants and their values does not need be of 2's power

crates/context_aware_config/Cargo.toml
pratikmishra356: Yes @Datron we need to calculate higher value

crates/context_aware_config/src/helpers.rs
pratikmishra356: @Datron this function calculates priority which is currently supported , 
for weight calculation there is different function `calculate_context_weightage`

crates/context_aware_config/src/helpers.rs
pratikmishra356: bigint wont support value larger than 2^64

crates/context_aware_config/migrations/2024-11-05-083200_add_priority_restructure/up.sql
pratikmishra356: we can , lets check with team also

crates/context_aware_config/src/api/config/handlers.rs
pratikmishra356: No , it does not need to be , i have separated it into 2 function and get_dimension_data_map this function returns only values which are needed

crates/context_aware_config/src/api/dimension/types.rs
pratikmishra356: yes we are allowing 0

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -83,40 +105,87 @@ async fn create(
...
-        }
-        Err(e) => {
-            log::error!("Dimension upsert failed with error: {e}");
-            Err(unexpected_error!(
-                "Something went wrong, failed to create/update dimension"
-            ))
+    conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+        match (prev_index, new_index.clone(), req_position_val.clone()) {
+            (Some(prev_val), Some(new_val), None) => {
+                println!("1");
```
Datron, line 111: Remove this

crates/context_aware_config/src/helpers.rs
Datron: @pratikmishra356 will this go away once we migrate all data and the frontend?

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -72,9 +72,31 @@ async fn create(
...
-    let new_dimension = Dimension {
+    let req_priority_val: i32 = create_req.priority.clone().into();
+    let dimension_name: String = create_req.dimension.clone().into();
+    let req_position_val = Into::<Option<i32>>::into(create_req.position.clone());
+
+    let mut results: Vec<(String, i32, i32)> = dimensions
+        .order(priority.asc())
+        .select((dimension, priority, position))
+        .load::<(String, i32, i32)>(&mut conn)
+        .expect("Error loading dimensions");
```
Datron, line 83: No expects, we should throw DB error

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -83,40 +105,87 @@ async fn create(
...
-            log::error!("Dimension upsert failed with error: {e}");
-            Err(unexpected_error!(
-                "Something went wrong, failed to create/update dimension"
-            ))
+    conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+        match (prev_index, new_index.clone(), req_position_val.clone()) {
+            (Some(prev_val), Some(new_val), None) => {
+                println!("1");
+                if prev_val < new_val {
+                    new_dimension.position = new_val.clone() as i32 - 1;
```
Datron, line 113: Why are we doing this check? @pratikmishra356 can we add some comments here and use better variable names to document whats happening?

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -35,6 +36,36 @@ impl TryFrom<i32> for Priority {
     }
 }
 
+#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[serde(try_from = "Option<i32>")]
+pub struct Position(Option<i32>);
```
Datron, line 41: Any reason for wrapping Option<i32> instead of i32 directly? You can do Option<Position> then, rather than deal with None and Some everywhere

Datron (changes requested):

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: Sorry , missed this will have to write comments in code to explain this

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: yes priority field will go away.

crates/context_aware_config/src/helpers.rs
```diff
@@ -38,6 +42,13 @@ use crate::db::{
     },
 };
 
+#[derive(Debug)]
+pub struct DimensionData {
+    pub schema: JSONSchema,
+    pub priority: i32,
+    pub position: i32,
```
ayushjain17, line 49: are we sure we want to call this `position` over here and not `weight` or something similar ?

crates/context_aware_config/src/api/context/handlers.rs
ayushjain17, line 1: pending review

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17, line 1: pending review

ayushjain17 (changes requested):

crates/context_aware_config/src/api/dimension/types.rs
ayushjain17: either case, it should be `Option<Position>`, where it is `Position(i64)`
because the parameter is optional, but currently it reads the parameter is mandatory which in turn takes optional value

ayushjain17: and when you would want to make it mandatory, you would want to change only the request type and not the properties of the type

crates/context_aware_config/src/helpers.rs
pratikmishra356: yes it will be position for dimension

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -72,9 +72,34 @@ async fn create(
...
+    let req_position_val = Into::<Option<i32>>::into(create_req.position.clone());
+
+    let mut results: Vec<(String, i32, i32)> = dimensions
+        .order(priority.asc())
+        .select((dimension, priority, position))
+        .load::<(String, i32, i32)>(&mut conn)
+        .map_err(|err| {
+            log::error!("failed to fetch dimensions with error: {}", err);
+            unexpected_error!("Something went wrong")
+        })?;
```
Datron, line 86: Should we throw a DB error. Also
```suggestion
        .map_err(|err| {
            log::error!("failed to fetch dimensions with error: {}", err);
            unexpected_error!("Could not process this request due to a database error. Please reach out to your admin")
        })?;
```

Datron (approved):
Small nitpick, can we resolve this quickly

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -10,6 +10,7 @@ use crate::db::models::Dimension;
 pub struct CreateReq {
     pub dimension: DimensionName,
     pub priority: Priority,
+    pub position: Position,
```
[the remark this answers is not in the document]
pratikmishra356: get_dimension_data_map is a util function on dimensions data 
so its better to have fetch dimensions function separately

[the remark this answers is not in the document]
pratikmishra356: > we can use LOG function to backfill
UPDATE public.contexts SET position = FLOOR(LOG(2, priority)) WHERE priority > 0;

during deployment priority value might need not be of power 2
not needed i have added a new api for this

crates/context_aware_config/src/api/config/handlers.rs
pratikmishra356: i have separated data fetch from db and business logic on db separate
lets we need to do multiple operation on dimension list in a single handler , then we wont have to do db call everytime

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -921,3 +935,84 @@ async fn priority_recompute(
...
+        })
+        .collect::<superposition::Result<Vec<Context>>>()?;
+
+    let config_version_id =
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            let insert = diesel::insert_into(contexts)
+                .values(&update_contexts)
+                .on_conflict(id)
+                .do_update()
+                .set(weight.eq(excluded(weight)))
```
ayushjain17, line 997: here, we also need to update `last_modified_at` and `last_modified_by` columns also

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -208,3 +214,33 @@ async fn delete_dimension(
...
+
+    let _ = for (index, (dim, pri, _)) in results.iter().enumerate() {
+        diesel::update(dimensions)
+            .filter(dimension.eq(dim))
+            .filter(priority.eq(*pri))
```
ayushjain17, line 236: why do we need to check the priority value over here ?
is checking for dimension name not enough ?

crates/context_aware_config/src/helpers.rs
```diff
@@ -195,6 +207,44 @@ pub fn calculate_context_priority(
     }
 }
 
+fn calculate_weight_from_index(index: u32) -> Result<BigDecimal, String> {
+    let base = BigUint::from(2u32);
+    let result = base.pow(index);
+    let biguint_str = &result.to_str_radix(10);
```
ayushjain17, line 213: `let biguint_str = &result.to_str_radix(10);` and `BigDecimal::from_str_radix(&biguint_str, 10)` seem complementary, we could directly use `result` ?

what is the format of `result`, is it not in some desired format, because of which we need to to these conversions ?

crates/superposition_types/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 # env
 actix-web = { workspace = true, optional = true }
 anyhow = { workspace = true, optional = true }
+bigdecimal = { version = "0.3.1" , features= ["serde"]}
```
ayushjain17, line 12: can we move this to workspace, being used in `context-aware-config` as well

ayushjain17 (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
ayushjain17: but then do we want to start separating db calls to separate functions yet  ?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -921,3 +935,84 @@ async fn priority_recompute(
...
+                .execute(transaction_conn);
```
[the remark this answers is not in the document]
pratikmishra356: audit log is inserted using db functions , it does not relates with code execution 
rather gets triggered when db level row changes happens

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -208,3 +214,33 @@ async fn delete_dimension(
         ))
     }
 }
+
+#[put("/position/update")]
+async fn temp_position_update(
+    user: User,
+    db_conn: DbConnection,
+) -> superposition::Result<HttpResponse> {
```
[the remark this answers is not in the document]
pratikmishra356: responder is not satisfied for Value

crates/context_aware_config/src/helpers.rs
pratikmishra356: result is of type BigUint
from_str_radix returns type Bigdecimal

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -81,6 +83,14 @@ pub struct PriorityRecomputeResponse {
     pub new_priority: i32,
 }
 
+#[derive(Serialize)]
+pub struct WeightRecomputeResponse {
+    pub id: String,
+    pub condition: Condition,
+    pub old_weight: BigDecimal,
```
ayushjain17, line 90: i think we should not send BigDecimal in the api response, we should convert it to String and send, as Javascript breaks on very large numbers and it rounds them off

crates/context_aware_config/src/api/config/handlers.rs
ayushjain17: no harm, just wanted clarification here

ayushjain17 (approved):

ayushjain17 (approved):

PR #280, opened 2024-11-08
Support searching and sorting experiments via API and sorting on UI

Datron:
## Problem
It is difficult to find experiments on different filters and sort them

## Solution
- Provide an API interface that allows a user to search and sort experiments based on multiple parameters.
- Add support in the UI to sort columns, and in experiments listing add the ability to sort on last_modified_at and created_at

PS: more frontend changes are to be expected to support searching. That is NOT present in this PR.

https://github.com/user-attachments/assets/20eff0c3-f38e-4540-8a64-5938ec9f4b72


crates/superposition_types/src/lib.rs
```diff
@@ -187,6 +187,32 @@ pub struct PaginatedResponse<T> {
...
+        match self {
+            SortBy::Desc => SortBy::Asc,
+            SortBy::Asc => SortBy::Desc,
+        }
+    }
+}
+
+impl Default for SortBy {
+    fn default() -> Self {
+        Self::Desc
```
pratikmishra356, line 212: I feel default should be Asc

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -550,9 +552,31 @@ async fn list_experiments(
...
+        if let Some(ref states) = filters.status {
             builder = builder.filter(experiments::status.eq_any(states.0.clone()));
         }
+        if let Some(ref experiment_name) = filters.experiment_name {
+            builder =
+                builder.filter(experiments::name.like(format!("%{}%", experiment_name)));
+        }
+        if let Some(ref context_search) = filters.context {
+            builder = builder.filter(
+                sql::<Bool>("context::text LIKE ")
```
pratikmishra356, line 564: are we type casting here ? context

crates/superposition_types/src/lib.rs
Datron: @pratikmishra356 you always want to see the latest data, especially if you've been using Superposition for years.

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: Yes we are, this just lets me modify the context type so I can use LIKE on it. It's currently an experiment to see if this is a viable way to search through contexts

crates/superposition_types/src/lib.rs
pratikmishra356: @Datron but it can differ based on entity , position/weight needs to be sort in asc order

Datron: You can set it in the sort function based on needs, don't need to call default

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -633,56 +633,54 @@ async fn list_contexts(
...
-            .load::<Context>(&mut conn)?;
+    let (data, total_items) =
+        if dimension_params.len() > 0 || filter_params.prefix.is_some() {
+            let mut all_contexts: Vec<Context> = builder.load(&mut conn)?;
+            if let Some(prefix) = filter_params.prefix {
+                let prefix_list = HashSet::from_iter(prefix.0);
+                all_contexts = all_contexts
+                    .into_iter()
+                    .filter_map(|mut context| {
+                        Context::filter_keys_by_prefix(&context, &prefix_list)
```
pratikmishra356, line 647: @Datron filter_keys_by_prefix uses try_from which results in lots of error logs 
`2024-11-25T07:06:27Z ERROR superposition_types::config] Override validation error: Override is empty`
Should we instead check if map is empty and then probably try_from

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -633,56 +633,54 @@ async fn list_contexts(
     }
```
pratikmishra356, line 633: should we separate Asc, Desc like you did in experiment/list ?

crates/context_aware_config/src/api/context/handlers.rs
Datron: @pratikmishra356 yes we can, but that is out of scope for this PR

crates/context_aware_config/src/api/context/handlers.rs
Datron: What would be the benefit? Could you share an example?

ayushjain17: @Datron https://github.com/juspay/superposition/pull/288#discussion_r1853885506

crates/superposition_types/src/custom_query.rs
```diff
@@ -197,3 +198,46 @@ impl<'de> Deserialize<'de> for PaginationParams {
...
     }
 }
+
+#[derive(Debug, Deserialize, Clone, Deref)]
+#[deref(forward)]
+pub struct StringArgs(
+    #[serde(deserialize_with = "deserialize_stringified_list")] pub Vec<String>,
+);
+
+pub fn deserialize_stringified_list<'de, D, I>(
```
ayushjain17, line 208: can we instead implement the `Deserialize` trait on `StringArgs` and write the logic in it instead of writing this as a function

docker-compose/postgres/Dockerfile
```diff
@@ -1,4 +1,4 @@
 FROM public.ecr.aws/docker/library/postgres:12-alpine
 
 COPY ./db_init.sql /docker-entrypoint-initdb.d/db_init.sql
-# CMD ["postgres", "-c", "log_statement=all"]
```
ayushjain17, line 4: is there any change here ?

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -165,14 +163,34 @@ pub struct StatusTypes(
...
 pub struct ExpListFilters {
     pub status: Option<StatusTypes>,
     pub from_date: Option<DateTime<Utc>>,
     pub to_date: Option<DateTime<Utc>>,
+    pub experiment_name: Option<String>,
+    pub experiment_ids: Option<StringArgs>,
+    pub created_by: Option<StringArgs>,
+    pub context: Option<String>,
+    pub sort_on: Option<ExperimentSortOn>,
+    pub sort_by: Option<SortBy>,
```
ayushjain17, line 189: do we want to have this as a separate filter ?
if so, then shouldn't we change it for `context/list `as well
there I had coupled `sort_on` and `sort_by` to a single enum list

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -550,9 +551,31 @@ async fn list_experiments(
...
```
ayushjain17, line 563: what is this query doing here ?

crates/frontend/src/components/table.rs
```diff
@@ -29,8 +30,36 @@ pub fn table(
...
+                                    } => {
+                                        view! {
+                                            <th
+                                                class="uppercase"
+                                                style="cursor: pointer"
+                                                on:click=move |_| sort_fn.call(())
+                                            >
+                                                {column_name}
+                                                    {
+                                                        if current {
```
ayushjain17, line 48: what is the meaning of the this `current`, I mean what does it signify ( I couldn't get it) ?

crates/frontend/src/types.rs
```diff
@@ -122,10 +126,71 @@ pub struct ExperimentsResponse {
 pub struct StatusTypes(pub Vec<ExperimentStatusType>);
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct ExpListFilters {
+pub struct ExperimentListFilters {
```
ayushjain17, line 129: is this type not there in `experimentation_platform` for the api type ?

crates/frontend/src/pages/experiment_list/utils.rs
```diff
@@ -115,20 +144,58 @@ pub fn experiment_table_columns() -> Vec<Column> {
...
+                sort_fn: Callback::new(move |_| {
+                    let filters = filters_rs.get();
+                    let sort_by = filters.sort_by.unwrap_or_default().flip();
+                    let new_filters = ExperimentListFilters {
+                        sort_on: Some(ExperimentSortOn::CreatedAt),
+                        sort_by: Some(sort_by),
+                        ..filters
+                    };
+                    filters_ws.set(new_filters);
+                }),
```
ayushjain17, line 177: we dont need to update the filters here
`sort_fn` can simple take a parameter which says what is the current state here
based on that things can be changed and rendered

alternatively, we can also check for `current_sort_on == ExperimentSortOn::CreatedAt` just like its being done for current value

crates/frontend/src/pages/experiment_list/utils.rs
```diff
@@ -115,20 +144,58 @@ pub fn experiment_table_columns() -> Vec<Column> {
...
+            ColumnSortable::Yes {
+                sort_fn: Callback::new(move |_| {
+                    let filters = filters_rs.get();
+                    let sort_by = filters.sort_by.as_ref().map(|i| i.flip());
+                    let new_filters = ExperimentListFilters {
+                        sort_on: Some(ExperimentSortOn::LastModifiedAt),
+                        sort_by,
+                        ..filters
+                    };
+                    filters_ws.set(new_filters);
```
ayushjain17, line 194: then we wont need this `filters_rws`

ayushjain17 (changes requested):

crates/experimentation_platform/src/api/experiments/types.rs
Datron: We should change context/list as well, but that is out of scope of this PR. Regarding this being a separate filter, I think for experiments, overrides and functions we will have different filters, all others should work with the common filter type

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -548,11 +549,33 @@ async fn list_experiments(
...
+        }
+        if let Some(ref created_by) = filters.created_by {
+            builder =
+                builder.filter(experiments::created_by.eq_any(created_by.0.clone()));
+        }
+        if let Some(experiment_ids) = filters.experiment_ids.clone() {
+            let experiment_ids: HashSet<i64> = experiment_ids
+                .0
+                .iter()
+                .map(|i| i.parse::<i64>().unwrap_or_default())
```
ayushjain17, line 575: shouldn't we remove those experiment ids which are not proper integers, instead of searching for `0` ?
```suggestion
                .filter_map(|i| i.parse::<i64>())
```

crates/superposition_types/src/custom_query.rs
Datron: I tried Deserialize, it was more verbose, and all examples I found did it this way. Can we continue with this method for now?

crates/frontend/src/types.rs
Datron: It is present, do you recommend moving this to superposition_types?

docker-compose/postgres/Dockerfile
Datron: Yes, adding a way to log queries in postgres but have commented it out

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: Yeah your suggestion makes way more sense. Will make this change

crates/frontend/src/types.rs
ayushjain17: its good to have, or we can wait till this PR goes in:
https://github.com/juspay/superposition/pull/291

after this PR is merged, I would raise a separate PR for moving the api types to `superposition_types` probably we can move it then

crates/superposition_types/src/custom_query.rs
```diff
@@ -199,6 +200,49 @@ impl<'de> Deserialize<'de> for PaginationParams {
...
+                let p: I = I::deserialize(param.into_deserializer())?;
+                query_vector.push(p);
+            }
+            Ok(query_vector)
+        }
+    }
+
+    deserializer.deserialize_any(StringVecVisitor(std::marker::PhantomData::<I>))
+}
+
```
ayushjain17, line 245: this should work, right ?
```suggestion
#[derive(Debug, Clone, Deref)]
#[deref(forward)]
pub struct StringArgs(pub Vec<String>);

impl<'de> Deserialize<'de> for StringArgs {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let s = String::deserialize(deserializer)?;
        let items = s.split(',').map(|item| item.trim().to_string()).collect();

        Ok(Self(items))
    }
}```

File: crates/superposition_types/src/custom_query.rs (hunk 5 above)
Comment (ayushjain17) [reply crates/superposition_types/src/custom_query.rs:209]: let me know if this works:
https://github.com/juspay/superposition/pull/280/files#r1865864524

File: crates/superposition_types/src/custom_query.rs (hunk 14 above)
Comment (Datron) [reply crates/superposition_types/src/custom_query.rs:245]: In the case you present, if the user passes a wrong value then there is no error, which I want.

File: crates/frontend/src/types.rs [hunk 15]
```diff
@@ -122,10 +126,71 @@ pub struct ExperimentsResponse {
...
     pub status: Option<StatusTypes>,
     pub from_date: Option<DateTime<Utc>>,
     pub to_date: Option<DateTime<Utc>>,
+    pub all: Option<bool>,
+    pub page: Option<i64>,
+    pub count: Option<i64>,
```
ShubhranshuSanjeev, line 135: `PaginationParams` can be used for this, the filters can be separate from pagination parameters.

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -165,14 +163,34 @@ pub struct StatusTypes(
...
+pub struct ExperimentListFilters {
     pub status: Option<StatusTypes>,
     pub from_date: Option<DateTime<Utc>>,
     pub to_date: Option<DateTime<Utc>>,
+    pub experiment_name: Option<String>,
+    pub experiment_ids: Option<StringArgs>,
+    pub created_by: Option<StringArgs>,
+    pub context: Option<String>,
+    pub sort_on: Option<ExperimentSortOn>,
+    pub sort_by: Option<SortBy>,
```
ShubhranshuSanjeev, line 189:
```suggestion
    pub order: Option<Order>,
```

crates/frontend/src/components/table/types.rs
```diff
@@ -10,11 +11,22 @@ pub struct TableSettings {
     pub redirect_prefix: Option<String>,
 }
 
+#[derive(Clone)]
+pub enum ColumnSortable {
+    Yes {
+        sort_fn: Callback<()>,
+        sort_by: SortBy,
```
ShubhranshuSanjeev, line 18:
```suggestion
        order: Order,
```

crates/frontend/src/components/table/types.rs
```diff
@@ -10,11 +11,22 @@ pub struct TableSettings {
...
+        current: bool,
```
ShubhranshuSanjeev, line 19:
```suggestion
        sorted: bool,
```
I am assuming `current` is for whether the respective column is the one according to which the table data is sorted.

crates/frontend/src/components/table.rs
```diff
@@ -29,8 +30,35 @@ pub fn table(
...
+                                                            match sort_by {
+                                                                SortBy::Desc => view! { <i class="ri-arrow-down-s-line"></i> },
+                                                                SortBy::Asc => view! { <i class="ri-arrow-up-s-line"></i> },
+                                                            }
+                                                        } else {
+                                                            view! { <i class="ri-expand-up-down-line"></i> }
+                                                        }
```
ShubhranshuSanjeev, line 54:
```suggestion
match (current, sort_by) {
(false, _) => ...
(true, Asc) =>
(true, Desc) => ... 
}
```

ShubhranshuSanjeev (dismissed):

crates/experimentation_platform/src/api/experiments/types.rs
Datron: No

crates/frontend/src/components/table.rs
Datron: No

crates/frontend/src/components/table/types.rs
Datron: No

crates/frontend/src/components/table/types.rs
Datron: How about currently_sorted?

crates/frontend/src/pages/experiment_list/utils.rs
```diff
@@ -1,15 +1,42 @@
...
 use web_sys::MouseEvent;
 
-pub fn experiment_table_columns() -> Vec<Column> {
+#[derive(Copy, Display, Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum ExperimentSortOn {
+    #[strum(to_string = "last_modified_at")]
+    LastModifiedAt,
+    #[strum(to_string = "created_at")]
+    CreatedAt,
```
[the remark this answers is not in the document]
Datron: Didn't get you

crates/frontend/src/types.rs
Datron: Sure, will make that change

crates/superposition_types/src/custom_query.rs
```diff
@@ -199,6 +200,49 @@ impl<'de> Deserialize<'de> for PaginationParams {
     }
 }
 
+#[derive(Debug, Deserialize, Clone, Deref)]
+#[deref(forward)]
+pub struct StringArgs(
```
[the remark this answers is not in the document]
Datron: It converts comma separated strings into string vectors

crates/superposition_types/src/lib.rs
```diff
@@ -192,6 +192,32 @@ pub struct PaginatedResponse<T> {
     pub data: Vec<T>,
 }
 
+#[derive(
+    Debug, Serialize, Deserialize, Clone, PartialEq, PartialOrd, derive_more::Display,
+)]
+#[serde(rename_all = "lowercase")]
+pub enum SortBy {
+    #[display(fmt = "desc")]
```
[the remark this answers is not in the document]
Datron: rename_all is for deserialization, the to_string representation would be different me thinks

crates/experimentation_platform/src/api/experiments/types.rs
ShubhranshuSanjeev: Isn't `Asc` and `Desc` an order in which the data has to be sorted, `sort_by` sounds more like on what you want to sort on

crates/frontend/src/components/table.rs
ShubhranshuSanjeev: Why 😅 ?

crates/superposition_types/src/custom_query.rs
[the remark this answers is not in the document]
ayushjain17: this makes me realise we should rename the type

crates/superposition_types/src/custom_query.rs
ayushjain17: this does return an error, if the data passed is not a string, which is more than enough to determine if the value is right or not
as the data in it must contain is not mandatory, because there can always be a single element in the array which does not require any sorts of delimiter

crates/experimentation_platform/src/api/experiments/types.rs
Datron: I went with the way the operation is spoken about, you sort by descending order. You can also order by descending order, but the meaning is the same, so making the change is irrelevant

crates/frontend/src/components/table.rs
Datron: They are similar enough that this isn't justified

crates/frontend/src/components/table.rs
Datron: current is the column which is presently being sorted

crates/superposition_types/src/custom_query.rs
Datron: @ayushjain17 I tried this, but this function is used to deserialize a lot of other vectors and type cast them. I don't want to manually impl deserialize on all of them so I won't change this for now.

crates/frontend/src/pages/experiment_list/utils.rs
Datron: Discussed this offline and resolved

crates/experimentation_platform/src/api/experiments/types.rs
Datron: Solved with PaginationParams

crates/frontend/src/pages/experiment_list/utils.rs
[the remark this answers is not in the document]
ayushjain17: he meant why do we need `strum to_string` here, is `serde rename_all` not enough here ?
just like [here](https://github.com/juspay/superposition/pull/280/files#diff-a5028c4e12052ae506e13bf0d2503ab00c8e05e803b28930a5c26d289b373ec5R166-R171)

crates/frontend/src/api.rs
```diff
@@ -129,37 +107,25 @@ pub async fn delete_context(
...
-
-    let url = format!("{}/experiments?{}", host, query_params.join("&"));
+    let url = if pagination.is_empty() {
+        format!("{}/experiments?{}", host, filters.to_string())
+    } else {
+        format!(
+            "{}/experiments?{}&{}",
+            host,
+            filters.to_string(),
+            pagination
```
ayushjain17, line 126: can we make it consistent over here, I got confused that how are these two not having the exact same syntax
```suggestion
            filters.to_string(),
            pagination.to_string()
```

crates/frontend/src/api.rs
```diff
@@ -129,37 +107,25 @@ pub async fn delete_context(
...
+        )
```
ayushjain17, line 127: this does not seem correct, if someone does not send `filters`,  the url will become: `base_url/experimnets?&exp_filt1=abc`

crates/frontend/src/components/table.rs
```diff
@@ -65,8 +91,10 @@ pub fn table(
                                         .filter(|column| !column.hidden)
                                         .map(|column| {
                                             let cname = &column.name;
-                                            let value: String =
-                                                row.get(cname).unwrap_or(&Value::String(String::new())).html_display();
+                                            let value: String = row
+                                                .get(cname)
+                                                .unwrap_or(&Value::String(String::new()))
+                                                .html_display();
```
ayushjain17, line 97: can we rather do something like this:
```suggestion
                                            let value: String = row.get(cname).map(|v| v.html_display()).unwrap_or_default();
```

crates/frontend/src/api.rs
Datron: There are always date filters

crates/frontend/src/components/table.rs
Datron: Why the difference?

ayushjain17: its better memory wise, right
in the default case we just need a String, and thats what would be get directly

else, currently in default case, you are getting a `Value` type which again is converted to `String` by `html_display`

crates/frontend/src/api.rs
ayushjain17: that is not being guaranteed by the type
all fields in `ExperimentListFilters` are optional, so one can simply avoid sending any params and then it breaks over here

crates/frontend/src/components/table.rs
Datron: Will look into that, can we take it as it is now?

crates/frontend/src/api.rs
Datron: It is guaranteed by code, 
```
from_date: Utc.timestamp_opt(0, 0).single(),
        to_date: Utc.timestamp_opt(4130561031, 0).single(),
```
https://github.com/juspay/superposition/pull/280/files/64469a660bb91eabe003b378e3a0ec64d2de3961#diff-a26bd230250ef10be78fd6369216b71e556f5b808a9cd12e2be3b6551c6e8cbbR41

Doesn't need to be guaranteed by the type

ShubhranshuSanjeev (approved):

ayushjain17 (approved):

PR #283, opened 2024-11-14
fix: fixed and improved webhooks

mahatoankitkumar:
## Problem
Improvement in webhook event.

## Solution
Added tenant_id, config_version in the webhook response.
And added custom_headers in webhook object too.

### DIFF
--- a/crates/superposition/Superposition.cac.toml
+++ b/crates/superposition/Superposition.cac.toml
@@ -4,16 +4,14 @@ mandatory_dimensions = { "value" = [
 experiments_webhook_config = { "value" = { "enabled" = false }, "schema" = { "type" = "object", "properties" = { "enabled" = { "type" = "boolean" }, "configuration" = { "type" = "object", "properties" = { "url" = { "type" = "string" }, "method" = { "enum" = [
     "Post",
     "Get",
-], "type" = "string" }, "headers" = { "type" = "array", "items" = { "type" = "string", "enum" = [
-    "ConfigVersion",
+], "type" = "string" }, "custom_headers" = { "type" = "object" }, "service_headers" = { "type" = "array", "items" = { "type" = "string", "enum" = [
+    "ConfigVersion", "TenantId"
 ] } }, "authorization" = { "type" = "object", "properties" = { "key" = { "type" = "string" }, "value" = { "type" = "string" } }, "required" = [
     "key",
     "value",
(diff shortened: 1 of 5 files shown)


crates/superposition_types/src/webhook.rs
```diff
@@ -43,9 +50,17 @@ pub enum WebhookEvent {
     ExperimentConcluded,
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct EventInfo {
```
Datron, line 54:
```suggestion
pub struct WebhookEventInfo {
```

We may have other system events in the future, so wanted to namespace this now

crates/superposition_types/src/webhook.rs
```diff
@@ -43,9 +50,17 @@ pub enum WebhookEvent {
...
+    pub webhook_event: WebhookEvent,
+    pub time: NaiveDateTime,
```
Datron, line 56: Why not string, then you don't need to import chrono

crates/superposition/Superposition.cac.toml
```diff
@@ -23,6 +21,6 @@ tenant = { schema = { "type" = "string", "enum" = ["test", "dev"] } }
 
 [context."$tenant == 'dev'"]
 mandatory_dimensions = []
-experiments_webhook_config = { "enabled" = false, "configuration" = { "url" = "http://localhost:8080/config/test", "method" = "Get", "headers" = [
-    "ConfigVersion",
+experiments_webhook_config = { "enabled" = false, "configuration" = { "url" = "http://localhost:8080/config/test", "method" = "Get", "service_headers" = [
```
Datron, line 24: Shouldn't you define custom_headers here or in default_config?

mahatoankitkumar: Yeah i'll define it
But i have kept it not required, so would be fine to leave it also

crates/superposition_types/src/webhook.rs
mahatoankitkumar: Why string?
importing chrono will cause any issue?

Datron: Yeah superposition_types is used in all other places, so it increases the size, compile time, etc. If we can avoid it, since it's used only to store the timestamp which is a string at the end - then fine

Datron (approved):

PR #284, opened 2024-11-14
feat: added description and comment columns in all existing tables and apis 

sauraww:
## Problem
Added description and comment columns in all existing tables and apis 

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -380,10 +380,31 @@ fn construct_new_payload(
             },
         )?;
 
-    return Ok(web::Json(PutReq {
-        context: context,
+    let description = match res.get("description") {
+        Some(Value::String(s)) => Some(s.clone()),
+        Some(_) => {
+            log::error!("construct new payload: Description is not a valid string");
+            return Err(bad_argument!("Description must be a string"));
+        }
+        None => None,
+    };
+
+    // Handle change_reason
+    let change_reason = res
+        .get("change_reason")
+        .and_then(|val| val.as_str())
+        .map(|s| s.to_string())
+        .ok_or_else(|| {
+            log::error!("construct new payload: Change reason not present or invalid");
+            bad_argument!("Change reason is required and must be a string")
+        })?;
+
+    Ok(web::Json(PutReq {
+        context,
         r#override: override_,
-    }));
+        description,
+        change_reason,
+    }))
 }
 
 #[allow(clippy::too_many_arguments)]
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -198,6 +198,15 @@ fn create_ctx_from_put_req(
     tenant_config: &TenantConfig,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
+    let description = if req.description.is_none() {
+        let ctx_condition_value = json!(ctx_condition);
+        ensure_description(ctx_condition_value, conn)?
+    } else {
+        req.description
+            .clone()
+            .ok_or_else(|| bad_argument!("Description should not be empty"))?
+    };
+    let change_reason = req.change_reason.clone();
     let condition_val = Value::Object(ctx_condition.clone().into());
     let r_override = req.r#override.clone().into_inner();
     let ctx_override = Value::Object(r_override.clone().into());
@@ -228,6 +237,8 @@ fn create_ctx_from_put_req(
         last_modified_at: Utc::now().naive_utc(),
         last_modified_by: user.get_email(),
         weight,
+        description: description,
+        change_reason,
     })
 }
 
@@ -297,6 +308,8 @@ fn db_update_override(
             dsl::override_id.eq(ctx.override_id),
             dsl::last_modified_at.eq(Utc::now().naive_utc()),
             dsl::last_modified_by.eq(user.get_email()),
+            dsl::description.eq(ctx.description),
+            dsl::change_reason.eq(ctx.change_reason),
         ))
         .get_result::<Context>(conn)?;
     Ok(get_put_resp(update_resp))
@@ -307,6 +320,8 @@ fn get_put_resp(ctx: Context) -> PutResp {
         context_id: ctx.id,
         override_id: ctx.override_id,
         weight: ctx.weight,
+        description: ctx.description,
+        change_reason: ctx.change_reason,
     }
 }
 
@@ -320,7 +335,6 @@ pub fn put(
 ) -> superposition::Result<PutResp> {
     use contexts::dsl::contexts;
     let new_ctx = create_ctx_from_put_req(req, conn, user, tenant_config)?;
-
     if already_under_txn {
         diesel::sql_query("SAVEPOINT put_ctx_savepoint").execute(conn)?;
     }
@@ -345,6 +359,33 @@ pub fn put(
     }
 }
 
+fn ensure_description(
+    context: Value,
+    transaction_conn: &mut diesel::PgConnection,
+) -> Result<String, superposition::AppError> {
+    use superposition_types::database::schema::contexts::dsl::{
+        contexts as contexts_table, id as context_id,
+    };
+
+    let context_id_value = hash(&context);
+
+    // Perform the database query
+    let existing_context = contexts_table
+        .filter(context_id.eq(context_id_value))
+        .first::<Context>(transaction_conn);
+
+    match existing_context {
+        Ok(ctx) => Ok(ctx.description), // If the context is found, return the description
+        Err(diesel::result::Error::NotFound) => Err(superposition::AppError::NotFound(
+            "Description not found in the existing context".to_string(),
+        )),
+        Err(e) => {
+            log::error!("Database error while fetching context: {:?}", e);
+            Err(superposition::AppError::DbError(e)) // Use the `DbError` variant for other Diesel-related errors
+        }
+    }
+}
+
 #[put("")]
 async fn put_handler(
     state: Data<AppState>,
@@ -356,18 +397,43 @@ async fn put_handler(
     tenant_config: TenantConfig,
 ) -> superposition::Result<HttpResponse> {
     let tags = parse_config_tags(custom_headers.config_tags)?;
+
     let (put_response, version_id) = db_conn
         .transaction::<_, superposition::AppError, _>(|transaction_conn| {
-            let put_response =
-                put(req, transaction_conn, true, &user, &tenant_config, false).map_err(
-                    |err: superposition::AppError| {
-                        log::info!("context put failed with error: {:?}", err);
-                        err
-                    },
-                )?;
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let mut req_mut = req.into_inner();
+
+            // Use the helper function to ensure the description
+            if req_mut.description.is_none() {
+                req_mut.description = Some(ensure_description(
+                    Value::Object(req_mut.context.clone().into_inner().into()),
+                    transaction_conn,
+                )?);
+            }
+            let put_response = put(
+                Json(req_mut.clone()),
+                transaction_conn,
+                true,
+                &user,
+                &tenant_config,
+                false,
+            )
+            .map_err(|err: superposition::AppError| {
+                log::info!("context put failed with error: {:?}", err);
+                err
+            })?;
+            let description = req_mut.description.unwrap_or_default();
+            let change_reason = req_mut.change_reason;
+
+            let version_id = add_config_version(
+                &state,
+                tags,
+                transaction_conn,
+                description,
+                change_reason,
+            )?;
             Ok((put_response, version_id))
         })?;
+
     let mut http_resp = HttpResponse::Ok();
 
     http_resp.insert_header((
@@ -396,14 +462,32 @@ async fn update_override_handler(
     let tags = parse_config_tags(custom_headers.config_tags)?;
     let (override_resp, version_id) = db_conn
         .transaction::<_, superposition::AppError, _>(|transaction_conn| {
-            let override_resp =
-                put(req, transaction_conn, true, &user, &tenant_config, true).map_err(
-                    |err: superposition::AppError| {
-                        log::info!("context put failed with error: {:?}", err);
-                        err
-                    },
-                )?;
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let mut req_mut = req.into_inner();
+            if req_mut.description.is_none() {
+                req_mut.description = Some(ensure_description(
+                    Value::Object(req_mut.context.clone().into_inner().into()),
+                    transaction_conn,
+                )?);
+            }
+            let override_resp = put(
+                Json(req_mut.clone()),
+                transaction_conn,
+                true,
+                &user,
+                &tenant_config,
+                true,
+            )
+            .map_err(|err: superposition::AppError| {
+                log::info!("context put failed with error: {:?}", err);
+                err
+            })?;
+            let version_id = add_config_version(
+                &state,
+                tags,
+                transaction_conn,
+                req_mut.description.unwrap().clone(),
+                req_mut.change_reason.clone(),
+            )?;
             Ok((override_resp, version_id))
         })?;
     let mut http_resp = HttpResponse::Ok();
@@ -431,6 +515,17 @@ fn r#move(
 ) -> superposition::Result<PutResp> {
     use contexts::dsl;
     let req = req.into_inner();
+
+    let ctx_condition = req.context.to_owned().into_inner();
+    let ctx_condition_value = Value::Object(ctx_condition.clone().into());
+    let description = if req.description.is_none() {
+        ensure_description(ctx_condition_value.clone(), conn)?
+    } else {
+        req.description
+            .ok_or_else(|| bad_argument!("Description should not be empty"))?
+    };
+
+    let change_reason = req.change_reason.clone();
     let ctx_condition = req.context.to_owned().into_inner();
     let ctx_condition_value = Value::Object(ctx_condition.clone().into());
     let new_ctx_id = hash(&ctx_condition_value);
@@ -471,6 +566,8 @@ fn r#move(
         last_modified_at: Utc::now().naive_utc(),
         last_modified_by: user.get_email(),
         weight,
+        description: description,
+        change_reason,
     };
 
     let handle_unique_violation =
@@ -534,7 +631,14 @@ async fn move_handler(
                 log::info!("move api failed with error: {:?}", err);
                 err
             })?;
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let version_id = add_config_version(
+                &state,
+                tags,
+                transaction_conn,
+                move_response.description.clone(),
+                move_response.change_reason.clone(),
+            )?;
+
             Ok((move_response, version_id))
         })?;
     let mut http_resp = HttpResponse::Ok();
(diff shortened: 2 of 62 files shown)


crates/context_aware_config/migrations/2024-11-06-123105_add_description_and_comment_columns/down.sql
```diff
@@ -0,0 +1,19 @@
+-- This file should undo anything in `up.sql`
+-- contexts table
```
mahatoankitkumar, line 2: Are these needed?

mahatoankitkumar (approved):

crates/context_aware_config/migrations/2024-11-06-123105_add_description_and_comment_columns/down.sql
sauraww: No , I will remove it

mahatoankitkumar (changes requested):
Lets make the fields non-nullable, as per the discussion.

crates/context_aware_config/migrations/2024-11-06-123105_add_description_and_comment_columns/up.sql
```diff
@@ -0,0 +1,20 @@
...
+ALTER TABLE public.dimensions ADD COLUMN IF NOT EXISTS description TEXT DEFAULT '' NOT NULL;
+ALTER TABLE public.dimensions ADD COLUMN IF NOT EXISTS comment TEXT DEFAULT '' NOT NULL;
+
+ALTER TABLE public.default_configs ADD COLUMN IF NOT EXISTS description TEXT DEFAULT '' NOT NULL;
+ALTER TABLE public.default_configs ADD COLUMN IF NOT EXISTS comment TEXT DEFAULT '' NOT NULL;
+
+ALTER TABLE public.type_templates ADD COLUMN IF NOT EXISTS description TEXT DEFAULT '' NOT NULL;
+ALTER TABLE public.type_templates ADD COLUMN IF NOT EXISTS comment TEXT DEFAULT '' NOT NULL;
+
+ALTER TABLE public.functions RENAME COLUMN function_description TO description;
```
pratikmishra356, line 17: why function does not have `comment`

crates/context_aware_config/migrations/2024-11-06-123105_add_description_and_comment_columns/up.sql
ayushjain17, line 1: why is comment and description mandatory ?
shouldn't they be totally optional ?
`DEFAULT ''` would also not be needed then

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -378,9 +378,25 @@ fn construct_new_payload(
...
+        .expect("Description not present")
+        .as_str()
+        .expect("Description is not a string")
+        .to_string();
+
+    let comment = res
+        .get("comment")
+        .expect("Comment not present")
+        .as_str()
+        .expect("Comment is not a string")
```
ayushjain17, line 392: do not use `expect` in `handlers` in it will `panic`

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -207,6 +207,8 @@ fn create_ctx_from_put_req(
     tenant_config: &TenantConfig,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
+    let description = req.description.to_owned();
+    let comment = req.comment.clone();
```
ayushjain17, line 211: these are just used once, you shouldn't require to clone them

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -445,6 +449,8 @@ fn r#move(
 ) -> superposition::Result<PutResp> {
     use contexts::dsl;
     let req = req.into_inner();
+    let description = req.description.clone();
+    let comment = req.comment.clone();
```
ayushjain17, line 453: same here, cloning should not be required here

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -6,12 +6,16 @@ use superposition_types::{Cac, Condition, Overrides};
 pub struct PutReq {
     pub context: Cac<Condition>,
     pub r#override: Cac<Overrides>,
+    pub description: String,
+    pub comment: String,
```
ayushjain17, line 10: shouldn't these be optional parameters ?

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -77,6 +77,8 @@ async fn create(
             ))
         }
     };
+    let description = req.description.clone();
+    let comment = req.comment.clone();
```
ayushjain17, line 81: here as well

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -6,12 +6,16 @@ use superposition_types::{Cac, Condition, Overrides};
...
 }
 
 #[cfg_attr(test, derive(Debug, PartialEq))] // Derive traits only when running tests
 #[derive(Deserialize, Clone)]
 pub struct MoveReq {
     pub context: Cac<Condition>,
+    pub description: String,
+    pub comment: String,
```
ayushjain17, line 18: same

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -10,6 +10,8 @@ pub struct CreateReq {
     pub schema: Option<Map<String, Value>>,
     #[serde(default, deserialize_with = "deserialize_option")]
     pub function_name: Option<Value>,
+    pub description: String,
+    pub comment: String,
```
ayushjain17, line 14: same, shouldn't these be optional ?

ayushjain17 (changes requested):
Did not comment in the all the places, had these issues across:
1. there are multiple place where `req.comment` and `req.description` are being used once but are still cloned?
2. shouldn't the input for `comment` and `description` be optional in all the APIs ?
3. I am not sure if `comment` as a column makes sense in the table when the data stored in the tables are not transactional. We will always be overwriting the `comment` column with the new data, which makes it behave exactly same as the `description` column. `comment` as a column only makes when if we add it AuditLog/EventLog table, where we are storing things as a transaction and then we can store the comment as a form of reason for that action (change-log) in particular. Here, having `comment` as mandatory makes sense for both the db type as well in the APIs.
4. Also, even if we intend to make `description` as mandatory, it should be mandatory in the db type and while creating an entity and not while updating it. While updating if the user forgets to give the input for `description` this would simply overwrite the column and make it empty, which is incorrect. It should not be needed again and again, as the `description` field serves to give a meaning to the existence of an entry in the db in words, this would not be intend to be updated every-time while making a change.

crates/context_aware_config/migrations/2024-11-06-123105_add_description_and_comment_columns/down.sql
ayushjain17, line 1: ### Highlighting point 4 from my previous review's summary

Also, even if we intend to make `description` as mandatory, it should be mandatory in the db type and while creating an entity and not while updating it. While updating if the user forgets to give the input for `description` this would simply overwrite the column and make it empty, which is incorrect. It should not be needed again and again, as the `description` field serves to give a meaning to the existence of an entry in the db in words, this would not be intend to be updated every-time while making a change.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -735,12 +827,26 @@ async fn delete_context(
...
+    };
     let ctx_id = path.into_inner();
     let tags = parse_config_tags(custom_headers.config_tags)?;
     let version_id =
         db_conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
-            delete_context_api(ctx_id, user, transaction_conn)?;
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let context = contexts_table
+                .filter(context_id.eq(ctx_id.clone()))
+                .first::<Context>(transaction_conn)?;
```
pratikmishra356, line 839: should we return 404 or 204

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -914,7 +1063,19 @@ async fn priority_recompute(
                 .do_update()
                 .set(priority.eq(excluded(priority)))
                 .execute(transaction_conn);
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            // Extract description and change_reason from the response
+            let description = response
+                .first()
+                .map(|r| r.description.clone())
+                .unwrap_or_else(|| "Default description".to_string());
+            let change_reason = "Priority recomputation".to_string(); // Use a constant or computed value
```
pratikmishra356, line 1071: shouldn't change reason come from api request

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -80,23 +80,51 @@ async fn create(
 
     let result = fetch_default_key(&key, &mut conn);
 
-    let (value, schema, function_name, created_at_val, created_by_val) = match result {
+    let description = match &result {
```
pratikmishra356, line 83: Ideally we should have separated create and update api

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -176,27 +220,26 @@ async fn delete_dimension(
 ) -> superposition::Result<HttpResponse> {
     let name: String = path.into_inner().into();
     let DbConnection(mut conn) = db_conn;
-    let dimension_data: Dimension = dimensions::dsl::dimensions
-        .filter(dimensions::dimension.eq(&name))
+    let dimension_data = dimension_table
```
pratikmishra356, line 223: there should be change reason for deleting dimension

pratikmishra356 (changes requested):

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -11,6 +11,8 @@ pub struct CreateReq {
     pub schema: Value,
     #[serde(default, deserialize_with = "deserialize_option")]
     pub function_name: Option<Value>,
+    pub description: Option<String>,
```
Datron, line 14: Is create and update combined here?

Datron (changes requested):
- Can you use map_err and `?` instead of match statements to unwrap Result
- Set a character limit to description and change_reason (not in database but in backend & frontend)
- Re-consider using textarea for input
- nitpick and out of scope (need not be done): separate out create and update for APIs so description is mandatory for creates but not updates

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -735,12 +827,26 @@ async fn delete_context(
...
     let version_id =
         db_conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
-            delete_context_api(ctx_id, user, transaction_conn)?;
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let context = contexts_table
+                .filter(context_id.eq(ctx_id.clone()))
+                .first::<Context>(transaction_conn)?;
+            delete_context_api(ctx_id.clone(), user.clone(), transaction_conn)?;
+            let description = context.description;
+            let change_reason = format!("Deleted context by {}", user.username);
```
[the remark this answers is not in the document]
sauraww: But in the delete request we don't have a reqBody . 
We can leave it for now .

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -74,6 +78,44 @@ async fn create(
...
+                log::error!(
+                    "No description provided and no existing dimension for {}.",
+                    dimension_name
+                );
+                return Err(bad_argument!(
+                    "No description provided for the new dimension {}.",
+                    dimension_name
+                ));
+            }
+        }
```
ayushjain17, line 116: this is should ideally be in a transaction, lets separate out the create and update requests, its difficult to verify these changes
separating these out will make all these logics way simpler

in insert, description is mandatory => create entity with the request as is
in update, description is optional => if description is sent use it, else fetch from db and use the old description in the update query

ayushjain17 (changes requested):
partial review

crates/context_aware_config/src/api/dimension/types.rs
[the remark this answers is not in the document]
sauraww: the same one was used for updating

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -374,18 +404,43 @@ async fn put_handler(
...
+                true,
+                &user,
+                &tenant_config,
+                false,
+            )
+            .map_err(|err: superposition::AppError| {
+                log::info!("context put failed with error: {:?}", err);
+                err
+            })?;
+            let description = req_mut.description.unwrap_or_default();
```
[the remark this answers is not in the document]
sauraww: We are taking this out for add_config_version , we need to pass description and change_reason there also.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -414,14 +469,32 @@ async fn update_override_handler(
...
+            )
+            .map_err(|err: superposition::AppError| {
+                log::info!("context put failed with error: {:?}", err);
+                err
+            })?;
+            let version_id = add_config_version(
+                &state,
+                tags,
+                transaction_conn,
+                req_mut.description.unwrap().clone(),
```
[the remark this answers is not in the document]
sauraww: unwrap is safe as we have already ensured that description is going to be there in the above code.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -356,18 +393,43 @@ async fn put_handler(
...
-                    |err: superposition::AppError| {
-                        log::info!("context put failed with error: {:?}", err);
-                        err
-                    },
-                )?;
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let mut req_mut = req.into_inner();
+
+            // Use the helper function to ensure the description
+            if req_mut.description.is_none() {
```
pratikmishra356, line 402: here can we do like this ,
let description = either from request_mut or if it is none then from ensure_description , also assign in req_mut.description
so that here description will be of type <string> and we will not have to use unwrap_or_default
also declare change_reason here only, so that we wont have to cline the whole req_mut

crates/context_aware_config/src/api/context/handlers.rs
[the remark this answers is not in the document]
pratikmishra356: i have suggested another approach aboce , 
yes from runtime i am making sure that description is there , but still i want to avoid this

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -431,6 +511,16 @@ fn r#move(
 ) -> superposition::Result<PutResp> {
     use contexts::dsl;
     let req = req.into_inner();
+
+    let ctx_condition = req.context.to_owned().into_inner();
+    let ctx_condition_value = Value::Object(ctx_condition.clone().into());
+    let description = if req.description.is_none() {
+        ensure_description(ctx_condition_value.clone(), conn)?
+    } else {
+        req.description.expect("Description should not be empty")
```
pratikmishra356, line 520: I know this will work , but if we can use match expression or abstract the some part without expect or unwrap 
would be great

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -864,35 +1025,51 @@ async fn weight_recompute(
...
+                    .execute(transaction_conn)
+                    .map_err(|err| {
                         log::error!(
                             "Failed to execute query while recomputing weight, error: {err}"
                         );
                         db_error!(err)
                     })?;
             }
-            let version_id = add_config_version(&state, tags, transaction_conn)?;
+            let description = contexts_new_weight.iter().map(|(_, _, description, _)| description.clone()).collect::<Vec<String>>().join(",");
```
pratikmishra356, line 1065: instead of using change_reason and description of all contexts,
lets add only this 
description: weight recomputed
change_reason: weight recomputed

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -380,10 +380,31 @@ fn construct_new_payload(
...
 
-    return Ok(web::Json(PutReq {
-        context: context,
+    let description = match res.get("description") {
+        Some(Value::String(s)) => Some(s.clone()),
+        Some(_) => {
+            log::error!("construct new payload: Description is not a valid string");
+            return Err(bad_argument!("Description must be a string"));
+        }
+        None => None,
```
Datron, line 389: @sauraww `construct_new_payload` function is used by reduce, I don't know if we can just get the description like this. I think we'll need to combine all the descriptions being reduced into one or 

insert a new description

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -198,6 +198,15 @@ fn create_ctx_from_put_req(
     tenant_config: &TenantConfig,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
+    let description = if req.description.is_none() {
+        let ctx_condition_value = Value::Object(ctx_condition.clone().into());
+        ensure_description(ctx_condition_value, conn)?
+    } else {
+        req.description
+            .clone()
+            .expect("Description should not be empty")
```
Datron, line 207: No expects, throw an error here

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -198,6 +198,15 @@ fn create_ctx_from_put_req(
     tenant_config: &TenantConfig,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
+    let description = if req.description.is_none() {
+        let ctx_condition_value = Value::Object(ctx_condition.clone().into());
```
Datron, line 202: nitpick: would json!(ctx_condition) work?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -345,6 +359,33 @@ pub fn put(
...
+    // Perform the database query
+    let existing_context = contexts_table
+        .filter(context_id.eq(context_id_value))
+        .first::<Context>(transaction_conn);
+
+    match existing_context {
+        Ok(ctx) => Ok(ctx.description), // If the context is found, return the description
+        Err(diesel::result::Error::NotFound) => Err(superposition::AppError::NotFound(
+            "Description not provided in existing context".to_string(),
+        )),
```
Datron, line 381: NotFound error message is confusing. A context was not found, but this message indicates only the description is missing. Probably based on the message the error code should be bad request

crates/context_aware_config/src/api/context/handlers.rs
Datron: expect will cause a panic, this error message won't be logged properly as well. We should throw an AppError here

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -765,12 +885,39 @@ async fn bulk_operations(
...
+
+                        let description = if put_req.description.is_none() {
+                            ensure_description(
+                                ctx_condition_value.clone(),
+                                transaction_conn,
+                            )?
+                        } else {
+                            put_req
+                                .description
+                                .expect("Description should not be empty")
```
Datron, line 901: No `expect`

crates/superposition_types/src/database/diesel.toml
```diff
@@ -1,6 +1,6 @@
 [print_schema]
 file = "schema.rs"
-patch_file = "schema.patch"
+#patch_file = "schema.patch"
```
Datron, line 3: we need schema.patch right?

Datron (approved):
Main concern: the error message for a context description not being found.

Nitpicks: no expect calls though they are safe since it throws off clippy checks

crates/context_aware_config/src/api/config/handlers.rs
sauraww: Are you suggesting that we take it as input in the reduce api itself ?

pratikmishra356 (approved):

PR #285, opened 2024-11-19
fix: properly parsing numbers that are strings

Datron:
## Problem
In cases where a user enters as input a string which is a number, our parsing fails

## Solution
add support for it

### DIFF
--- a/crates/frontend/src/components/context_form/utils.rs
+++ b/crates/frontend/src/components/context_form/utils.rs
@@ -7,6 +7,7 @@ use crate::{
     },
 };
 use anyhow::Result;
+use leptos::logging;
 use serde_json::{json, Map, Value};
 
 pub fn get_condition_schema(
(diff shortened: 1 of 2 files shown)


crates/frontend/src/utils.rs
```diff
@@ -313,6 +313,7 @@ pub fn parse_value(val: &Value, config_type: ConfigValueType) -> Result<Value, S
 
         ConfigValueType::String => match val {
             Value::String(_) => Ok(val.clone()),
+            Value::Number(i) => Ok(Value::String(i.to_string())),
```
ayushjain17, line 316: can we also add for bool as well

crates/frontend/src/components/context_form/utils.rs
```diff
@@ -186,7 +187,15 @@ pub fn construct_context(
     } else {
         let condition_schemas = conditions
             .iter()
-            .map(|condition| get_condition_schema(condition, dimensions.clone()).unwrap())
+            .map(
```
ayushjain17, line 190: shouldn't this be a filter_map ?
for `Ok(ans)` we can return `Some(ans)`
and for `Err` we can return `None` directly, returning `Null` [here](https://github.com/juspay/superposition/pull/285/files#diff-564227801c44858d95faa622d6194be18193e660848f20a138956b1fd2a15efbR195) seemed wrong

ayushjain17 (changes requested):

crates/frontend/src/utils.rs
Datron: Sure makes sense

crates/frontend/src/components/context_form/utils.rs
Datron: I want the parsing to fail with Null, I don't want to filter out failed values and still proceed. I want to fail on wrong inputs

mahatoankitkumar (approved):

ayushjain17 (approved):

PR #286, opened 2024-11-19
feat: db model types migration

ayushjain17:
## Problem
Model types are very tightly tied down to cac and exp crates via diesel, due to which usage of these types in any other crate requires redefining the type again in that crate

## Solution
Move out the migrations to separate crates (`cac_db_config` and `experimentation_db_config`) for migrations only, which include the corresponding schema.rs file
using this schema.rs defined the models in `superposition_types` which conditionally implement the diesel traits, when `diesel_derives` feature is enabled

## Note
The change-log seems big, because the migration files (sql files) where moved along with few other files which increased the number of effected files

### This PR is not complete, it only introduces the separation, the usage of the types will be raised a separate PR


crates/cac_db_config/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "cac_db_config"
```
ShubhranshuSanjeev, line 2: The package name can be just `cac_db`

ShubhranshuSanjeev: Same for `experimentation_db_config` -> `experimentation_db`

crates/superposition_types/Cargo.toml
```diff
@@ -8,23 +8,34 @@ edition = "2021"
...
 
 [features]
 default = ["server"]
-diesel_derives = ["dep:diesel", "dep:superposition_derives"]
+diesel_derives = [
+    "dep:diesel",
+    "dep:superposition_derives",
+    "dep:cac_db_config",
+    "cac_db_config/diesel_derives",
+    "experimentation_db_config?/diesel_derives",
```
ShubhranshuSanjeev, line 33: What does `?` do here?

crates/superposition_types/src/cac_models.rs
ShubhranshuSanjeev, line 1: The models can structured as such `models/cac.rs` and `modes/exp.rs`

ShubhranshuSanjeev (changes requested):

crates/superposition_types/Cargo.toml
ayushjain17: it is used while enabling certain feature of another dependency which in turn in optional and is dependent on some other feature to enable the usage of that dependency
check the section before the second Note [here](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features)

Datron (approved):

Datron (approved):

PR #296, opened 2024-12-03
fix: webhook optional field

mahatoankitkumar:
## Problem
Make custom_headers and service_headers optional

## Solution

### DIFF
--- a/crates/service_utils/src/helpers.rs
+++ b/crates/service_utils/src/helpers.rs
@@ -429,6 +429,7 @@ where
     let mut header_array = webhook_config
         .service_headers
         .clone()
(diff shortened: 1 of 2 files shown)


crates/service_utils/src/helpers.rs
```diff
@@ -15,6 +15,7 @@ use serde::{
 };
 use serde_json::{Map, Value};
 use std::{
+    collections::HashMap,
```
ayushjain17, line 18: this too wont be needed

ayushjain17 (approved):

ShubhranshuSanjeev (approved):

PR #297, opened 2024-12-04
feat: JsonLogic algebraic representation

ShubhranshuSanjeev:
## Problem
No concrete type representation for JsonLogic in Rust, leading to clumsy parsing and custom logic to generate context conditions.

## Solution
This PR introduces types such as:
- Operands
- Operator
- Condition

which aims towards representing JsonLogic correctly in an algebraic fashion.
Along with the type, behaviors such as generating context JsonLogic, conversion from JsonLogic to these types are also added in this PR.

Alongside, generating forms from JsonLogic is handled in this PR, additionally logic to parse based on operators all defined at one place, for ease of use.

### DIFF
--- a/crates/frontend/src/logic.rs
+++ b/crates/frontend/src/logic.rs
@@ -1,0 +1,357 @@
+use std::fmt::Display;
+
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Map};
+
+use crate::schema::{HtmlDisplay, SchemaType};
+use superposition_types::Context;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum Expression {
+    Is(serde_json::Value),
+    In(serde_json::Value),
+    Has(serde_json::Value),
+    Between(serde_json::Value, serde_json::Value),
+    Other(String, Vec<serde_json::Value>),
+}
+
+pub fn is_variable(v: &serde_json::Value) -> bool {
+    v.is_object() && v.as_object().unwrap().contains_key("var")
+}
+
+pub fn is_constant(v: &serde_json::Value) -> bool {
+    !is_variable(v)
+}
+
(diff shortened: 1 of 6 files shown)


crates/frontend/src/logic.rs
```diff
@@ -0,0 +1,278 @@
...
+use serde_json::{json, Map, Value};
+
+use crate::schema::{HtmlDisplay, SchemaType};
+use derive_more::{Deref, DerefMut};
+use superposition_types::Context;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Operand {
+    Value(serde_json::Value),
+    Dimension(serde_json::Value),
```
Datron, line 13: dimension will always be a string, can it be typed like that?

crates/frontend/src/schema.rs
```diff
@@ -81,14 +81,18 @@ pub enum SchemaType {
 impl SchemaType {
     pub fn default_value(&self) -> Value {
         match self {
-            SchemaType::Multiple(_) => json!(""),
-            SchemaType::Single(JsonSchemaType::String) => json!(""),
+            SchemaType::Multiple(_) => Value::String(String::default()),
```
Datron, line 84: Can we write a comment on how to use SchemaType::Mulitple with some documentation at the type location?

Datron (approved):

crates/frontend/src/logic.rs
Datron: Would it be better to take it out? Since "var" is useless in parsing later on

crates/frontend/src/schema.rs
Datron: @ShubhranshuSanjeev this definitely needs documentation. I assumed this represented a way to define multiple schemas for a type

PR #299, opened 2024-12-06
Replace priority with position

pratikmishra356:
## Problem
Remove priority and replace it with position in dimension table and weight in context table
only get config api will have both priority and weight in the response
weight in get config api response will contain the order of that context sorted by actual weights

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+        .get_result::<Dimension>(&mut conn)?;
+
+    let update_req = req.into_inner();
+
+    if let Some(schema_value) = update_req.schema {
+        validate_jsonschema(&state.meta_schema, &schema_value)?;
+
+        let schema_compile_result = JSONSchema::options()
+            .with_draft(Draft::Draft7)
+            .compile(&schema_value);
```
ayushjain17, line 163: this JSONSchema creation is common in a lot of places along with `validate_jsonschema`, should we not couple these to avoid missing out either of this

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+            log::error!("Expected a string or null as the function name.");
+            return Err(bad_argument!(
+                "Expected a string or null as the function name."
+            ));
         }
-        Err(e) => {
-            log::error!("Dimension upsert failed with error: {e}");
-            Err(unexpected_error!(
-                "Something went wrong, failed to create/update dimension"
+    };
```
ayushjain17, line 183: why is the type of `function_name` in update request as `option<Value>`, it should simply be `option<String>`

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+                    dimensions::function_name.eq(dimension_row.function_name),
+                    dimensions::schema.eq(dimension_row.schema),
+                    dimensions::position.eq(new_position),
+                ))
+                .get_result::<Dimension>(transaction_conn)?;
+            let is_mandatory = tenant_config
+                .mandatory_dimensions
+                .contains(&result.dimension);
+            Ok(HttpResponse::Created()
+                .json(DimensionWithMandatory::new(result, is_mandatory)))
```
ayushjain17, line 228: why is the response `Created` here ?
we are just updating in this api, create should not be possible here

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+    db_conn: DbConnection,
+    tenant_config: TenantConfig,
+) -> superposition::Result<HttpResponse> {
+    let name: String = path.clone().into();
+    use dimensions::dsl;
+    let DbConnection(mut conn) = db_conn;
+
+    let mut dimension_row: Dimension = dsl::dimensions
+        .filter(dimensions::dimension.eq(name.clone()))
+        .get_result::<Dimension>(&mut conn)?;
```
ayushjain17, line 154: ideally, this should also be in the transaction, right ?

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
-        Err(e) => {
-            log::error!("Dimension upsert failed with error: {e}");
-            Err(unexpected_error!(
-                "Something went wrong, failed to create/update dimension"
+    };
+    dimension_row.function_name = fun_name.clone();
+
+    if let Some(position_val) = update_req.position.clone() {
+        let new_position: i32 = position_val.clone().into();
+        validate_dimension_position(path.into_inner(), position_val)?;
```
ayushjain17, line 188: should this is be a part of `UpdateRequest` deserialiser itself ?

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+                dimensions::schema.eq(dimension_row.schema),
             ))
-        }
+            .get_result::<Dimension>(&mut conn)?;
+        let is_mandatory = tenant_config
+            .mandatory_dimensions
+            .contains(&result.dimension);
+        Ok(HttpResponse::Created()
+            .json(DimensionWithMandatory::new(result, is_mandatory)))
     }
```
ayushjain17, line 245: we can optimise this repetition here,
we can simply set the position here as well, only difference being the if the user is sending a new value then that would be used, else whatever previously value which was earlier found you can use that itself

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+                diesel::update(dsl::dimensions)
+                    .filter(dimensions::position.gt(dimension_row.position))
+                    .filter(dimensions::position.le(new_position.clone()))
+                    .set((
+                        dsl::last_modified_at.eq(Utc::now().naive_utc()),
+                        dsl::last_modified_by.eq(user.get_email()),
+                        dimensions::position.eq(dimensions::position - 1),
+                    ))
+                    .execute(transaction_conn)?
+            } else {
```
ayushjain17, line 202: if previous_position == new_position then we don't need to execute any position update query
```suggestion
            } else if  previous_position > new_position {
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+
+    if let Some(position_val) = update_req.position.clone() {
+        let new_position: i32 = position_val.clone().into();
+        validate_dimension_position(path.into_inner(), position_val)?;
+
+        let previous_position = dimension_row.position.clone();
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            if previous_position < new_position {
+                diesel::update(dsl::dimensions)
+                    .filter(dimensions::position.gt(dimension_row.position))
```
ayushjain17, line 194: to avoid confusion, can we rather use this
```suggestion
                    .filter(dimensions::position.gt(previous_position)
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+    };
+    dimension_row.function_name = fun_name.clone();
+
+    if let Some(position_val) = update_req.position.clone() {
+        let new_position: i32 = position_val.clone().into();
+        validate_dimension_position(path.into_inner(), position_val)?;
+
+        let previous_position = dimension_row.position.clone();
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            if previous_position < new_position {
```
ayushjain17, line 192: we also need to check that the position given is not greater than the already existing number of dimensions

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -50,6 +54,10 @@ async fn create(
     let create_req = req.into_inner();
     let schema_value = create_req.schema;
 
+    validate_dimension_position(
```
ayushjain17, line 57: we also need to check that the position given is not greater than the already existing number of dimensions

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+                    .filter(dimensions::position.le(new_position.clone()))
+                    .set((
+                        dsl::last_modified_at.eq(Utc::now().naive_utc()),
+                        dsl::last_modified_by.eq(user.get_email()),
+                        dimensions::position.eq(dimensions::position - 1),
+                    ))
+                    .execute(transaction_conn)?
+            } else {
+                diesel::update(dsl::dimensions)
+                    .filter(dimensions::position.lt(dimension_row.position))
```
ayushjain17, line 204:
```suggestion
                    .filter(dimensions::position.lt(previous_position))
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+    if let Some(position_val) = update_req.position.clone() {
+        let new_position: i32 = position_val.clone().into();
+        validate_dimension_position(path.into_inner(), position_val)?;
+
+        let previous_position = dimension_row.position.clone();
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            if previous_position < new_position {
+                diesel::update(dsl::dimensions)
+                    .filter(dimensions::position.gt(dimension_row.position))
+                    .filter(dimensions::position.le(new_position.clone()))
```
ayushjain17, line 195: the filter param which is having the previous value should have equal check, and not in the other filter

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -86,39 +93,155 @@ async fn create(
...
+                    .set((
+                        dsl::last_modified_at.eq(Utc::now().naive_utc()),
+                        dsl::last_modified_by.eq(user.get_email()),
+                        dimensions::position.eq(dimensions::position - 1),
+                    ))
+                    .execute(transaction_conn)?
+            } else {
+                diesel::update(dsl::dimensions)
+                    .filter(dimensions::position.lt(dimension_row.position))
+                    .filter(dimensions::position.ge(new_position.clone()))
```
ayushjain17, line 205: the filter param which is having the previous value should have equal check, and not in the other filter

ayushjain17 (changes requested):

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: why ?

crates/context_aware_config/migrations/2024-12-05-172909_dimension_priority_default/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+ALTER TABLE public.dimensions
+ALTER COLUMN priority SET DEFAULT 1;
```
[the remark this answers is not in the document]
pratikmishra356: that will be done later , after all this migrations gets settled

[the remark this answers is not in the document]
pratikmishra356: post migration , we will not insert into priority any values via code , 
so we need a default value otherwise it will give db error

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: in create it will be option<string>
in update it will be option<value>

crates/context_aware_config/src/api/dimension/handlers.rs
sauraww, line 57: Can we change the http method for create api to post instead of put ?

sauraww (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -180,10 +181,9 @@ pub fn generate_config_from_version(
         {
             Ok((latest_version, config)) => {
                 *version = Some(latest_version);
-
-                serde_json::from_value::<Config>(config).map_err(|err| {
+                serde_json::from_value::<Config>(config).or_else(|err| {
                     log::error!("failed to decode config: {}", err);
-                    unexpected_error!("failed to decode config")
+                    generate_cac(conn)
```
Datron, line 186: `blocker`: we should fail here, rather than generate_cac. The user expects the CAC for the version they requested and may act based on that assumption. If we are sending the latest cac regardless of failure, they may act on the wrong assumptions.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -75,39 +73,37 @@ pub fn endpoints() -> Scope {
...
+) -> superposition::Result<()> {
+    let check_dimension = |key: &String, val: &Value| -> superposition::Result<()> {
         if key == "var" {
             let dimension_name = val
                 .as_str()
                 .ok_or(bad_argument!("Dimension name should be of `String` type"))?;
             dimension_schema_map
                 .get(dimension_name)
-                .map(|dimension_val| &dimension_val.priority)
+                .map(|_| Ok(()))
```
Datron, line 93: nab (not a blocker): we can remove this?

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -50,76 +52,175 @@ async fn create(
...
-        .compile(&schema_value);
-
-    if let Err(e) = schema_compile_result {
-        return Err(bad_argument!(
-            "Invalid JSON schema (failed to compile): {:?}",
-            e
-        ));
-    };
+    let num_rows = dimensions
+        .select(count_star())
```
Datron, line 56: nab: you get all columns if you remove this right? This is redundant I think.

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -50,76 +52,175 @@ async fn create(
...
+            .values(&dimension_data)
+            .get_result::<Dimension>(transaction_conn);
+
+        match insert_resp {
+            Ok(inserted_dimension) => {
+                let is_mandatory = tenant_config
+                    .mandatory_dimensions
+                    .contains(&inserted_dimension.dimension);
+                Ok(HttpResponse::Created().json(DimensionWithMandatory::new(
+                    inserted_dimension,
```
Datron, line 100: nab but curious, if someone changes the tenant config to remove mandatory status for a dimension do we set false here @mahatoankitkumar ? If not we can add it in workspace settings

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -59,7 +39,15 @@ impl Default for Position {
     }
 }
 
-#[derive(Debug, Deserialize, AsRef, Deref, DerefMut, Into)]
+#[derive(Debug, Deserialize, Clone)]
+pub struct UpdateReq {
+    pub position: Option<Position>,
+    pub schema: Option<Value>,
```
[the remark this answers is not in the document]
Datron: @ayushjain17 I think serde_json does this automatically for optional fields. I'm not sure though

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -105,7 +105,7 @@ pub enum ContextAction {
 pub struct ContextPutResp {
     pub context_id: String,
     pub override_id: String,
-    pub priority: i32,
+    pub weight: String,
```
Datron, line 108: Weight is string?

crates/frontend/src/components/dimension_form.rs
```diff
@@ -31,7 +31,7 @@ where
 {
     let tenant_rs = use_context::<ReadSignal<String>>().unwrap();
 
-    let (priority, set_priority) = create_signal(priority);
+    let (position, set_position) = create_signal(position);
```
Datron, line 34: nab: 
```suggestion
    let (position_rs, position_ws) = create_signal(position);
```

crates/frontend/src/components/dimension_form.rs
```diff
@@ -85,25 +85,31 @@ where
     let on_submit = move |ev: MouseEvent| {
         req_inprogress_ws.set(true);
         ev.prevent_default();
-        let f_priority = priority.get();
+        let f_position = position.get();
         let f_name = dimension_name_rs.get();
         let fun_name = function_name.get();
-
         let f_schema = dimension_schema_rs.get();
```
Datron, line 92: nab: I think we can just replace `f` with `function`

crates/frontend/src/types.rs
```diff
@@ -271,7 +271,7 @@ impl From<ExperimentResponse> for Experiment {
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct Dimension {
     pub dimension: String,
-    pub priority: i32,
+    pub position: i32,
```
Datron, line 274: should this have been u32 as well?

Datron (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
pratikmishra356: @Datron , generate_cac is in else case where user is not providing the version

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: @Datron here I want the total number of rows , that's why count_star()

Datron:
When merging, squash and merge

crates/experimentation_platform/src/api/experiments/types.rs
pratikmishra356: yes @Datron when it is send as a response or on serialize it will get converted into string as client/frontend wont be able to handle it as a number

crates/frontend/src/types.rs
pratikmishra356: @Datron its been handled in deserialization of Position value
if i change here to u32, in handler i have to do as i32 in few places , since in db to use u32 , i will have to write deserialization there

crates/context_aware_config/src/api/context/handlers.rs
pratikmishra356: since in the else case it is a recursive call which has result<()> type

crates/context_aware_config/src/api/dimension/handlers.rs
pratikmishra356: @Datron currently from appstate we are injecting in the request and thats being passed on to handler 
we will have to add in workspace_settings and use db calls probably

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: @pratikmishra356 then you can call count() function on it

crates/context_aware_config/src/api/config/handlers.rs
Datron: Hmm okay, I will create a ticket for this and we can remove both later

sauraww (approved):

PR #300, opened 2024-12-09
feat: position migration update for demo up

pratikmishra356:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/docker-compose/postgres/db_init.sql
+++ b/docker-compose/postgres/db_init.sql
@@ -19275,3 +19275,18 @@ VALUES (
 -- PostgreSQL database dump complete
 --
 
+ALTER TABLE dev_cac.dimensions
+add column position integer DEFAULT 0 NOT NULL;
+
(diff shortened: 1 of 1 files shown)


docker-compose/postgres/db_init.sql
```diff
@@ -19275,3 +19275,18 @@ VALUES (
 -- PostgreSQL database dump complete
 --
 
+ALTER TABLE dev_cac.dimensions
```
Datron, line 19278: @pratikmishra356 when we do the flip, we will have to regenerate the data in this file right?

Datron (approved):

docker-compose/postgres/db_init.sql
pratikmishra356: yes , we will regenerate the data along with my second pr

ayushjain17 (approved):

PR #303, opened 2024-12-12
feat: Use dimension[] for dimension params in context/list

ayushjain17:
## Problem
`platform[key_name]` format for passing params forces to change all endpoints to create a consistency across all the endpoints

## Solution
1. `dimension[dimension_name]` rather to get the dynamic dimension params from the query request and keep the other params outside of any prefix notation
2. split current `sort_by` filter into `sort_by` and `sort_on` filters, following the new convention

### DIFF
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -38,7 +38,7 @@ use superposition_types::{
             default_configs::dsl,
         },
     },
-    custom_query::{self as superposition_query, CustomQuery, PlatformQuery, QueryMap},
+    custom_query::{self as superposition_query, CustomQuery, DimensionQuery, QueryMap},
     result as superposition, Cac, Contextual, Overridden, Overrides, PaginatedResponse,
     TenantConfig, User,
 };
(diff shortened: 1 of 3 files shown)


pratikmishra356 (approved):

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -633,15 +633,15 @@ async fn list_contexts(
...
-        ContextFilterSortBy::PriorityAsc => builder = builder.order(priority.asc()),
-        ContextFilterSortBy::PriorityDesc => builder = builder.order(priority.desc()),
-        ContextFilterSortBy::CreatedAtAsc => builder = builder.order(created_at.asc()),
-        ContextFilterSortBy::CreatedAtDesc => builder = builder.order(created_at.desc()),
-    }
-
+    let builder = contexts.into_boxed();
+
+    #[rustfmt::skip]
+    let mut builder = match (filter_params.sort_on.unwrap_or_default(), filter_params.sort_by.unwrap_or(SortBy::Asc)) {
```
mahatoankitkumar, line 639:
```suggestion
    let mut builder = match (filter_params.sort_on.unwrap_or_default(), filter_params.sort_by.unwrap_or_default) {
```
We already have a default for this to `Desc`, can use it for consistency.
Now if we want Asc to be default, better to change there first.

mahatoankitkumar (changes requested):

mahatoankitkumar (approved):

PR #308, opened 2024-12-16
feat: add key filter to default config page and API

Datron:
## Problem
Now that we have added pagination, it is difficult for users to find default configs in pages

## Solution
Provide a way to search/filter through default configs

Note: This is an incomplete set of filters and does not include sorting columns. These features will come in a future PR, but this will solve some customer issues right now

https://github.com/user-attachments/assets/cc0ef9be-1f4a-46e9-a90c-e4f7d3a3b107


crates/frontend/src/types.rs
```diff
@@ -174,6 +174,27 @@ impl Display for ExperimentListFilters {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DefaultConfigFilters {
```
mahatoankitkumar, line 178: Can we not move this to superposition_types?
Rest looks fine.

PR #310, opened 2024-12-16
feat: added update api for default config

sauraww:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -1,7 +1,7 @@
 extern crate base64;
 
 use actix_web::{
-    delete, get, put,
+    delete, get, post, put,
     web::{self, Data, Json, Path, Query},
     HttpResponse, Scope,
 };
@@ -42,16 +42,19 @@ use crate::{
     helpers::add_config_version,
 };
 
-use super::types::CreateReq;
+use super::types::{CreateReq, FunctionNameEnum, UpdateReq};
 
 pub fn endpoints() -> Scope {
-    Scope::new("").service(create).service(get).service(delete)
+    Scope::new("")
+        .service(create_default_config)
+        .service(update_default_config)
+        .service(get)
+        .service(delete)
 }
 
-#[put("/{key}")]
-async fn create(
+#[post("")]
+async fn create_default_config(
     state: Data<AppState>,
-    key: web::Path<DefaultConfigKey>,
     custom_headers: CustomHeaders,
     request: web::Json<CreateReq>,
     db_conn: DbConnection,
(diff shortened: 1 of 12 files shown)


crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -42,14 +42,18 @@ use crate::{
...
+    Scope::new("")
+        .service(create_config)
+        .service(update_config)
+        .service(get)
+        .service(delete)
 }
 
-#[put("/{key}")]
-async fn create(
+#[post("/{key}")]
```
pratikmishra356, line 55: can we take in requestbody itself 
as its like that in dimension, type_templates, function

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -201,6 +172,118 @@ async fn create(
...
+                &f_code,
+                &updated_config.key,
+                &updated_config.value,
+            )?;
+        }
+    }
+
+    let version_id =
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            diesel::insert_into(dsl::default_configs)
```
pratikmishra356, line 260: lets use update instead of insert

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -5,6 +5,14 @@ use superposition_types::RegexEnum;
 
 #[derive(Debug, Deserialize)]
 pub struct CreateReq {
+    pub value: Value,
+    pub schema: Map<String, Value>,
```
pratikmishra356, line 9: can we validate that schema should not be empty , currently it is not there ,should we allow empty schema , if one dont want type safety ?

pratikmishra356 (changes requested):

ayushjain17 (changes requested):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -108,27 +111,48 @@ where
...
+                    .await
+                } else {
+                    // Call create_default_config when edit is false
+                    create_default_config(
+                        f_name.clone(),
+                        tenant_rs.get(),
+                        payload.clone(),
+                    )
+                    .await
+                };
```
ayushjain17, line 134:
```suggestion
                let result = if is_edit {
                    // Call update_default_config when edit is true
                    update_default_config(
                        f_name,
                        tenant_rs.get(),
                        payload,
                    )
                    .await
                } else {
                    // Call create_default_config when edit is false
                    create_default_config(
                        f_name,
                        tenant_rs.get(),
                        payload,
                    )
                    .await
                };
```

crates/context_aware_config/src/api/default_config/handlers.rs
ayushjain17: yeah I wanted to raise this as well
we should keep the formats/structures consistent across endpoints

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -201,6 +172,118 @@ async fn create(
...
+            )
+        }
+        _ => {
+            log::error!("Failed to fetch {key_str}: {e}");
+            unexpected_error!("Something went wrong.")
+        }
+    })?;
+
+    let value = req.value.unwrap_or_else(|| existing.value.clone());
+    let schema = match req.schema {
```
ayushjain17, line 204:
```suggestion
    let schema = req.schema.map(Value::Object).unwrap_or_else(|| existing.schema.clone());
```

ayushjain17, line 204:
```suggestion
    let mut schema = existing.schema.clone();
    if let Some(sch) = req.value {
        schema = sch;
    };
```

ayushjain17, line 204: should we try either of these ?
not mandatory

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -201,6 +172,118 @@ async fn create(
...
+        Some(sch) => Value::Object(sch),
+        None => existing.schema.clone(),
+    };
+    let func_name = match req.function_name {
+        Some(Value::String(s)) => Some(s),
+        Some(Value::Null) => None,
+        None => existing.function_name.clone(),
+        Some(_) => return Err(bad_argument!("function_name must be string or null.")),
+    };
```
ayushjain17, line 213: can we please use the same format/type as being used in `dimension/update` endpoint

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -201,6 +172,118 @@ async fn create(
...
+            .map_err(|_| bad_argument!("Function {} doesn't exist.", f_name))?;
+        if let Some(f_code) = function_code {
+            validate_value_with_function(
+                f_name,
+                &f_code,
+                &updated_config.key,
+                &updated_config.value,
+            )?;
+        }
+    }
```
ayushjain17, line 256: can we make a function for this, it is being used as is in create also

crates/context_aware_config/src/api/default_config/handlers.rs
ayushjain17: yeah its better to keep use update
but its cleaner like this if we use insert_into

ayushjain17 (changes requested):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -201,6 +172,118 @@ async fn create(
     Ok(http_resp.json(default_config))
 }
 
+#[put("/{key}")]
```
mahatoankitkumar, line 175: see if we want partial updates, then change it to
```suggestion
#[patch("/{key}")]
```
otherwise remove the partial update thing and accept the all the properties as input.

mahatoankitkumar (changes requested):

sauraww:
Not putting FunctionNameEnum in superposition types as CI checks for Linux were failing : 
image

crates/context_aware_config/src/api/default_config/handlers.rs
sauraww: I agree but we would need to do the similar change in dimension update api and other apis too. 
Can we make that change in a seperate PR , that will keep it consistent .

pratikmishra356 (approved):

pratikmishra356 (approved):

mahatoankitkumar (approved):
Changing PUT to PATCH left for the upcoming PR

ayushjain17 (approved):

ayushjain17 (approved):

pratikmishra356 (approved):

PR #321, opened 2024-12-19
feat: Add auth via OAUTH2

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/superposition/src/auth/oidc.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        let cookie_result = serde_json::to_string(&protection)
+            .map_err(|err| {
+                log::error!("Unable to stringify data: {err}");
+                ErrorInternalServerError(format!("Unable to stringify data"))
+            })
+            .map(|cookie| {
+                Cookie::build("protection", cookie)
+                    .max_age(Duration::days(7))
+                    // .http_only(true)
```
ShreyBana, line 95: Should we un-comment this?

crates/superposition/src/auth/oidc.rs
```diff
@@ -0,0 +1,327 @@
...
+        .cookie("user")
+        .and_then(|user_cookie| {
+            auth.decode_token(user_cookie.value())
+                .map_err(|e| log::error!("Error in decoding user : {e}"))
+                .ok()
+        })
+        .map(|claims| claims.additional_claims().organisations.clone());
+
+    match organisations {
+        Some(organisations) => HttpResponse::Ok().json(serde_json::json!(organisations)),
```
ShreyBana, line 233: nit: Could just be `Some(orgs)`

crates/superposition/src/auth/oidc.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+#[get("/organisations/switch/{organisation_id}")]
+async fn switch_organisation(
+    auth: Data<OIDCAuthenticator>,
+    req: HttpRequest,
+    path: Path<SwitchOrgParams>,
+) -> actix_web::Result<HttpResponse> {
+    let issuer_endpoint = get_from_env_unsafe::<String>("OIDC_ISSUER_ENDPOINT_FORMAT")
```
ShreyBana, line 244: Should we make these fields in the OIDC struct?

crates/superposition/src/auth/oidc.rs
```diff
@@ -0,0 +1,327 @@
...
+        Some(ClientSecret::new(auth.client_secret.clone())),
+    )
+    .set_redirect_uri(
+        RedirectUrl::new(format!("{}/", auth.base_url.clone())).map_err(|e| {
+            log::error!("Unable to create redirect url {e}");
+            ErrorInternalServerError("Unable to create redirect url")
+        })?,
+    );
+
+    let (username, switch_pass) = req
```
ShreyBana, line 272: Since we're already setting the `User` in the middleware, should we just add these fields as well?

crates/superposition/src/auth/oidc/utils.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub(super) fn verify_presence(n: Option<&Nonce>) -> Result<(), String> {
+    if n.is_some() {
+        Ok(())
+    } else {
+        Err("missing nonce claim".to_string())
+    }
+}
+
+pub(super) fn presence_no_check(_: Option<&Nonce>) -> Result<(), String> {
```
ShreyBana, line 14: What does this mean?

crates/context_aware_config/Cargo.toml
```diff
@@ -17,9 +17,9 @@ cac_client = { path = "../cac_client" }
 cfg-if = { workspace = true }
 chrono = { workspace = true }
 derive_more = { workspace = true }
-diesel = { workspace = true , features = ["numeric"]}
+diesel = { workspace = true, features = ["numeric"] }
 fred = { workspace = true, optional = true, features = ["metrics"] }
-futures-util = "0.3.28"
+futures-util.workspace = true
```
Datron, line 22: nab:  
```suggestion
futures-util = { workspace = true }
```

ShreyBana:
We've added some handlers directly via the authenticator, that is, under the `oidc` scope. For login it makes sense, but for get & switch organization, may not be right as it should be independent of the authenticator. For example if I'm using some other auth-provider, let's say `DISABLED`, those routes won't exist. 
I'm thinking that this functionality can move to functions in the `Authenticator` trait and the routes+handler can be hardcoded in `main.rs`. What do you think?

crates/superposition/src/auth.rs
```diff
@@ -0,0 +1,150 @@
...
+use url::Url;
+
+use self::authenticator::Authenticator;
+
+pub struct AuthMiddleware<S> {
+    service: S,
+    auth_handler: AuthHandler,
+}
+
+impl<S, B> Service<ServiceRequest> for AuthMiddleware<S>
```
Datron, line 33: nab:
```suggestion
impl<S, B> Service<ServiceRequest> for OIDCAuthMiddleware<S>
```

Datron (approved):

crates/superposition/src/auth.rs
ShreyBana: This is a generic middleware, the type/protocol is not going to just be OIDC.

ShreyBana (approved):

ShreyBana (approved):

PR #323, opened 2024-12-20
Workspace UI

mahatoankitkumar:
Workspace UI

### DIFF
--- a/crates/frontend/src/api.rs
+++ b/crates/frontend/src/api.rs
@@ -2,7 +2,10 @@ use leptos::ServerFnError;
 use superposition_types::{
     custom_query::PaginationParams,
     database::{
-        models::cac::{ConfigVersion, DefaultConfig, Function, TypeTemplate},
+        models::{
+            cac::{ConfigVersion, DefaultConfig, Function, TypeTemplate},
+            Workspace,
+        },
         types::DimensionWithMandatory,
     },
     Config, PaginatedResponse,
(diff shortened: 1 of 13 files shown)


crates/frontend/src/app.rs
```diff
@@ -237,6 +238,14 @@ pub fn app(app_envs: Envs) -> impl IntoView {
                                 }
                             />
 
+                            <Route
+                                ssr=SsrMode::Async
+                                path="/:tenant/workspace"
```
Datron, line 243:
```suggestion
                                path="/:org/workspace"
```

crates/frontend/src/api.rs
```diff
@@ -276,3 +278,21 @@ pub async fn fetch_types(
...
+
+pub async fn fetch_workspaces(
+    filters: &PaginationParams,
+) -> Result<PaginatedResponse<Workspace>, ServerFnError> {
+    let client = reqwest::Client::new();
+    let host = use_host_server();
+    let url = format!("{}/workspaces?{}", host, filters.to_string());
+    let response: PaginatedResponse<Workspace> = client
+        .get(url)
+        .header("x-tenant", "dev")
```
Datron: @mahatoankitkumar you can remove this and add /workspace to tenant-middleware-exclusion-list env. nab

crates/frontend/src/app.rs
mahatoankitkumar, line 243: Have to do this till the tenant middleware is fixed
```suggestion
                                path="/workspace"
```

Datron: Okay that works, we can also add this to exclusion list too

Datron (approved):

PR #325, opened 2024-12-20
Refactor/diesel juspay fork

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


flake.nix
```diff
@@ -29,7 +29,7 @@
         devShells.default = pkgs.mkShell {
           inputsFrom = [
             self'.devShells.rust
-            # self'.devShells.haskell
+            self'.devShells.haskell
```
ShreyBana, line 32: Should we just keep this comment out? :thinking:

PR #330, opened 2025-01-02
fix: add unique position contraint as deferred

pratikmishra356:
## Problem
During position update statement , we increment/decrement the values , which conflicts with the unique constraint within the transaction

## Solution
Making constraint as deferred which validates on transaction commit instead of statement

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/migrations/2024-12-05-172909_dimension_priority_default/up.sql
+++ b/crates/context_aware_config/migrations/2024-12-05-172909_dimension_priority_default/up.sql
@@ -3,4 +3,6 @@ ALTER TABLE public.dimensions
 ALTER COLUMN priority SET DEFAULT 1;
 
 ALTER TABLE public.dimensions
-ADD CONSTRAINT dimension_unique_position UNIQUE (position);
+ADD CONSTRAINT dimension_unique_position UNIQUE (position) DEFERRABLE INITIALLY IMMEDIATE;
+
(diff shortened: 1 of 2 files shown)


crates/context_aware_config/migrations/2024-12-05-172909_dimension_priority_default/up.sql
```diff
@@ -3,4 +3,6 @@ ALTER TABLE public.dimensions
 ALTER COLUMN priority SET DEFAULT 1;
 
 ALTER TABLE public.dimensions
-ADD CONSTRAINT dimension_unique_position UNIQUE (position);
\ No newline at end of file
+ADD CONSTRAINT dimension_unique_position UNIQUE (position) DEFERRABLE INITIALLY IMMEDIATE;
```
mahatoankitkumar, line 6: Guess you can make a new migration file for this one?

mahatoankitkumar (changes requested):

sauraww (approved):

crates/context_aware_config/migrations/2024-12-05-172909_dimension_priority_default/up.sql
pratikmishra356: @mahatoankitkumar no need to make a new migration for this , its just a minor upgradation in the constraint

mahatoankitkumar (approved):

PR #332, opened 2025-01-06
fix: webhook kms decrypt auth key

mahatoankitkumar:
## Problem
kms decrypt auth key in webhook call.

## Solution
Added kms client in Appstate.

### DIFF
--- a/crates/service_utils/src/helpers.rs
+++ b/crates/service_utils/src/helpers.rs
@@ -1,4 +1,7 @@
-use crate::service::types::{AppState, Tenant};
+use crate::{
+    aws::kms,
+    service::types::{AppEnv, AppState, Tenant},
+};
 use actix_web::{error::ErrorInternalServerError, web::Data, Error};
 use anyhow::anyhow;
 use chrono::Utc;
@@ -23,6 +26,7 @@ use superposition_types::{
     },
     Condition,
 };
+use urlencoding::encode;
 
 const CONFIG_TAG_REGEX: &str = "^[a-zA-Z0-9_-]{1,64}$";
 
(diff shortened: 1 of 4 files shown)


crates/service_utils/src/service/types.rs
```diff
@@ -54,6 +55,7 @@ pub struct AppState {
     #[cfg(feature = "high-performance-mode")]
     pub redis: fred::clients::RedisPool,
     pub http_client: reqwest::Client,
+    pub kms_client: Option<Client>,
```
Datron, line 58: Why `Option<Client>` ?

Datron: Would it better to make this `Arc<Mutex<Client>>`?

Datron (approved):

crates/service_utils/src/helpers.rs
```diff
@@ -411,13 +417,16 @@ where
...
-                    AppError::WebhookError(
-                        String::from("Failed to retrieve authentication token for the webhook. Please verify the credentials in TenantConfig.")
-                    )
-                })?;
+        let auth_token_value: String = match app_env {
+            AppEnv::DEV | AppEnv::TEST => {
+                get_from_env_or_default(&auth.value, "1234".into())
+            }
+            _ => {
+                let kms_client = kms_client.clone().unwrap();
```
pratikmishra356, line 425: why unwrap() , can we have error and log

sauraww (approved):

crates/service_utils/src/helpers.rs
```diff
@@ -411,13 +417,21 @@ where
...
+        let auth_token_value: String = match app_env {
+            AppEnv::DEV | AppEnv::TEST => {
+                get_from_env_or_default(&auth.value, "1234".into())
+            }
+            _ => {
+                let kms_client = kms_client.clone().ok_or_else(|| {
+                    log::error!("Failed to retrieve kms client: KMS client is None");
+                    AppError::WebhookError(String::from(
+                        "Something went wrong. Please check the logs.",
+                    ))
```
ShubhranshuSanjeev, line 429: Why clone?

crates/service_utils/src/service/types.rs
ShubhranshuSanjeev: Is this has been kept optional for case when `webhook` is not enabled?

ShubhranshuSanjeev (changes requested):

PR #339, opened 2025-01-11
Refactored Makefile

ShreyBana:
- Removed nested/recursive make calls in targets.
- Made docker conditonal, w/ podman as an alternative.
- Fixed deprecation warning in compose.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,3 @@
-version: "3.4"
-
 services:
   postgres:
     build: ./docker-compose/postgres/
--- a/makefile
+++ b/makefile
@@ -3,6 +3,29 @@ DOCKER_DNS ?= localhost
 TENANT ?= dev
 SHELL := /usr/bin/env bash
 FEATURES ?= ssr
+CARGO_FLAGS := --color always --no-default-features
+HAS_DOCKER := $(shell command -v docker > /dev/null; echo $$?)
+HAS_PODMAN := $(shell command -v podman > /dev/null; echo $$?)
+ifeq ($(HAS_DOCKER),0)
+  DOCKER := docker
+else ifeq ($(HAS_PODMAN),0)
+  DOCKER := podman
+  export PODMAN_COMPOSE_WARNING_LOGS = false
+else
+	$(error "Neither docker nor podman found, please install one of them.")
+endif
+COMPOSE := $(DOCKER) compose
+
+define read-container-name
+	yq -r '.services.$(1).container_name' docker-compose.yaml
+endef
+define check-container
+	$(DOCKER) ps | grep $(1) 2>&1 > /dev/null; echo $$?
+endef
+DB_CONTAINER_NAME = $(shell $(call read-container-name,postgres))
+DB_UP = $(shell $(call check-container,$(DB_CONTAINER_NAME)))
+LSTACK_CONTAINER_NAME = $(shell $(call read-container-name,localstack))
+LSTACK_UP = $(shell $(call check-container,$(LSTACK_CONTAINER_NAME)))
 
 .PHONY:
 	db-init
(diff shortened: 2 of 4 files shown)


docker-compose.yaml
```diff
@@ -1,5 +1,3 @@
-version: "3.4"
```
ShreyBana, line 1: This is to fix a deprecation warning.

ayushjain17: thanks!

makefile
```diff
@@ -110,22 +125,20 @@ backend:
...
-	make superposition_legacy -e DOCKER_DNS=$(DOCKER_DNS)
+run_legacy: kill build db localstack superposition_legacy
 
-ci-test: ci-setup
+test:
 	cargo test
+	npm ci
+	@echo "Awaiting superposition boot..."
+	@timeout 200s bash -c "while ! curl --silent 'http://localhost:8080/health' 2>&1 > /dev/null; do sleep 0.5; done"
 	npm run test
```
ayushjain17:
```suggestion
test:
	cargo test
	@echo "Awaiting superposition boot..."
	@timeout 200s bash -c "while ! curl --silent 'http://localhost:8080/health' 2>&1 > /dev/null; do sleep 0.5; done"
	npm run test
```

makefile
```diff
@@ -14,22 +25,42 @@ FEATURES ?= ssr
...
-
-migration: cleanup
-	docker-compose up -d postgres
-	cp .env.example .env
-	while ! make validate-psql-connection; \
-		do echo "waiting for postgres bootup"; \
-		sleep 0.5; \
-		done
+	-$(DOCKER) rm -f $$($(DOCKER) container ls --filter name=^context-aware-config -a -q)
+	-$(DOCKER) rmi -f $$($(DOCKER) images | grep context-aware-config-postgres | cut -f 10 -d " ")
```
ayushjain17, line 59:
```suggestion
	-$(DOCKER) rm -f $$($(DOCKER) container ls --filter name=^superposition -a -q)
	-$(DOCKER) rmi -f $$($(DOCKER) images | grep superposition-postgres | cut -f 10 -d " ")
```

makefile
```diff
@@ -44,52 +75,43 @@ validate-psql-connection:
...
-	# NOTE: The container spinned-up here is the actual container being used in development
-	echo setup completed successfully!!!
-
-setup: env-setup
+SETUP_DEPS = env-file db localstack
+ifdef CI
+	SETUP_DEPS += test-tenant
+endif
+setup: $(SETUP_DEPS)
+	npm ci
```
ayushjain17, line 89: are we sure we want to show all logs ?
previously we were logging only error logs here

makefile
```diff
@@ -110,22 +132,18 @@ backend:
...
 build: frontend backend
 
-run: kill frontend
-	while ! make validate-psql-connection validate-aws-connection; \
-		do echo "waiting for postgres, localstack bootup"; \
-		sleep 0.5; \
-		done
-	make superposition -e DOCKER_DNS=$(DOCKER_DNS)
+run: kill db localstack frontend superposition
+	@./target/debug/superposition
```
ayushjain17, line 136: lets not suppress the logs of these command executions, it will always lead to the person debugging to cross refer with makefile that which command was the issue in the scenario when there no logs appearing for the execution of this command

makefile
```diff
@@ -3,6 +3,17 @@ DOCKER_DNS ?= localhost
 TENANT ?= dev
 SHELL := /usr/bin/env bash
 FEATURES ?= ssr
+HAS_DOCKER := $(shell command -v docker > /dev/null; echo $$?)
+HAS_PODMAN := $(shell command -v podman > /dev/null; echo $$?)
+CARGO_FLAGS := --color always --no-default-features
```
ayushjain17, line 8: can we move this above `HAS_DOCKER`, it kinda became invisible while being present amidst un-related variables

makefile
```diff
@@ -44,52 +75,43 @@ validate-psql-connection:
...
 
+superposition: CARGO_FLAGS += --features=$(FEATURES)
 superposition:
-	cargo run --color always --bin superposition --no-default-features --features=$(FEATURES)
+	cargo build $(CARGO_FLAGS) --bin superposition
 
 superposition-example:
 	cargo run --bin cac-demo-app
 
+superposition_legacy: CARGO_FLAGS += --features='ssr superposition_types/disable_db_data_validation
```
ayushjain17, line 104:
```suggestion
superposition_legacy: CARGO_FLAGS += --features='$(FEATURES)
```

makefile
```diff
@@ -44,52 +75,43 @@ validate-psql-connection:
...
+superposition_legacy: CARGO_FLAGS += superposition_types/disable_db_data_validation
+superposition_legacy: CARGO_FLAGS += context_aware_config/disable_db_data_validation
+superposition_legacy: CARGO_FLAGS += experimentation_platform/disable_db_data_validation'
 superposition_legacy:
-	cargo run --color always --bin superposition --no-default-features --features='ssr superposition_types/disable_db_data_validation context_aware_config/disable_db_data_validation experimentation_platform/disable_db_data_validation'
+	cargo run $(CARGO_FLAGS) --bin superposition
```
ayushjain17, line 109: lets keep it same as `superposition`
```suggestion
	cargo build $(CARGO_FLAGS) --bin superposition
```

makefile
```diff
@@ -110,22 +132,18 @@ backend:
...
+run: kill db localstack frontend superposition
+	@./target/debug/superposition
 
-run_legacy: kill build
-	while ! make validate-psql-connection validate-aws-connection; \
-		do echo "waiting for postgres, localstack bootup"; \
-		sleep 0.5; \
-		done
-	make superposition_legacy -e DOCKER_DNS=$(DOCKER_DNS)
+run_legacy: kill build db localstack superposition_legacy
```
ayushjain17, line 138: same as run

ayushjain17 (changes requested):

makefile
ShubhranshuSanjeev, line 109: This could also have been a `build` command, followed by `@./target/debug/superposition`

makefile
ShreyBana: For debugging reasons I changed the log-level.

makefile
ShreyBana: I personally feel that this is not a very interesting command, plus it's the last one. So, if it so happens that there are no logs after the build, I think one will know the problem is w/ the app.

ayushjain17 (approved):

ShubhranshuSanjeev (approved):

PR #344, opened 2025-01-16
feat: added description and comment (#284)

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -198,6 +198,15 @@ fn create_ctx_from_put_req(
     tenant_config: &TenantConfig,
 ) -> superposition::Result<Context> {
     let ctx_condition = req.context.to_owned().into_inner();
+    let description = if req.description.is_none() {
+        let ctx_condition_value = json!(ctx_condition);
+        ensure_description(ctx_condition_value, conn)?
+    } else {
+        req.description
+            .clone()
+            .ok_or_else(|| bad_argument!("Description should not be empty"))?
+    };
+    let change_reason = req.change_reason.clone();
     let condition_val = Value::Object(ctx_condition.clone().into());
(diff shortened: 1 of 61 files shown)


crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -864,35 +1030,51 @@ async fn weight_recompute(
...
         })
-        .collect::<superposition::Result<Vec<(BigDecimal, String)>>>()?;
+        .collect::<superposition::Result<Vec<(BigDecimal, String, String, String)>>>()?;
 
+    // Update database and add config version
     let last_modified_time = Utc::now().naive_utc();
     let config_version_id =
         conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
-            for (context_weight, context_id) in contexts_new_weight {
+            for (context_weight, context_id, _description, _change_reason) in contexts_new_weight.clone() {
```
Datron, line 1055: Why _description and _change_reason? Are we not using these variables anywhere?

crates/context_aware_config/src/api/type_templates/handlers.rs
```diff
@@ -87,6 +112,7 @@ async fn update_type(
             type_templates::type_schema.eq(request.clone()),
```
Datron, line 112: Hey so there is an issue here. Request now contains description and change reason, which cannot directly be inserted into the column `type_schema`. We need to extract type schema. It would be better to create a type `TypeTemplateUpdateRequest`. This can be taken from org__id-test branch, or when we rebase we can keep that code

Datron (approved):
The additions to ramp and conclude form UI to take `change reason` will come in another PR?

ShubhranshuSanjeev:
Closing this PR, the commit is there in the `org__id-test` branch.

PR #374, opened 2025-01-21
fix: webhook ui failure handling

mahatoankitkumar:
## Problem
Webhook failures in frontend not handled properly

## Solution
added a condition to ignore the 512 status code

### DIFF
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -429,7 +429,7 @@ where
         enqueue_alert(error_msg.clone(), AlertType::Error, 5000);
         return Err(error_msg);
     }
-    if status.is_server_error() {
+    if status.is_server_error() && status != 512 {
         let error_msg = response
             .json::<ErrorResponse>()
             .await


crates/frontend/src/utils.rs
```diff
@@ -429,7 +429,7 @@ where
         enqueue_alert(error_msg.clone(), AlertType::Error, 5000);
         return Err(error_msg);
     }
-    if status.is_server_error() {
+    if status.is_server_error() && status != 512 {
```
ShubhranshuSanjeev, line 432: Why are we ignoring this very specific error code?

ShubhranshuSanjeev (changes requested):

crates/frontend/src/utils.rs
mahatoankitkumar: This is a custom error code sent by the webhook.
When the webhook fails, the steps involving the experiment and CAC are already completed. Users receive a false signal that the experiment wasn't created, prompting them to attempt creation multiple times, potentially leading to duplicate experiments.

ShubhranshuSanjeev (approved):

ayushjain17 (approved):

ayushjain17:
why is this pointed to main ?
can we point to saas instead

mahatoankitkumar:
> why is this pointed to main ? can we point to saas instead

saas might take a while to go to prod

Datron (approved):
We shouldn't ignore the webhook failure going forward. Rather this should be displayed in the experiments screen otherwise the user assumes even the webhook trigger succeeded.

Issue #367, opened 2025-01-20
General pending issues


ayushjain17:
https://github.com/juspay/superposition/pull/333#discussion_r1922238537
https://github.com/juspay/superposition/pull/333#discussion_r1922336689
https://github.com/juspay/superposition/pull/333/files#r1922342977
https://github.com/juspay/superposition/pull/333/files#r1922344272
https://github.com/juspay/superposition/pull/333#discussion_r1922355299

PR #375, opened 2025-01-21
fix: Resolving general pending issues

ayushjain17:
# Problem
Resolving #367 

## Solution

### DIFF
--- a/crates/superposition/src/organisation/handlers.rs
+++ b/crates/superposition/src/organisation/handlers.rs
@@ -7,7 +7,7 @@ use chrono::Utc;
 use diesel::prelude::*;
 use idgenerator::IdInstance;
 use service_utils::service::types::DbConnection;
-use superposition_macros::db_error;
+use superposition_macros::{db_error, unexpected_error};
 use superposition_types::database::{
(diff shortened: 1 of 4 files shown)


crates/superposition/src/organisation/handlers.rs
```diff
@@ -124,9 +124,7 @@ pub async fn update_organisation(
         .get_result(&mut conn)
         .map_err(|e| {
             log::error!("Failed to update organisation: {:?}", e);
-            superposition::AppError::UnexpectedError(anyhow::anyhow!(
-                "Failed to update organisation"
-            ))
+            unexpected_error!("Failed to update organisation")
```
ShreyBana, line 127: What about logging ?

mahatoankitkumar (approved):

crates/superposition/src/organisation/handlers.rs
ayushjain17: I just removed usage of `anyhow` over here, nothing else
there is no wording changes of any sort

ShreyBana (approved):

Issue #361, opened 2025-01-20
import extract org function from service utils


ayushjain17:
_Originally posted by @ayushjain17 in https://github.com/juspay/superposition/pull/333#discussion_r1922291851_

Issue #356, opened 2025-01-20
Cleanup error response for tenant middleware


ayushjain17:
_Originally posted by @ShreyBana in https://github.com/juspay/superposition/pull/333#discussion_r1921739481_

Issue #353, opened 2025-01-20
Club OrgId, WorkspaceId, and its combination together in request extensions


ayushjain17:
Club data:
_Originally posted by @ShreyBana in https://github.com/juspay/superposition/pull/333#discussion_r1920069590_
Rename tenant to workspace:
_Originally posted by @ShreyBana https://github.com/juspay/superposition/pull/333#discussion_r1921745205

PR #379, opened 2025-01-22
refactor: OrgId, WorkspaceId, SchemaName cleanup and refactor

ayushjain17:
Resolving the following issues
#353
#356  
#361

### DIFF
--- a/crates/experimentation_platform/src/api/experiments/handlers.rs
+++ b/crates/experimentation_platform/src/api/experiments/handlers.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{HashMap, HashSet},
-    ops::Deref,
-};
+use std::collections::{HashMap, HashSet};
 
 use actix_http::header::{self};
 use actix_web::{
@@ -9,7 +6,6 @@ use actix_web::{
     web::{self, Data, Json, Query},
     HttpRequest, HttpResponse, HttpResponseBuilder, Scope,
 };
-use anyhow::anyhow;
 use chrono::{DateTime, Duration, NaiveDateTime, Utc};
 use diesel::{
     dsl::sql,
@@ -20,14 +16,13 @@ use diesel::{
 };
 use reqwest::{Method, Response, StatusCode};
 use serde_json::{json, Map, Value};
-use service_utils::service::types::{
-    AppHeader, AppState, CustomHeaders, DbConnection, Tenant,
-};
 use service_utils::{
     helpers::{
         construct_request_headers, execute_webhook_call, generate_snowflake_id, request,
     },
-    service::types::OrganisationId,
+    service::types::{
+        AppHeader, AppState, CustomHeaders, DbConnection, SchemaName, WorkspaceContext,
+    },
 };
 use superposition_macros::{bad_argument, response_error, unexpected_error};
 use superposition_types::{
@@ -138,10 +133,9 @@ async fn create(
     custom_headers: CustomHeaders,
     req: web::Json<ExperimentCreateRequest>,
     db_conn: DbConnection,
-    tenant: Tenant,
+    workspace_request: WorkspaceContext,
     user: User,
     tenant_config: TenantConfig,
-    org_id: OrganisationId,
 ) -> superposition::Result<HttpResponse> {
     use superposition_types::database::schema::experiments::dsl::experiments;
     let mut variants = req.variants.to_vec();
@@ -196,7 +190,7 @@ async fn create(
         &unique_override_keys,
         None,
         flags,
-        &tenant,
+        &workspace_request.schema_name,
         &mut conn,
     )?;
     if !valid {
@@ -244,25 +238,18 @@ async fn create(
             err
         )
     })?;
-    let org_id = org_id.clone().deref().to_owned();
-    log::info!("Organisation ID {org_id}");
+
     let extra_headers = vec![
         ("x-user", Some(user_str)),
         ("x-config-tags", custom_headers.config_tags),
-        ("x-org-id", Some(org_id)),
     ]
     .into_iter()
     .filter_map(|(key, val)| val.map(|v| (key, v)))
     .collect::<Vec<_>>();
 
     let headers_map = construct_header_map(
-        tenant
-            .get_tenant_name()
-            .map_err(|err| {
-                log::error!("{err}");
-                unexpected_error!("failed to decode tenant")
-            })?
-            .as_str(),
+        &workspace_request.workspace_id,
+        &workspace_request.organisation_id,
         extra_headers,
     )?;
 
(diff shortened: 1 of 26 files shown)


crates/superposition_types/src/webhook.rs
```diff
@@ -7,14 +7,14 @@ use std::{
...
-    TenantId,
+    WorkspaceId,
 }
 
 impl fmt::Display for HeadersEnum {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::ConfigVersion => write!(f, "x-config-version"),
-            Self::TenantId => write!(f, "x-tenant"),
+            Self::WorkspaceId => write!(f, "x-tenant"),
```
Datron, line 17: Something for later: should we start supporting `x-workspace-id` below?

crates/service_utils/src/service/types.rs
```diff
@@ -78,117 +75,93 @@ pub enum AppScope {
...
+    type Future = Ready<Result<Self, Self::Error>>;
+
+    fn from_request(
+        req: &actix_web::HttpRequest,
+        _: &mut actix_web::dev::Payload,
+    ) -> Self::Future {
+        let result = req.extensions().get::<Self>().cloned().ok_or_else(|| {
+            log::error!("Schema name not found, please check that the organisation id and workspace id are being properly sent");
+            actix_web::error::ErrorInternalServerError("Schema name not found, please check that the organisation id and workspace id are being properly sent")
+        });
```
Datron, line 139: saying schema_name not found and then asking for org id and workspace id is confusing, can we drop
`Schema name not found`

crates/service_utils/src/helpers.rs
```diff
@@ -454,7 +456,8 @@ where
             event_info: WebhookEventInfo {
                 webhook_event: event,
                 time: Utc::now().naive_utc().to_string(),
-                tenant_id: tenant.to_string(),
+                workspace_id: workspace_request.workspace_id.to_string(),
+                organisation_id: workspace_request.organisation_id.to_string(),
```
Datron, line 460: @mahatoankitkumar do you have to make any changes to our webhook receiver? Are you dependent on the tenant_id?

Datron (approved):

crates/superposition_types/src/webhook.rs
ayushjain17: we need to support `x-org-id` here, I had earlier added it but later removed, thinking it has some DB dependency for which I think I would need to update the list of enums in DB, if that is not the case then I can add it

also, we soon need to migrate `x-tenant` to `x-workspace` or `x-workspace-id`

crates/service_utils/src/service/types.rs
ayushjain17: Sure

crates/superposition_types/src/webhook.rs
Datron: migration is difficult, we'll need to be backwards compatible for a bit. I need @mahatoankitkumar to weigh in on how this is being used

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -462,24 +442,20 @@ pub async fn conclude(
                     let url = format!("{}/default-config/{}", state.cac_host, key);
 
                     let headers = construct_request_headers(&[
-                        ("x-tenant", tenant_name.as_str()),
+                        ("x-tenant", &workspace_request.workspace_id),
```
ShreyBana, line 445: We have to call this x-tenant?

crates/service_utils/src/middlewares/tenant.rs
```diff
@@ -230,3 +179,58 @@ where
         })
     }
 }
+
+pub trait ServiceRequestExt {
+    fn get_organisation_id(&self) -> Option<OrganisationId>;
+    fn get_workspace_id(&self) -> Option<WorkspaceId>;
+}
+
+impl ServiceRequestExt for ServiceRequest {
```
ShreyBana, line 188: These functions are almost identical.

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: yes, this has not yet been migrated to something like `x-workspace` or `x-workspace-id`

crates/service_utils/src/middlewares/tenant.rs
ShreyBana: And I think this should be in some helpers file somewhere, otherwise people might end-up repeating functions.

crates/service_utils/src/service/types.rs
```diff
@@ -78,117 +75,93 @@ pub enum AppScope {
...
-            .last()
-            .map(str::to_string)
-            .ok_or(String::from("failed to decode tenant"))
+        ready(result)
     }
 }
-impl FromRequest for Tenant {
+
+#[derive(Clone)]
+pub struct WorkspaceRequest {
```
ShreyBana, line 146:
```suggestion
pub struct WorkspaceContext {
```
Couldn't think of a better name :disappointed:

ShreyBana (changes requested):

crates/experimentation_platform/src/api/experiments/handlers.rs
ShreyBana: Recap: This header is externally exposed, so we have to migrate those services as well.

crates/service_utils/src/middlewares/tenant.rs
ayushjain17: no one needs to repeat these, they will simply call the function on ServiceRequest, I didnt get how this can be repeated ?

ShreyBana: Not in that sense, but if everyone starts writing utils/extensions for service-request, it may happen that people end up repeating some basic utility.

crates/service_utils/src/extensions.rs
```diff
@@ -0,0 +1,52 @@
...
+pub trait ServiceRequestExt {
+    fn get_header(&self, header_name: &str) -> Option<&str>;
+    fn get_path_param(&self, param: &str) -> Option<&str>;
+    fn get_query_param(&self, query_param: &str) -> Option<&str>;
+
+    fn get_organisation_id(&self) -> Option<OrganisationId>;
+    fn get_workspace_id(&self) -> Option<WorkspaceId>;
+}
+
+impl ServiceRequestExt for ServiceRequest {
```
ShreyBana, line 14: CLEAN!

ShreyBana (approved):

PR #385, opened 2025-01-24
fix: using schema_name in redis key

Datron:
## Problem
The variable schema_name is de-referenced only once, doing it twice grants access to the string it's wrapping

## Solution
De-reference twice

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4334,9 +4334,9 @@ dependencies = [
 
 [[package]]
 name = "serde_with"
-version = "3.11.0"
+version = "3.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e28bdad6db2b8340e449f7108f020b3b092e8583a9e3fb82713e1d4e71fe817"
+checksum = "d6b6f7f2fcb69f747921f79f3926bd1e203fce4fef62c268dd3abfb6d86029aa"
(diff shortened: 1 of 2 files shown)


mahatoankitkumar (approved):

Cargo.lock
```diff
@@ -4334,9 +4334,9 @@ dependencies = [
 
 [[package]]
 name = "serde_with"
-version = "3.11.0"
+version = "3.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e28bdad6db2b8340e449f7108f020b3b092e8583a9e3fb82713e1d4e71fe817"
+checksum = "d6b6f7f2fcb69f747921f79f3926bd1e203fce4fef62c268dd3abfb6d86029aa"
```
ayushjain17, line 4339: do we need cargo lock change for this ?

ShreyBana:
Is this causing a compilation failure?

Cargo.lock
Datron: It made the change after I compiled with `high-performance-mode` flag

Datron:
@ShreyBana it will if you do `make run -e FEATURES=high-performance-mode`. The custom version we run internally runs with this feature flag

ayushjain17 (approved):

PR #390, opened 2025-01-29
fix: update queries refactor

pratikmishra356:
## Problem
Redundant fetch and all columns update 

## Solution
Created new changeset request type for normal api request and added a function to map to changeset type

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -67,21 +67,23 @@ async fn put_handler(
     schema_name: SchemaName,
 ) -> superposition::Result<HttpResponse> {
     let tags = parse_config_tags(custom_headers.config_tags)?;
+    let description = match req.description.clone() {
+        Some(val) => val,
+        None => ensure_description(
```
ayushjain17, line 72: not reviewing the changes around `ensure_description`

ayushjain17 (changes requested):

crates/context_aware_config/src/api/functions/handlers.rs
```diff
@@ -116,47 +116,18 @@ async fn update(
...
-        change_reason: req.change_reason,
-    };
-
     let mut updated_function = diesel::update(functions)
         .filter(schema::functions::function_name.eq(f_name))
-        .set(new_function)
+        .set((
+            req.as_changeset(),
+            dsl::draft_edited_by.eq(user.get_email()),
+            dsl::draft_edited_at.eq(Utc::now().naive_utc()),
```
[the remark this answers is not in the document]
pratikmishra356: @ayushjain17 it is not meant to be updated here, 
for published code changes it gets updated

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -48,6 +49,35 @@ pub struct UpdateReq {
     pub change_reason: String,
 }
 
+impl UpdateReq {
+    pub fn as_changeset(self) -> UpdateReqChangeset {
+        UpdateReqChangeset {
+            position: self.position.map(|x| x.into()),
```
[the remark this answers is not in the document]
pratikmishra356: yes its needed , deref does not removes the wrapper type

crates/context_aware_config/src/api/functions/handlers.rs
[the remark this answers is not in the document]
ayushjain17: this is the update endpoint
and previously we were very much updating these two fields
[ref](https://github.com/juspay/superposition/pull/390/files/73a9e34b0ff065bdfc92bd69ef20b1877c867bc4#diff-0ed7935c111bc1f85535d40643e36a37aaf4416285e1e1c6b97fc06a4de97b79L152-L153)

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -230,31 +215,34 @@ async fn update_default_config(
         ));
     }
 
+    let function_name = match req.function_name.clone() {
+        Some(FunctionNameEnum::Name(func_name)) => Some(func_name),
+        Some(FunctionNameEnum::Remove) => None,
+        None => existing.function_name.clone(),
+    };
```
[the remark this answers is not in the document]
pratikmishra356: this is being used for validation in another function , thats why need to extract it

crates/context_aware_config/src/api/context/operations.rs
```diff
@@ -75,6 +76,7 @@ pub fn put(
 pub fn r#move(
     old_ctx_id: String,
     req: Json<MoveReq>,
+    req_description: String,
```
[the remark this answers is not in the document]
pratikmishra356: it is of optional type , 
in move handler , at multiple places we needed non-option description 
thats why we first check request or from db and then pass it to other utils function

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -495,8 +498,19 @@ async fn reduce_config_key(
                     if is_approve {
                         let _ = context::delete(cid.clone(), user, conn, schema_name);
                         if let Ok(put_req) = construct_new_payload(request_payload) {
+                            let description = match put_req.description.clone() {
+                                Some(val) => val,
+                                None => ensure_description(
```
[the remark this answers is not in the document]
pratikmishra356: The pr was to remove redundant call not ensure description , 
this pr removes redundant update query and redundant ensure description call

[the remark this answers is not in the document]
ayushjain17: All that I am saying is that `ensure_description` should not at all be needed after removing these redundant queries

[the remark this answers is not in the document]
pratikmishra356: @ayushjain17 ensure_description is being used in context handlers only 
context apis are different like put and move request , where we need to have description either from the request 
or from the db , ensure_description checks if contexts exist then fetch description from db 
and if not then consider the request description and if missing in request then throw error

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -225,17 +214,15 @@ async fn update(
                         .execute(transaction_conn)?
                 };
             }
-
+            let update_position =
+                Position::try_from(new_position).map_err(|err| unexpected_error!(err))?;
```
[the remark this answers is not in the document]
ayushjain17: why is this still needed ?

crates/context_aware_config/src/api/default_config/handlers.rs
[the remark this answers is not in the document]
ayushjain17: in that case, can we use this:
```suggestion
    let function_name = req.function_name.map_or_else(|| existing.function_name.clone(), |f| f.to_option());
```

crates/context_aware_config/src/api/context/operations.rs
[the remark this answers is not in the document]
ayushjain17: can we not fix all of these and clean all these mess

crates/context_aware_config/src/api/context/handlers.rs
ayushjain17: this is still pending on my part

ayushjain17 (changes requested):

crates/superposition_types/src/config.rs
```diff
@@ -208,6 +215,91 @@ impl Config {
...
+    fn from_sql(bytes: PgValue<'_>) -> DResult<Self> {
+        let value = <i32 as FromSql<Integer, Pg>>::from_sql(bytes)?;
+        Ok(Position(value))
+    }
+}
+
+#[cfg(feature = "diesel_derives")]
+impl ToSql<Integer, Pg> for Position {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> SResult {
+        let value = self.to_owned().into();
```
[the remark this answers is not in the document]
pratikmishra356: No this doesn't work as Into trait implies for Position and not referenced values

crates/context_aware_config/src/api/context/operations.rs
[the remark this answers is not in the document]
pratikmishra356: Its not the scope of this PR, 
anyway context apis are critical and a bit complex comparing to other apis , we can take it as an individual task to refactor implementation of context apis

crates/superposition_types/src/config.rs
```diff
@@ -208,6 +215,91 @@ impl Config {
...
+    PartialEq,
+    Eq,
+    Ord,
+    PartialOrd,
+    Into,
+)]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Integer))]
+#[serde(try_from = "i32")]
+pub struct Position(i32);
```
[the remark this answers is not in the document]
pratikmishra356: since Position is a part of Context also , and context reside in config.rs file , 
so Position should be in this file only 
FunctionNameEnum moved to function.rs file inside api

crates/context_aware_config/src/api/config/handlers.rs
ShreyBana, line 503: I know that `ensure_description` wasn't written in this PR, but please rename it to `query_description`.

crates/superposition_types/src/config.rs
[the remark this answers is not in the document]
ayushjain17: see, In Api we re-use stuff from other places which are more core in nature
models is the very core type on top which everything else is structured, so Apis do use types from other places and `Position` is not being used at that core level itself

and `FunctionNameEnum` is not something is needed for the functionality around `Functions` to work, it just is an API contract for handlers related to functions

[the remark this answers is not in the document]
pratikmishra356: Position cant be in models/cac.rs , I think models folder should refer to only table struct , 
other core types can be moved to other folders or files like api, config etc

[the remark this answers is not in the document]
ShreyBana: One more thing we can do is keep them in both places. You can define them in `models` & then re-export it from `api` & then use the `api` import to have the same effect.

ShreyBana (changes requested):

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -23,30 +26,31 @@ pub struct UpdateReq {
...
 
+//changeset type for update request type
+#[derive(AsChangeset)]
+#[diesel(table_name = default_configs)]
+pub struct UpdateReqChangeset {
+    pub value: Option<Value>,
+    pub schema: Option<Value>,
+    //function_name is nullable column, so to support null update with changeset
+    //we had to make it Option<Option<String>>
+    pub function_name: Option<Option<String>>,
```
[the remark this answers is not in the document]
pratikmishra356: Yes , it has been tested
that's subjective decision , if we let user to remove or only replace
Option<Option<String>>> wont work on the request type since we cant derive option<option type after deserliaze

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -48,30 +25,31 @@ pub struct UpdateReq {
...
-        let map: Value = Deserialize::deserialize(deserializer)?;
-        match map {
-            Value::String(func_name) => Ok(Self::Name(func_name)),
-            Value::Null => Ok(Self::Remove),
-            _ => {
-                log::error!("Expected a string or null literal as the function name.");
-                Err("Expected a string or null literal as the function name.")
-                    .map_err(serde::de::Error::custom)
-            }
+impl From<UpdateReq> for UpdateReqChangeset {
```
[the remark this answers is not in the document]
pratikmishra356: We can use AsChangeset on UpdateReq Directly as its feilds types doesnt correspond to table's column types

crates/superposition_types/src/api/function.rs
```diff
@@ -0,0 +1,36 @@
+use serde::{Deserialize, Deserializer};
+use serde_json::Value;
+
+#[derive(Debug, Clone)]
+pub enum FunctionNameEnum {
```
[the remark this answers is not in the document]
pratikmishra356: No we cant

crates/superposition_types/src/config.rs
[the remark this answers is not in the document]
ayushjain17: `api` and `config` are not core types, thats the misunderstanding we are having here
the very basis of both `api` and `config` types are `models`
not vice versa

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -159,3 +170,119 @@ pub struct TypeTemplate {
...
+}
+
+pub fn deserialize_function_name<'de, D>(
+    deserializer: D,
+) -> Result<Option<Option<String>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let map: Option<Value> = Deserialize::deserialize(deserializer)?;
+    log::error!("function xxx {:?}", map.clone());
```
ayushjain17, line 276: Return of Xander Cage

crates/superposition_types/src/lib.rs
```diff
@@ -16,6 +16,7 @@ use std::future::{ready, Ready};
 
 #[cfg(feature = "server")]
 use actix_web::{dev::Payload, FromRequest, HttpMessage, HttpRequest};
+use base64 as _;
```
ayushjain17, line 19: why this ?
simply make `bse64` option dependency and add it in the list of dependencies for `diesel_derives`

crates/superposition_types/src/database/schema.patch
ayushjain17, line 1: what is the diff in this file ?
ideally there should not be any change in this file as there is no db level change being made in this PR

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -159,3 +170,119 @@ pub struct TypeTemplate {
...
+    PartialOrd,
+    Into,
+    Copy,
+)]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Integer))]
+#[serde(try_from = "i32")]
+pub struct Position(i32);
+impl Position {
+    fn validate_data(position_val: i32) -> Result<Self, String> {
```
ayushjain17, line 194: do we need this separately ?
should we move this directly inside `try_from` ?

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -1,9 +1,21 @@
...
+    expression::AsExpression,
+    pg::{Pg, PgValue},
+    serialize::{Output, Result as SResult, ToSql},
+    sql_types::Integer,
+    AsChangeset, Insertable, Queryable, Selectable,
+};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
+#[cfg(feature = "diesel_derives")]
+use std::str;
```
ayushjain17, line 18: is this being used ?

also, can we follow this:
```
std package imports
{{new_line}}
package imports
{{new_line}}
crate imports
{{new_line}}
super imports
{{new_line}}
self imports
```

crates/superposition_types/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 # env
 actix-web = { workspace = true, optional = true }
 anyhow = { workspace = true, optional = true }
+base64 = { workspace = true }
```
ayushjain17, line 12:
```suggestion
base64 = { workspace = true, optional = true }
```

crates/superposition_types/src/database/schema.patch
ayushjain17: oh just checked, we are removing a column, but the patch file does not seem to be correct for that change, @pratikmishra356 can you please verify

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -159,3 +170,119 @@ pub struct TypeTemplate {
...
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = diesel::sql_types::Text))]
+pub struct FunctionCode(pub String);
+
+#[cfg(feature = "diesel_derives")]
+impl FromSql<diesel::sql_types::Text, Pg> for FunctionCode {
+    fn from_sql(bytes: PgValue<'_>) -> DResult<Self> {
+        let value = <String as FromSql<diesel::sql_types::Text, Pg>>::from_sql(bytes)?;
+        let decode_val = decode_base64_to_string(&value)?;
+        Ok(FunctionCode(decode_val))
```
ayushjain17, line 236:
```suggestion
        Ok(Self(decode_val))
```

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -159,3 +170,119 @@ pub struct TypeTemplate {
...
+    }
+}
+
+#[cfg(feature = "diesel_derives")]
+impl ToSql<diesel::sql_types::Text, Pg> for FunctionCode {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> SResult {
+        let value: String = BASE64_STANDARD.encode(self.0.clone());
```
ayushjain17, line 243: can you please remind me once, why did we need all these `BASE64` ops on the function name ?

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -159,3 +170,119 @@ pub struct TypeTemplate {
...
+}
+
+#[cfg(feature = "diesel_derives")]
+fn decode_base64_to_string(code: &String) -> Result<String, String> {
+    BASE64_STANDARD
+        .decode(code)
+        .map_err(|e: base64::DecodeError| {
+            log::info!("Error while decoding function: {}", e);
+            e.to_string()
+        })
```
ayushjain17, line 258: ?
```suggestion
        })?
```

crates/frontend/src/pages/function.rs
```diff
@@ -377,13 +377,13 @@ pub fn function_page() -> impl IntoView {
...
                                                                 let (fun_code_rs, fun_code_ws) = create_signal(
-                                                                    function_rs
+                                                                    Into::<String>::into(function_rs
                                                                         .get()
                                                                         .published_code
                                                                         .unwrap_or(
-                                                                            "// Code not published yet, publish function to see it here!"
-                                                                                .to_string(),
-                                                                        ),
+                                                                            FunctionCode("// Code not published yet, publish function to see it here!"
```
ayushjain17, line 384: this should be an `Error` , right ?

crates/context_aware_config/src/helpers.rs
```diff
@@ -310,7 +309,6 @@ pub fn add_config_version(
         tags,
         created_at: Utc::now().naive_utc(),
         description,
```
ayushjain17, line 311:
```suggestion
        description: change_reason,
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -294,7 +294,6 @@ pub fn add_config_version(
     state: &Data<AppState>,
     tags: Option<Vec<String>>,
     description: String,
-    change_reason: String,
```
ayushjain17, line 297: this should be kept and the other removed, right ?

crates/context_aware_config/src/api/functions/types.rs
```diff
@@ -1,20 +1,27 @@
...
 pub struct UpdateFunctionRequest {
-    pub function: Option<String>,
-    pub runtime_version: Option<String>,
+    #[serde(rename = "function")]
+    pub draft_code: Option<FunctionCode>,
+    #[serde(rename = "runtime_version")]
+    pub draft_runtime_version: Option<String>,
     pub description: Option<String>,
     pub change_reason: String,
 }
```
ayushjain17, line 19: should we avoid such stuff?
since we dont have an api doc, these types are our only source for api spec and this sort of makes it complicated

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -13,60 +18,17 @@ pub struct CreateReq {
...
-}
-
-#[derive(Debug, Deserialize, Clone)]
+#[derive(Debug, Deserialize, Clone, AsChangeset)]
+#[diesel(table_name = dimensions)]
 pub struct UpdateReq {
     pub position: Option<Position>,
     pub schema: Option<Value>,
-    pub function_name: Option<FunctionNameEnum>,
+    #[serde(default, deserialize_with = "deserialize_function_name")]
```
ayushjain17, line 26: default is not needed, right ?
```suggestion
    #[serde(deserialize_with = "deserialize_function_name")]
```

crates/context_aware_config/src/api/dimension/types.rs
```diff
@@ -13,60 +18,17 @@ pub struct CreateReq {
...
+    pub function_name: Option<Option<String>>,
```
ayushjain17, line 27: again, same point, changing here its like, when these types become common, which they are on the its way of becoming common

in the future sdk, the user would have to send data like: `Some(None)`, `Some(Some(function_name))`
does not make sense in that way

over here, as long as we are seeing this type as something which is constructed only while deserialising, it can make sense, but in the longer run it just simply does not look correct

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -13,41 +17,18 @@ pub struct CreateReq {
...
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, AsChangeset)]
+#[diesel(table_name = default_configs)]
 pub struct UpdateReq {
     #[serde(default, deserialize_with = "deserialize_option")]
     pub value: Option<Value>,
-    pub schema: Option<Map<String, Value>>,
-    pub function_name: Option<FunctionNameEnum>,
+    pub schema: Option<Value>,
+    #[serde(deserialize_with = "deserialize_function_name")]
```
ayushjain17, line 26: and then this requires us to keep writing this everywhere, 
I would still say that there should be a separate type for this use case, as this is an `action` item which needs to be specify and not simply a param

making a separate type for this, clarifies the intent as well

ayushjain17 (changes requested):

crates/context_aware_config/src/helpers.rs
pratikmishra356: we are removing description 
https://github.com/juspay/superposition/pull/390/files#r1952583143

crates/superposition_types/src/database/models/cac.rs
pratikmishra356: we are not encoding decoding function name , 
its function code

crates/superposition_types/src/database/models/cac.rs
pratikmishra356: it does not need ?, we are using and then after that
at the end error will get returned

crates/frontend/src/pages/function.rs
pratikmishra356: its not an error , it is shown in pubslish tab as code in monaco editor

crates/context_aware_config/src/helpers.rs
pratikmishra356: I have made description as param and not change_reason

crates/superposition_types/src/database/models/cac.rs
pratikmishra356: yes its being used

crates/superposition_types/src/database/schema.patch
pratikmishra356: I have verified we do not need any change in schema.patch 
and this mandatory dimensions change is already there in main , the changes are same ideally it should not show something like this , will check after commit squash , not a breaking change

crates/superposition_types/src/database/models/cac.rs
ayushjain17: I meant, can we rather flatten the operations which are happening here

crates/superposition_types/src/database/models/cac.rs
ayushjain17: it is not right, std::str is a standard package, which should be moved at the top

ShreyBana: Personally I feel that maintaining such things w/o CI checks is moot. You can't really rely on it being there through out the code base, just in files you were able to manage. Moreover it's there today & somebody may forget/screw it up  tomorrow. To me it seems like fruitless work. How about we add the necessary checks via rustfmt instead?

crates/context_aware_config/src/api/default_config/types.rs
ShreyBana: Since were making a type, what's a good name for the field? I think something like:
`function_association: Option<FunctionAssociation>`

pratikmishra356: Will do this in next Pr , 
created the issue 
https://github.com/juspay/superposition/issues/428

ayushjain17 (approved):
Pending points being picked up later

ShreyBana (approved):

ShreyBana (approved):

PR #395, opened 2025-01-30
fix: toggle default value

mahatoankitkumar:
## Problem
The toggle component defaults to null instead of a boolean value when initially rendered, causing unexpected behavior until user interaction occurs.

## Solution
Called the onChange event handler to properly handle the initial null state, ensuring the toggle always has a valid boolean value (true/false).

### DIFF
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -57,6 +57,8 @@ where
     let (req_inprogess_rs, req_inprogress_ws) = create_signal(false);
     let (description_rs, description_ws) = create_signal(description);
     let (change_reason_rs, change_reason_ws) = create_signal(change_reason);
+    let (schema_type_rs, schema_type_ws) = create_signal(Err("".to_string()));
+    let (enum_variants_rs, enum_variants_ws) = create_signal(Err("".to_string()));
 
     let functions_resource: Resource<(String, String), Vec<Function>> =
         create_blocking_resource(
@@ -249,8 +251,27 @@ where
                                     dropdown_options=options
                                     on_select=Callback::new(move |selected_item: TypeTemplate| {
                                         logging::log!("selected item {:?}", selected_item);
+                                        let type_schema = selected_item.type_schema.clone();
+                                        let parsed_schema_type = SchemaType::try_from(
+                                            type_schema.clone(),
+                                        );
+                                        let parsed_enum_variants = EnumVariants::try_from(
+                                            type_schema.clone(),
+                                        );
                                         config_type_ws.set(selected_item.type_name);
-                                        config_schema_ws.set(selected_item.type_schema);
+                                        config_schema_ws.set(type_schema);
+                                        schema_type_ws.set(parsed_schema_type.clone());
+                                        enum_variants_ws.set(parsed_enum_variants.clone());
+                                        if let (Ok(schema_type), Ok(enum_variants)) = (
+                                            parsed_schema_type,
+                                            parsed_enum_variants,
+                                        ) {
+                                            if InputType::from((schema_type, enum_variants))
+                                                == InputType::Toggle
+                                            {
+                                                config_value_ws.set(Value::Bool(false));
+                                            }
+                                        }
                                     })
                                 />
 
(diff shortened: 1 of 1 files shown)


Datron (approved):

crates/frontend/src/components/input.rs
```diff
@@ -488,14 +487,17 @@ pub fn input(
     match r#type {
         InputType::Toggle => match value.as_bool() {
             Some(ref v) => {
-                view! { <Toggle value=v.clone() on_change class name/> }.into_view()
+                view! { <Toggle value=v.clone() on_change class name /> }.into_view()
             }
-            None => view! { <Toggle value=false on_change class name/> }.into_view(),
+            None => {
+                on_change.call(Value::Bool(false));
```
ayushjain17, line 493: this would get triggered on every re-render
this should be dealt at the moment when the type is being set

ayushjain17 (changes requested):

crates/frontend/src/components/input.rs
ShubhranshuSanjeev: +1

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -278,6 +278,9 @@ where
                     ) {
                         (Ok(schema_type), Ok(enum_variants)) => {
                             let input_type = InputType::from((schema_type.clone(), enum_variants));
+                            if input_type == InputType::Toggle {
+                                config_value_ws.set(Value::Bool(false));
```
ayushjain17, line 282: this is also at render time

ayushjain17 (changes requested):

Datron (approved):

crates/frontend/src/components/default_config_form.rs
ayushjain17: pending comment

ayushjain17 (changes requested):

Datron (approved):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -251,6 +251,16 @@ where
                                         logging::log!("selected item {:?}", selected_item);
                                         config_type_ws.set(selected_item.type_name);
                                         config_schema_ws.set(selected_item.type_schema);
+                                        if let (Ok(schema_type), Ok(enum_variants)) = (
+                                            SchemaType::try_from(config_schema_rs.get()),
+                                            EnumVariants::try_from(config_schema_rs.get()),
+                                        ) {
+                                            if InputType::from((schema_type, enum_variants))
```
ayushjain17, line 258: are we not going to store it to optimise the repeated computations ?

crates/frontend/src/components/default_config_form.rs
```diff
@@ -249,8 +251,22 @@ where
                                     dropdown_options=options
                                     on_select=Callback::new(move |selected_item: TypeTemplate| {
                                         logging::log!("selected item {:?}", selected_item);
+                                        let type_schema = selected_item.type_schema.clone();
                                         config_type_ws.set(selected_item.type_name);
-                                        config_schema_ws.set(selected_item.type_schema);
+                                        config_schema_ws.set(type_schema.clone());
+                                        schema_type_ws
+                                            .set(SchemaType::try_from(type_schema.clone()));
+                                        enum_variants_ws.set(EnumVariants::try_from(type_schema));
```
ayushjain17, line 259: can we rather store in local variable and then set and then use the local variable in the subsequent lines

crates/frontend/src/components/default_config_form.rs
```diff
@@ -249,8 +251,22 @@ where
...
+                                        if let (Ok(schema_type), Ok(enum_variants)) = (
+                                            schema_type_rs.get(),
+                                            enum_variants_rs.get(),
+                                        ) {
```
ShubhranshuSanjeev, line 263: Please assign the value of schema type and enum variants to some variable and then use it set the signal and the if check.
Please refrain from setting the signal and then immediately reading from it.

ShubhranshuSanjeev (dismissed):

ayushjain17 (approved):

ShubhranshuSanjeev (approved):

PR #398, opened 2025-01-30
feat: Discard experiment

ayushjain17:
## Problem
No option to discard a created experiment

## Solution
Allow discarding of experiment only when it is in created state

Needs update in `experiment_status_type` DB enum

## API changes
| /experiments/{experiment_id}/discard"     | PATCH | `{change_reason: String}` | `ExperimentResponse` |


ShreyBana (changes requested):

ShreyBana (approved):

crates/frontend/src/components/experiment_discard_form.rs
```diff
@@ -0,0 +1,77 @@
...
+            )
+            .await;
+            req_inprogress_ws.set(false);
+            handle_submit_clone()
+        });
+    };
+
+    view! {
+        <h3 class="font-bold text-lg">Discard Experiment</h3>
+        <p class="py-4">Safely discard the experiment without affecting any pre-existing overrides</p>
```
Datron, line 51: Smart sentence

crates/frontend/src/components/experiment_discard_form.rs
```diff
@@ -0,0 +1,77 @@
...
+                </label>
+                <textarea
+                    placeholder="Enter a reason for this change"
+                    class="textarea textarea-bordered w-full max-w-md"
+                    value=change_reason.get_untracked()
+                    on:change=move |ev| {
+                        let value = event_target_value(&ev);
+                        set_change_reason.set(value);
+                    }
+                />
```
Datron, line 65: Lets use the generic input component for this?

crates/frontend/src/components/experiment_discard_form/utils.rs
```diff
@@ -0,0 +1,31 @@
...
+    match client
+        .patch(format!("{host}/experiments/{}/discard", exp_id))
+        .header("x-tenant", tenant)
+        .header("x-org-id", org_id)
+        .json(&json!({"change_reason": change_reason }))
+        .send()
+        .await
+    {
+        Ok(experiment) => {
+            log!("experiment response {:?}", experiment);
```
Datron, line 23: nitpick: 
```suggestion
            log!("experiment discard response {:?}", experiment);
```

crates/frontend/src/pages/default_config.rs
```diff
@@ -78,12 +78,10 @@ pub fn default_config() -> impl IntoView {
...
-        let redirect_url = match key_name {
-            Some(prefix) => {
-                format!("admin/{org_id}/{tenant}/default-config?prefix={prefix}")
-            }
-            None => format!("admin/{org_id}/{tenant}/default-config"),
-        };
+        let mut redirect_url = format!("admin/{org_id}/{tenant}/default-config");
+        if let Some(prefix) = key_name {
+            redirect_url = format!("{redirect_url}?prefix={prefix}")
+        }
```
Datron, line 84: instead of mutation redirect_url, you could just overwrite it. Try this sample:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=049ec5fec47ae8a1253bd871838ed69a

Datron (approved):

workspace_template.sql
```diff
@@ -450,3 +450,5 @@ VALUES (
         null,
         'variantIds are used by experimentation module to manage and select variations'
 );
+
+ALTER TYPE public.experiment_status_type ADD VALUE 'DISCARDED';
```
Datron, line 454: Can we add a failure catch here? This would fail in the future for schemas that already have this Value.

Datron (approved):

crates/frontend/src/components/experiment_discard_form.rs
ayushjain17: @sauraww is making a generic component for description and reason, it is expected to be used here as well

workspace_template.sql
ayushjain17: why would this fail ?
this is anyways going to used only when a new workspace is going to be created, in which case, the type would not exist at all

PR #400, opened 2025-01-31
feat: wrapper connection implementation

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


ShreyBana:
A couple of things.
1. What is happening to the prepared statement objects in postgres? Are we creating them everytime? And if so are they being GC'd?
2. How are we planning to handle schema changes during the execution of flow? For instance, during workspace setup we want to first create the workspace schema, ran from the `public` schema & then switch to the `superposition` schema for inserting into the `workspaces` table, all in one transaction.
3.

crates/service_utils/src/db/types.rs
```diff
@@ -0,0 +1,142 @@
+use diesel::connection::{
+    AnsiTransactionManager, Connection, ConnectionSealed, DefaultLoadingMode,
+    LoadConnection, SimpleConnection, TransactionManager,
+};
+use diesel::pg::{Pg, PgQueryBuilder};
+use diesel::r2d2::{ConnectionManager, PooledConnection};
+use diesel::PgConnection;
+use diesel::RunQueryDsl;
+
+pub struct SuperTransactionManager;
```
ShreyBana, line 10: How about `TransactionManagerImpl`?

crates/service_utils/src/db/types.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    fn transaction_manager_status_mut(
+        conn: &mut SuperConnection,
+    ) -> &mut diesel::connection::TransactionManagerStatus {
+        AnsiTransactionManager::transaction_manager_status_mut(&mut *conn.conn)
+    }
+}
+
+pub struct SuperConnection {
```
ShreyBana, line 45: `ConnectionImpl`?

crates/service_utils/src/service/types.rs
```diff
@@ -198,6 +199,40 @@ impl FromRequest for DbConnection {
...
+            Some(state) => state,
+            None => {
+                log::info!(
+                    "DbConnection-FromRequest: Unable to get app_data from request"
+                );
+                return ready(Err(error::ErrorInternalServerError("")));
+            }
+        };
+
+        let schema_name = req.extensions().get::<SchemaName>().cloned().unwrap().0;
```
ShreyBana, line 222: We might have to update `tenant.rs` to insert `superposition` schema in case we want to add APIs for just querying things from those tables. I don't think we have that today.

ShreyBana:
I am seeing that we don't need the `.schema(..)` DSL w/ this, we might just be able to move back to standard diesel!

ShreyBana:
> ...
>     4. Can you add tests that verify that things are working as intended?

I can do the tests part if something is taking time.

Datron:
Closing due to age

PR #407, opened 2025-02-04
fix: horizontal scroll

mahatoankitkumar:
## Problem
Needed to scroll horizontally for larger values.

## Solution
Added a new component to expand and condense the text and fixed the first column.

https://github.com/user-attachments/assets/61f27df4-1ab2-4356-a1f7-3decdfa38f59

### DIFF
--- a/crates/frontend/src/components/table.rs
+++ b/crates/frontend/src/components/table.rs
@@ -2,10 +2,54 @@ pub mod types;
 
 use crate::{components::pagination::Pagination, schema::HtmlDisplay};
 
-use self::types::{Column, TablePaginationProps};
+use self::types::{Column, Expandable, TablePaginationProps};
 use leptos::*;
 use serde_json::{json, Map, Value};
 use superposition_types::SortBy;
+
+#[component]
+pub fn expandable_text(
+    value: String,
+    formatter: types::CellFormatter,
+    row: Map<String, Value>,
+    class_name: String,
+    is_expandable: Expandable,
+) -> impl IntoView {
+    let stored_value = StoredValue::new(value);
+    let get_child = move |val: &str| formatter(val, &row);
+    let (is_expanded_rs, is_expanded_ws) = create_signal(false);
+    view! {
+        <td class=format!("{} align-top", class_name)>
+            {move || {
+                let value = stored_value.get_value();
+                let is_expanded = is_expanded_rs.get();
+                let (should_expand, displayed_text) = match is_expandable {
+                    Expandable::Enabled(len) if value.len() > len => {
+                        (true, if is_expanded { &value } else { &value[..len] })
+                    }
+                    _ => (false, value.as_str()),
+                };
+                view! {
+                    {get_child(displayed_text).into_view()}
+                    {should_expand
+                        .then(|| {
+                            view! {
+                                <div
+                                    class=format!(
+                                        "inline w-fit text-[#4a00ff] cursor-pointer {}",
+                                        if is_expanded { "pl-2" } else { "" },
+                                    )
+                                    on:click=move |_| is_expanded_ws.update(|val| *val = !*val)
+                                >
+                                    {if is_expanded { "less" } else { "...more" }}
+                                </div>
+                            }
+                        })}
+                }
+            }}
+        </td>
+    }
+}
 
 #[component]
 pub fn table(
@@ -19,18 +63,27 @@ pub fn table(
 ) -> impl IntoView {
     let pagination_props = StoredValue::new(pagination);
     let container_style = format!("{} overflow-x-auto", class);
+    let get_sticky_position_classes = |index: usize| {
+        match index {
+        0 => "sticky left-20 z-20 bg-inherit after:content-[''] after:absolute after:right-0 after:top-0 after:bottom-[-1px] after:w-[5px] after:bg-[linear-gradient(90deg,rgba(0,0,0,0.1)_0%,rgba(0,0,0,0)_100%)]",
+        _ => "",
+    }
+    };
+
     view! {
         <div class=container_style>
             <table class="table table-zebra">
                 <thead class=head_class>
-                    <tr>
-                        <th></th>
+                    <tr class="bg-white">
+                        <th class="sticky left-0 bg-inherit min-w-[5rem] px-3"></th>
 
                         {columns
                             .iter()
                             .filter(|column| !column.hidden)
-                            .map(|column| {
+                            .enumerate()
+                            .map(|(index, column)| {
                                 let column_name = column.name.replace('_', " ");
+                                let sticky_class = get_sticky_position_classes(index);
                                 match column.sortable.clone() {
                                     types::ColumnSortable::Yes {
                                         sort_fn,
@@ -39,7 +92,10 @@ pub fn table(
                                     } => {
                                         view! {
                                             <th
-                                                class="uppercase cursor-pointer"
+                                                class=format!(
+                                                    "uppercase cursor-pointer px-3 {}",
+                                                    sticky_class,
+                                                )
                                                 on:click=move |_| sort_fn.call(())
                                             >
                                                 {column_name}
@@ -59,7 +115,12 @@ pub fn table(
                                         }
                                     }
                                     types::ColumnSortable::No => {
-                                        view! { <th class="uppercase">{column_name}</th> }
+                                        view! {
+                                            <th class=format!(
+                                                "uppercase px-3 {}",
+                                                sticky_class,
+                                            )>{column_name}</th>
+                                        }
                                     }
                                 }
                             })
@@ -86,31 +147,39 @@ pub fn table(
                             } else {
                                 index as i64 + 1
                             };
+                            let cell_class_clone = cell_class.clone();
                             view! {
-                                <tr id=row_id>
-                                    <th class="w-20">{row_num}</th>
+                                <tr id=row_id class="odd:bg-white even:bg-[#f2f2f2]">
+                                    <td class="sticky z-20 bg-inherit left-0 min-w-[5rem] px-3 align-top">
+                                        {row_num}
+                                    </td>
 
                                     {columns
                                         .iter()
                                         .filter(|column| !column.hidden)
-                                        .map(|column| {
-                                            let cname = &column.name;
+                                        .enumerate()
+                                        .map(move |(index, column)| {
+                                            let column = column.clone();
                                             let value: String = row
-                                                .get(cname)
+                                                .get(&column.name)
                                                 .unwrap_or(&Value::String(String::new()))
                                                 .html_display();
+                                            let sticky_class = get_sticky_position_classes(index);
                                             view! {
-                                                <td class=cell_class
-                                                    .to_string()>{(column.formatter)(&value, row)}</td>
+                                                <ExpandableText
+                                                    value
+                                                    formatter=column.formatter
+                                                    row=row.clone()
+                                                    class_name=format!("min-w-48 max-w-106 px-3 break-words font-mono {cell_class_clone} {sticky_class}")
+                                                    is_expandable=column.expandable
+                                                />
                                             }
                                         })
                                         .collect_view()}
-
                                 </tr>
                             }
                         })
                         .collect_view()}
-
                 </tbody>
             </table>
         </div>
--- a/crates/frontend/src/components/table/types.rs
+++ b/crates/frontend/src/components/table/types.rs
@@ -22,11 +22,18 @@ pub enum ColumnSortable {
 }
 
 #[derive(Clone)]
+pub enum Expandable {
+    Disabled,
+    Enabled(usize),
+}
+
+#[derive(Clone)]
 pub struct Column {
     pub name: String,
     pub hidden: bool,
     pub formatter: CellFormatter,
     pub sortable: ColumnSortable,
+    pub expandable: Expandable,
 }
 
 impl PartialEq for Column {
@@ -46,6 +53,16 @@ impl Column {
             hidden: false,
             formatter: Box::new(Rc::new(default_formatter)),
             sortable: ColumnSortable::No,
+            expandable: Expandable::Enabled(100),
+        }
+    }
+    pub fn default_no_collapse(name: String) -> Column {
+        Column {
+            name,
+            hidden: false,
+            formatter: Box::new(Rc::new(default_formatter)),
+            sortable: ColumnSortable::No,
+            expandable: Expandable::Disabled,
         }
     }
     pub fn default_with_sort(name: String, sortable: ColumnSortable) -> Column {
@@ -54,22 +71,25 @@ impl Column {
             hidden: false,
             formatter: Box::new(Rc::new(default_formatter)),
             sortable,
+            expandable: Expandable::Enabled(100),
         }
     }
     pub fn new<NF>(
         name: String,
-        hidden: Option<bool>,
+        hidden: bool,
         formatter: NF,
         sortable: ColumnSortable,
+        expandable: Expandable,
     ) -> Column
     where
         NF: Fn(&str, &Map<String, Value>) -> View + 'static,
     {
         Column {
             name,
-            hidden: hidden.unwrap_or(false),
+            hidden,
             formatter: Box::new(Rc::new(formatter)),
             sortable,
+            expandable,
         }
     }
 }
--- a/crates/frontend/src/components/context_card.rs
+++ b/crates/frontend/src/components/context_card.rs
@@ -44,7 +44,7 @@ pub fn context_card(
     let overrides = store_value(overrides);
 
     let table_columns = vec![
-        Column::default("KEY".to_string()),
+        Column::default_no_collapse("KEY".to_string()),
         Column::default("VALUE".to_string()),
     ];
 
@@ -110,7 +110,6 @@ pub fn context_card(
                     class="xl:w-[400px] h-fit"
                 />
                 <Table
-                    cell_class="min-w-48 font-mono".to_string()
                     rows=override_table_rows
                     key_column="KEY".to_string()
                     columns=table_columns
--- a/crates/frontend/src/components/experiment.rs
+++ b/crates/frontend/src/components/experiment.rs
@@ -21,13 +21,13 @@ fn badge_class(status_type: ExperimentStatusType) -> &'static str {
     }
 }
 
-use super::table::types::ColumnSortable;
+use super::table::types::{ColumnSortable, Expandable};
 
 #[allow(clippy::type_complexity)]
 pub fn gen_variant_table(
     variants: &[Variant],
 ) -> Result<(Vec<Map<String, Value>>, Vec<Column>), String> {
-    let mut columns = vec![Column::default("Config Key".into())];
+    let mut columns = vec![Column::default_no_collapse("Config Key".into())];
     let mut row_map: HashMap<String, Map<String, Value>> = HashMap::new();
     for (i, variant) in variants.iter().enumerate() {
         let name = match variant.variant_type {
@@ -36,9 +36,10 @@ pub fn gen_variant_table(
         };
         columns.push(Column::new(
             name.clone(),
-            None,
+            false,
             |value: &str, _| view! { <span>{value.to_string()}</span> }.into_view(),
             ColumnSortable::No,
+            Expandable::Enabled(100),
         ));
         for (config, value) in variant.overrides.clone().into_inner().into_iter() {
             match row_map.get_mut(&config) {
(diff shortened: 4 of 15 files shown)


crates/frontend/src/components/table.rs
```diff
@@ -23,14 +64,20 @@ pub fn table(
...
                             .filter(|column| !column.hidden)
-                            .map(|column| {
+                            .enumerate()
+                            .map(|(index, column)| {
                                 let column_name = (&column.name).replace('_', " ");
+                                let sticky_class = if index == 0 {
+                                    "sticky left-20 z-20 bg-inherit"
+                                } else {
+                                    ""
+                                };
```
Datron, line 80: This can be a future change, but can the person who creates the table component specify the columns to be sticky, rather than assume the first column should be sticky always?

crates/frontend/src/components/table.rs
```diff
@@ -83,21 +134,38 @@ pub fn table(
...
                                                 .unwrap_or(&Value::String(String::new()))
                                                 .html_display();
-                                            view! {
-                                                <td class=cell_class
-                                                    .to_string()>{(column.formatter)(&value, row)}</td>
+                                            let sticky_class = if index == 0 {
+                                                "sticky left-20 z-20 bg-inherit"
+                                            } else {
+                                                ""
+                                            };
```
Datron, line 154: This seems to be defined twice, anyway to simplify?

crates/frontend/src/components/table.rs
```diff
@@ -83,21 +134,38 @@ pub fn table(
...
+                                                        .to_string()>{(column.formatter)(&value, row)}</td>
+                                                }
+                                                    .into_view()
+                                            } else {
+                                                view! {
+                                                    <ExpandableText
+                                                        text=value
+                                                        cell_class=format!("{} {}", cell_class.clone(), sticky_class)
+                                                        max_length=50
+                                                    />
```
Datron, line 167: ExpandableText is still a span right? This should be wrapped in a td. Maybe you can bring the if condition down.

crates/frontend/src/components/table.rs
```diff
@@ -95,9 +143,20 @@ pub fn table(
                                                 .get(cname)
                                                 .unwrap_or(&Value::String(String::new()))
                                                 .html_display();
-                                            view! {
-                                                <td class=cell_class
-                                                    .to_string()>{(column.formatter)(&value, row)}</td>
+                                            if cname == "actions" {
```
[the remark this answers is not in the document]
Datron: Don't do the if condition, let it all columns be wrapped with ExpandableText inside the td

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,47 @@ use leptos::*;
 use serde_json::{json, Map, Value};
 use superposition_types::SortBy;
 
+#[component]
+pub fn expandable_text(
+    #[prop(into)] text: String,
+    #[prop(into, default = String::new())] cell_class: String,
+    #[prop(default = 100)] max_length: usize,
+) -> impl IntoView {
+    let (is_expanded, set_is_expanded) = create_signal(false);
```
Datron, line 16:
```suggestion
    let (is_expanded_rs, is_expanded_ws) = create_signal(false);
```

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,47 @@ use leptos::*;
...
+                    view! {
+                        <>
+                            {displayed_text}
+                            {if text.len() > max_length {
+                                view! {
+                                    <span
+                                        class="text-blue-500 cursor-pointer ml-2"
+                                        on:click=move |_| set_is_expanded.set(!is_expanded.get())
+                                    >
+                                        {if is_expanded.get() { "Read less" } else { "Read more" }}
```
Datron, line 37: Should we simplify and just show "more" and "less"?
```suggestion
                                        {if is_expanded.get() { "less" } else { "more" }}
```

Datron (changes requested):

crates/frontend/src/components/table/types.rs
```diff
@@ -70,6 +74,7 @@ impl Column {
             hidden: hidden.unwrap_or(false),
             formatter: Box::new(Rc::new(formatter)),
             sortable,
+            expandable: expandable.unwrap_or(true),
```
Datron, line 77: Should it be false by default?

crates/frontend/src/components/table/types.rs
```diff
@@ -54,13 +56,15 @@ impl Column {
...
             sortable,
+            expandable: true,
         }
     }
     pub fn new<NF>(
         name: String,
         hidden: Option<bool>,
         formatter: NF,
         sortable: ColumnSortable,
+        expandable: Option<bool>,
```
Datron, line 67: Is there a way we can avoid option<bool>? In these cases an enum might help
Something like
EXPANDED,
UNEXPANDED,
CANNOT_EXPAND

crates/frontend/src/components/table/types.rs
mahatoankitkumar: I guess default true should be fine as most of the stuff would be expandable, rare cases to not keep it as unexpandable.

crates/frontend/src/components/table/types.rs
mahatoankitkumar: Can be done, but EXPANDED and UNEXPANDED are fine, why CANNOT_EXPAND (didn't get the meaning of this).

Datron: To represent the None case

mahatoankitkumar: But None case defaults to false, so 2 states are fine

Datron: Then why not just bool?

mahatoankitkumar: Makes sense, then hidden can also be changed to bool from option<bool>.

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,59 @@ use leptos::*;
...
+                                {if value_clone_inner.len() > max_length && is_expandable {
+                                    view! {
+                                        <span
+                                            class="text-[#4a00ff] cursor-pointer ml-2"
+                                            on:click=move |_| {
+                                                logging::log!("<<>> clicked");
+                                                set_is_expanded_ws.update(|val| *val = !*val)
+                                            }
+                                        >
+                                            {if is_expanded_rs.get() { "less" } else { "more" }}
```
ShubhranshuSanjeev, line 47: The reactivity won't here, either put this in `Show` or a move closure is required

crates/frontend/src/components/table.rs
```diff
@@ -65,48 +134,61 @@ pub fn table(
...
-                                            }
-                                        })
-                                        .collect_view()}
-
-                                </tr>
-                            }
-                        })
-                        .collect_view()}
+                    {
+                        let rows = rows.to_vec();
```
ShubhranshuSanjeev, line 138: What's the need of `rows.to_vec`?

crates/frontend/src/components/table.rs
```diff
@@ -65,48 +134,61 @@ pub fn table(
...
+                        rows.iter()
+                            .enumerate()
+                            .map(|(index, row)| {
+                                let row = row.clone();
```
ShubhranshuSanjeev, line 142: Is this clone required?

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/table.rs
mahatoankitkumar: But it works, and the outer move block won't help here?

ShubhranshuSanjeev (dismissed):

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+    class_name: String,
+    is_expandable: bool,
+) -> impl IntoView {
+    let value = StoredValue::new(value);
+    let get_child = move |val: &str| (formatter)(val, &row);
+    let (is_expanded_rs, set_is_expanded_ws) = create_signal(false);
+    view! {
+        <td class=class_name>
+            <span>
+                {move || {
```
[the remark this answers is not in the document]
mahatoankitkumar: Its required right? as i am using `is_expandable` signal for reactivity.

crates/frontend/src/components/table/types.rs
```diff
@@ -27,6 +27,7 @@ pub struct Column {
     pub hidden: bool,
     pub formatter: CellFormatter,
     pub sortable: ColumnSortable,
+    pub expandable: bool,
```
ayushjain17, line 30: do we need this ?
saw its `true` in all the places, its not false at all anywhere

crates/frontend/src/components/table.rs
```diff
@@ -19,18 +70,25 @@ pub fn table(
...
+    };
+
     view! {
         <div class=container_style>
             <table class="table table-zebra">
                 <thead class=head_class>
-                    <tr>
-                        <th></th>
+                    <tr class="bg-white">
+                        <th class="sticky left-0 bg-inherit min-w-[5rem]"></th>
```
ayushjain17, line 83: are we pinning the `sl no` column, only ?
shouldn't we also consider pinning the name column also ?
like take this as a config for the table, where like whether to pin the first column or not, along with the `sl no` column ?

crates/frontend/src/components/table.rs
```diff
@@ -39,23 +97,32 @@ pub fn table(
...
                                         }
                                     }
                                     types::ColumnSortable::No => {
-                                        view! { <th class="uppercase">{column_name}</th> }
+                                        view! {
+                                            <th class=format!(
+                                                "uppercase {}",
+                                                sticky_class,
+                                            )>{column_name}</th>
+                                        }
```
ayushjain17, line 125:
```suggestion
                                        view! { <th class=format!("uppercase {sticky_class}")> {column_name} </th> }
```

crates/frontend/src/components/table.rs
```diff
@@ -19,18 +70,25 @@ pub fn table(
...
+        _ => "",
+    };
+
     view! {
         <div class=container_style>
             <table class="table table-zebra">
                 <thead class=head_class>
-                    <tr>
-                        <th></th>
+                    <tr class="bg-white">
```
ayushjain17, line 82: this would make the header row and the first odd row both white, right ?
can we use a third colour over here for the header instead of the 2 being used to represent the data inside

crates/frontend/src/components/table.rs
```diff
@@ -82,31 +149,41 @@ pub fn table(
                             } else {
                                 index as i64 + 1
                             };
+                            let cell_class_clone = cell_class.clone();
                             view! {
-                                <tr id=row_id>
-                                    <th class="w-20">{row_num}</th>
+                                <tr id=row_id class="odd:bg-white even:bg-[#f2f2f2]">
+                                    <th class="sticky z-20 bg-inherit left-0 min-w-[5rem]">
```
ayushjain17, line 155: why is this is an header ?

crates/frontend/src/components/table.rs
```diff
@@ -82,31 +149,41 @@ pub fn table(
...
 
                                     {columns
                                         .iter()
                                         .filter(|column| !column.hidden)
-                                        .map(|column| {
-                                            let cname = &column.name;
+                                        .enumerate()
+                                        .map(move |(index, column)| {
+                                            let column = column.clone();
+                                            let cname = column.name.clone();
```
ayushjain17, line 165: these clones should not be needed right ?

crates/frontend/src/components/table.rs
```diff
@@ -82,31 +149,41 @@ pub fn table(
...
+                                            let sticky_class = get_sticky_position_classes(index);
                                             view! {
-                                                <td class=cell_class
-                                                    .to_string()>{(column.formatter)(&value, row)}</td>
+                                                <ExpandableText
+                                                    value
+                                                    max_length=50
+                                                    formatter=column.formatter
+                                                    row=row.clone()
+                                                    class_name=format!("{} {}", cell_class_clone, sticky_class)
```
ayushjain17, line 177: ?
```suggestion
                                                    class_name=format!("{cell_class_clone} {sticky_class}")
```

crates/frontend/src/components/table.rs
```diff
@@ -82,31 +149,41 @@ pub fn table(
...
+                                                .get(&cname)
                                                 .unwrap_or(&Value::String(String::new()))
                                                 .html_display();
+                                            let sticky_class = get_sticky_position_classes(index);
                                             view! {
-                                                <td class=cell_class
-                                                    .to_string()>{(column.formatter)(&value, row)}</td>
+                                                <ExpandableText
+                                                    value
+                                                    max_length=50
```
ayushjain17, line 174: should this be configurable ?
instead of hardcoding it here ?

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+pub fn expandable_text(
+    value: String,
+    max_length: usize,
+    formatter: types::CellFormatter,
+    row: Map<String, Value>,
+    class_name: String,
+    is_expandable: bool,
+) -> impl IntoView {
+    let value = StoredValue::new(value);
+    let get_child = move |val: &str| (formatter)(val, &row);
```
ayushjain17, line 20: is this bracket needed here ?
is this some mandatory requirement here for the function to be used ?
```suggestion
    let get_child = move |val: &str| formatter(val, &row);
```

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+#[component]
+pub fn expandable_text(
+    value: String,
+    max_length: usize,
+    formatter: types::CellFormatter,
+    row: Map<String, Value>,
+    class_name: String,
+    is_expandable: bool,
+) -> impl IntoView {
+    let value = StoredValue::new(value);
```
ayushjain17, line 19: the `value` variable should be directly usable right ?
why do we need to wrap it in `StoredValue` ?

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+    let get_child = move |val: &str| (formatter)(val, &row);
+    let (is_expanded_rs, set_is_expanded_ws) = create_signal(false);
```
ayushjain17, line 21: since the setter already has `set` in the name, we can drop the `_rs` and `_ws` part, right ?
```suggestion
    let (is_expanded, set_is_expanded) = create_signal(false);
```

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+    view! {
+        <td class=class_name>
+            <span>
+                {move || {
+                    view! {
+                        <span class="flex">
```
ayushjain17, line 27: dont wrap `span` with `span`, many responsiveness things dont work
better to use `div` here

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+                                        )
+                                        on:click=move |_| {
+                                            set_is_expanded_ws.update(|val| *val = !*val)
+                                        }
+                                    >
+                                        {if is_expanded_rs.get() { "less" } else { "...more" }}
+                                    </span>
+                                }
+                            } else {
+                                view! { <span></span> }
```
ayushjain17, line 51: why not this ?
```suggestion
                                view! { <></> }
```

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+                                            if is_expanded_rs.get() { " ml-2" } else { "" },
+                                        )
+                                        on:click=move |_| {
+                                            set_is_expanded_ws.update(|val| *val = !*val)
+                                        }
+                                    >
+                                        {if is_expanded_rs.get() { "less" } else { "...more" }}
+                                    </span>
+                                }
+                            } else {
```
ayushjain17, line 50: can we simplify this, seems way too raw

has quite a few conditions which is bit confusing to understand that what exactly is happening here

ayushjain17 (changes requested):

crates/frontend/src/components/table.rs
```diff
@@ -7,6 +7,57 @@ use leptos::*;
...
+                                get_child(&(value.get_value()[..max_length].to_string()))
+                                    .into_view()
+                            }}
+                            {if value.get_value().len() > max_length && is_expandable {
+                                view! {
+                                    <span
+                                        class=format!(
+                                            "text-[#4a00ff] cursor-pointer{}",
+                                            if is_expanded_rs.get() { " ml-2" } else { "" },
```
ayushjain17, line 41: why is this optional ?
can we please use `gap-2` instead of using margins anywhere

crates/frontend/src/components/table/types.rs
mahatoankitkumar: https://github.com/juspay/superposition/blob/eec7fc366d969e90323ae8d20ce1c55020fd3fef/crates/frontend/src/pages/experiment_list/utils.rs#L153

crates/frontend/src/components/table.rs
mahatoankitkumar: pinning both

crates/frontend/src/components/table.rs
mahatoankitkumar: leptosfmt does this

crates/frontend/src/components/table.rs
mahatoankitkumar: Had to clone at multiple places, @ShubhranshuSanjeev suggested this instead.

crates/frontend/src/components/table.rs
ayushjain17: suggested in-lining `sticky_class` inside format

crates/frontend/src/components/table.rs
mahatoankitkumar: Doesn't work

crates/frontend/src/components/table.rs
mahatoankitkumar: i guess this should be fine for the first cut.

crates/frontend/src/components/table.rs
mahatoankitkumar: Yes mandatory

crates/frontend/src/components/table.rs
mahatoankitkumar: Had to use this.

crates/frontend/src/components/table.rs
mahatoankitkumar: Optional for UI purpose, as it seemed better to have ml-2 when expanded.

crates/frontend/src/components/table.rs
```diff
@@ -19,18 +63,27 @@ pub fn table(
 ) -> impl IntoView {
     let pagination_props = StoredValue::new(pagination);
     let container_style = format!("{} overflow-x-auto", class);
+    let get_sticky_position_classes = |index: usize| {
+        match index {
+        0 => "sticky left-20 z-20 bg-inherit after:content-[''] after:absolute after:right-0 after:top-0 after:bottom-[-1px] after:w-[5px] after:bg-[linear-gradient(90deg,rgba(0,0,0,0.1)_0%,rgba(0,0,0,0)_100%)]",
+        _ => "",
+    }
+    };
```
Datron, line 71: nitpick: would it be better to do:
```suggestion
    // this constant is at the top somewhere
    const STICKY_CLASS_FOR_COL = "sticky left-20 z-20 bg-inherit after:content-[''] after:absolute after:right-0 after:top-0 after:bottom-[-1px] after:w-[5px] after:bg-[linear-gradient(90deg,rgba(0,0,0,0.1)_0%,rgba(0,0,0,0)_100%)]";
    // then use it where ever needed
    let get_sticky_position_classes = if index == 0 { STICKY_CLASS_FOR_COL } else { "" };
    };
```

crates/frontend/src/components/table.rs
Datron: better would be the pin icon to indicate sticky columns

Datron (approved):

crates/frontend/src/components/table.rs
```diff
@@ -86,31 +147,39 @@ pub fn table(
                             } else {
                                 index as i64 + 1
                             };
+                            let cell_class_clone = cell_class.clone();
```
ayushjain17, line 150: why do we need to clone this ?
just want to understand this

ayushjain17 (approved):

crates/frontend/src/components/table.rs
mahatoankitkumar: `cannot move out of `cell_class`, a captured variable in an `FnMut` closure
`cell_class` is moved here`

PR #409, opened 2025-02-07
fix: show delete prompt for delete actions

### LINKED ISSUE #399: No delete popup appears for default config
No delete popup appears for default config

Datron:
## Problem
Delete modal is not displayed in default config and type templates pages. Also, the type template form shows the description and change reason twice

## Solution
Show a delete prompt and remove duplicate fields

https://github.com/user-attachments/assets/d29819bf-97e1-4248-aa6c-09c13247ee8c

closes #399

### DIFF
--- a/crates/frontend/src/components/type_template_form.rs
+++ b/crates/frontend/src/components/type_template_form.rs
@@ -119,36 +119,6 @@ where
                 <textarea
                     placeholder="Enter description"
                     class="textarea textarea-bordered w-full max-w-md"
-                    value=description_rs.get_untracked()
-                    on:change=move |ev| {
-                        let value = event_target_value(&ev);
-                        description_ws.set(value);
-                    }
-                />
-            </div>
-
-            <div class="form-control">
-                <label class="label">
-                    <span class="label-text">change_reason</span>
-                </label>
-                <textarea
-                    placeholder="Enter change_reason"
-                    class="textarea textarea-bordered w-full max-w-md"
-                    value=change_reason_rs.get_untracked()
-                    on:change=move |ev| {
-                        let value = event_target_value(&ev);
-                        change_reason_ws.set(value);
-                    }
-                />
-            </div>
-
-            <div class="form-control">
-                <label class="label">
-                    <span class="label-text">Description</span>
-                </label>
-                <textarea
-                    placeholder="Enter description"
-                    class="textarea textarea-bordered w-full max-w-md"
                     value=description_rs.get()
                     on:change=move |ev| {
                         let value = event_target_value(&ev);
--- a/crates/frontend/src/pages/custom_types.rs
+++ b/crates/frontend/src/pages/custom_types.rs
@@ -6,12 +6,15 @@ use superposition_types::custom_query::PaginationParams;
 use crate::components::table::types::ColumnSortable;
 use crate::components::type_template_form::utils::delete_type;
 use crate::components::{
+    alert::AlertType,
+    delete_modal::DeleteModal,
     drawer::{close_drawer, open_drawer, Drawer, DrawerBtn},
     skeleton::Skeleton,
     stat::Stat,
     table::Table,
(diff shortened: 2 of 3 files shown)


crates/frontend/src/components/type_template_form.rs
```diff
@@ -112,36 +112,6 @@ where
 
             </div>
 
-            <div class="form-control">
```
mahatoankitkumar, line 115: @Datron We don't have this description change for type_templates?

crates/frontend/src/pages/custom_types.rs
```diff
@@ -42,6 +45,34 @@ pub fn types_page() -> impl IntoView {
                 )
         },
     );
+
+    let (delete_modal_visible_rs, delete_modal_visible_ws) = create_signal(false);
```
mahatoankitkumar, line 49: @Datron Don't you think this signal should be inside the DeleteModal component?

crates/frontend/src/pages/custom_types.rs
```diff
@@ -42,6 +45,34 @@ pub fn types_page() -> impl IntoView {
...
+    let (delete_row_rs, delete_row_ws) = create_signal::<Option<TypeTemplateRow>>(None);
+
+    let confirm_delete = Callback::new(move |_| {
+        if let Some(row_data) = delete_row_rs.get().clone() {
+            spawn_local(async move {
+                let tenant = tenant_rws.get().0;
+                let org = org_rws.get().0;
+                let api_response =
+                    delete_type(tenant, row_data.clone().type_name, org).await;
+                match api_response {
```
mahatoankitkumar, line 59: Same for this?
And we can pass the api call function as a callback?

mahatoankitkumar (changes requested):

crates/frontend/src/pages/custom_types.rs
sauraww: Hey @Datron , both of the @mahatoankitkumar's comment make a lot of sense . We can add the show/hide method inside the component itself so that visibility state can be managed via component itself rather than being passed from the parent . Can we do these changes right now or log a ticket for the refactor which can be picked up later on ?

sauraww (changes requested):

crates/frontend/src/components/type_template_form.rs
Datron: It's there, duplicated twice. Removed one copy

crates/frontend/src/pages/custom_types.rs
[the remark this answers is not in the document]
Datron: The feature was called custom types in the beginning

crates/frontend/src/pages/custom_types.rs
Datron: @mahatoankitkumar the API call is different for each delete right? Should the whole of the function be a callback?

crates/frontend/src/pages/custom_types.rs
Datron: @mahatoankitkumar the modal has this inside, but since the modal is controlled by the outside page by these signals - they still need to be present right? Otherwise how do you trigger the display and hide of the modal?

crates/frontend/src/pages/custom_types.rs
mahatoankitkumar: Yeah that can be, as rest all is common for all, and the api call would take only 3-4 lines as a callback function, rest can be common

Datron: The types for each function call are different, so the Callback needs to be generic. There maybe a better way to solve this.

mahatoankitkumar: As per our discussion, we can push this refactor for later as this might have some complexities and drawbacks.

mahatoankitkumar (approved):

ShreyBana (approved):

PR #411, opened 2025-02-10
fix: update rust-monaco version

Datron:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/frontend/Cargo.toml
+++ b/crates/frontend/Cargo.toml
@@ -17,7 +17,7 @@ js-sys = "0.3.65"
 leptos = { workspace = true }
 leptos_meta = { version = "0.6.11" }
 leptos_router = { version = "0.6.11" }
-monaco = { git = "https://github.com/datron/rust-monaco.git" }
+monaco = { git = "https://github.com/datron/rust-monaco.git", rev = "5c9c93a" }
 once_cell = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/makefile
+++ b/makefile
@@ -4,6 +4,7 @@ TENANT ?= dev
 SHELL := /usr/bin/env bash
 FEATURES ?= ssr
 CARGO_FLAGS := --color always --no-default-features
+WASM_PACK_MODE ?= --dev
 HAS_DOCKER := $(shell command -v docker > /dev/null; echo $$?)
 HAS_PODMAN := $(shell command -v podman > /dev/null; echo $$?)
 ifeq ($(HAS_DOCKER),0)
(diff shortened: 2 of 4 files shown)


crates/frontend/Cargo.toml
```diff
@@ -17,7 +17,7 @@ js-sys = "0.3.65"
 leptos = { workspace = true }
 leptos_meta = { version = "0.6.11" }
 leptos_router = { version = "0.6.11" }
-monaco = { git = "https://github.com/datron/rust-monaco.git" }
+monaco = { git = "https://github.com/datron/rust-monaco.git", version = "0.5.1" }
```
ayushjain17, line 20: let's use commit hash over here, using version number is a pain for repos which do not have all the versions pushed as proper releases
the moment, a version change is done in such repo's cargo fails to fetch the given version, as the version as per HEAD is different for that repo

for commit hashes the issue is not there, as it can track all commits via git

Datron: Either way works, but this is easier to track. We just need the discipline to update the version when we make a change -  and changes are rare. Also we own this repo, so can maintain that discipline.

ayushjain17: that is fine, but the moment the version is changed, it starts breaking for others who do not have a local cache/build for that version

ayushjain17 (approved):

ShreyBana (approved):

makefile
```diff
@@ -4,6 +4,7 @@ TENANT ?= dev
 SHELL := /usr/bin/env bash
 FEATURES ?= ssr
 CARGO_FLAGS := --color always --no-default-features
+WASM_PACK_MODE := --dev
```
ShreyBana, line 7: You'll have to use `?=` if you want it to be overridable.

.github/workflows/ci_check_pr.yaml
```diff
@@ -129,3 +129,4 @@ jobs:
                   make test CI=1
```
ShreyBana, line 129: For it to take effect you'll have to do `make test CI=1 WASM_PACK_MODE=--profiling`

ShreyBana (changes requested):

.github/workflows/ci_check_pr.yaml
ShreyBana: Alternatively, you can add
`test: WASM_PACK_MODE=--profiling`
above the `test` recipe so as to override it, this way, whenever someone runs `test`, it automatically changes rather than having to pass it through.

Datron: Done

ShreyBana (approved):

PR #416, opened 2025-02-13
feat: add key filter for default config

Datron:
## Problem
Default config screen is paginated, but provides no way to search for a key

## Solution
Provide a filter to find config keys

https://github.com/user-attachments/assets/d52d23dd-b7de-4954-bb38-3e1801365508

### DIFF
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -4,7 +4,10 @@ use actix_web::{
     HttpResponse, Scope,
 };
 use chrono::Utc;
-use diesel::{Connection, ExpressionMethods, QueryDsl, RunQueryDsl, SelectableHelper};
+use diesel::{
+    Connection, ExpressionMethods, QueryDsl, RunQueryDsl, SelectableHelper,
+    TextExpressionMethods,
+};
 use jsonschema::{Draft, JSONSchema, ValidationError};
 use serde_json::Value;
 use service_utils::{
@@ -15,6 +18,7 @@ use superposition_macros::{
     bad_argument, db_error, not_found, unexpected_error, validation_error,
 };
 use superposition_types::{
+    api::default_config::DefaultConfigFilters,
     custom_query::PaginationParams,
     database::{
         models::cac::{self as models, Context, DefaultConfig},
@@ -40,7 +44,7 @@ pub fn endpoints() -> Scope {
     Scope::new("")
         .service(create_default_config)
         .service(update_default_config)
-        .service(get)
+        .service(list_default_configs)
         .service(delete)
 }
 
@@ -294,17 +298,28 @@ fn fetch_default_key(
 }
 
 #[get("")]
-async fn get(
+async fn list_default_configs(
     db_conn: DbConnection,
-    filters: Query<PaginationParams>,
+    pagination: Query<PaginationParams>,
+    filters: Query<DefaultConfigFilters>,
     schema_name: SchemaName,
 ) -> superposition::Result<Json<PaginatedResponse<DefaultConfig>>> {
     let DbConnection(mut conn) = db_conn;
 
-    if let Some(true) = filters.all {
-        let result: Vec<DefaultConfig> = dsl::default_configs
-            .schema_name(&schema_name)
-            .get_results(&mut conn)?;
+    let filters = filters.into_inner();
+
+    let query_builder = |filters: &DefaultConfigFilters| {
+        let mut builder = dsl::default_configs.schema_name(&schema_name).into_boxed();
+        if let Some(ref config_name) = filters.name {
+            builder = builder
+                .filter(schema::default_configs::key.like(format!["%{}%", config_name]));
+        }
+        builder
+    };
+
+    if let Some(true) = pagination.all {
+        let result: Vec<DefaultConfig> =
+            query_builder(&filters).get_results(&mut conn)?;
         return Ok(Json(PaginatedResponse {
             total_pages: 1,
             total_items: result.len() as i64,
@@ -312,17 +327,13 @@ async fn get(
         }));
     }
 
-    let n_default_configs: i64 = dsl::default_configs
-        .count()
-        .schema_name(&schema_name)
-        .get_result(&mut conn)?;
-    let limit = filters.count.unwrap_or(10);
-    let mut builder = dsl::default_configs
-        .order(dsl::created_at.desc())
-        .limit(limit)
-        .schema_name(&schema_name)
-        .into_boxed();
-    if let Some(page) = filters.page {
+    let base_query = query_builder(&filters);
+    let count_query = query_builder(&filters);
+
+    let n_default_configs: i64 = count_query.count().get_result(&mut conn)?;
+    let limit = pagination.count.unwrap_or(10);
+    let mut builder = base_query.order(dsl::created_at.desc()).limit(limit);
+    if let Some(page) = pagination.page {
         let offset = (page - 1) * limit;
         builder = builder.offset(offset);
     }
--- a/crates/frontend/src/api.rs
+++ b/crates/frontend/src/api.rs
@@ -1,5 +1,6 @@
 use leptos::ServerFnError;
 use superposition_types::{
+    api::default_config::DefaultConfigFilters,
     custom_query::PaginationParams,
     database::{
         models::{
@@ -45,14 +46,15 @@ pub async fn fetch_dimensions(
 
 // #[server(GetDefaultConfig, "/fxn", "GetJson")]
 pub async fn fetch_default_config(
-    filters: &PaginationParams,
+    pagination: &PaginationParams,
+    filters: &DefaultConfigFilters,
     tenant: String,
     org_id: String,
 ) -> Result<PaginatedResponse<DefaultConfig>, ServerFnError> {
     let client = reqwest::Client::new();
     let host = use_host_server();
 
-    let url = format!("{}/default-config?{}", host, filters);
+    let url = format!("{}/default-config?{}&{}", host, pagination, filters);
     let response: PaginatedResponse<DefaultConfig> = client
         .get(url)
         .header("x-tenant", tenant)
--- a/crates/frontend/src/pages/default_config.rs
+++ b/crates/frontend/src/pages/default_config.rs
@@ -3,12 +3,15 @@ use std::collections::HashSet;
 use leptos::*;
 use leptos_router::{use_navigate, use_query_map};
 use serde_json::{json, Map, Value};
-use superposition_types::custom_query::PaginationParams;
+use superposition_types::{
+    api::default_config::DefaultConfigFilters, custom_query::PaginationParams,
+};
 
 use crate::api::{delete_default_config, fetch_default_config};
 use crate::components::table::types::Expandable;
 use crate::components::{
     alert::AlertType,
+    button::Button,
     default_config_form::DefaultConfigForm,
     delete_modal::DeleteModal,
     drawer::{close_drawer, open_drawer, Drawer, DrawerBtn},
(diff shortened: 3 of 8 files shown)


mahatoankitkumar:
Attach video please

crates/frontend/src/api.rs
```diff
@@ -45,14 +45,21 @@ pub async fn fetch_dimensions(
...
     let client = reqwest::Client::new();
     let host = use_host_server();
 
-    let url = format!("{}/default-config?{}", host, filters.to_string());
+    let filters = filters.to_string();
+    let url = if filters.is_empty() {
+        format!("{}/default-config?{}", host, pagination)
+    } else {
+        format!("{}/default-config?{}&{}", host, pagination, filters)
+    };
```
ayushjain17, line 61: this will simply work
```suggestion
    let url = format!("{host}/default-config?{pagination}&{filters}");
```

crates/context_aware_config/src/api/default_config/types.rs
```diff
@@ -73,3 +73,8 @@ where
     let value: Value = Deserialize::deserialize(deserializer)?;
     Ok(Some(value))
 }
+
+#[derive(Deserialize, Debug)]
+pub struct DefaultConfigFilters {
+    pub name: Option<String>,
+}
```
ayushjain17, line 80: you have already created a type in `superposition_types`, can we directly use that

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -316,35 +319,42 @@ fn fetch_default_key(
...
-        let result: Vec<DefaultConfig> = dsl::default_configs
-            .schema_name(&schema_name)
-            .get_results(&mut conn)?;
+    let filters = filters.into_inner();
+
+    let query_builder = |filters: &DefaultConfigFilters| {
+        let mut builder = dsl::default_configs.schema_name(&schema_name).into_boxed();
+        if let Some(ref config_name) = filters.name {
+            builder = builder
+                .filter(schema::default_configs::key.like(format!["%{}%", config_name]));
```
ayushjain17, line 336: can you please share the query which is going to be generated here ?
with the if let condition, this does not seem much extensible as well

crates/frontend/src/pages/default_config.rs
```diff
@@ -482,6 +496,90 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+                                    Some(key_name)
+                                };
+                                let filters = filters_buffer_rws.get();
+                                let filters = DefaultConfigFilters { name: key_name, ..filters };
+                                filters_buffer_rws.set(filters);
+                            }
+                        />
+                    </div>
+                </div>
+                <div class="flex justify-start mt-8">
```
ayushjain17, line 555: please do not use margin
rather make parent `flex flex-col gap-8`

crates/frontend/src/pages/default_config.rs
```diff
@@ -482,6 +496,90 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+                    <Button
+                        class="pl-[70px] pr-[70px] w-48 h-12".to_string()
```
ayushjain17, line 557:
```suggestion
                        class="h-12 w-48 px-[70px]".to_string()
```

crates/frontend/src/pages/default_config.rs
```diff
@@ -482,6 +496,90 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+                        text="Submit".to_string()
+                        on_click=move |event| {
+                            event.prevent_default();
+                            let filter = filters_buffer_rws.get();
+                            filters_rws.set(filter);
+                            close_drawer("default_config_filter_drawer")
+                        }
+                    />
+                    <Button
+                        class="pl-[70px] pr-[70px] w-48 h-12".to_string()
```
ayushjain17, line 567:
```suggestion
                        class="h-12 w-48 px-[70px]".to_string()
```

crates/frontend/src/pages/default_config.rs
```diff
@@ -482,6 +496,90 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+                            id="default-config-name-filter"
+                            placeholder="eg: city"
+                            class="input input-bordered rounded-md resize-y w-full max-w-md"
+                            value=move || {
+                                let prefix = key_prefix.get();
+                                if prefix.is_some() {
+                                    prefix
+                                } else {
+                                    filters_buffer_rws.get().name
+                                }
```
ayushjain17, line 539: shouldn't it be this ?
because if the prefix is some value, but I update the filters value, I would still be showing prefix value itself
```suggestion
                                if prefix.is_some() && filters_buffer_rws.get().name.is_empty() {
                                    prefix
                                } else {
                                    filters_buffer_rws.get().name
                                }
```

crates/frontend/src/pages/default_config.rs
```diff
@@ -378,6 +386,12 @@ pub fn default_config() -> impl IntoView {
                                         </DrawerBtn>
                                     </div>
                                 </div>
+                                <div class="flex justify-start m-1">
```
ayushjain17, line 389: please avoid using margin

ayushjain17 (dismissed):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -316,35 +319,42 @@ fn fetch_default_key(
...
+        }
+        builder
+    };
+
+    if let Some(true) = pagination.all {
```
pratikmishra356, line 341: A suggestion for all list apis , even if user wants all the rows , we  should set some hard limit on backend side 
lets se 1000 or 10,000 and we can provide total pages like not always 1

crates/frontend/src/pages/default_config.rs
```diff
@@ -482,6 +496,90 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+            style="cursor-pointer btn btn-purple-outline m-1".to_string()
+        >
+            Filters
+            <i class="ri-filter-3-line"></i>
+        </DrawerBtn>
+        <Drawer
+            id="default_config_filter_drawer".to_string()
+            header="default_config Filters"
+            drawer_width="w-[50vw]"
+            handle_close=move || {
```
pratikmishra356, line 518: clicking outside does not closes the drawer

crates/superposition_types/src/custom_query.rs
```diff
@@ -286,6 +286,21 @@ where
...
+impl Display for DefaultConfigFilters {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut query_params = vec![];
+        if let Some(key_name) = &self.name {
+            query_params.push(format!("name={}", key_name));
+        }
+        write!(f, "{}", query_params.join("&"))
+    }
+}
+
```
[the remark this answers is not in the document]
Datron: Would that make sense? It would be a file with just this code.

crates/frontend/src/pages/default_config.rs
Datron: Can you give a reason for not using margin?

Datron: nevermind figured it out

crates/frontend/src/pages/default_config.rs
Datron: Good catch

crates/frontend/src/pages/default_config.rs
```diff
@@ -480,6 +492,89 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+                            <span class="label-text">Configuration Name</span>
+                        </label>
+                        <input
+                            type="text"
+                            id="default-config-name-filter"
+                            placeholder="eg: city"
+                            class="input input-bordered rounded-md resize-y w-full max-w-md"
+                            value=move || {
+                                let prefix = key_prefix.get();
+                                if prefix.is_some() && filters_buffer_rws.get().name.is_none() {
```
ShreyBana, line 531: So how's this supposed to work? The prefix is the default value that the input starts w/?

ShreyBana (changes requested):

crates/superposition_types/src/custom_query.rs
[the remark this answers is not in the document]
ayushjain17: no that is for now, that file is supposed to have all such api types in it related to default config, which is in progress and this is a deviation from it

crates/frontend/src/pages/default_config.rs
ayushjain17: yep, always use `gap` and `padding` and never use single child div unless there is no other option
because not following either of these leads to a lot of issues in responsiveness of the page

if you follow these, often the desired responsiveness is given by the browser without any extra effort

crates/context_aware_config/src/api/default_config/handlers.rs
Datron: @ShreyBana I think if someone has enough default configs that fetching all of them impacts the database (which is itself a very large number of configs), it would happen from years of use of superposition. We can probably wait this out and see if it actually happens. Either way, it is out of scope of this PR

crates/frontend/src/pages/default_config.rs
Datron: Right, if someone wants to search within a prefix this simplifies their input entry. The behaviour is visible in the video example.

crates/superposition_types/src/custom_query.rs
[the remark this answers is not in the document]
Datron: Should we wait on this till it becomes relevant?

crates/frontend/src/api.rs
```diff
@@ -45,14 +45,15 @@ pub async fn fetch_dimensions(
 
 // #[server(GetDefaultConfig, "/fxn", "GetJson")]
 pub async fn fetch_default_config(
-    filters: &PaginationParams,
+    pagination: &PaginationParams,
```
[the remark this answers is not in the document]
Datron: The type provides some hint as well? `pagination` seems sufficient to me.

crates/context_aware_config/src/api/default_config/handlers.rs
ShreyBana: I agree default config is the least likely to cause this sort of a problem. But, even so, we should have still have a hard limit. Not requesting you to change it right now, it's a broader issue w/ list APIs, should be dealt separately.

crates/frontend/src/pages/default_config.rs
ShreyBana: Didn't catch it in the video 😬, my bad.

crates/frontend/src/api.rs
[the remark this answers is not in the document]
ShreyBana: Yeah this is also fine ig, though I would personally prefer `page_params`/`pagep`, or some kind of an abbreviation of the type, as otherwise the name sounds a bit incomplete. Though one can get used to it, not a big deal.

ShreyBana (approved):
UI seems, good nothing from me. But @ayushjain17 has a point, sooner we move to correctly organized code the better.

crates/superposition_types/src/custom_query.rs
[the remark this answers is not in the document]
ayushjain17: this is in types files itself, why should we unnecessarily wait to only move it, when we know it has to be there only

crates/frontend/src/pages/default_config.rs
```diff
@@ -480,6 +492,89 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+            id="default_config_filter_drawer".to_string()
+            header="default_config Filters"
+            drawer_width="w-[50vw]"
+            handle_close=move || {
+                close_drawer("default_config_filter_drawer");
+            }
+        >
+            <div class="card-body">
+                <div class="flex flex-col gap-9 justify-between">
+                    <div class="form-control">
```
ayushjain17, line 520: can we please merge these two divs
```suggestion
                <div class="form-control flex flex-col gap-9 justify-between">
```

crates/frontend/src/pages/default_config.rs
```diff
@@ -480,6 +492,89 @@ pub fn get_bread_crums(key_prefix: Option<String>) -> Vec<BreadCrums> {
...
+        <DrawerBtn
+            drawer_id="default_config_filter_drawer".into()
+            style="cursor-pointer btn btn-purple-outline m-1 w-[8rem]".to_string()
+        >
+            Filters
+            <i class="ri-filter-3-line"></i>
+        </DrawerBtn>
+        <Drawer
+            id="default_config_filter_drawer".to_string()
+            header="default_config Filters"
```
ayushjain17, line 512: shouldn't the header be something like: (you wouldn't want to use snake case in the displayed text in UI)
```suggestion
            header="Default Config Filters"
```

crates/superposition_types/src/custom_query.rs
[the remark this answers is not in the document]
ayushjain17: this is in types files itself, why should we unnecessarily wait to only move it, when we know it has to be there only

ayushjain17 (approved):

PR #419, opened 2025-02-14
fix: Update experiments page

ayushjain17:
## Problem
Fields not being passed while updating experiments

## Solution
Updating contract in frontend (Not the best fix)

### DIFF
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -439,24 +439,30 @@ async fn bulk_operations(
                             err
                         })?;
 
-                        let ctx_condition = put_req.context.to_owned().into_inner();
-                        let ctx_condition_value =
-                            Value::Object(ctx_condition.clone().into());
-
-                        let description = if put_req.description.is_none() {
-                            ensure_description(
-                                ctx_condition_value.clone(),
-                                transaction_conn,
-                                &schema_name,
-                            )?
-                        } else {
-                            put_req
-                                .description
-                                .expect("Description should not be empty")
-                        };
-                        all_descriptions.push(description);
+                        all_descriptions.push(put_resp.description.clone());
                         all_change_reasons.push(put_req.change_reason.clone());
                         response.push(ContextBulkResponse::Put(put_resp));
+                    }
+                    ContextAction::Replace(put_req) => {
+                        let put_resp = operations::put(
+                            Json(put_req.clone()),
+                            transaction_conn,
+                            true,
+                            &user,
+                            &schema_name,
+                            true,
+                        )
+                        .map_err(|err| {
+                            log::error!(
+                                "Failed at insert into contexts due to {:?}",
+                                err
+                            );
+                            err
+                        })?;
+
+                        all_descriptions.push(put_resp.description.clone());
+                        all_change_reasons.push(put_req.change_reason.clone());
+                        response.push(ContextBulkResponse::Replace(put_resp));
                     }
                     ContextAction::Delete(ctx_id) => {
                         let context: Context = contexts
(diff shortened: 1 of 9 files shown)


Datron (approved):

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -439,22 +439,7 @@ async fn bulk_operations(
...
-                                ctx_condition_value.clone(),
-                                transaction_conn,
-                                &schema_name,
-                            )?
-                        } else {
-                            put_req
-                                .description
-                                .expect("Description should not be empty")
-                        };
-                        all_descriptions.push(description);
```
ayushjain17, line 457: this was redundant
this function [here](https://github.com/juspay/superposition/pull/419/files#diff-d134f3fb88f77342e5f42aaac3b57af7631de3f0926a1c0cd0e6b702d0e97c36R426) does all the `ensure_description` handling already

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1049,18 +1049,6 @@ async fn update_overrides(
...
-        let context_id = existing_variant.context_id.ok_or_else(|| {
-            log::error!(
-                "context id not available for variant {:?}",
-                existing_variant.id
-            );
-            unexpected_error!("Something went wrong, failed to update experiment")
-        })?;
-        cac_operations.push(ContextAction::DELETE(context_id.to_string()));
-    }
-
```
ayushjain17, line 1063: ancient logic:
logic here earlier was delete context and then recreate context (which might have been earlier been need earlier)
but in the current state of the service this is no longer needed (this was creating a bug related to description)

crates/frontend/src/components/experiment_form.rs
```diff
@@ -117,7 +117,15 @@ where
...
-                    update_experiment(experiment_id, f_variants, tenant, org).await
+                    update_experiment(
+                        experiment_id,
+                        f_variants,
+                        tenant,
+                        org,
+                        description_rs.get(),
+                        change_reason_rs.get(),
+                    )
+                    .await
```
ayushjain17, line 128: `create_experiment` was updated [here](https://github.com/juspay/superposition/pull/419/files#diff-d5654b362e303f1a000869ca92f73d818ec4c009b22298a9abd3acc9c4028d43R130-R138) but the update was never handled

crates/frontend/src/components/experiment_form/types.rs
```diff
@@ -37,4 +37,6 @@ impl FromIterator<VariantFormT> for Vec<VariantUpdateRequest> {
 #[derive(Serialize, Debug)]
 pub struct ExperimentUpdateRequest {
     pub variants: Vec<VariantUpdateRequest>,
+    pub description: Option<String>,
+    pub change_reason: String,
 }
```
ayushjain17, line 42: I will be moving this to `superposition_types`

crates/experimentation_platform/src/api/experiments/types.rs
ayushjain17, line 1: backwards compatibility changes

crates/experimentation_platform/src/api/experiments/handlers.rs
ShubhranshuSanjeev: If we don't delete the context, then the overrides will be merged.

Old override:
Key1, Key2

New override:
Key2, Key3

**This will result in**

Merged Override:
Key1, Key2, Key3

crates/experimentation_platform/src/api/experiments/types.rs
mahatoankitkumar: We could have updated the UI by introducing the change reason directly, instead of ensuring backward compatibility?

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: yeah, added REPLACE option

ShubhranshuSanjeev (approved):

mahatoankitkumar (approved):

PR #424, opened 2025-02-19
feat: experiment filters in UI

Datron:
## Problem
It is hard to search for experiments

## Solution
Add the ability to filter and manage experiments. Search is done on the backend, so even paginated experiments can be searched for

https://github.com/user-attachments/assets/cc9f1344-fc7b-4868-885f-65f2a3085703

### DIFF
--- a/crates/frontend/src/components/input.rs
+++ b/crates/frontend/src/components/input.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use chrono::{DateTime, Utc};
 use leptos::*;
 use serde_json::{json, Map, Value};
 
@@ -476,6 +477,58 @@ pub fn monaco_input(
                 }}
 
             </Show>
+        </div>
+    }
+}
+
+#[component]
+pub fn date_input(
+    id: String,
+    class: String,
+    name: String,
+    on_change: Callback<DateTime<Utc>, ()>,
+    #[prop(into, default = Utc::now().format("%Y-%m-%d").to_string())] value: String,
+    #[prop(default = false)] disabled: bool,
+    #[prop(default = false)] required: bool,
+    #[prop(into, default = (Utc::now() - chrono::Duration::weeks(4)).format("%Y-%m-%d").to_string())]
+    min: String,
+    #[prop(into, default = Utc::now().format("%Y-%m-%d").to_string())] max: String,
+) -> impl IntoView {
+    let (error_rs, error_ws) = create_signal::<String>(String::new());
+    view! {
+        <div class="flex flex-col gap-1">
+            <input
+                id=id
+                name=name
+                type="date"
+                class=format!("input input-bordered {}", class)
+                required=required
+                disabled=disabled
+                min=min
+                max=max
+                value=value
+                on:change=move |e| {
+                    let date = format!("{}T00:00:00Z", event_target_value(&e));
+                    logging::log!("The date selected is: {}", date);
+                    match DateTime::parse_from_rfc3339(&date) {
+                        Ok(v) => {
+                            error_ws.set(String::new());
+                            on_change.call(v.to_utc());
+                        }
+                        Err(e) => {
+                            logging::log!("error occurred: {:?}", e);
+                            error_ws.set(e.to_string());
+                        },
+                    }
+                }
+            />
+            <Show when=move || !error_rs.get().is_empty() >
+                <span class="flex gap-2 px-4 text-xs font-semibold text-red-600">
+                    <i class="ri-close-circle-line"></i>
+                    {move || error_rs.get()}
+                </span>
+            </Show>
+
         </div>
     }
 }
--- a/crates/frontend/src/pages/experiment_list.rs
+++ b/crates/frontend/src/pages/experiment_list.rs
@@ -1,23 +1,39 @@
 pub mod utils;
 
-use chrono::{prelude::Utc, TimeZone};
+use std::collections::HashSet;
+
+use chrono::{prelude::Utc, DateTime};
 use futures::join;
 use leptos::*;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Map, Value};
+use strum::IntoEnumIterator;
 use superposition_types::{
     api::default_config::DefaultConfigFilters,
     custom_query::PaginationParams,
-    database::{models::cac::DefaultConfig, types::DimensionWithMandatory},
-    PaginatedResponse, SortBy,
+    database::{
+        models::{cac::DefaultConfig, experimentation::ExperimentStatusType},
+        types::DimensionWithMandatory,
+    },
+    PaginatedResponse,
 };
 use utils::experiment_table_columns;
 
-use crate::components::drawer::{close_drawer, Drawer, DrawerBtn};
-use crate::components::skeleton::Skeleton;
-use crate::components::table::types::TablePaginationProps;
-use crate::components::{experiment_form::ExperimentForm, stat::Stat, table::Table};
 use crate::logic::Conditions;
+use crate::{
+    components::{
+        button::Button,
+        context_form::ContextForm,
+        drawer::{close_drawer, Drawer, DrawerBtn},
+        dropdown::DropdownDirection,
+        experiment_form::ExperimentForm,
+        input::DateInput,
+        skeleton::Skeleton,
+        stat::Stat,
+        table::{types::TablePaginationProps, Table},
+    },
+    types::StatusTypes,
+};
 
 use crate::providers::condition_collapse_provider::ConditionCollapseProvider;
 use crate::providers::editor_provider::EditorProvider;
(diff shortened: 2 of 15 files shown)


crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,22 +49,338 @@ struct CombinedResource {
...
+                </div>
+                <div class="flex justify-start mt-8">
+                    <Button
+                        class="pl-[70px] pr-[70px] w-48 h-12".to_string()
+                        text="Submit".to_string()
+                        on_click=move |event| {
+                            event.prevent_default();
+                            let filter = filters_buffer_rws.get();
+                            filters_rws.set(filter);
+                            close_drawer("experiment_filter_drawer")
```
ShreyBana, line 358: This string probably should either an enum or a constant.

ShubhranshuSanjeev (changes requested):

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,22 +49,338 @@ struct CombinedResource {
...
+                        }
+                    />
+                    <Button
```
ShreyBana, line 361: This should be styled as some kind of cancellation button, maybe a cross?

crates/frontend/src/types.rs
```diff
@@ -88,6 +88,17 @@ pub struct ExperimentResponse {
 #[derive(Serialize, Deserialize, Debug, Clone, Deref, DerefMut, PartialEq)]
 pub struct StatusTypes(pub Vec<ExperimentStatusType>);
 
+impl Default for StatusTypes {
+    fn default() -> Self {
+        Self(vec![
+            ExperimentStatusType::CREATED,
```
ShreyBana, line 94: For listing out all the variants there is a macro, & if add a new type here, we will probably miss it if we do it this way. Can you use that macro? Here's the link: https://docs.rs/strum/latest/strum/derive.EnumIter.html

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,22 +49,338 @@ struct CombinedResource {
...
+                                };
+                                filters_buffer_rws.set(new_filters);
+                            })
+                        />
+                    </div>
+                </div>
+
+                <div class="form-control w-full">
+                    <label class="label">
+                        <span class="label-text">Experiment Status</span>
```
ShreyBana, line 173: Could have looped over the variants to get this out, in-fact, if some body adds a new variant, then we might forget it. Can you loop over the variants instead?

ShreyBana (changes requested):

crates/superposition_types/Cargo.toml
```diff
@@ -22,6 +22,7 @@ log = { workspace = true }
 regex = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+strum = { workspace = true }
```
ayushjain17, line 25: also, need to add it in deps for `experimentation` feature here
```suggestion
strum = { workspace = true, optional = true }
```

crates/frontend/src/pages/experiment_list.rs
ayushjain17, line 1: need to re-review

ayushjain17 (changes requested):

crates/frontend/src/components/input.rs
```diff
@@ -473,6 +474,56 @@ pub fn monaco_input(
...
+                class=format!("input input-bordered {}", class)
+                required=required
+                disabled=disabled
+                min=min
+                max=max
+                value=value
+                on:change=move |e| {
+                    let date = format!("{}T00:00:00Z", event_target_value(&e));
+                    logging::log!("The date selected is: {}", date);
+                    match DateTime::parse_from_rfc3339(&date) {
```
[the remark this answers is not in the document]
ShreyBana: I don't think there's any point to it in this case, it's like the `unchecked_into` call, it can panic but it won't, user would have to opt into breaking the UI themselves. Not blocking on this, but IMO this one of those cases where error handling is unnecessary/needless.

crates/frontend/src/components/input.rs
```diff
@@ -473,6 +474,56 @@ pub fn monaco_input(
...
+#[component]
+pub fn date_input(
+    id: String,
+    class: String,
+    name: String,
+    on_change: Callback<DateTime<Utc>, ()>,
+    #[prop(into, default = Utc::now().format("%Y-%m-%d").to_string())] value: String,
+    #[prop(default = false)] disabled: bool,
+    #[prop(default = false)] required: bool,
+    #[prop(into, default = String::from("2020-01-01"))] min: String,
```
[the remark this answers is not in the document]
ShreyBana: I think @ayushjain17 has a point, why not have a sane default. IMO it should be `Utc::now() - (24*60*60)`, basically one day prior, that's a pretty sane default in most cases.

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,22 +49,338 @@ struct CombinedResource {
...
-        from_date: Utc.timestamp_opt(0, 0).single(),
-        to_date: Utc.timestamp_opt(4130561031, 0).single(),
-        experiment_name: None,
-        experiment_ids: None,
-        created_by: None,
-        context: None,
-        sort_on: Some(utils::ExperimentSortOn::LastModifiedAt),
-        sort_by: Some(SortBy::Desc),
-    });
+    let filters_rws = create_rw_signal(ExperimentListFilters::default());
```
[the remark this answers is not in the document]
Datron: Actually the original was too verbose, sort_on and sort_by defaults in API are the same as the ones here. The timestamp code was improved

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -105,6 +436,7 @@ pub fn experiment_list() -> impl IntoView {
     );
 
     let handle_submit_experiment_form = move || {
+        filters_rws.set(ExperimentListFilters::default());
```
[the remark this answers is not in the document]
Datron: On creation of a new experiment, it may not show up in the listing since a filter was previously applied. This would confuse the user who was expecting to see it

crates/frontend/src/components/input.rs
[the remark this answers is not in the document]
Datron: Lets compromise, I set a month as the default. A day wouldn't suffice

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -196,13 +527,20 @@ pub fn experiment_list() -> impl IntoView {
                                     };
                                     view! {
                                         <ConditionCollapseProvider>
+                                            <div class="flex justify-start m-1">
```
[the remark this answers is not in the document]
Datron: Done

crates/frontend/src/components/input.rs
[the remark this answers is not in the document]
ShreyBana: image
Fine for now, we can change later if somethings up.

ShreyBana (approved):

ShubhranshuSanjeev (approved):

crates/frontend/src/components/input.rs
```diff
@@ -475,6 +476,58 @@ pub fn monaco_input(
...
+    name: String,
+    on_change: Callback<DateTime<Utc>, ()>,
+    #[prop(into, default = Utc::now().format("%Y-%m-%d").to_string())] value: String,
+    #[prop(default = false)] disabled: bool,
+    #[prop(default = false)] required: bool,
+    #[prop(into, default = (Utc::now() - chrono::Duration::weeks(4)).format("%Y-%m-%d").to_string())]
+    min: String,
+    #[prop(into, default = Utc::now().format("%Y-%m-%d").to_string())] max: String,
+) -> impl IntoView {
+    let (error_rs, error_ws) = create_signal::<String>(String::new());
```
ayushjain17, line 492: instead of using a empty string, should we rather not use `Option<String>`?
this type models the requirement best

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,36 +49,334 @@ struct CombinedResource {
...
+                        text="Submit".to_string()
+                        on_click=move |event| {
+                            event.prevent_default();
+                            let filter = filters_buffer_rws.get();
+                            filters_rws.set(filter);
+                            close_drawer("experiment_filter_drawer")
+                        }
+                    />
+                    <Button
+                        class="pl-[70px] pr-[70px] w-48 h-12".to_string()
```
ayushjain17, line 346: can we directly use `px-[70px]` instead
```suggestion
                        class="h-12 w-48 px-[70px]".to_string()
```

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,36 +49,334 @@ struct CombinedResource {
...
+                                };
+                                filters_buffer_rws.set(filters);
+                            }
+                        />
+
+                    </div>
+                </div>
+                <div class="flex justify-start mt-8">
+                    <Button
+                        class="pl-[70px] pr-[70px] w-48 h-12".to_string()
```
ayushjain17, line 336: can we directly use `px-[70px]` instead
```suggestion
                        class="h-12 w-48 px-[70px]".to_string()
```

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,36 +49,334 @@ struct CombinedResource {
...
+                                let filters = filters_buffer_rws.get();
+                                let filters = ExperimentListFilters {
+                                    experiment_name,
+                                    ..filters
+                                };
+                                filters_buffer_rws.set(filters);
+                            }
+                        />
+                    </div>
+                    <div class="form-control mt-4">
```
ayushjain17, line 286: instead of separately adding `mt-4`, can we directly increase the gap value in the parent from `gap-1` to `gap-5` ?
and if we still need to add some spacing in the top, can we please use padding, instead ?

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,36 +49,334 @@ struct CombinedResource {
...
+                                let checked = event_target_checked(&event);
+                                status_filter_management(checked, ExperimentStatusType::DISCARDED)
+                            }
+                        />
+                        <label
+                            for="discard-checkbox"
+                            class="px-6 h-[30px] py-2 badge peer-checked/discard-checkbox:badge-neutral cursor-pointer transition duration-300 ease-in-out"
+                        >
+                            Discarded
+                        </label>
```
ayushjain17, line 256: instead of copy pasting this entire section, can we rather loop through the options to be generated and populate the UI, rather ?

ayushjain17, line 256: also, its better to wrap the input and label of a particular item into some div, allows easier and better styling of elements
for example, by doing so, you can avoid the need of using a named peer tailwind class and can achieve the above comment in a much easier way

crates/frontend/src/pages/experiment_list.rs
ayushjain17: > The colors change for each variant which isn't captured by the enum, how would I apply that here?

@Datron you can iterate on an array of tuples for this

```
[
   (Variant1, Variant1Color, Variant1OtherProp),
   (Variant2, Variant2Color, Variant2OtherProp),
   (Variant3, Variant3Color, Variant3OtherProp),
   (Variant4, Variant4Color, Variant4OtherProp)
]
```

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -34,36 +49,334 @@ struct CombinedResource {
...
+            <i class="ri-filter-3-line"></i>
+        </DrawerBtn>
+        <Drawer
+            id="experiment_filter_drawer".to_string()
+            header="Experiment Filters"
+            drawer_width="w-[50vw]"
+            handle_close=move || close_drawer("experiment_filter_drawer")
+        >
+            <div class="card-body">
+                <div class="my-4">
```
ayushjain17, line 96: this div can be avoided, we dont need single child divs and please lets not use margins

ayushjain17 (changes requested):

ayushjain17:
@ShubhranshuSanjeev @ShreyBana 
I and @Datron could not come into a conclusion regarding my comments, can you guys please chip into it and help conclude it

crates/frontend/src/components/input.rs
ShubhranshuSanjeev: I get the data model point, but having just `String` will also work, we can avoid verbosity if not required.

crates/frontend/src/pages/experiment_list.rs
ShubhranshuSanjeev: unnecessary nested divs prevent inheriting styles like width, which is annoying to deal with if we want something to span for the same width as a parent. I have faced this issue multiple times, it would be better if this change can be made.

crates/frontend/src/pages/experiment_list.rs
ShubhranshuSanjeev: I agree this is important.

ShubhranshuSanjeev: Good point
Something similar, what has been done at line 260.

crates/frontend/src/pages/experiment_list.rs
ShubhranshuSanjeev: I think this should be changed, as we are already defining gap in the parent which is a flexbox, since this is already provides the spacing why not just use that property, it will be consistent.

crates/frontend/src/pages/experiment_list.rs
ShubhranshuSanjeev: +1

crates/frontend/src/pages/experiment_list.rs
ShubhranshuSanjeev: +1

ShubhranshuSanjeev (changes requested):

ayushjain17:
@juspay/sdk-backend 
Just noticed this and wanted to point it out for all, in the contextual search over here we are doing an exact match search not a partial search, as in we are matching over the exact operator here, if the operator matches along with the data only then the search would work

ShubhranshuSanjeev (approved):

mahatoankitkumar (approved):

PR #426, opened 2025-02-20
fix: removed err as a default value for schema type and enum variants

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/frontend/src/components/default_config_form.rs
+++ b/crates/frontend/src/components/default_config_form.rs
@@ -17,7 +17,7 @@ use crate::{
         dropdown::{Dropdown, DropdownBtnType, DropdownDirection},
         input::{Input, InputType},
     },
-    schema::{EnumVariants, JsonSchemaType, SchemaType},
+    schema::{EnumVariants, HtmlDisplay, JsonSchemaType, SchemaType},
     types::FunctionsName,
 };
 use crate::{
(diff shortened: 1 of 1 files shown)


crates/frontend/src/components/default_config_form.rs
```diff
@@ -57,8 +57,15 @@ where
...
-    let (enum_variants_rs, enum_variants_ws) = create_signal(Err("".to_string()));
+
+    let schema_type_s = Signal::derive(move || {
+        SchemaType::try_from(config_schema_rs.get())
+            .map_err(|err| format!("Failed to get schema type: {}", err))
+    });
+    let enum_variants_s = Signal::derive(move || {
+        EnumVariants::try_from(config_schema_rs.get())
+            .map_err(|err| format!("Failed to get enum variants: {}", err))
+    });
```
Datron, line 68: So these signals are read-only?

Datron (approved):

crates/frontend/src/components/default_config_form.rs
```diff
@@ -285,70 +290,82 @@ where
...
+                                    <textarea
+                                        type="text"
+                                        placeholder="Value"
+                                        class="input input-bordered w-full pt-3"
+                                        disabled=true
+                                    >
+                                        {config_value_rs.get().html_display()}
+                                    </textarea>
+                                </div>
+                            </div>
```
ShubhranshuSanjeev, line 333: Rendering a disabled field until SchemaType is figured out for the default config.

crates/frontend/src/components/default_config_form.rs
ShubhranshuSanjeev: Yes

mahatoankitkumar (approved):

PR #433, opened 2025-03-06
fix: function fetch with type

pratikmishra356:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/context/helpers.rs
+++ b/crates/context_aware_config/src/api/context/helpers.rs
@@ -10,7 +10,7 @@ use service_utils::{helpers::extract_dimensions, service::types::SchemaName};
 use superposition_macros::{unexpected_error, validation_error};
 use superposition_types::{
     database::{
-        models::cac::Context,
+        models::cac::{Context, FunctionCode},
         schema::{contexts, default_configs::dsl, dimensions},
     },
     result as superposition, Cac, Condition, DBConnection, Overrides, User,
@@ -124,7 +124,7 @@ fn get_functions_map(
     keys_function_array: Vec<(String, String)>,
     schema_name: &SchemaName,
 ) -> superposition::Result<HashMap<String, FunctionsInfo>> {
-    let functions_map: HashMap<String, Option<String>> =
+    let functions_map: HashMap<String, Option<FunctionCode>> =
         get_published_functions_by_names(
             conn,
             keys_function_array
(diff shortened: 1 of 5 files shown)


crates/context_aware_config/src/api/functions/helpers.rs
```diff
@@ -24,26 +24,27 @@ pub fn get_published_function_code(
     f_name: String,
     schema_name: &SchemaName,
 ) -> superposition::Result<Option<String>> {
```
ayushjain17, line 26: ?
```suggestion
) -> superposition::Result<Option<FunctionCode>> {
```

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -144,7 +144,11 @@ fn get_functions_map(
                     key.clone(),
                     FunctionsInfo {
                         name: function_name.clone(),
-                        code: functions_map.get(&function_name).cloned().flatten(),
+                        code: functions_map
+                            .get(&function_name)
+                            .cloned()
+                            .flatten()
+                            .map(|val| val.0),
```
ayushjain17, line 151: instead of changing here shouldn't we update the type of `code` in `FunctionsInfo` with `Option<FunctionCode>`

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -154,11 +154,11 @@ fn get_functions_map(
 
 pub fn validate_value_with_function(
     _fun_name: &str,
-    function: &str,
+    function: FunctionCode,
     key: &String,
     value: &Value,
 ) -> superposition::Result<()> {
-    if let Err((err, stdout)) = execute_fn(function, key, value.to_owned()) {
+    if let Err((err, stdout)) = execute_fn(&function.0, key, value.to_owned()) {
```
ayushjain17, line 161: my point was to pass `FunctionCode` till the core levels, which here seems to be `execute_fn`

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -154,11 +154,11 @@ fn get_functions_map(
 
 pub fn validate_value_with_function(
     _fun_name: &str,
-    function: &str,
+    function: FunctionCode,
```
ayushjain17, line 157: also, shouldn't the ref be passed along  ? 
```suggestion
    function: &FunctionCode,
```

ayushjain17 (approved):

PR #435, opened 2025-03-06
fix: Params being sent in resolve page

ayushjain17:
## Problem
params for resolve being sent with the operator like: `key==value`, `keyinvalue`, instead of a simple `key=value`.
For resolve a simple assign is expected and no other operator

example:
`http://localhost:8080/config/resolve?clientId==studio1&platform==web&show_reasoning=true`

## Solution
Generate `key=value` while generating the params only, operators other than `==` are not expected in this flow

### DIFF
--- a/crates/frontend/src/logic.rs
+++ b/crates/frontend/src/logic.rs
@@ -116,34 +116,10 @@ impl Expression {
             }
         }
     }
-    fn to_expression_query_str(&self, var: &str) -> String {
+    fn to_expression_query_str(&self, var: &str) -> Option<String> {
         match self {
-            Expression::Is(c) => {
-                format!("{}{}{}", var, "==", c.html_display())
-            }
-            Expression::In(c) => {
-                format!("{}{}{}", var, "in", c.html_display())
-            }
-            Expression::Has(c) => {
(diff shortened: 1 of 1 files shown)


ShubhranshuSanjeev (approved):

crates/frontend/src/logic.rs
```diff
@@ -304,7 +280,7 @@ impl Condition {
         self.expression.to_expression_json(&self.variable)
     }
 
-    pub fn to_condition_query_str(&self) -> String {
+    pub fn to_condition_query_str(&self) -> Option<String> {
```
ShreyBana, line 283: Is this wrapper really necessary now?

ShreyBana (approved):

crates/frontend/src/logic.rs
ayushjain17: Do you mean the condition type ?

ShreyBana: I meant this function, it just has one line, & that too a call another function.

ShreyBana: Is this to implement a trait?

PR #436, opened 2025-03-07
fix: z-index of delete modal

ayushjain17:
## Problem
sticky content of the tables are overlapping with the delete modal, as no z-index has been provided to the backdrop of the delete modal

## Solution
Add a high z-index value to the modal backdrop

### DIFF
--- a/crates/frontend/src/components/delete_modal.rs
+++ b/crates/frontend/src/components/delete_modal.rs
@@ -11,7 +11,7 @@ pub fn delete_modal(
 
     view! {
         <Show when=move || modal_visible.get()>
-            <div class="fixed inset-0 bg-black bg-opacity-50 backdrop-blur-sm flex items-center justify-center">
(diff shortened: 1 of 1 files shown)


crates/frontend/src/components/delete_modal.rs
```diff
@@ -11,7 +11,7 @@ pub fn delete_modal(
 
     view! {
         <Show when=move || modal_visible.get()>
-            <div class="fixed inset-0 bg-black bg-opacity-50 backdrop-blur-sm flex items-center justify-center">
+            <div class="fixed inset-0 bg-black bg-opacity-50 backdrop-blur-sm flex items-center justify-center z-[99999999]">
```
Datron, line 14: nitpick:
Can we keep it at 999? The columns have a z-index of 20. This seems overkill. 
Also TL;DR https://stackoverflow.com/questions/491052/minimum-and-maximum-value-of-z-index

Datron (approved):

PR #443, opened 2025-03-17
feat: add autocomplete functions

Datron:
## Problem
Some dimensions or configs are hard to figure out values for since they maybe complex, depend on external systems that superposition does not have context of and should not.

## Solution
Introduce autocomplete functions as discussed here:
https://github.com/juspay/superposition/discussions/431

This PR introduces the syntax, and refactors functions to handle more than validation functions.

demo video: https://drive.google.com/file/d/10tVTNZOV4tn29kcAiFjnou6oRiT7qcbM/view?usp=sharing


crates/context_aware_config/src/validation_functions.rs
```diff
@@ -1,128 +1,163 @@
...
         const script = new vm.Script(\`
 
-        {}
-        Promise.resolve(validate({}, {})).then((output) => {{
-
-            if(output!=true){{
+        {replaceme-with-code}
+        Promise.resolve({function-invocation}).then((output) => {
+            if({condition}) {
                 throw new Error("The function did not return true as expected. Check the conditions or logic inside the function.")
```
mahatoankitkumar, line 41: @Datron I guess this error message should change according to the type of functions.

mahatoankitkumar (changes requested):

crates/superposition_types/migrations/2025-03-12-061636_function_types/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
pratikmishra356, line 1: can we write query for undo up.sql

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -1,128 +1,168 @@
...
+            FunctionTypes::Autocomplete
+                .get_fn_signature()
+                .replace("{name}", format!("\"{}\"", &name).as_str())
+                .replace("{prefix}", format!("\"{}\"", &prefix).as_str())
+                .replace("{environment}", &environment.to_string()),
+            "!(output instanceof Array)",
+        ),
+    };
+    FUNCTION_EXECUTION_SNIPPET
+        .replace(CODE_TOKEN, code_str)
```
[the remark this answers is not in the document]
Datron: @pratikmishra356 Would the order of replace matter? Since the code is executed in order of CODE_TOKEN, function-invocation, condition?

crates/superposition_types/migrations/2025-03-12-061636_function_types/up.sql
```diff
@@ -0,0 +1,7 @@
+-- Your SQL goes here
+CREATE TYPE public.function_types AS ENUM (
```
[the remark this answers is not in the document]
Datron: Sadly CREATE TYPE does not support if not exists
https://www.postgresql.org/docs/current/sql-createtype.html

crates/superposition_types/migrations/2025-03-12-061636_function_types/down.sql
Datron: Sure

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -1,128 +1,168 @@
...
+    code_str: &FunctionCode,
+    function_args: &FunctionExecutionRequest,
+) -> String {
+    let (function_invocation, output_check) = match function_args {
+        FunctionExecutionRequest::ValidateFunctionRequest { key, value } => (
+            FunctionTypes::Validation
+                .get_fn_signature()
+                .replace("{key}", format!("\"{}\"", &key).as_str())
+                .replace("{value}", &value.to_string()),
+            "output!=true",
```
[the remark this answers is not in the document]
Datron: @pratikmishra356 that seems unnecessary, I wrote constants for tokens that don't change and people will re-use in multiple places. Its better that this is defined here so its easily understood

crates/frontend/src/components/function_form.rs
```diff
@@ -4,26 +4,68 @@ pub mod utils;
...
+    fn key(&self) -> String {
+        self.to_string()
+    }
+
+    fn label(&self) -> String {
+        self.to_string()
+    }
+}
+
+#[wasm_bindgen(module = "/src-js/utils.js")]
```
pratikmishra356, line 54: @Datron can you check this , its panicking when i visit  function detail page

crates/superposition_types/src/api/functions.rs
```diff
@@ -71,8 +82,41 @@ pub struct TestParam {
...
+        Self::ValidateFunctionRequest {
+            key: String::new(),
+            value: Value::String(String::new()),
+        }
+    }
+    pub fn autocomplete_default() -> Self {
+        Self::AutocompleteFunctionRequest {
+            name: String::new(),
+            prefix: String::new(),
+            environment: json!({
```
pratikmishra356, line 109: `environment` name seems a bit odd when we are expecting context and overrides

Datron: It captures the environment of the config or dimension that requires completion.

crates/frontend/src/components/function_form.rs
Datron: Hard refresh your browser, or make sure the code is present in utils.js. If it is not, take a pull

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -1,128 +1,168 @@
...
+        });
+        worker.on("exit", (code) => {
             clearTimeout(tl);
-            if (code !== 0) {{
-                console.error(`Script stopped with exit code ${{code}}`);
-                process.exit(code);
-            }} else {{
+            if (code != 0) {
+                console.error(`Script stopped with exit code ${code}`);
                 worker.terminate();
```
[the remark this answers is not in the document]
Datron: @pratikmishra356 I checked this, in earlier implementations we were not throwing an error here:
```
if (code !== 0) {{
                console.error(`Script stopped with exit code ${{code}}`);
                process.exit(code);
            }} else {{
                worker.terminate();
            }}
```

I believe on termination the right error message is provided by the rust side.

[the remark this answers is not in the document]
pratikmishra356: @Datron earlier if code is 1 , we are exiting with that code that resulted in error 
and if it is 0 then we are terminate the worker

in current implementation , if code is not 0, then we are just terminating the worker threads which is not resulting in error , either we should exit with 1 or throw err

pratikmishra356 (approved):

mahatoankitkumar (approved):
Looks good

PR #444, opened 2025-03-18
fix: config version alter query

Datron:
## Problem
When running diesel migration command, I was getting errors as there was a discrepancy between database schemas and models defined in rust

## Solution
Updated the query causing the issue and added a make target that can help setup the schema


crates/superposition_types/migrations/2025-02-11-100518_remove_description_config_version/up.sql
```diff
@@ -1,3 +1,3 @@
 -- Your SQL goes here
 ALTER TABLE public.config_versions
-DROP COLUMN description;
\ No newline at end of file
+DROP COLUMN change_reason;
```
pratikmishra356, line 3: @Datron change is correct , 
can we rename migration file from remove_description -> remove_change_reason

pratikmishra356 (approved):

crates/superposition_types/src/database/diesel.toml
```diff
@@ -3,8 +3,8 @@ file = "schema.rs"
 patch_file = "schema.patch"
 
 [print_schema.superposition]
-file = "src/superposition_schema.rs"
-patch_file = "schema.patch"
+file = "superposition_schema.rs"
+patch_file = "superposition_schema.patch"
```
ayushjain17, line 7: why do we have schema.patch file ?
if its not being used ?

Datron: I'm just keeping it consistent incase a diesel migration is run

crates/superposition_types/migrations/2025-02-11-100518_remove_changereason_config_version/up.sql
```diff
@@ -1,3 +1,3 @@
 -- Your SQL goes here
 ALTER TABLE public.config_versions
```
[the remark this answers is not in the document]
Datron: Done

makefile
```diff
@@ -197,3 +197,10 @@ amend-no-edit: COMMIT_FLAGS += --no-edit
 amend-no-edit: amend
 
 default: dev-build
+
+schema-file:
```
ayushjain17, line 201: should we not add this to .PHONY ? is it returning some usable file ?

Datron: No it just does migration

crates/superposition_types/migrations/2025-02-11-100518_remove_changereason_config_version/up.sql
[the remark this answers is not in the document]
Datron: Done

makefile
ayushjain17: then it should be in .PHONY list also na ?

crates/superposition_types/src/database/diesel.toml
ayushjain17: oh, so now via diesel migration, we are intending to only create database stuff from `superposition` schema only and not from `public` schema which are also there in the migration file ?

Datron: The diesel migration creates all the database stuff, the reason this is there is to ensure superposition schema goes to superposition_schema.rs and public schema goes to schema.rs. Without this, diesel writes public to superposition_schema.rs as well

ayushjain17 (approved):

PR #445, opened 2025-03-24
feat: generate multiple binaries for clients based on platform

Datron:
## Problem
We don't generate distributive binaries for our clients

## Solution
Add a workflow to generate builds for linux, macOS and update interface.rs to support windows builds in the future.

### DIFF
--- a/.github/workflows/generate_client_bundles.yaml
+++ b/.github/workflows/generate_client_bundles.yaml
@@ -1,0 +1,83 @@
+name: Build and distribute clients for multiple platforms
+
+on: workflow_dispatch
+
+jobs:
+  generate-builds:
+    strategy:
+      max-parallel: 5
+      matrix:
+        platform:
+            - os: ubuntu-latest
+              target: x86_64-unknown-linux-gnu
+              cac_bin_name: libcac_client.so
+              exp_bin_name: libexperimentation_client.so
+              cac_zip_name: superposition_cac_client-x86_64-unknown-linux-gnu.zip
+              exp_zip_name: superposition_experimentation_client-x86_64-unknown-linux-gnu.zip
+            - os: macos-latest
+              target: x86_64-apple-darwin
+              cac_bin_name: libcac_client.dylib
+              exp_bin_name: libexperimentation_client.dylib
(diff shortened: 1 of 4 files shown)


.github/workflows/generate_client_bundles.yaml
```diff
@@ -0,0 +1,89 @@
...
+              exp_bin_name: libexperimentation_client.dylib
+              cac_zip_name: superposition_cac_client-x86_64-apple-darwin.zip
+              exp_zip_name: superposition_experimentation_client-x86_64-apple-darwin.zip
+            - os: macos-latest
+              target: aarch64-apple-darwin
+              cac_bin_name: libcac_client.dylib
+              exp_bin_name: libexperimentation_client.dylib
+              cac_zip_name: superposition_cac_client-aarch64-apple-darwin.zip
+              exp_zip_name: superposition_experimentation_client-aarch64-apple-darwin.zip
+            # - os: windows-latest
```
sauraww, line 35: why is this commented out ?

crates/cac_client/src/interface.rs
```diff
@@ -130,19 +131,21 @@ pub extern "C" fn cac_new_client_with_cache_properties(
     cache_ttl: c_ulong,
     cache_tti: c_ulong,
 ) -> c_int {
-    let duration = Duration::new(update_frequency, 0);
+    #[allow(clippy::useless_conversion)] // done for windows support
```
sauraww, line 134: #[allow(clippy::useless_conversion)]
this is only used to suppress the warnings, right ?
It would have still done the type conversions regardless.

sauraww (approved):

.github/workflows/generate_client_bundles.yaml
Datron: You need special C compiler to run for aarch64, I was looking into how to get it installed so this is commented out

crates/cac_client/src/interface.rs
Datron: @sauraww no for windows compilation it fails

ayushjain17: @Datron didn't get the issue here

PR #448, opened 2025-03-27
feat: dependent dimensions ui

mahatoankitkumar:
## Problem
UI Changes for dependent dimensions

## Solution
Add the dependent dimensions in the context_form while adding the dimensions in the context.

https://github.com/user-attachments/assets/024096ac-50d5-4290-b0be-94c018926ae5

### DIFF
--- a/crates/frontend/src/components/context_form.rs
+++ b/crates/frontend/src/components/context_form.rs
@@ -28,7 +28,7 @@ pub fn condition_input(
 ) -> impl IntoView {
     let (dimension, operator): (String, Operator) =
         condition.with_value(|v| (v.variable.clone(), v.into()));
-
+    let tooltip_text = StoredValue::new(tooltip_text);
     let inputs: Vec<(Value, Callback<Value, ()>)> = match condition.get_value().expression
     {
         Expression::Is(c) => {
@@ -171,8 +171,8 @@ pub fn condition_input(
                     >
                         <i class="ri-delete-bin-2-line text-2xl font-bold"></i>
                     </button>
-                </Show> <Show when=move || !allow_remove>
-                    <div class="tooltip" data-tip=tooltip_text.clone()>
+                </Show> <Show when=move || !tooltip_text.get_value().is_empty()>
+                    <div class="tooltip" data-tip=tooltip_text.get_value()>
                         <i class="ri-information-line text-2xl"></i>
                     </div>
                 </Show>
@@ -211,45 +211,43 @@ where
             .collect::<HashSet<String>>(),
     );
 
-    let update_context_with_dimension =
-        move |context: &mut Conditions, dimension_info: &DimensionWithMandatory| {
-            if !context.includes(&dimension_info.dimension) {
+    let insert_dimension =
+        move |context: &mut Conditions, dimension: &DimensionWithMandatory| {
+            if !context.includes(&dimension.dimension) {
                 context.push(Condition::new_with_default_expression(
-                    dimension_info.dimension.clone(),
-                    SchemaType::try_from(dimension_info.schema.clone()).unwrap(),
+                    dimension.dimension.clone(),
+                    SchemaType::try_from(dimension.schema.clone()).unwrap(),
                 ));
             }
-            let dependencies: Vec<String> = dimension_info
+            dimension
                 .dependency_graph
                 .keys()
-                .filter(|key| **key != dimension_info.dimension)
-                .cloned()
-                .collect();
-            dependencies.into_iter().for_each(|dependency| {
-                if let Some(r#type) = dimension_map
-                    .get_value()
-                    .get(&dependency)
-                    .and_then(|d| SchemaType::try_from(d.schema.clone()).ok())
-                {
-                    if !context.includes(&dependency) {
-                        context.push(Condition::new_with_default_expression(
-                            dependency.clone(),
-                            r#type.clone(),
-                        ));
+                .filter(|key| **key != dimension.dimension)
+                .for_each(|dependency| {
+                    if let Some(r#type) = dimension_map
+                        .get_value()
+                        .get(dependency)
+                        .and_then(|d| SchemaType::try_from(d.schema.clone()).ok())
+                    {
+                        if !context.includes(dependency) {
+                            context.push(Condition::new_with_default_expression(
+                                dependency.clone(),
+                                r#type.clone(),
+                            ));
+                        }
                     }
-                }
-            });
+                });
         };
 
     let context_data = {
         let mut context = context;
-        if !disabled {
+        if !disabled && !resolve_mode {
             dimensions
                 .get_value()
                 .into_iter()
                 .filter(|dim| dim.mandatory)
-                .for_each(|dimension_info| {
-                    update_context_with_dimension(&mut context, &dimension_info);
+                .for_each(|dimension| {
+                    insert_dimension(&mut context, &dimension);
                 });
         }
         context
@@ -289,7 +287,14 @@ where
     let on_select_dimension =
         Callback::new(move |selected_dimension: DimensionWithMandatory| {
             let mut context = context_rs.get();
-            update_context_with_dimension(&mut context, &selected_dimension);
+            if !resolve_mode {
+                insert_dimension(&mut context, &selected_dimension);
+            } else {
+                context.push(Condition::new_with_default_expression(
+                    selected_dimension.dimension.clone(),
+                    SchemaType::try_from(selected_dimension.schema.clone()).unwrap(),
+                ));
+            }
             context_ws.update(|v| {
                 *v = context;
             });
@@ -379,7 +384,7 @@ where
                                 condition.variable,
                                 idx,
                                 condition.expression.to_operator(),
-                                context_dependencies.get().contains(&condition.variable),
+                                context_dependencies.with(|v| v.contains(&condition.variable)),
                             )
                         }
 
@@ -398,9 +403,15 @@ where
                                     .into_view();
                             }
                             let schema_type = store_value(schema_type.unwrap());
-                            let allow_remove = !disabled
-                                && !mandatory_dimensions.get_value().contains(&condition.variable)
-                                && !context_dependencies.get().contains(&condition.variable);
+                            let is_mandatory = mandatory_dimensions
+                                .with_value(|v| v.contains(&condition.variable));
+                            let has_context_dependency = context_dependencies
+                                .with(|v| v.contains(&condition.variable));
+                            let allow_remove = if resolve_mode {
+                                !disabled
+                            } else {
+                                !disabled && !is_mandatory && !has_context_dependency
+                            };
                             let input_type = store_value(
                                 InputType::from((
                                     schema_type.get_value(),
--- a/crates/frontend/src/components/context_form.rs
+++ b/crates/frontend/src/components/context_form.rs
@@ -24,6 +24,7 @@ pub fn condition_input(
     #[prop(into)] on_remove: Callback<String, ()>,
     #[prop(into)] on_value_change: Callback<Expression, ()>,
     #[prop(into)] on_operator_change: Callback<Operator, ()>,
+    #[prop(default = String::new())] tooltip_text: String,
 ) -> impl IntoView {
     let (dimension, operator): (String, Operator) =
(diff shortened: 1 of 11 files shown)


crates/frontend/src/pages/context_override.rs
```diff
@@ -235,23 +235,8 @@ pub fn context_override() -> impl IntoView {
...
-
-            default_ctx.push(Condition::new_with_default_expression(
-                dim.dimension.clone(),
-                r#type.unwrap(),
-            ));
-        }
-
         selected_context_ws.set(Some(Data {
-            context: default_ctx,
+            context: Conditions(vec![]),
```
ayushjain17, line 239: is this change correct ?
seems like a dummy change

or am I missing something here ?

mahatoankitkumar: No, its a relevant change

crates/frontend/src/components/context_form.rs
```diff
@@ -164,15 +164,18 @@ pub fn condition_input(
...
                         }
                     >
-
                         <i class="ri-delete-bin-2-line text-2xl font-bold"></i>
                     </button>
+                </Show> <Show when=move || !allow_remove>
+                    <div class="tooltip" data-tip=tooltip_text.clone()>
+                        <i class="ri-information-line text-2xl"></i>
+                    </div>
                 </Show>
```
Datron, line 179: nitpick: Needs to be formatted

crates/frontend/src/components/context_form.rs
```diff
@@ -24,6 +25,7 @@ pub fn condition_input(
     #[prop(into)] on_remove: Callback<String, ()>,
     #[prop(into)] on_value_change: Callback<Expression, ()>,
     #[prop(into)] on_operator_change: Callback<Operator, ()>,
+    tooltip_text: String,
```
Datron, line 28: Should there be a default prop on tooltip_text? Makes it easier to ignore if you don't need it

```suggestion
    #[prop(default = String::new())]  tooltip_text: String,
```

crates/frontend/src/components/context_form.rs
```diff
@@ -216,19 +218,86 @@ where
             .collect::<HashSet<String>>(),
     );
 
+    let (dependent_dimension_locks_rs, dependent_dimension_locks_ws) =
+        create_signal(used_dimensions_rs.get());
```
Datron, line 222: I don't think reactivity works here. Check this for the proper way to do this:
https://book.leptos.dev/reactivity/working_with_signals.html#making-signals-depend-on-each-other

crates/frontend/src/components/context_form.rs
```diff
@@ -216,19 +218,86 @@ where
...
+                .dependency_graph
+                .keys()
+                .filter(|key| **key != dimension_info.dimension.clone())
+                .cloned()
+                .collect();
+            dependent_dimensions.iter().for_each(|dependent| {
+                if !used_dimensions_rs.get().contains(dependent) {
+                    used_dimensions_ws.update(|value: &mut HashSet<String>| {
+                        value.insert(dependent.clone());
+                    });
```
Datron, line 237: You could just blindly insert the dimension into used_dimension_ws and let the hashset ignore it if the dimension already exists

crates/frontend/src/components/context_form.rs
```diff
@@ -266,8 +344,48 @@ where
...
+        {
+            removed_dimension.dependencies.iter().for_each(|dependent| {
+                dependent_dimension_locks_ws.update(|value| {
+                    value.remove(dependent);
+                });
+            })
+        }
     });
 
+    let get_tool_tip_text = move |condition: StoredValue<Condition>| {
```
Datron, line 360: Can we define the return type with this closure?
```suggestion
    let get_tool_tip_text = move |condition: StoredValue<Condition>| -> String {
```

Datron (changes requested):
Just take a look at dependent signals, I don't know if that is working properly. Also anyway to reduce the number of clones in the context_form when working with dependent dimensions?

crates/frontend/src/components/context_form.rs
mahatoankitkumar: Reactivity not needed here, its just for initialization

crates/frontend/src/components/context_form.rs
```diff
@@ -266,8 +344,49 @@ where
...
+                        .any(|parent| used_dimensions_rs.get().contains(parent))
+                    {
+                        return;
+                    }
+                    dependent_dimension_locks_ws.update(|locks| {
+                        locks.remove(dependent);
+                    });
+                }
+            })
+        }
```
ayushjain17, line 365: instead of handling this manually, shouldn't we rather make this a derived signal, avoids the requirement of maintaining the logic in 3 places (initial signal creation, adding new dimension, removing dimension)

crates/frontend/src/components/context_form.rs
```diff
@@ -216,19 +218,86 @@ where
...
+                        &|d_name, r#type| {
+                            context.borrow_mut().push(
+                                Condition::new_with_default_expression(d_name, r#type),
+                            );
+                        },
+                    );
+                });
+        }
+        context.into_inner()
+    };
```
ayushjain17, line 284: does this logic take into consideration the fact that there will be/can be data which exist today without the mandatory/dependency dimensions ?

ayushjain17 (changes requested):
ideally, the change should have been just a derived signal to track the currently required list of dependencies and the logic get the string for tooltip

crates/frontend/src/pages/context_override.rs
```diff
@@ -235,23 +235,8 @@ pub fn context_override() -> impl IntoView {
...
-                return;
-            }
-
-            default_ctx.push(Condition::new_with_default_expression(
-                dim.dimension.clone(),
-                r#type.unwrap(),
-            ));
-        }
-
         selected_context_ws.set(Some(Data {
```
ayushjain17, line 238:
```suggestion
        selected_context_ws.set(Some(Data::default()))
```

crates/frontend/src/components/context_form.rs
ShubhranshuSanjeev, line 365: Can you explain the logic?

crates/frontend/src/components/context_form.rs
```diff
@@ -266,8 +344,49 @@ where
...
+                    dependent_dimension_locks_ws.update(|locks| {
+                        locks.remove(dependent);
+                    });
+                }
+            })
+        }
     });
 
+    let get_tool_tip_text = move |condition: StoredValue<Condition>| -> String {
+        let variable = condition.get_value().variable;
```
ShubhranshuSanjeev, line 369: Can you use `with_value` for this operation?

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/context_form.rs
```diff
@@ -216,19 +218,86 @@ where
             .collect::<HashSet<String>>(),
     );
 
+    let (dependent_dimension_locks_rs, dependent_dimension_locks_ws) =
```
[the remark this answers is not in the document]
mahatoankitkumar: Using it to block the removal of dependencies of the added dimension.

crates/frontend/src/components/context_form.rs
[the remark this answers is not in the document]
mahatoankitkumar: Just for initialization

crates/frontend/src/components/context_form.rs
```diff
@@ -216,19 +218,86 @@ where
...
+                .dependency_graph
+                .keys()
+                .filter(|key| **key != dimension_info.dimension.clone())
+                .cloned()
+                .collect();
+            dependent_dimensions.into_iter().for_each(|dependent| {
+                if !used_dimensions_rs.get().contains(&dependent) {
+                    used_dimensions_ws.update(|value: &mut HashSet<String>| {
+                        value.insert(dependent.clone());
+                    });
```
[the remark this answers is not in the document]
mahatoankitkumar: Will try keeping a local mutable variable here

ShubhranshuSanjeev (changes requested):

crates/frontend/src/components/context_form.rs
```diff
@@ -216,30 +211,88 @@ where
             .collect::<HashSet<String>>(),
     );
 
+    let update_context_with_dimension =
```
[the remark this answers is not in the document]
mahatoankitkumar: Because it actually just adds dimensions

crates/frontend/src/components/context_form.rs
```diff
@@ -216,30 +211,88 @@ where
...
+                    dimension_info.dimension.clone(),
+                    SchemaType::try_from(dimension_info.schema.clone()).unwrap(),
+                ));
+            }
+            let dependencies: Vec<String> = dimension_info
+                .dependency_graph
+                .keys()
+                .filter(|key| **key != dimension_info.dimension)
+                .cloned()
+                .collect();
```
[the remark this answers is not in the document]
mahatoankitkumar: Yes

crates/frontend/src/components/context_form.rs
```diff
@@ -216,30 +211,88 @@ where
             .collect::<HashSet<String>>(),
     );
 
+    let update_context_with_dimension =
+        move |context: &mut Conditions, dimension_info: &DimensionWithMandatory| {
```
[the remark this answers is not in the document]
mahatoankitkumar: dimensions here would mean multiple dimensions right
But here its the data for just one dimension at a time

crates/frontend/src/components/context_form.rs
```diff
@@ -259,15 +312,34 @@ where
...
+                let parents_in_context = parents
+                    .into_iter()
+                    .filter(|parent| used_dimensions.get().contains(parent))
+                    .collect::<Vec<String>>()
+                    .join(", ");
+                return format!("Required by: {parents_in_context}",);
+            }
+        }
+        String::new()
+    };
```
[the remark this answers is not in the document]
mahatoankitkumar: But isn't it whats happening now
this is being called inside the For each loop in the view
which will trigger it when context changes

ShubhranshuSanjeev (approved):

crates/frontend/src/components/dropdown.rs
```diff
@@ -24,6 +24,8 @@ pub enum DropdownDirection {
 pub fn dropdown<T>(
     dropdown_options: Vec<T>,
     on_select: Callback<T, ()>,
+    #[prop(into, default = Vec::new())] selected: Vec<T>,
+    #[prop(into, default = Callback::from(|_| {}))] on_remove: Callback<T, ()>,
```
ayushjain17, line 28: fine, but does not seem like the best way to change the mode from  single item to multiple items

crates/frontend/src/components/dropdown.rs
```diff
@@ -59,81 +65,116 @@ where
...
+                            let label = option.label();
+                            view! {
+                                <li
+                                    class="w-full"
+                                    on:click=move |_| {
+                                        let selected_option = option.clone();
+                                        selected_ws
+                                            .update(|selected| {
+                                                selected.push(selected_option.clone());
+                                            });
```
ayushjain17, line 136: shouldn't multi-select be a boolean param and this to executed only if the boolean is true ?

crates/frontend/src/pages/dimensions.rs
```diff
@@ -155,51 +167,57 @@ pub fn dimensions() -> impl IntoView {
...
+                                header="Edit Dimension"
+                                handle_close=handle_close
+                            >
+                                <DimensionForm
+                                    edit=true
+                                    position=selected_dimension_data.position
+                                    dimension_name=selected_dimension_data.dimension
+                                    dimension_schema=selected_dimension_data.schema
+                                    dependencies=selected_dimension_data.dependencies
+                                    function_name=selected_dimension_data.function_name
```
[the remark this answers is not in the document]
ayushjain17: this can be taken up later

ayushjain17 (approved):
re-initialization of signal needs to be checked

Datron (approved):

PR #450, opened 2025-04-01
feat: smithy models

pratikmishra356:
## Problem
Generate sdk for superposition in multiple languages

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -100,12 +100,11 @@ pub fn add_audit_id_to_header(
 }
 
 fn add_last_modified_to_header(
-    max_created_at: Option<NaiveDateTime>,
+    max_created_at: Option<DateTime<Utc>>,
     resp_builder: &mut HttpResponseBuilder,
 ) {
-    if let Some(ele) = max_created_at {
-        let datetime_utc: DateTime<Utc> = TimeZone::from_utc_datetime(&Utc, &ele);
-        let value = HeaderValue::from_str(&DateTime::to_rfc2822(&datetime_utc));
```
Datron, line 108: Is the default for DateTime<Utc> RFC 2822? If not the cac_client will fail since it has this code:
`DateTime::parse_from_rfc2822(header_str)`

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -130,29 +129,29 @@ fn add_config_version_to_header(
...
-fn is_not_modified(max_created_at: Option<NaiveDateTime>, req: &HttpRequest) -> bool {
-    let nanosecond_erasure = |t: NaiveDateTime| t.with_nanosecond(0);
+fn is_not_modified(max_created_at: Option<DateTime<Utc>>, req: &HttpRequest) -> bool {
+    let nanosecond_erasure = |t: DateTime<Utc>| t.with_nanosecond(0);
     let last_modified = req
         .headers()
         .get("If-Modified-Since")
         .and_then(|header_val| {
             let header_str = header_val.to_str().ok()?;
             DateTime::parse_from_rfc2822(header_str)
```
Datron, line 148: The header comes from the client as far as I know, if we rework that you don't need to this.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -413,21 +414,31 @@ async fn delete_context_handler(
 async fn bulk_operations(
     state: Data<AppState>,
     custom_headers: CustomHeaders,
-    reqs: Json<Vec<ContextAction>>,
+    req: Either<Json<Vec<ContextAction>>, Json<BulkOperation>>,
```
Datron, line 417: This is cool

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -105,6 +105,16 @@ pub struct WeightRecomputeResponse {
...
+#[derive(Deserialize)]
+pub struct BulkOperation {
+    pub operations: Vec<ContextAction>,
+}
+
+#[derive(Serialize)]
+pub struct BulkOperationResponse {
+    pub output: Vec<ContextBulkResponse>,
+}
+
```
Datron, line 117: Should these just be type aliases?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -8,12 +8,12 @@ use actix_web::http::header::ContentType;
...
 use actix_web::web::Data;
 use actix_web::{
-    get, put,
+    get, put, route,
     web::{self, Json, Query},
     HttpRequest, HttpResponse, HttpResponseBuilder, Scope,
 };
 use cac_client::{eval_cac, eval_cac_with_reasoning, MergeStrategy};
-use chrono::{DateTime, NaiveDateTime, TimeZone, Timelike, Utc};
+use chrono::{DateTime, Timelike, Utc};
```
sauraww, line 16: Timelike is redundant , we are not using it anywhere . 
Can we remove it ?

sauraww (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -709,26 +710,37 @@ async fn get_config(
...
-    if !query_params_map.is_empty() {
-        config = config.filter_by_dimensions(&query_params_map)
+    let context = if req.method() == actix_web::http::Method::GET {
+        query_params_map
+    } else {
+        match body {
+            Some(b) => b.into_inner(),
+            None => {
+                return Err(bad_argument!(
+                    "When using POST, context needs to be provided in the body."
```
sauraww, line 724: Why this error message ? 
This is assuming that the method cannot be PUT or any other thing ?

sauraww (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -763,14 +775,28 @@ async fn get_resolved_config(
...
+        Some(Value::String(_))
+    );
+    let context = if req.method() == actix_web::http::Method::GET {
+        query_params_map
+    } else {
+        match body {
+            Some(b) => b.into_inner(),
+            None => {
+                return Err(bad_argument!(
+                    "When using POST, context needs to be provided in the body."
```
sauraww, line 789: Same here

sauraww (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -709,26 +710,37 @@ async fn get_config(
...
-        config = config.filter_by_dimensions(&query_params_map)
+    let context = if req.method() == actix_web::http::Method::GET {
+        query_params_map
+    } else {
+        match body {
+            Some(b) => b.into_inner(),
+            None => {
+                return Err(bad_argument!(
+                    "When using POST, context needs to be provided in the body."
+                ))
```
ShubhranshuSanjeev, line 725:
```suggestion
        body.ok_or(bad_argument!(
            "When using POST, context needs to be provided in the body."
        ))?
        .into_inner()
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -763,14 +775,28 @@ async fn get_resolved_config(
...
+            }
```
ShubhranshuSanjeev, line 791:
```suggestion
        body.ok_or(bad_argument!(
            "When using POST, context needs to be provided in the body."
        ))?
        .into_inner()
```

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -105,6 +105,16 @@ pub struct WeightRecomputeResponse {
     pub new_weight: BigDecimal,
 }
 
+#[derive(Deserialize)]
+pub struct BulkOperation {
+    pub operations: Vec<ContextAction>,
+}
```
ShubhranshuSanjeev, line 111: BulkOperation could have been a tuple struct.

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -245,11 +246,16 @@ pub struct ContextMoveReq {
...
     pub action: Option<CommaSeparatedStringQParams>,
     pub username: Option<String>,
     pub count: Option<i64>,
     pub page: Option<i64>,
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ApplicableVriantsOutput {
+    pub applicable_variants: Vec<Variant>,
+}
```
ShubhranshuSanjeev, line 261: For all the list kind of responses, can we create a wrapper response type, something like 
```
struct ListResponse<T> { data: Vec<T> }
```

ShubhranshuSanjeev (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
ShreyBana: Great catch @Datron! Let's evaluate how to handle this one.

smithy/models/context.smithy
```diff
@@ -0,0 +1,385 @@
...
+@http(method: "GET", uri: "/context/list")
+operation ListContexts {
+    input := with [WorkspaceMixin] {
+        @httpQuery("page")
+        @notProperty
+        page: Integer
+
+        @httpQuery("size")
+        @notProperty
+        size: Integer
```
ShubhranshuSanjeev, line 226: In some of the list endpoints, `count` has been used in place of `size`. This can be unified.

smithy/models/context.smithy
```diff
@@ -0,0 +1,385 @@
...
+        @notProperty
+        condition: Condition
+
+        @notProperty
+        old_weight: Integer
+
+        @notProperty
+        new_weight: Integer
+    }
+}
```
ShubhranshuSanjeev, line 297: I believe this operation recomputes the weight for whole list of contexts in the DB, so the output should be a list, but this seems like a single object response. Correct me if I am wrong.

crates/context_aware_config/src/api/config/handlers.rs
ShreyBana: Yes it is, if you scroll up a bit you'll see that only GET & POST are allowed.

crates/context_aware_config/src/api/config/handlers.rs
ShreyBana: Same here

smithy/models/context.smithy
```diff
@@ -0,0 +1,385 @@
...
+
+    $value
+
+    $override
+
+    $override_id
+
+    $weight
+
+    $override_with_keys
```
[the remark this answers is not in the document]
ShreyBana: Yeah my mistake, removing it.

smithy/models/context.smithy
```diff
@@ -0,0 +1,385 @@
...
+    input := for Context with [WorkspaceMixin] {
+        // TODO Find re-name functionality.
+        @required
+        @notProperty
+        context: Condition
+
+        // REVIEW Should this be made a property?
+        @httpHeader("x-config-tags")
+        @notProperty
+        config_tags: String
```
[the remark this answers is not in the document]
ShreyBana: Sadly that will require us to have config_tags to be a property of `Context` resource.

crates/context_aware_config/src/api/config.rs
```diff
@@ -1,3 +1,10 @@
 mod handlers;
 pub use handlers::endpoints;
+use serde::Deserialize;
+use superposition_types::custom_query::QueryMap;
 mod helpers;
+
+#[derive(Deserialize)]
+pub struct ContextPayload {
+    pub context: QueryMap,
+}
```
ayushjain17, line 10: we dont add types to this file

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -100,12 +100,18 @@ pub fn add_audit_id_to_header(
 }
 
 fn add_last_modified_to_header(
-    max_created_at: Option<NaiveDateTime>,
+    max_created_at: Option<DateTime<Utc>>,
+    is_smithy: bool,
```
ayushjain17, line 104: instead of calling it `is_smithy` which is tying the code down to smithy in a way, can we rather make it like an enum saying which RFC to follow for time

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -591,16 +602,22 @@ async fn bulk_operations(
...
     // Commit the transaction
     #[cfg(feature = "high-performance-mode")]
     put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
-    Ok(http_resp.json(response))
+
+    let http_resp = if is_v2 {
+        resp_builder.json(BulkOperationResponse { output: response })
+    } else {
+        resp_builder.json(response)
+    };
```
ayushjain17, line 619: instead of v2 check, could we have rather used the match case on request itself ?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -688,5 +705,6 @@ async fn weight_recompute(
         AppHeader::XConfigVersion.to_string(),
         config_version_id.to_string(),
     ));
-    Ok(http_resp.json(response))
+    // FIXME Should go a in more generic struct like `ListResponse<Response>`.
+    Ok(http_resp.json(json!({ "results": response })))
```
ayushjain17, line 709: is this is the case (as per the comment) should we rather use `PaginationResponse` here directly, instead of migrating to it later on ?

makefile
```diff
@@ -164,13 +167,29 @@ test: setup frontend superposition
...
 
+smithy-build:
+	cd smithy && smithy build
+
+ts-client: smithy-build
+	mkdir -p $(SMITHY_CLIENT_DIR)
+# cp -r $(SMITHY_BUILD_SRC)/typescript-client-codegen $(SMITHY_CLIENT_DIR)/ts
+## Skipping package.json as we have installed jest, probably will have to change
+## this once we move to `smoke-tests`.
+	rsync -av --exclude="package.json" $(SMITHY_BUILD_SRC)/typescript-client-codegen/ $(SMITHY_CLIENT_DIR)/ts/
```
ayushjain17, line 189: what is `rsync` ?

ayushjain17 (changes requested):
review of smithy files and 2 rust files remaining

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -673,44 +677,22 @@ async fn get_applicable_variants(
...
+async fn get_applicable_variants_v2(
+    db_conn: DbConnection,
+    query_data: Query<ApplicableVariantsQuery>,
+    schema_name: SchemaName,
+) -> superposition::Result<HttpResponse> {
+    let variants = fetch_applicable_variants(db_conn, query_data, schema_name)?;
+    Ok(HttpResponse::Ok().json(ApplicableVariantsOutput {
+        applicable_variants: variants,
+    }))
+}
```
ayushjain17, line 694: are we not going to combine the handlers and use `route` here as well ?

smithy/models/dimension.smithy
```diff
@@ -0,0 +1,144 @@
...
+        position: Integer
+        schema: Document
+        function_name: String
+        description: String
+        change_reason: String
+        created_at: DateTime
+        created_by: String
+        last_modified_at: DateTime
+        last_modified_by: String
+    }
```
ayushjain17, line 22: 3 new fields have been added to it, can we get that change reflected here as well
also, the spec for create and update has a new key, can we add that as well

makefile
ShreyBana: [https://linux.die.net/man/1/rsync](rsync)

smithy/models/experiments.smithy
```diff
@@ -0,0 +1,317 @@
...
+    output : ExperimentListResponse
+}
+
+
+
+@http(method: "POST", uri: "/experiments/applicable-variants")
+operation ApplicableVariants {
+    input: ApplicableVariantsInput
+    output : ApplicableVariantsOutput
+}
```
ayushjain17, line 317: the api needs to be fixed first though, it is currently taking no BODY and only in query params

sauraww (changes requested):

smithy/models/organisation.smithy
```diff
@@ -0,0 +1,164 @@
...
+
+
+@httpError(404)
+@error("client")
+structure OrganisationNotFound {}
+
+
+// Operations
+@http(method: "POST", uri: "/superposition/organisations")
+operation CreaterOrganisation {
```
sauraww, line 124: Why is this named CreaterOrganisation ? 
Shouldn't it be CreateOrganisation ?
@ShreyBana

smithy/models/organisation.smithy
```diff
@@ -0,0 +1,164 @@
...
+    $contact_phone
+
+    $admin_email
+
+    $sector
+
+    $status
+
+}
+
```
sauraww, line 75: Can use mixin to reduce DRY ?

sauraww: `@mixin 
structure OrganisationFields {
    $name: String,
    $country_code: String,
    $contact_email: String,
    $contact_phone: String,
    $admin_email: String,
    $sector: String
}

structure CreateOrganisationRequest for Organisation with [OrganisationFields] {
    @required $admin_email
    @required $name
}

structure UpdateOrganisationRequest for Organisation with [OrganisationFields] {
    @httpLabel @required $id
    $status: OrgStatus
}
`

crates/superposition_types/src/database/superposition_schema.rs
```diff
@@ -51,8 +51,8 @@ pub mod superposition {
             workspace_admin_email -> Text,
             created_by -> Text,
             last_modified_by -> Text,
-            last_modified_at -> Timestamp,
-            created_at -> Timestamp,
+            last_modified_at -> Timestamptz,
+            created_at -> Timestamptz,
```
sauraww, line 55: Was this autogenerated ?

sauraww (changes requested):
Can you guys address all the comments ?

crates/superposition_types/src/database/superposition_schema.rs
ShreyBana: No,I wrote this.

sauraww: this should have been generated by some diesel command , right ?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -707,7 +714,11 @@ async fn get_applicable_variants(
             variants.push(v)
         }
     }
-
+    if req.method() == actix_web::http::Method::POST {
```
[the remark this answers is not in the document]
ShreyBana: Good idea. Btw is there any chance that a lint rule will help us catch this?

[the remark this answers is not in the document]
ayushjain17: in that case, we can also use Either as Response type, that way the request and response would get better documented

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -667,14 +669,19 @@ pub async fn discard(
     Ok((updated_experiment, config_version_id))
 }
 
-#[get("/applicable-variants")]
+#[route("/resolve", method = "GET", method = "POST")]
```
ayushjain17, line 672:
```suggestion
#[route("/applicable-variants", method = "GET", method = "POST")]
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -667,14 +669,19 @@ pub async fn discard(
...
+    req_body: web::Json<ApplicableVariantsQuery>,
     query_data: Query<ApplicableVariantsQuery>,
     schema_name: SchemaName,
 ) -> superposition::Result<HttpResponse> {
     let DbConnection(mut conn) = db_conn;
-    let query_data = query_data.into_inner();
+    let mut req_data = query_data.into_inner();
+    if req.method() == actix_web::http::Method::POST {
+        req_data = req_body.into_inner();
+    }
```
ayushjain17, line 684: should we consider checking that Query params is needed for GET and request body for POST, for other cases we give proper error, so that it is less confusing as to why things didnt work when someone is using with wrong combination

as with the separate route approach, it would have resulted in a well known `Not Found` error

crates/superposition_types/src/database/superposition_schema.rs
ShreyBana: No we have to use patch files to maintain such alters. Diesel doesn't manage them.

smithy/models/functions.smithy
```diff
@@ -0,0 +1,229 @@
...
+        last_modified_by: String
+        change_reason: String
+    }
+
+    read: GetFunction
+    put: UpdateFunction
+    delete: DeleteFunction
+    list: ListFunction
+    operations: [CreateFunction, Test, Publish]
+}
```
ayushjain17, line 34: this also requires some change

smithy/models/organisation.smithy
ShreyBana: @sauraww - I eat my words, I didn't know you could re-define the fields after using a mixin. Which part of the documentation did you get this from?
CC: @pratikmishra356

crates/context_aware_config/src/api/config/handlers.rs
ShubhranshuSanjeev: I think the only reason to use the other RFC3339 is because Smithy follows that, there is no other usecase as far as I know (correct me if I am wrong). Important question to ask is, "Is it really necessary to have an ADT to represent this state?". My answer: I don't really care.

crates/superposition_types/src/database/schema-timestamp-migration.patch
ayushjain17, line 1: this file should not be needed anymore

crates/context_aware_config/src/api/config/handlers.rs
ShubhranshuSanjeev: Just to clarify, I mean that there is not a good enough case to use an ADT, so I don't care if an ADT is being used or not.

ayushjain17 (approved):

ShubhranshuSanjeev (approved):

sauraww (approved):

PR #451, opened 2025-04-03
feat: auto desearlizing resolved config

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/clients/haskell/hs-cac-client/src/Client.hs
+++ b/clients/haskell/hs-cac-client/src/Client.hs
@@ -15,8 +15,10 @@ module Client
 ) where
 
 import           Data.Aeson
+import           Data.Text.Encoding    (encodeUtf8)
 import           Data.Functor          (($>))
 import           Data.List             (intercalate)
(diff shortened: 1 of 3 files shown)


ayushjain17 (approved):

clients/haskell/hs-cac-client/src/Client.hs
```diff
@@ -131,7 +133,7 @@ getFullConfigStateWithFilter client mbFilters mbPrefix = do
         then Left <$> getError
         else do
             fptrConfig <- newForeignPtr c_free_string config
-            Right . toJSON <$> withForeignPtr fptrConfig peekCString
+            (maybe (Left "Failed to decode resolved config") Right . decodeStrict <$> encodeUtf8) . pack <$> withForeignPtr fptrConfig peekCString
```
Datron, line 136: Can we break this up into steps?

Datron (approved):

PR #458, opened 2025-04-11
feat: Link autocomplete functions

Datron:
## Problem
Users cannot yet use the autocomplete functions they create

## Solution
Provide a new kind of input component that can show a dropdown that shows potential values returned by the autocomplete function

## Pre-deployment activity
Need to run the database queries to add new columns to dimension and default config tables

https://github.com/user-attachments/assets/1fe00b3e-9961-4f6d-8e88-8732628760fe

### Example completion functions

City prediction 

```
// name: string - dimension or config name
// prefix: string - characters entered in the input field
// environment: object { context: Object, overrides: Object } - captures out elements in the form like context, overrides etc.
// returns: [string]
async function autocomplete(name, prefix, environment) {
    // console.log("Environment = ", JSON.stringify(environment));
    let state = null;
    for(const s of environment.context) {
        if (s.variable == "state")
            state = Object.values(s.expression)[0];
    }
    console.log("State value = ", state);
    let url = state == undefined || state == null ? 'https://countriesnow.space/api/v0.1/countries/cities' : 'https://countriesnow.space/api/v0.1/countries/state/cities';
    let payload = state == undefined || state == null ?
        {
            country: "India"
        } :
        {
            country: "India",
            state: state
        };
    let config = {
        method: 'post',
        maxBodyLength: Infinity,
        url: url,
        headers: { 'Content-Type': 'application/json' },
        data: payload
    };
    try {
        let response = await axios(config);
        let cities = response.data.data;
        let city_arr = cities.filter(city =>
            city.toLowerCase().startsWith(prefix.toLowerCase())
        );
        console.log(city_arr);
        return Array.from(city_arr);
    } catch {
        console.error('Error fetching cities:', error);
        return [];

    }
}
```

States prediction

```
// name: string - dimension or config name
// prefix: string - characters entered in the input field
// environment: object { context: Object, overrides: Object } - captures out elements in the form like context, overrides etc.
// returns: [string]
async function autocomplete(name, prefix, environment) {
    console.log("environment = ", environment);
    var config = {
        method: 'post',
        maxBodyLength: Infinity,
        url: 'https://countriesnow.space/api/v0.1/countries/states',
        headers: { 'Content-Type': 'application/json' },
        data: {
            country: "India"
        }
    };
    try {
        let response = await axios(config);
    if (!response.data || !response.data.data || !response.data.data.states) {
            console.log('Invalid response format');
            return [];
        }

        // Extract state names from the response
        let states = response.data.data.states.map(state => state.name);
        // Filter states that start with the prefix (case-insensitive)
        let state_arr = states.filter(state => 
            typeof state === 'string' && 
            state.toLowerCase().startsWith(prefix.toLowerCase())
        );
        console.log("state check ", state_arr);
        return Array.from(state_arr);

    } catch (error) {
        console.log('Error in autocompleteStates function:', error);
        if (error.response) {
            console.log('Error response:', error.response.data);
        }
        return [];
    }

}
```


crates/context_aware_config/src/api/functions/helpers.rs
```diff
@@ -35,10 +38,14 @@ pub fn get_published_function_code(
 pub fn get_published_functions_by_names(
     conn: &mut DBConnection,
     function_names: Vec<String>,
+    fntype: FunctionTypes,
```
pratikmishra356, line 41: why we need functiontype here , we can fetch by name itself ?

ayushjain17 (changes requested):
need to review again once - probably once the changes are pushed

clients/generated/smithy/typescript/.gitignore
```diff
@@ -5,5 +5,4 @@
 /dist-*
 *.tsbuildinfo
 *.tgz
-*.log
-package-lock.json
```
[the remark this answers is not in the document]
Datron: package lock guarantees the builds are the same since the dependencies won't get randomly bumped up every time we run the typescript client

crates/frontend/src/components/context_form.rs
```diff
@@ -184,8 +190,10 @@ pub fn condition_input(
 #[component]
 pub fn context_form<NF>(
     handle_change: NF,
-    context: Conditions,
+    context_rs: ReadSignal<Conditions>,
```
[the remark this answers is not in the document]
Datron: Why not just do ReadSignal?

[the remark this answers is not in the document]
ayushjain17: that forces, that only ReadSignal is allowed for this, passing a derived signal would not work in that case

using this Signal, works for both the cases

crates/frontend/src/api.rs
```diff
@@ -528,3 +530,41 @@ pub async fn resolve_config(
...
+    let host = use_host_server();
+    let client = reqwest::Client::new();
+    let url = format!("{}/function/{}/PUBLISHED/test", host, fn_name);
+    let err_handler = |e: reqwest::Error| ServerFnError::new(e.to_string());
+    let payload = FunctionExecutionRequest::AutocompleteFunctionRequest {
+        name: name.to_owned(),
+        prefix: value.to_owned(),
+        environment: environment.clone(),
+    };
+    let response = client
```
mahatoankitkumar, line 551: Can we please the request fn here, it handles the error there itself

mahatoankitkumar (changes requested):

crates/frontend/src/api.rs
```diff
@@ -528,3 +530,41 @@ pub async fn resolve_config(
...
+        .send()
+        .await
+        .map_err(err_handler)?
+        .json::<FunctionExecutionResponse>()
+        .await
+        .map_err(err_handler)?;
+    let result = response
+        .fn_output
+        .as_array()
+        .unwrap()
```
sauraww, line 565: Can we avoid plain unwrap here and rather use expect or something ?

crates/frontend/src/api.rs
```diff
@@ -528,3 +530,41 @@ pub async fn resolve_config(
...
+        .iter()
+        .map(|v| v.as_str().unwrap().to_string())
```
sauraww, line 567: `.filter_map(|v| v.as_str().map(String::from))`

sauraww (changes requested):

crates/frontend/src/pages/default_config.rs
```diff
@@ -169,11 +170,13 @@ pub fn default_config() -> impl IntoView {
             let schema_object =
                 serde_json::from_str::<Value>(&schema).unwrap_or(Value::Null);
 
-            let function_name = row["function_name"].to_string();
-            let fun_name = match function_name.as_str() {
-                "null" => None,
-                _ => Some(function_name.replace('"', "")),
-            };
+            // keeping the function_name field the same for backwards compatibility
```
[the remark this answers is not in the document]
Datron: > half and half is going to cause a lot of problems

What lot of problems will it cause?

crates/frontend/src/pages/dimensions.rs
```diff
@@ -103,6 +103,7 @@ pub fn dimensions() -> impl IntoView {
             let schema = row["schema"].clone().to_string();
             let schema = serde_json::from_str::<Value>(&schema).unwrap_or(Value::Null);
 
+            // keeping the function_name field the same for backwards compatibility
```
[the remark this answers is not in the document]
Datron: Not picking this up now, we can create an issue to track it

crates/frontend/src/components/context_form.rs
```diff
@@ -345,6 +347,26 @@ where
...
+                .0;
+            autocomplete_fn_generator(
+                d.dimension.clone(),
+                d.autocomplete_function_name.clone().unwrap(),
+                fn_environment,
+                tenant,
+                org_id,
+            )
+        })
+        .collect::<AutoCompleteCallbacks>();
```
sauraww, line 368: let tenant = use_context::<RwSignal<Tenant>>()
    .expect("Tenant context not found")
    .get_untracked()
    .0;

let org_id = use_context::<RwSignal<OrganisationId>>()
    .expect("OrganisationId context not found")
    .get_untracked()
    .0;
    
let autocomplete_callbacks = dimensions
    .get_value()
    .iter()
    .filter_map(|d| {
        d.autocomplete_function_name.as_ref().map(|function_name| {
            autocomplete_fn_generator(
                d.dimension.clone(),
                function_name.clone(),
                fn_environment,
                tenant.clone(),
                org_id.clone(),
            )
        })
    })
    .collect::<AutoCompleteCallbacks>();

crates/frontend/src/components/menu.rs
```diff
@@ -0,0 +1,49 @@
...
+    #[prop(into, default = String::new())] id: String,
+) -> impl IntoView
+where
+    T: DropdownOption + Clone + 'static,
+{
+    let all_options = StoredValue::new(options.clone());
+    let menu_options =
+        Signal::derive(move || all_options.get_value().into_iter().collect::<Vec<T>>());
+    view! {
+        <div class="relative">
```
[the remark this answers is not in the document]
Datron: Screenshot 2025-05-12 at 3 53 19 PM
Without this div the menu shifts things around it. If you still feel its an issue lets create a ticket and pick it up at a later point. I don't see this as a PR blocker

crates/frontend/src/components/context_form.rs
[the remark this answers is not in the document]
Datron: I want to force read_signal, no? Are you suggesting this so we can use RWSignals with this component?

crates/frontend/src/api.rs
Datron: I want to fail if an element is not a string, rather than ignoring it

crates/frontend/src/api.rs
Datron: Its is guaranteed to be an array of strings since the function producing it has been tested before being stored in the DB. This unwrap will fail in scenarios of massive data corruption

crates/frontend/src/components/override_form.rs
```diff
@@ -173,6 +179,23 @@ where
         };
     });
 
+    let autocomplete_callbacks = default_config
+        .get_value()
+        .iter()
+        .filter(|default_config| default_config.autocomplete_function_name.is_some())
```
sauraww, line 185: Can we use filter map here as well and avoid using unwrap ?
I have added a similar suggestion above also:[ https://github.com/juspay/superposition/pull/458/files#r2086455070](url)

crates/frontend/src/api.rs
Datron: It doesn't though, it returns a String as error which I will still have to wrap as a ServerFnError, but agree on using request function

crates/frontend/src/api.rs
sauraww: In that case also , I guess it makes sense to add expect

crates/frontend/src/components/default_config_form.rs
```diff
@@ -357,20 +357,23 @@ where
...
+                        functions.sort_by(|a, b| a.function_name.cmp(&b.function_name));
+                        functions.iter()
                             .for_each(|ele| {
-                                function_names.push(ele.function_name);
+                                if ele.function_type == FunctionType::Validation {
+                                    validation_function_names.push(ele.function_name.clone());
+                                } else {
+                                    autocomplete_function_names.push(ele.function_name.clone());
+                                }
                             });
```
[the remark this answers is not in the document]
Datron: Could you elaborate? What is a deselect component?

crates/frontend/src/components/override_form.rs
ayushjain17: +1

Datron: @sauraww here I only want to generate functions for items that have autocomplete_function_name, so inside the filter_map I would still check if `default_config.autocomplete_function_name.is_some()` is the case, and return Some(fn). Doesn't really make sense for this case

clients/generated/smithy/typescript/.gitignore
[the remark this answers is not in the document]
Datron: It would be right? The package-lock is generated when `npm ci` is run inside the typescript client, then the client is built and used in our tests. Our tests could fail if a package is updated with a breaking change

crates/frontend/src/api.rs
ayushjain17: that is not how you can handle this, you would need to send back an error from here and show the error somewhere probably as a snackbar

using unwrap simply crashes the frontend application giving feedback, so this technically worse than ignoring the error

crates/frontend/src/api.rs
ayushjain17: please dont expect or unwrap anywhere in frontend, it is a feedback application crash - with no recovery

crates/frontend/src/components/context_form.rs
Datron: Tenant and org_id are used nowhere else except inside this closure, and I would keep moving it inside each time and cloning it. Seems redundant.

ayushjain17: +1

crates/frontend/src/components/context_form.rs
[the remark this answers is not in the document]
ayushjain17: you dont want to force read_signal, you want to force Signal which is meant to be the types for read related signals

> Signal is an enumerated type that represents any kind of readable reactive signal, or a plain value. It can be useful when defining APIs for components you’ll want to reuse while passing different sorts of signals.
https://book.leptos.dev/view/03_components.html#into-props

crates/frontend/src/components/context_form.rs
ayushjain17: but we can do this
```suggestion
            let tenant = use_context::<RwSignal<Tenant>>().unwrap();
            let org_id = use_context::<RwSignal<OrganisationId>>().unwrap();
    let autocomplete_callbacks = dimensions
        .get_value()
        .iter()
        .filter(|d| d.autocomplete_function_name.is_some())
        .map(|d| {
            autocomplete_fn_generator(
                d.dimension.clone(),
                d.autocomplete_function_name.clone().unwrap(),
                fn_environment,
                tenant.get_untracked().0,
                org_id.get_untracked().0,
            )
        })
        .collect::<AutoCompleteCallbacks>();
```

crates/frontend/src/components/override_form.rs
```diff
@@ -173,6 +179,26 @@ where
...
+                .0;
+            autocomplete_fn_generator(
+                d.key.clone(),
+                d.autocomplete_function_name.clone().unwrap(),
+                fn_environment,
+                tenant,
+                org_id,
+            )
+        })
+        .collect::<AutoCompleteCallbacks>();
```
ayushjain17, line 200:
```suggestion
            let tenant = use_context::<RwSignal<Tenant>>().unwrap();
            let org_id = use_context::<RwSignal<OrganisationId>>().unwrap();
    let autocomplete_callbacks = default_config
        .get_value()
        .iter()
        .filter_map(|d| {
           d.autocomplete_function_name.as_ref().map(|function_name| {
            autocomplete_fn_generator(
                d.key.clone(),
                function_name,
                fn_environment,
                tenant.get_untracked().0,
                org_id.get_untracked().0,
            )
            })
        })
        .collect::<AutoCompleteCallbacks>();;
```

crates/frontend/src/components/override_form.rs
ayushjain17: what @sauraww is simply saying is to do this:
https://github.com/juspay/superposition/pull/458/files#r2086626709
```
            let tenant = use_context::<RwSignal<Tenant>>().unwrap();
            let org_id = use_context::<RwSignal<OrganisationId>>().unwrap();
    let autocomplete_callbacks = default_config
        .get_value()
        .iter()
        .filter_map(|d| {
           d.autocomplete_function_name.as_ref().map(|function_name| {
            autocomplete_fn_generator(
                d.key.clone(),
                function_name,
                fn_environment,
                tenant.get_untracked().0,
                org_id.get_untracked().0,
            )
            })
        })
        .collect::<AutoCompleteCallbacks>();
```

crates/frontend/src/pages/dimensions.rs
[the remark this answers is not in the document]
ayushjain17: we already a have of pending TODOs for this PR, and anyways this should be a small change only

crates/frontend/src/pages/default_config.rs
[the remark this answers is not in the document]
ayushjain17: its a tech debt, which can be totally avoided

crates/frontend/src/components/override_form.rs
Datron: But why? The filter then map looks cleaner than doing filter_map and then a map inside the filter_map. Regardless, this is an opinion and is not a PR blocker. Can we compromise on this, please?

crates/frontend/src/components/input.rs
```diff
@@ -294,6 +321,39 @@ fn basic_input(
...
+                                <SelectionMenu
+                                    id=dropdown_id
+                                    options=suggestions.to_owned()
+                                    on_select=Callback::new(move |selected: String| {
+                                        let input = get_element_by_id::<
+                                            web_sys::HtmlInputElement,
+                                        >(&target_id);
+                                        if let Some(input) = input {
+                                            input.set_value(&selected);
+                                        }
```
ayushjain17, line 343: pending comment:

> this should not be needed, the input component can be made reactive to value change rather

https://github.com/juspay/superposition/pull/458#discussion_r2083564732

crates/frontend/src/components/input.rs
```diff
@@ -294,6 +321,39 @@ fn basic_input(
                 }
             }}
 
+            {move || {
+                suggestions_rs
+                    .with(|suggestions| {
+                        if autocomplete_function.is_some() && !suggestions.is_empty()
+                            && show_suggestions_rws.get()
```
ayushjain17, line 328: pending thread:
https://github.com/juspay/superposition/pull/458#discussion_r2083563988

crates/frontend/src/components/override_form.rs
ayushjain17: to me, usage of any unwrap usage seems a blocker
the point of the comment is to move use_context outside map, and write code in such a way which simply does not require you to use unwrap

crates/frontend/src/components/context_form.rs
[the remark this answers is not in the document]
Datron: Actually I just want a read signal, and I don't want to complicate this any further. I'm not seeing the benefit of what you're suggesting.

crates/frontend/src/components/context_form.rs
knutties: Folks went through the code.  My 2 cents on this topic:

1. We should avoid `unwrap()` where-ever possible.  I agree with this completely
2. In the alternative that @ayushjain17 and @sauraww has posted - can we do one better to remove the inner map where the `autocomplete_fn_generator` function can return an Err wrapped inside a Result.  That the outer `filter_map` itself will drop that entry.  That implementation would be even more succinct right?  Basically the return type of `autocomplete_fn_generator` can be a `Result<(String, AutoCompleteCallback), Err>` instead of just `(String, AutoCompleteCallback)`.

Based on the example at https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map - the filter_map implementation should be smaller and should not ideally need the inner map.

ayushjain17: @knutties so basically instead of taking String as function name, that function would rather take Option<String>, right ?

In this case the autocomplete generator would have to internally either write a match case or write the map there

crates/frontend/src/components/context_form.rs
```diff
@@ -345,6 +347,22 @@ where
         String::new()
     };
 
+    let autocomplete_callbacks = dimensions
+        .get_value()
+        .iter()
+        .filter_map(|d| {
+            let tenant = use_context::<RwSignal<Tenant>>().unwrap();
+            let org_id = use_context::<RwSignal<OrganisationId>>().unwrap();
```
ayushjain17, line 355: can we please move these out of the loop, in the places

crates/frontend/src/api.rs
```diff
@@ -618,3 +619,48 @@ pub async fn get_context_from_condition(
...
 }
+
+pub async fn execute_autocomplete_function(
+    name: &str,
+    value: &str,
+    environment: &Value,
+    fn_name: &String,
+    tenant: &String,
+    org_id: &String,
+) -> Result<Vec<String>, ServerFnError> {
```
ayushjain17, line 630: why do we need to use `ServerFnError` here ? does `String` as error type not suffice ?
asking this, cause we have been using String in all other places for the functions which handle api call, does this type offer any extra information as well ?

crates/frontend/src/components/context_form.rs
Datron: Done

crates/frontend/src/api.rs
Datron: Done

ayushjain17 (approved):
comments still pending though

mahatoankitkumar (approved):

sauraww (approved):

PR #468, opened 2025-04-16
fix: added tests for contexts (along with bug-fixes)

ShubhranshuSanjeev:
- ✅  Tests for contexts (list and delete endpoint tests are pending)
- ✅  Fixed buggy mandatory dimension check in move operation
- ✅  Fixed type-mismatch for bulk operation payload between smithy and backend (context field wrongly names condition)
- ✅  Incase of move operation, description was being updated in the context conflict flow, so without a conflict the moved context never had the updated description
- ⚠️  Not updating change_reason in a move operation (to be fixed)
- ⚠️  Payload type Mismatch for Move operation in Bulk API (to be fixed) [Will fail bulk operation made by client having a move operation]

### DIFF
--- a/clients/tests/ts/tests/context.test.ts
+++ b/clients/tests/ts/tests/context.test.ts
@@ -1,0 +1,1050 @@
+import {
+    SuperpositionClient,
+    UpdateWorkspaceCommand,
+    CreateContextCommand,
+    UpdateOverrideCommand,
+    MoveContextCommand,
+    DeleteContextCommand,
+    GetContextCommand,
+    ListContextsCommand,
+    BulkOperationCommand,
+    CreateDimensionCommand,
+    CreateDefaultConfigCommand,
+    type CreateContextCommandOutput,
+    type GetConfigCommandOutput,
+    type GetContextCommandOutput,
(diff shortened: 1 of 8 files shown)


ayushjain17 (approved):

clients/tests/ts/tests/context.test.ts
```diff
@@ -0,0 +1,1082 @@
+import { SuperpositionClient, UpdateWorkspaceCommand, CreateContextCommand, UpdateOverrideCommand, MoveContextCommand, DeleteContextCommand, GetContextCommand, ListContextsCommand, BulkOperationCommand, CreateDimensionCommand, CreateDefaultConfigCommand, type CreateContextCommandOutput, type GetConfigCommandOutput, type GetContextCommandOutput, WeightRecomputeCommand, WorkspaceStatus } from "@io.juspay/superposition-sdk";
+import { ENV } from "./env.ts";
```
ayushjain17, line 2: this does not exist

ayushjain17 (changes requested):

ayushjain17 (approved):

PR #470, opened 2025-04-17
feat: add strict mode support for workspaces

Datron:
## Problem
Non deterministic resolutions of configs

## Solution
Create a workspace setting called strict mode that limits operators to `==`

https://github.com/user-attachments/assets/b7dfbcef-0d31-495d-9127-5f33c25ab716

### DIFF
--- a/crates/context_aware_config/src/api/context/helpers.rs
+++ b/crates/context_aware_config/src/api/context/helpers.rs
@@ -6,15 +6,18 @@ use cac_client::utils::json_to_sorted_string;
 use chrono::Utc;
 use diesel::{ExpressionMethods, QueryDsl, RunQueryDsl, SelectableHelper};
 use serde_json::{Map, Value};
-use service_utils::{helpers::extract_dimensions, service::types::SchemaName};
-use superposition_macros::{unexpected_error, validation_error};
+use service_utils::{
+    helpers::{extract_dimensions, get_variable_name_and_value},
+    service::types::SchemaName,
+};
+use superposition_macros::{bad_argument, unexpected_error, validation_error};
 use superposition_types::{
     api::functions::{FunctionExecutionRequest, FunctionExecutionResponse},
     database::{
         models::cac::{Context, DependencyGraph, FunctionCode, FunctionType},
         schema::{contexts, default_configs::dsl, dimensions},
     },
-    result as superposition, Cac, DBConnection, Overrides, User,
+    result as superposition, Cac, Condition, DBConnection, Overrides, User,
 };
 
 use crate::validation_functions::execute_fn;
@@ -115,6 +118,56 @@ pub fn validate_condition_with_functions(
             }
         }
     }
+    Ok(())
+}
+
+pub fn validate_condition_with_strict_mode(
+    context: &Condition,
+    strict_mode: bool,
+) -> superposition::Result<()> {
+    if !strict_mode {
+        return Ok(());
+    }
+
+    let conditions: Vec<Value> = match (*context).get("and") {
+        Some(conditions_json) => conditions_json
+            .as_array()
+            .ok_or(bad_argument!("Error extracting dimensions, failed parsing conditions as an array. Ensure the context provided obeys the rules of JSON logic"))?
+            .clone(),
+        None => vec![Value::Object(context.to_owned().into())],
+    };
+
+    for condition in &conditions {
+        let condition_obj =
+            condition
+                .as_object()
+                .ok_or(bad_argument!(
+                    "Failed to parse condition as an object. Ensure the context provided obeys the rules of JSON logic"
+                ))?;
+        let operators = condition_obj.keys();
+
+        for operator in operators {
+            let operands = condition_obj[operator].as_array().ok_or(bad_argument!(
+                    "Failed to parse operands as an arrays. Ensure the context provided obeys the rules of JSON logic"
+            ))?;
+
+            let (dimension_name, _) = get_variable_name_and_value(operands)?;
+            // variantIds use 'HAS', and since they are managed internally allow them to bypass strict mode
+            if dimension_name == "variantIds" && operator != "==" && operator != "in" {
+                return Err(bad_argument!(
+                    "Strict mode is enabled, and variantIds is a system dimension that supports IS and IN. Found operator: {}",
+                    operator
+                ));
+            }
+            if dimension_name != "variantIds" && operator != "==" {
+                return Err(bad_argument!(
+                    "Strict mode is enabled, but the context contains unsupported operators. Please use IS operator only. Found operator: {}",
+                    operator
+                ));
+            }
+        }
+    }
+
     Ok(())
 }
 
@@ -267,6 +320,8 @@ pub fn create_ctx_from_put_req(
     let ctx_override = Value::Object(r_override.clone().into());
 
     let workspace_settings = get_workspace(schema_name, conn)?;
+
+    validate_condition_with_strict_mode(&ctx_condition, workspace_settings.strict_mode)?;
 
     let change_reason = req.change_reason.clone();
     let context_map = extract_dimensions(&ctx_condition)?;
--- a/crates/frontend/src/components/experiment_form.rs
+++ b/crates/frontend/src/components/experiment_form.rs
@@ -3,7 +3,8 @@ pub mod utils;
 
 use leptos::*;
 use superposition_types::database::{
-    models::cac::DefaultConfig, types::DimensionWithMandatory,
+    models::{cac::DefaultConfig, Workspace},
+    types::DimensionWithMandatory,
 };
 use utils::{create_experiment, update_experiment};
 use web_sys::MouseEvent;
@@ -46,6 +47,7 @@ where
     let default_config = StoredValue::new(default_config);
     let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
     let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
+    let workspace_settings = use_context::<StoredValue<Workspace>>().unwrap();
 
     let (experiment_name, set_experiment_name) = create_signal(name);
     let (f_context, set_context) = create_signal(context.clone());
@@ -192,6 +194,7 @@ where
                             dimensions=dimensions.get_value()
                             context=context
                             handle_change=handle_context_form_change
+                            resolve_mode=workspace_settings.get_value().strict_mode
                             disabled=edit
                             heading_sub_text=String::from(
                                 "Define rules under which this experiment would run",
--- a/crates/frontend/src/components/workspace_form.rs
+++ b/crates/frontend/src/components/workspace_form.rs
@@ -35,6 +35,7 @@ pub fn workspace_form(
     let (mandatory_dimensions_rs, mandatory_dimensions_ws) =
         create_signal(mandatory_dimensions);
     let (req_inprogess_rs, req_inprogress_ws) = create_signal(false);
+    let (strict_mode_rs, strict_mode_ws) = create_signal(true);
 
     let on_submit = move |ev: MouseEvent| {
         req_inprogress_ws.set(true);
@@ -43,6 +44,7 @@ pub fn workspace_form(
             workspace_admin_email: workspace_admin_email_rs.get(),
             workspace_name: workspace_name_rs.get(),
             workspace_status: Some(workspace_status_rs.get()),
+            workspace_strict_mode: strict_mode_rs.get(),
         };
 
         let update_payload = UpdateWorkspaceRequest {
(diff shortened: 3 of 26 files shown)


crates/frontend/src/components/workspace_form.rs
```diff
@@ -168,6 +170,24 @@ pub fn workspace_form(
                     />
                 </div>
 
+                <Show when=move || !edit>
+                    <div class="form-control">
+                        <label class="label">
+                            <div class="tooltip" data-tip="Strict Mode limits the operators available to just ==. This is the recommended mode for production environments">
```
knutties, line 176: Does this just limit `==` in comparison ?  It should also limit the use of other logical operators except for `&&`.

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,51 @@ pub fn validate_condition_with_functions(
...
+
+    let conditions: Vec<Value> = match (*context).get("and") {
+        Some(conditions_json) => conditions_json
+            .as_array()
+            .ok_or(superposition::AppError::BadArgument("Error extracting dimensions, failed parsing conditions as an array. Ensure the context provided obeys the rules of JSON logic".into()))?
+            .clone(),
+        None => vec![Value::Object(context.to_owned().into())],
+    };
+
+    for condition in &conditions {
```
knutties, line 140: We need to check for `&&` between conditions as well right?  I think the UI today just allows `&&` and hence we might not have other conditions in use - but valid JSON logic allows it and can be used via the API.

crates/frontend/src/components/workspace_form.rs
```diff
@@ -168,6 +170,24 @@ pub fn workspace_form(
...
+                                <span class="label-text">Strict Mode</span>
+                            </div>
+                        </label>
+                        <Toggle
+                            name="workspace-strict-mode"
+                            value=strict_mode_rs.get()
+                            on_change=Callback::new(move |_| {
+                                strict_mode_ws.update(|v| *v = !*v);
+                            })
+                            class=String::from("mt-2")
```
ayushjain17, line 186: can we please rather add `flex flex-col gap-2` to the parent div instead of using margin

crates/frontend/src/components/workspace_form.rs
```diff
@@ -35,6 +35,7 @@ pub fn workspace_form(
     let (mandatory_dimensions_rs, mandatory_dimensions_ws) =
         create_signal(mandatory_dimensions);
     let (req_inprogess_rs, req_inprogress_ws) = create_signal(false);
+    let (strict_mode_rs, strict_mode_ws) = create_signal(true);
```
ayushjain17, line 38: just an advice, can we start using the newer standards of the leptos signal functions names
for example, `create_signal` is deprecated, and instead there is `signal` now to replace it - exact same functionality - just change in naming convention

crates/frontend/src/components/workspace_form.rs
ayushjain17: should we rather consider storing the context as straight forward key value pair object, instead of jsonlogic?
making it implicit `==` everywhere and no other need ?

this would require changes in multiple places, but would lead to simpler and faster operations, with availability to do many operations via sql itself

but on a second thought, we should avoid it if we are going to support workspaces with operators as it would become too many if elses everywhere - but it is kinda tempting as well to have to maybe consider the cost of doing it this way

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,51 @@ pub fn validate_condition_with_functions(
...
 
+pub fn validate_condition_with_strict_mode(
+    context: &Condition,
+    strict_mode: bool,
+) -> superposition::Result<()> {
+    if !strict_mode {
+        return Ok(());
+    }
+
+    let conditions: Vec<Value> = match (*context).get("and") {
```
ayushjain17, line 132: instead of doing this manually, can we leverage the functions from JsonLogic, if we don't have an exact function for this requirement, we can consider adding the logic over there, as that would be a more structured and generic way of doing this

PS: we are anyways using our own fork of jsonlogic

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,51 @@ pub fn validate_condition_with_functions(
...
+        let operators = condition_obj.keys();
+
+        for operator in operators {
+            let operands = condition_obj[operator].as_array().ok_or(superposition::AppError::BadArgument(
+                    "Failed to parse operands as an arrays. Ensure the context provided obeys the rules of JSON logic"
+                            .into()
+            ))?;
+
+            let (dimension_name, _) = get_variable_name_and_value(operands)?;
+            // variantIds use 'HAS', and since they are managed internally allow them to bypass strict mode
```
ayushjain17, line 156: but the following check does not explicitly check for that, it is just saying that for `variantIds` any operator is allowed

crates/frontend/src/components/experiment_form.rs
```diff
@@ -2,6 +2,7 @@ pub mod types;
 pub mod utils;
 
 use leptos::*;
+use superposition_types::database::models::Workspace;
```
ayushjain17, line 5: can we combine the imports

crates/frontend/src/pages/workspace.rs
```diff
@@ -101,6 +103,21 @@ pub fn workspace() -> impl IntoView {
...
+                .find(|w| w.workspace_name == workspace_name)
+            else {
+                enqueue_alert(
+                    "Invalid workspace selected, please contact an admin".into(),
+                    AlertType::Error,
+                    2000,
+                );
+                return;
+            };
+            provide_context(create_rw_signal(workspace_settings));
```
ayushjain17, line 120: @ShubhranshuSanjeev can you confirm if this is the correct way of doing this or not?
AFAIK, `provide_context` works in a hierarchal  way, only the children of the current component would get access to the value of this context, right ?

and over here, the entire route itself is getting changed, so won't this data not be there at all ?

ayushjain17 (changes requested):

crates/context_aware_config/src/api/context/helpers.rs
knutties: > should we rather consider storing the context as straight forward key value pair object, instead of jsonlogic?
making it implicit == everywhere and no other need ?

@ayushjain17 - we might do this - but maybe latter - too much of migration involved and closes the door on the other rule based engine approach too fast.  For now just dis-allow via setting while we can take time to figure out the right way.

crates/frontend/src/components/workspace_form.rs
Datron: @ayushjain17 I was thinking that we could do something like this in jsonlogic-rs, where we can a function to check if it obeys our requirements of strict mode. Similar for checking for only `and`

crates/frontend/src/pages/workspace.rs
ayushjain17: this is not going to work at all based on the above mentioned logic

ayushjain17: also, when I open any workspace's page directly this page would not be visited, so the provide context logic itself would not work

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,57 @@ pub fn validate_condition_with_functions(
...
+    strict_mode: bool,
+) -> superposition::Result<()> {
+    if !strict_mode {
+        return Ok(());
+    }
+
+    let conditions: Vec<Value> = match (*context).get("and") {
+        Some(conditions_json) => conditions_json
+            .as_array()
+            .ok_or(superposition::AppError::BadArgument("Error extracting dimensions, failed parsing conditions as an array. Ensure the context provided obeys the rules of JSON logic".into()))?
```
ayushjain17, line 135: instead of writing `superposition::AppError::BadArgument` in all the places, we can directly use the `bad_argument` macro

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,57 @@ pub fn validate_condition_with_functions(
...
+            let operands = condition_obj[operator].as_array().ok_or(superposition::AppError::BadArgument(
+                    "Failed to parse operands as an arrays. Ensure the context provided obeys the rules of JSON logic"
+                            .into()
+            ))?;
+
+            let (dimension_name, _) = get_variable_name_and_value(operands)?;
+            // variantIds use 'HAS', and since they are managed internally allow them to bypass strict mode
+            if dimension_name == "variantIds" && operator != "==" && operator != "in" {
+                return Err(superposition::AppError::BadArgument(format!(
+                    "Strict mode is enabled, and variantIds is a system dimension that supports IS and IN. Found operator: {}",
```
ayushjain17, line 159: `IS` is not possible for `variantIds`
also, this check currently allows `IS`, `IN` as well as `HAS`, but all that we need to allow is `HAS` only

ayushjain17 (changes requested):

ayushjain17:
also, needs to be rebased, there are conflicts

crates/context_aware_config/src/api/context/helpers.rs
Datron: We discussed this already, there is no operator called `has`, its just `in`. `IS` or `==` should be fine.

crates/frontend/src/pages/workspace.rs
Datron: Hmm do we add this to every single page then? Sounds not DRY.

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,57 @@ pub fn validate_condition_with_functions(
...
+                    "Strict mode is enabled, but the context contains unsupported operators. Please use IS operator only. Found operator: {}",
+                    operator
+                )));
+            }
+        }
+    }
+
+    Ok(())
+}
+
```
sauraww, line 174: Instead of writing if logic at two places , we can structure it better to reduce DRY : 

```
if (dimension_name == "variantIds" && operator != "==" && operator != "in")
    || (dimension_name != "variantIds" && operator != "==") 
{
    let error_message = if dimension_name == "variantIds" {
        format!(
            "Strict mode is enabled, and variantIds is a system dimension that supports IS and IN. Found operator: {}",
            operator
        )
    } else {
        format!(
            "Strict mode is enabled, but the context contains unsupported operators. Please use IS operator only. Found operator: {}",
            operator
        )
    };

    return Err(superposition::AppError::BadArgument(error_message));
}
```

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -118,6 +121,57 @@ pub fn validate_condition_with_functions(
...
+            let (dimension_name, _) = get_variable_name_and_value(operands)?;
+            // variantIds use 'HAS', and since they are managed internally allow them to bypass strict mode
+            if dimension_name == "variantIds" && operator != "==" && operator != "in" {
+                return Err(superposition::AppError::BadArgument(format!(
+                    "Strict mode is enabled, and variantIds is a system dimension that supports IS and IN. Found operator: {}",
+                    operator
+                )));
+            }
+            if dimension_name != "variantIds" && operator != "==" {
+                return Err(superposition::AppError::BadArgument(format!(
```
sauraww, line 164: We can use bad_argument macro here

sauraww (changes requested):

crates/context_aware_config/src/api/context/helpers.rs
Datron: I kept it separate for better readability

sauraww: In that case we can do this : 
```
let is_invalid_operator = match dimension_name {
    "variantIds" => operator != "==" && operator != "in",
    _ => operator != "=="
};

if is_invalid_operator {
    let error_message = match dimension_name {
        "variantIds" => format!(
            "Strict mode is enabled, and variantIds is a system dimension that supports IS and IN. Found operator: {}",
            operator
        ),
        _ => format!(
            "Strict mode is enabled, but the context contains unsupported operators. Please use IS operator only. Found operator: {}",
            operator
        )
    };

    return Err(superposition::AppError::BadArgument(error_message));
}
```

sauraww:
@Datron Can you rebase it ? We can merge it quickly .

Datron:
@sauraww there are no conflicts, we can merge

sauraww (approved):

crates/frontend/src/hoc/layout.rs
```diff
@@ -41,38 +47,70 @@ pub fn use_org() -> OrganisationId {
...
+        {move || {
+            let route_context = use_route();
+            let original_path = String::from(route_context.original_path());
+            let path = route_context.path();
+            if let Some(workspace) = workspaces.get() {
+                let tenant = tenant_rws.get().0;
+                let settings = workspace.data.into_iter().find(|w| w.workspace_name == tenant);
+                match settings {
+                    Some(settings) => {
+                        provide_context(StoredValue::new(settings));
```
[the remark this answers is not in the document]
Datron: How so?

ayushjain17 (approved):

PR #472, opened 2025-04-17
test: Experiment

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/superposition_types/src/api/experiments.rs
```diff
@@ -260,6 +260,7 @@ pub struct VariantUpdateRequest {
 
 #[derive(Deserialize, Serialize, Debug)]
 pub struct OverrideKeysUpdateRequest {
+    #[serde(alias = "variant_list")]
```
Datron, line 263: Why this change? Does it need a change in frontend as well?

Datron (approved):

PR #475, opened 2025-04-17
test: added pre-test setup

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


ShreyBana (approved):
Great!

makefile
```diff
@@ -169,16 +169,18 @@ run: kill db localstack frontend superposition
 run_legacy: kill build db localstack superposition_legacy
 	@./target/debug/superposition
 
-test: WASM_PACK_MODE=--profiling
+#test: WASM_PACK_MODE=--profiling
```
Datron, line 172: Can we uncomment this?

makefile
```diff
@@ -169,16 +169,18 @@ run: kill db localstack frontend superposition
...
 test: setup frontend superposition
 	cargo test
+	@cd target && ln -s ../node_modules node_modules || true
```
ayushjain17, line 175: can we please move this to the setup command instead, after `npm ci` is done, as this is needed for normal local development as well

ayushjain17 (approved):

makefile
ShubhranshuSanjeev: Done

ShreyBana (approved):

PR #476, opened 2025-04-21
feat: Add Pagination Params to frontend url

ayushjain17:
## Problem
Pagination params of each page are not tracked in the url of the pages

## Solution
Added pagination params support to frontend urls

### DIFF
--- a/crates/frontend/src/pages/function/function_list.rs
+++ b/crates/frontend/src/pages/function/function_list.rs
@@ -3,14 +3,18 @@ use leptos::*;
 use leptos_router::A;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Map, Value};
+use superposition_macros::box_params;
 use superposition_types::{
-    custom_query::PaginationParams, database::models::cac::Function, PaginatedResponse,
+    custom_query::{CustomQuery, PaginationParams, Query},
+    database::models::cac::Function,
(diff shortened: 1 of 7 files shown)


crates/frontend/src/pages/function/function_list.rs
```diff
@@ -24,7 +27,15 @@ struct CombinedResource {
 pub fn function_list() -> impl IntoView {
     let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
     let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
-    let (filters, set_filters) = create_signal(PaginationParams::default());
+    let (filters, set_filters) = use_signal_from_query(move |query_string| {
```
Datron, line 30:
```suggestion
    let (pagination_params_rs, pagination_params_ws) = use_signal_from_query(move |query_string| {
```
This but everywhere that (filters, set_filters) are used

Datron (approved):

crates/frontend/src/pages/function/function_list.rs
ayushjain17: I too wanted to do that, but didn't want to increase the change log, but sure, adding the change as well

ShubhranshuSanjeev (approved):

PR #481, opened 2025-04-23
feat: Update override using context id

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -504,43 +504,24 @@ async fn bulk_operations(
                         all_change_reasons.push(put_req.change_reason.clone());
                         response.push(ContextBulkResponse::Put(put_resp));
                     }
-                    ContextAction::Replace(put_req) => {
-                        let ctx_condition = put_req.context.to_owned().into_inner();
-                        let ctx_condition_value =
-                            Value::Object(ctx_condition.clone().into());
-                        let description = if put_req.description.is_none() {
-                            query_description(
-                                ctx_condition_value,
-                                transaction_conn,
-                                &schema_name,
-                            )?
-                        } else {
-                            put_req
-                                .description
-                                .clone()
-                                .expect("Description should not be empty")
-                        };
-
-                        let put_resp = operations::upsert(
-                            put_req.clone(),
-                            description.clone(),
+                    ContextAction::Replace(update_request) => {
+                        all_change_reasons
+                            .push((*update_request.change_reason).to_string());
+                        let update_resp = operations::update(
+                            update_request,
                             transaction_conn,
-                            true,
                             &user,
                             &schema_name,
-                            true,
                         )
                         .map_err(|err| {
                             log::error!(
-                                "Failed at insert into contexts due to {:?}",
+                                "Failed at update into contexts due to {:?}",
                                 err
                             );
                             err
                         })?;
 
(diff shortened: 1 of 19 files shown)


crates/superposition_types/src/api/context.rs
```diff
@@ -72,3 +76,18 @@ impl Display for ContextListFilters {
         write!(f, "{}", parts.join("&"))
     }
 }
+
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+pub enum Identifier {
```
pratikmishra356, line 81: can we add snake_case rename 
and change Context -> Condition

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -515,43 +504,24 @@ async fn bulk_operations(
                         all_change_reasons.push(put_req.change_reason.clone());
                         response.push(ContextBulkResponse::Put(put_resp));
                     }
-                    ContextAction::Replace(put_req) => {
```
pratikmishra356, line 518: can we change enum to  ContextAction::Update
Replace now seems misleading

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -131,40 +131,29 @@ async fn put_handler(
...
             .map_err(|err: superposition::AppError| {
-                log::info!("context put failed with error: {:?}", err);
+                log::info!("context update failed with error: {:?}", err);
                 err
             })?;
             let version_id = add_config_version(
                 &state,
                 tags,
-                req_change_reason.clone(),
+                req_change_reason.deref().to_string(),
```
pratikmishra356, line 156: why deref needed here ?

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -30,7 +33,7 @@ pub struct ContextPutResp {
 #[derive(Serialize, Deserialize, Debug)]
 pub enum ContextBulkResponse {
     PUT(ContextPutResp),
-    REPLACE(ContextPutResp),
+    REPLACE(Context),
```
pratikmishra356, line 36: can we also change this to pdate ?

crates/context_aware_config/src/api/context/handlers.rs
ayushjain17: this was actually doing a replace itself of the overrides being provided, it does not have a update the values of those keys which are being sent kind of behaviour, so rather having it still named as `Replace` helps us avoid the confusion of this kind of behaviour

pratikmishra356 (approved):

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -1,10 +1,31 @@
...
+    },
     Cac, Condition, Overrides,
 };
 
+#[derive(Serialize, AsChangeset)]
+#[diesel(table_name = contexts)]
+pub(crate) struct UpdateContextOverridesChangeset {
+    #[serde(rename = "override")]
+    pub override_id: String,
+    pub override_: Overrides,
```
Datron, line 22:
```suggestion
    pub override_id: String,
     #[serde(rename = "override")]
    pub override_: Overrides,
```

Datron (changes requested):

Datron (approved):

PR #482, opened 2025-04-23
fix delete dimension bug

pratikmishra356:
## Problem
Delete dimension is not working and it gives , issue was in constraint instead of initially immediate
it should be initially deferred.

also without deferrable constraints create/update/delete apis not working , so deferrable is necessary

Constraint Type | Can Be Deferred? | Enforced When? | Behavior in Your Case
-- | -- | -- | --
UNIQUE (position) (non-deferrable) | ❌ No | Always immediately | ❌ Fails due to temporary duplicates
UNIQUE (position) DEFERRABLE INITIALLY IMMEDIATE | ✅ Yes | By default, immediately  (but deferrable by planner if safe) | ✅ Works, because PG planner handles it smartly

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/dimension/handlers.rs
+++ b/crates/context_aware_config/src/api/dimension/handlers.rs
@@ -357,15 +357,18 @@ async fn delete_dimension(
                 .returning(Dimension::as_returning())
                 .schema_name(&schema_name)
                 .execute(transaction_conn)?;
+
+            let deleted_row = delete(dsl::dimensions.filter(dsl::dimension.eq(&name)))
+                .schema_name(&schema_name)
+                .execute(transaction_conn);
+
             diesel::update(dimensions::dsl::dimensions)
                 .filter(dimensions::position.gt(dimension_data.position))
                 .set(dimensions::position.eq(dimensions::position - 1))
                 .returning(Dimension::as_returning())
                 .schema_name(&schema_name)
                 .execute(transaction_conn)?;
-            let deleted_row = delete(dsl::dimensions.filter(dsl::dimension.eq(&name)))
-                .schema_name(&schema_name)
-                .execute(transaction_conn);
+
             match deleted_row {
                 Ok(0) => Err(not_found!("Dimension `{}` doesn't exists", name)),
                 Ok(_) => Ok(HttpResponse::NoContent().finish()),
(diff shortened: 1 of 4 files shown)


crates/superposition_types/migrations/2024-12-05-172909_dimension_priority_default/up.sql
```diff
@@ -3,6 +3,4 @@ ALTER TABLE public.dimensions
 ALTER COLUMN priority SET DEFAULT 1;
 
 ALTER TABLE public.dimensions
-ADD CONSTRAINT dimension_unique_position UNIQUE (position) DEFERRABLE INITIALLY IMMEDIATE;
-
-SET CONSTRAINTS dimension_unique_position DEFERRED;
\ No newline at end of file
+ADD CONSTRAINT dimension_unique_position UNIQUE (position);
```
mahatoankitkumar, line 6: Shouldn't we create a new migration for this?

docker-compose/postgres/db_init.sql
```diff
@@ -469,9 +469,7 @@ ALTER TABLE localorg_test.dimensions
 ALTER COLUMN priority SET DEFAULT 1;
 
 ALTER TABLE localorg_test.dimensions
-ADD CONSTRAINT dimension_unique_position UNIQUE (position) DEFERRABLE INITIALLY IMMEDIATE;
-
-SET CONSTRAINTS localorg_test.dimension_unique_position DEFERRED;
+ADD CONSTRAINT dimension_unique_position UNIQUE (position) DEFERRABLE INITIALLY DEFERRED;
```
ayushjain17, line 472: please modify it for `localord_dev` as well

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -357,15 +357,18 @@ async fn delete_dimension(
                 .returning(Dimension::as_returning())
                 .schema_name(&schema_name)
                 .execute(transaction_conn)?;
+
+            let deleted_row = delete(dsl::dimensions.filter(dsl::dimension.eq(&name)))
```
ayushjain17, line 361: while we are at it, should we consider changing `position + 100` logic of update dimension to better solution ?

ayushjain17 (changes requested):

ayushjain17 (approved):

mahatoankitkumar (approved):

PR #486, opened 2025-04-24
fix: added validation for and wrapper over context json logic

ShubhranshuSanjeev:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/superposition_types/src/config.rs
```diff
@@ -79,29 +80,53 @@ impl_try_from_map!(Exp, Overrides, Overrides::validate_data);
...
     }
 
     fn validate_data_for_exp(condition_map: Map<String, Value>) -> Result<Self, String> {
         let condition_val = json!(condition_map);
+        logic_validation::validate(&condition_val, &EXP_CONDITION_VALIDATION_CONFIG)
+            .map_err(|err| {
+                let msg = format!("Condition validation error: {}", err.message);
+                log::error!("{}", msg);
+                msg
+            })?;
```
ayushjain17, line 120: last we discussed, there were some changes needed for this in experiments handlers also, right, for everything to remain working correctly ?

PR #489, opened 2025-04-28
fix: Empty check for keys of struct

ayushjain17:
## Problem
When reading params from url and no params are being provided in frontend, since all keys of the filter being `Option` type leads to the type getting successfully created with all as `None` values.
This leads to the default value not being used as the extract was a success.
For experiments, default time range is last `30days` in the type and last `24hours` in the backend, so when no params are sent to frontend, it is not going to the type default. Thus `experiments` of last 24 hours are being shown.

## Solution

1. Added `extract_non_empty` function to extract query and fallback to default if all keys are `None`
2. Added `IsEmpty` proc_macro_derive to auto derive `IsEmpty` trait for structs having `Named` fields
3. Added check in `use_param_updater` to consider `no params` and `default params` generated via default as same

### DIFF
--- a/crates/frontend/src/query_updater.rs
+++ b/crates/frontend/src/query_updater.rs
@@ -1,26 +1,37 @@
 use std::fmt::Display;
 
 use leptos::*;
-use leptos_router::{use_location, use_navigate, use_query_map};
+use leptos_router::{use_location, use_navigate, use_query_map, NavigateOptions};
 
 use crate::utils::use_service_prefix;
 
-pub fn use_param_updater(source: impl Fn() -> Vec<Box<dyn Display>> + 'static) {
+pub trait DisplayDefault: Display {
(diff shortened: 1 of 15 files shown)


ShreyBana (approved):

crates/frontend/src/query_updater.rs
```diff
@@ -1,26 +1,37 @@
...
-pub fn use_param_updater(source: impl Fn() -> Vec<Box<dyn Display>> + 'static) {
+pub trait DisplayDefault: Display {
+    fn default(&self) -> String;
+}
+
+impl<T: Display + Default> DisplayDefault for T {
+    fn default(&self) -> String {
+        T::default().to_string()
+    }
+}
```
Datron, line 16: Do we need this? Seems superfluous

Datron (approved):

PR #490, opened 2025-04-28
fix: Make last_modified non-mandatory in experiments/list and fix UI …

ayushjain17:
## Problem
filter based on time range was mandatory params in experiment/list API via `from_date` and `to_date` params, which even if not provided was falling back to go time range of last 24 hours

## Solution
1. filter on time range only if either of `from_date` or `to_date` are being used
2. updated default value of `ExperimentListFilters` in accordance to the above change
3. fix behaviour of `date input` when clicked on `clear`
4. removed redundant `get` and `set` calls of signal by using `update`

### DIFF
--- a/crates/frontend/src/pages/experiment_list.rs
+++ b/crates/frontend/src/pages/experiment_list.rs
@@ -2,7 +2,7 @@ pub mod utils;
 
 use std::collections::HashSet;
 
-use chrono::{prelude::Utc, DateTime};
+use chrono::{prelude::Utc, DateTime, Days, Duration};
 use futures::join;
 use leptos::*;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


sauraww (approved):

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -253,17 +250,16 @@ fn experiment_table_filter_widget(
                             type="text"
                             id="experiment-id-filter"
                             class="input input-bordered rounded-md resize-y w-full max-w-md"
-                            value=move || filters_rws.get().experiment_ids.map(|d| d.to_string())
+                            value=move || {
+                                filters_buffer_rws.get().experiment_ids.map(|d| d.to_string())
+                            }
```
Datron, line 255: auto formatter?

Datron (approved):

PR #495, opened 2025-05-05
feat: introduce compare UI

Datron:
## Problem
It is difficult to view differences between multiple contexts and compare them

## Solution

Introduce a compare UI to see diffs

https://github.com/user-attachments/assets/fa616921-abac-4235-9c8a-5487e76cc64f

### DIFF
--- a/crates/frontend/src/components/table/types.rs
+++ b/crates/frontend/src/components/table/types.rs
@@ -5,6 +5,7 @@ use serde_json::{Map, Value};
 use superposition_types::SortBy;
 
 pub type CellFormatter = Box<Rc<dyn Fn(&str, &Map<String, Value>) -> View>>;
+pub type ColumnFormatter = Box<Rc<dyn Fn(&str) -> View>>;
 
 #[derive(Clone, Debug)]
 pub struct TableSettings {
@@ -34,6 +35,7 @@ pub struct Column {
     pub formatter: CellFormatter,
     pub sortable: ColumnSortable,
     pub expandable: Expandable,
+    pub column_formatter: ColumnFormatter,
 }
 
 impl PartialEq for Column {
@@ -42,8 +44,13 @@ impl PartialEq for Column {
     }
 }
 
-fn default_formatter(value: &str, _row: &Map<String, Value>) -> View {
+pub fn default_formatter(value: &str, _row: &Map<String, Value>) -> View {
     view! { <span>{value.to_string()}</span> }.into_view()
+}
+
+pub fn default_column_formatter(value: &str) -> View {
+    let column_name = value.replace('_', " ");
+    view! { <span>{column_name}</span> }.into_view()
 }
 
 impl Column {
@@ -54,6 +61,7 @@ impl Column {
             formatter: Box::new(Rc::new(default_formatter)),
             sortable: ColumnSortable::No,
             expandable: Expandable::Enabled(100),
+            column_formatter: Box::new(Rc::new(default_column_formatter)),
         }
     }
     pub fn default_no_collapse(name: String) -> Column {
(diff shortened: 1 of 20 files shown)


crates/frontend/src/pages/compare_overrides.rs
```diff
@@ -0,0 +1,218 @@
...
+                        </div>
+                        <Drawer
+                            id="add_comparison_drawer".to_string()
+                            header="Add a context to compare"
+                            handle_close=move || {
+                                close_drawer("add_comparison_drawer");
+                            }
+                        >
+
+                            <EditorProvider>
```
ayushjain17, line 182: this is not needed here, we are not using Monaco anywhere in this page

crates/frontend/src/pages/compare_overrides.rs
```diff
@@ -0,0 +1,218 @@
...
+                    let resolved_config_map = resolved_config_resource.get().unwrap_or_default();
+                    let table_columns = table_columns(contexts_vector_rws);
+                    let dimensions = dimension_resource.get().unwrap_or_default();
+                    let data = resolved_config_map.into_values().collect();
+                    view! {
+                        <div class="card rounded-xl w-full bg-base-100 shadow">
+                            <div class="card-body">
+                                <div class="flex justify-between">
+                                    <h2 class="card-title">Compare Overrides</h2>
+                                    <div>
```
ayushjain17, line 160: can we please avoid this div nesting here

crates/frontend/src/pages/compare_overrides.rs
```diff
@@ -0,0 +1,218 @@
...
+                                row_vector.insert(context.clone(), resolved_value);
+                                continue;
+                            }
+                            let mut row_vector = Map::new();
+                            row_vector.insert(
+                                "config_key".into(),
+                                Value::String(config_key.clone()),
+                            );
+                            row_vector.insert(context.clone(), resolved_value);
+                            contexts_config_vector_map.insert(config_key, row_vector);
```
ayushjain17, line 124: simply doing this would also work, right ?
```suggestion
                            let mut row_vector = contexts_config_vector_map.get_mut(&config_key).unwrap_or_default();
                            row_vector.insert(
                                "config_key".into(),
                                Value::String(config_key.clone()),
                            );
                            row_vector.insert(context.clone(), resolved_value);
                            contexts_config_vector_map.insert(config_key, row_vector);
```

ayushjain17: in that case this would work, right ?
we just dont use the `Default` trait
```suggestion
                            let mut row_vector = contexts_config_vector_map.get_mut(&config_key).unwrap_or(&Map::new());
                            row_vector.insert(
                                "config_key".into(),
                                Value::String(config_key.clone()),
                            );
                            row_vector.insert(context.clone(), resolved_value);
                            contexts_config_vector_map.insert(config_key, row_vector);
```

crates/frontend/src/pages/compare_overrides.rs
ayushjain17: but we are restricting operator to `IS` only, right ?
we can ignore this though

crates/frontend/src/components/table/types.rs
```diff
@@ -42,10 +44,15 @@ impl PartialEq for Column {
     }
 }
 
-fn default_formatter(value: &str, _row: &Map<String, Value>) -> View {
+pub fn default_formatter(value: &str, _row: &Map<String, Value>) -> View {
     view! { <span>{value.to_string()}</span> }.into_view()
 }
 
+pub fn default_column_formatter(value: &str) -> View {
```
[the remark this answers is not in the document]
ayushjain17: fair enough

ayushjain17 (approved):

mahatoankitkumar (approved):

mahatoankitkumar (changes requested):
Use Condition pill
Use hashset to store conditions instead of vector

ayushjain17 (approved):

PR #496, opened 2025-05-05
feat: webhooks ui

mahatoankitkumar:
https://github.com/user-attachments/assets/2879f8d8-d707-4492-bceb-58f59be902f1

### DIFF
--- a/crates/frontend/src/components/webhook_form.rs
+++ b/crates/frontend/src/components/webhook_form.rs
@@ -1,0 +1,292 @@
+use leptos::*;
+use serde_json::Value;
+use strum::IntoEnumIterator;
+use superposition_types::database::models::others::{
+    HttpMethod, PayloadVersion, WebhookEvent,
+};
+use web_sys::MouseEvent;
+
+use crate::{
+    api::{create_webhook, update_webhook},
+    components::{
+        alert::AlertType,
+        button::Button,
+        change_form::ChangeForm,
+        dropdown::{Dropdown, DropdownBtnType, DropdownDirection},
+        input::{Input, InputType, Toggle},
+    },
+    providers::{alert_provider::enqueue_alert, editor_provider::EditorProvider},
+    schema::{JsonSchemaType, SchemaType::Single},
+    types::{OrganisationId, Tenant},
+};
+
+#[component]
+pub fn webhook_form<NF>(
+    #[prop(default = false)] edit: bool,
+    #[prop(default = String::new())] webhook_name: String,
+    #[prop(default = String::new())] description: String,
+    #[prop(default = false)] enabled: bool,
+    #[prop(default = String::new())] url: String,
+    #[prop(default = HttpMethod::default())] method: HttpMethod,
+    #[prop(default = PayloadVersion::default())] payload_version: PayloadVersion,
+    #[prop(default = serde_json::json!({}))] custom_headers: Value,
+    #[prop(default = Vec::new())] events: Vec<WebhookEvent>,
+    handle_submit: NF,
+) -> impl IntoView
+where
+    NF: Fn() + 'static + Clone,
+{
+    let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
+    let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
+
+    let (webhook_name_rs, webhook_name_ws) = create_signal(webhook_name);
+    let (description_rs, description_ws) = create_signal(description);
+    let (enabled_rs, enabled_ws) = create_signal(enabled);
+    let (url_rs, url_ws) = create_signal(url);
+    let (method_rs, method_ws) = create_signal(method);
+    let (payload_version_rs, payload_version_ws) = create_signal(payload_version);
+    let (custom_headers_rs, custom_headers_ws) = create_signal(custom_headers);
+    let (change_reason_rs, change_reason_ws) = create_signal(String::new());
+    let (events_rs, events_ws) = create_signal(events);
+    let (req_inprogess_rs, req_inprogress_ws) = create_signal(false);
+
+    let handle_select_webhook_event_dropdown_option =
+        Callback::new(move |selected_event: WebhookEvent| {
+            events_ws.update(|value| value.push(selected_event));
+        });
+
+    let handle_remove_webhook_event_dropdown_option =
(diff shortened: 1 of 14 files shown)


crates/frontend/src/pages/webhooks.rs
```diff
@@ -0,0 +1,298 @@
...
+        };
+        vec![
+            Column::default_no_collapse("name".to_string()),
+            Column::default("description".to_string()),
+            Column::default("enabled".to_string()),
+            Column::default("url".to_string()),
+            Column::default("method".to_string()),
+            Column::default("payload_version".to_string()),
+            Column::default("custom_headers".to_string()),
+            Column::default("events".to_string()),
```
Datron, line 168: nitpick: Can you format the events to be badge component?

crates/frontend/src/pages/webhooks.rs
```diff
@@ -0,0 +1,298 @@
...
+            Column::default("max_retries".to_string()),
+            Column::default("last_triggered_at".to_string()),
+            Column::default("change_reason".to_string()),
```
Datron, line 171: Change reason and description should be a modal popup, since they can be very large and won't work well in a table

crates/frontend/src/pages/webhooks.rs
```diff
@@ -0,0 +1,298 @@
...
+                        <i class="ri-delete-bin-5-line ri-xl text-red-500"></i>
+                    </span>
+                </div>
+            }
+            .into_view()
+        };
+        vec![
+            Column::default_no_collapse("name".to_string()),
+            Column::default("description".to_string()),
+            Column::default("enabled".to_string()),
```
Datron, line 163: nitpick: Instead of showing this flag, can we grey out or show a red dot for disabled wehooks next to their name?

crates/frontend/src/components/webhook_form.rs
```diff
@@ -0,0 +1,338 @@
...
+                                    .unwrap_or_default()
+                                on_change=Callback::new(move |value: String| {
+                                    logging::log!(
+                                        "Metrics Editor Value: {} {:#?}", value, serde_json::from_str::<Value>(&value)
+                                    );
+                                    custom_headers_ws
+                                        .set(serde_json::from_str(&value).unwrap_or_default())
+                                })
+                                classes=vec!["w-[400px]", "h-[100px]"]
+                            />
```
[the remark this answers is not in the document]
mahatoankitkumar: It requires schema to be passed.

crates/frontend/src/components/webhook_form.rs
```diff
@@ -0,0 +1,338 @@
...
+                    on:change=move |ev| {
+                        let value = event_target_value(&ev);
+                        url_ws.set(value);
+                    }
+                >
+                    {url_rs.get()}
+                </textarea>
+            </div>
+
+            {move || {
```
ayushjain17, line 216: we can avoid move in all of these 3 dropdowns

crates/frontend/src/pages/webhooks.rs
```diff
@@ -0,0 +1,298 @@
...
+                            <Drawer
+                                id="webhook_drawer".to_string()
+                                header="Create New Webhook"
+                                handle_close=handle_close
+                            >
+                                <WebhookForm handle_submit=move || {
+                                    webhooks_resource.refetch();
+                                    selected_webhook.set(None);
+                                    close_drawer("webhook_drawer");
+                                } />
```
ayushjain17, line 233:
```suggestion
                                <WebhookForm handle_submit=move || {
                                    pagination_params_rws.update(|f| f.reset_page());
                                    webhooks_resource.refetch();
                                    selected_webhook.set(None);
                                    close_drawer("webhook_drawer");
                                } />
```

ayushjain17 (approved):

crates/frontend/src/components/webhook_form.rs
```diff
@@ -0,0 +1,297 @@
...
+                        data=serde_json::to_string_pretty(&custom_headers_rs.get_untracked())
+                            .unwrap_or_default()
+                        on_change=Callback::new(move |value: String| {
+                            logging::log!(
+                                "Metrics Editor Value: {} {:#?}", value, serde_json::from_str::<Value>(&value)
+                            );
+                            custom_headers_ws.set(serde_json::from_str(&value).unwrap_or_default())
+                        })
+                        classes=vec!["w-[400px]", "h-[100px]"]
+                    />
```
Datron, line 276: Can we replace this with <Input>. Also the node ID and log say metrics editor

crates/frontend/src/components/webhook_form.rs
```diff
@@ -0,0 +1,297 @@
...
+                <textarea
+                    placeholder="Enter the webhook URL"
+                    class="textarea textarea-bordered w-full max-w-md"
+                    on:change=move |ev| {
+                        let value = event_target_value(&ev);
+                        url_ws.set(value);
+                    }
+                >
+                    {url_rs.get_untracked()}
+                </textarea>
```
Datron, line 204: This can also be Input component

crates/frontend/src/components/webhook_form.rs
```diff
@@ -0,0 +1,297 @@
...
+    let (description_rs, description_ws) = create_signal(description);
+    let (enabled_rs, enabled_ws) = create_signal(enabled);
+    let (url_rs, url_ws) = create_signal(url);
+    let (method_rs, method_ws) = create_signal(method);
+    let (payload_version_rs, payload_version_ws) = create_signal(payload_version);
+    let (custom_headers_rs, custom_headers_ws) = create_signal(custom_headers);
+    let (change_reason_rs, change_reason_ws) = create_signal(String::new());
+    let (events_rs, events_ws) = create_signal(events);
+    let (req_inprogess_rs, req_inprogress_ws) = create_signal(false);
+    let loaded_in_csr_rws = RwSignal::new(false);
```
Datron, line 52: Why this variable? Should we use create_effect instead?

ayushjain17: @Datron `create_effect` is also being used, right in the next line
we need to check if is client side rendering or not, so this is needed
there is monaco code which cannot be executed in ssr and cfg if was not possible here

if we see more such use case in future, would probably make a proper construct then, just like `use_param_updater` to auto update query params

Datron (approved):

PR #500, opened 2025-05-06
feat: delete override via experiment

pratikmishra356:
## Problem
Delete overrides via experiment

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/superposition_types/migrations/2025-05-06-064904_add_experiment_type/up.sql/down.sql
ayushjain17, line 1: please add the proper down queries

crates/superposition_types/migrations/2025-05-06-064904_add_experiment_type/up.sql/up.sql
ayushjain17, line 1: can we format this document

crates/superposition_types/src/api/experiments.rs
```diff
@@ -71,6 +72,7 @@ pub struct ExperimentCreateRequest {
     pub name: String,
     pub context: Exp<Condition>,
     pub variants: Vec<Variant>,
+    pub experiment_type: ExperimentType,
```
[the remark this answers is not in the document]
ayushjain17: smithy model also needs to be updated

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -68,8 +68,8 @@ impl ExperimentStatusType {
 
     pub fn discardable(&self) -> bool {
         match self {
-            Self::CREATED => true,
-            Self::INPROGRESS | Self::CONCLUDED | Self::DISCARDED => false,
+            Self::CREATED | Self::INPROGRESS => true,
```
ayushjain17, line 71: frontend change is also needed for this

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -118,6 +180,18 @@ pub fn check_variants_override_coverage(
...
+    variant_delete_overrides: &[Option<Vec<String>>],
+    override_keys: &[String],
+) -> bool {
+    for variant_override in variant_delete_overrides.iter().flatten() {
+        if !check_variant_delete_override_coverage(variant_override, override_keys) {
+            return false;
+        }
+    }
+    true
+}
```
ayushjain17, line 193: having this check forces that there can technically be only 2 Variants, 1 Control and 1 Exp, if you try to have multiple Exp variants, then this check forces that all of the Exp variants to be exactly same, thus only 2 Variants, 1 Control and 1 Exp

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -187,9 +140,86 @@ async fn create(
...
+        };
+        config.contexts = above_context;
+        let exp_context_dimension_value = extract_dimensions(&exp_context)?;
+        let evaled_cac = eval_cac(
+            config.default_configs.to_owned(),
+            &config.contexts,
+            &config.overrides,
+            &exp_context_dimension_value,
+            MergeStrategy::MERGE,
+        )
```
ayushjain17, line 169: why not call api for resolve, why do we need to get the raw config ?

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17, line 1: review not completed

crates/experimentation_platform/src/api/experiments/cac_api.rs
ayushjain17, line 1: review pending

ayushjain17 (changes requested):

crates/experimentation_platform/src/api/experiments/helpers.rs
pratikmishra356: this function is checking in all the variants the keys that needs to be deleted should exist in the actual overrides

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -83,6 +83,42 @@ impl ExperimentStatusType {
...
+    derive(diesel_derive_enum::DbEnum, QueryId)
+)]
+#[cfg_attr(feature = "diesel_derives", DbValueStyle = "UPPERCASE")]
+#[cfg_attr(
+    feature = "diesel_derives",
+    ExistingTypePath = "crate::database::schema::sql_types::ExperimentType"
+)]
+#[allow(non_camel_case_types)]
+pub enum ExperimentType {
+    DEFAULT,
```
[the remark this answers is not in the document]
pratikmishra356: please suggest

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -192,9 +144,86 @@ async fn create(
...
+                && variant.variant_type == VariantType::CONTROL
+            {
+                return Err(bad_argument!(
+                    "Control variant should not contain delete keys",
+                ));
+            }
+            if let Some(delete_keys) = &variant.overrides_delete_keys {
+                for key in delete_keys {
+                    let mut variant_override: Map<String, Value> =
+                        variant.overrides.clone().into_inner().into();
```
ayushjain17, line 207: shouldn't this be outside the for loop

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -192,9 +144,86 @@ async fn create(
...
+
+                    if let Some(variant_config) = evaled_cac.get(key) {
+                        if variant_override.get(key).is_some() {
+                            variant_override
+                                .insert(key.clone(), variant_config.to_owned());
+                        }
```
ayushjain17, line 213: this check does not make sense

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -473,12 +505,44 @@ pub async fn conclude(
...
+                        workspace_request,
+                        exp_context_id.clone(),
+                    )
+                    .await?;
+                    let mut context_override: Map<String, Value> =
+                        current_context.override_.into();
+
+                    for key in variant.overrides_delete_keys.unwrap_or(vec![]) {
+                        context_override.remove(&key);
+                    }
```
ayushjain17, line 521: before doing this, you need to verify if it control variant or not, i am not sure if relying on the check that for `overrides_delete_keys` is always going to be empty a good enough check or not

ayushjain17 (changes requested):

crates/superposition_types/migrations/2025-05-06-064904_add_experiment_type/up.sql/down.sql
ayushjain17, line 1: please add relevant down queries
also, file path: `/up.sql/down.sql`

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+        .iter()
+        .map(|variant| variant.overrides_delete_keys.clone())
+        .collect();
+
+    let are_valid_variants = validate_variants_delete_keys(
+        &variant_delete_overrides,
+        &current_context
+            .override_
+            .keys()
+            .map(|x| x.to_string())
```
ayushjain17, line 300:
```suggestion
            .cloned()
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+        exp_context,
+        context_id,
+        workspace_request,
+    )
+    .await?;
+
+    let variant_delete_overrides: Vec<Option<Vec<String>>> = variants
+        .iter()
+        .map(|variant| variant.overrides_delete_keys.clone())
+        .collect();
```
ayushjain17, line 293:
```suggestion
    let variant_delete_overrides: Vec<Vec<String>> = variants
        .iter()
        .filter_map(|variant| variant.overrides_delete_keys.clone())
        .collect();
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+            }
+        }
+    }
+    true
+}
+
+pub fn validate_variant_delete_override_value(
+    variant_override: &Overrides,
+    delete_override_keys: Vec<String>,
+    config_value: &Map<String, Value>,
```
ayushjain17, line 206:
```suggestion
    resolved_config: &Map<String, Value>,
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+        } else {
+            return false;
+        }
+    }
+    true
+}
+
+pub fn validate_variants_delete_override_value(
+    variants: &Vec<Variant>,
+    config_value: &Map<String, Value>,
```
ayushjain17, line 222:
```suggestion
    resolved_config: &Map<String, Value>,
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+    exp_context: &Condition,
+    context_id: &str,
+    workspace_request: &WorkspaceContext,
+    variants: &Vec<Variant>,
+) -> superposition::Result<()> {
+    let current_context =
+        get_context_override(user, state, workspace_request, context_id.to_owned())
+            .await?;
+
+    let evaled_cac = get_partial_resolve_config(
```
ayushjain17, line 281:
```suggestion
    let partial_resolved_config = get_partial_resolve_config(
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+    );
+    if !are_valid_variants {
+        return Err(bad_argument!(
+                "Variant delete keys are not valid. Ensure the keys are present in the context"
+            ));
+    }
+
+    if !(validate_variants_delete_override_value(variants, &evaled_cac)) {
+        return Err(bad_argument!("Variant delete key's value are not valid"));
+    }
```
ayushjain17, line 311: this check allows me to create an experimental variant with no delete keys, should we allow

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+        } else {
+            return false;
+        }
+    }
+    true
+}
+
+pub fn validate_variants_control_override_value(
+    variants: &Vec<Variant>,
+    config_value: &Map<String, Value>,
```
ayushjain17, line 255:
```suggestion
    context_overrides: &Map<String, Value>,
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+            return false;
+        }
+    }
+    true
+}
+
+pub fn validate_variant_control_override_value(
+    variant_override: &Overrides,
+    delete_override_keys: Vec<String>,
+    config_value: &Map<String, Value>,
```
ayushjain17, line 239:
```suggestion
    context_overrides: &Map<String, Value>,
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,136 @@ pub fn check_variants_override_coverage(
...
+) -> bool {
+    for (override_key, value) in variant_override.iter() {
+        if let Some(val) = config_value.get(override_key) {
+            if !(delete_override_keys.contains(override_key)) && val != value {
+                return false;
+            }
```
ayushjain17, line 245: is this check correct ?
should delete list be at all considered here ?

instead of iterating over all the variants, shouldn't we only pick the CONTROL variant directly and check if `val != value`, as we only allow one CONTROL variant

ayushjain17: not sure if we should also add the check that the delete list should be exclusively empty for this

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: by doing this, essentially on the CAC side, concluding with control would become a NO_OP change

or alternatively we could keep the current code, with a check on creation that the delete list is always kept empty, but the caveat for that is that on concluding with control, it would not delete any keys, but it has a potential to modify the data on CAC side as, if the overrides for might be stale as compared to the current actual values for the context's overrides

on a second thought, should we rather consider making concluding with CONTROL, as an actual NO_OP change in CAC side (other than just deleting the overrides for the variants, as the exp is over now)

ayushjain17 (changes requested):

crates/experimentation_platform/src/api/experiments/handlers.rs
pratikmishra356: question1 : should we validate here that if it is a control variant or not and if it is so then its override_delete_keys should be empty
Answer : although during creation or update it should be checked that control variant should have empty in override_delete keys , we can add it here but it will be redundant

question2? Currently I am taking the current state of overrides and updating the value accordingly , now should we consider current state or the variants value as the source of truth
Answer : this needs discussion @knutties what should be done here ?

crates/experimentation_platform/src/api/experiments/helpers.rs
pratikmishra356: Yes , its not for control variant , 
its for all the variants which have keys that does not need to be deleted 
those keys value should be same as current state of override value

crates/experimentation_platform/src/api/experiments/helpers.rs
pratikmishra356: yes

crates/experimentation_platform/src/api/experiments/helpers.rs
pratikmishra356: changed the logic added some validation

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +189,146 @@ pub fn check_variants_override_coverage(
...
+) -> superposition::Result<()> {
+    let current_context =
+        get_context_override(user, state, workspace_request, context_id.to_owned())
+            .await?;
+
+    let evaled_cac = get_partial_resolve_config(
+        user,
+        state,
+        exp_context,
+        context_id,
```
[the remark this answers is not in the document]
pratikmishra356: lets discuss what how this should go in query params

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -745,6 +746,86 @@ async fn get_config(
...
+    config.contexts = if let Some(index) =
+        config.contexts.iter().position(|ctx| ctx.id == context_id)
+    {
+        config.contexts[..index].to_vec()
+    } else {
+        return Err(bad_argument!(
+            "context with id {} not found in CAC",
+            context_id
+        ));
+    };
```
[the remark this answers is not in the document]
pratikmishra356: show reasoning added , another thing which we removed is not modified response ,
we could club both the apis but not doing as of now

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -745,6 +746,86 @@ async fn get_config(
...
+
+    let context_id = if let Some(context_id) = context.get("context_id") {
+        context_id
+            .as_str()
+            .ok_or_else(|| bad_argument!("context_id is not a string"))?
+            .to_string()
+    } else {
+        return Err(bad_argument!("context_id is not present"));
+    };
```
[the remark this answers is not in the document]
pratikmishra356: let me know if we want to do this

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +186,140 @@ pub fn check_variants_override_coverage(
...
+        if let Some(val) = current_context_overrides.get(override_key) {
+            if val != value {
+                return false;
+            }
+        } else {
+            return false;
+        }
+    }
+    true
+}
```
ayushjain17, line 246:
```suggestion
fn validate_keys_from_source(
    variant_overrides: &Overrides,
    source: &Map<String, Value>,
) -> bool {
    for (key, value) in variant_overrides.iter() {
        if let Some(val) = source.get(key) {
            if val != value {
                return false;
            }
        } else {
            return false;
        }
    }
    true
}

pub fn validate_variants_delete_override_value(
    delete_variant_overrides: &Vec<Overrides>,
    resolved_config: &Map<String, Value>,
) -> bool {
    for override_ in delete_variant_overrides {
        if !validate_keys_from_source(override_, resolved_config) {
            return false;
        }
    }
    true
}

pub fn validate_variants_control_override_value(
    control_variant_overrides: &Overrides,
    current_context_overrides: &Map<String, Value>,
) -> bool {
    validate_keys_from_source(control_variant_overrides, current_context_overrides)
}
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -134,6 +141,141 @@ pub fn check_variants_override_coverage(
...
+            delete_keys.insert(key);
+        }
+    }
+    if delete_keys.len() != control_variant_overrides.len() {
+        return false;
+    }
+    true
+}
+
+pub fn validate_keys_from_source(
```
ayushjain17, line 169: not a blcoker
shouldn't all of these helper validate function not be public

clients/generated/smithy/typescript/.gitignore
ayushjain17, line 1: this should not have been deleted

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -448,6 +418,7 @@ pub async fn conclude(
         .schema_name(&workspace_request.schema_name)
         .get_result::<Experiment>(conn)?;
 
+    let exp_context_id = hash(&Value::Object(experiment.context.clone().into()));
```
ayushjain17, line 421: a big edge case has been missed here, for actions ramp(including both start and normal ramp) and conclude, the API need to check if the context still exists or not

we should not allow these experiments to continue on contexts which do not exist anymore, a probably a proper error should be sent saying that `the context does not exist anymore` and only discard is allowed

or should we auto discard seeing this situation ?

crates/experimentation_platform/src/api/experiments/cac_api.rs
```diff
@@ -0,0 +1,228 @@
...
+        match err {
+            superposition::AppError::ResponseError(val) => {
+                let mut msg: String =
+                    "failed while checking cac context for given experiment with error:  "
+                        .to_string() + &val.message.clone();
+
+                if val.status_code == StatusCode::NOT_FOUND {
+                    msg = "Context not found in CAC for given experiment, you should discard this experiment".to_string() + &val.message.clone();
+                }
+                response_error!(val.status_code, msg)
```
ayushjain17, line 222: ideally, we should not forward the status code here, an error here, should be an `Internal Server Error` with respect to the API using this, right ?
the only difference being that the message being forwarded is different, just in the case of 404 from this particular API
```suggestion
                match val.status_code {
                   StatusCode::NOT_FOUND => response_error!(503, "Context not found in CAC for given experiment, you should discard this experiment: {}", val.message),
                   _ => err
                }
```

ayushjain17 (approved):

ayushjain17 (approved):

crates/experimentation_platform/src/api/experiments/cac_api.rs
```diff
@@ -0,0 +1,221 @@
...
+        HeaderValue::from_str(value.as_str())
+            .map(|header_val| headers.insert(header_name, header_val))
+            .map_err(|err| {
+                log::error!("failed to set header: {}", err);
+                unexpected_error!("Something went wrong")
+            })?;
+    }
+
+    Ok(headers)
+}
```
Datron, line 52: Isn't this function already defined in src/api/experiments/helpers.rs, any reason to re-define this?

Datron: So this is because we have a cyclic dependency between files?

ayushjain17 (approved):

Datron (approved):

crates/experimentation_platform/src/api/experiments/cac_api.rs
pratikmishra356: No @Datron it was duplicate , I have removed it from helper

ayushjain17 (approved):

PR #507, opened 2025-05-13
fix: String splice in table component

ayushjain17:
## Problem
string slicing fails with non ASCII characters

## Solution
use `take` on slice of `chars`

### DIFF
--- a/crates/frontend/src/components/table.rs
+++ b/crates/frontend/src/components/table.rs
@@ -19,18 +19,23 @@ pub fn expandable_text(
     let get_child = move |val: &str| formatter(val, &row);
     let (is_expanded_rs, is_expanded_ws) = create_signal(false);
     view! {
-        <td class=format!("{} align-top", class_name)>
+        <td class=format!("{class_name} align-top")>
             {move || {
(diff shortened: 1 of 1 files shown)


Datron (approved):

crates/frontend/src/components/table.rs
```diff
@@ -19,18 +19,23 @@ pub fn expandable_text(
...
     view! {
-        <td class=format!("{} align-top", class_name)>
+        <td class=format!("{class_name} align-top")>
             {move || {
                 let value = stored_value.get_value();
                 let is_expanded = is_expanded_rs.get();
                 let (should_expand, displayed_text) = match is_expandable {
                     Expandable::Enabled(len) if value.len() > len => {
-                        (true, if is_expanded { &value } else { &value[..len] })
+                        let displayed_text = if is_expanded {
```
sauraww, line 28: let displayed_text = value.chars().take(if is_expanded { value.len() } else { len }).collect::<String>();
(true, displayed_text)

sauraww (approved):

PR #511, opened 2025-05-16
feat: Frontend changes for delete experiment

ayushjain17:
## Changes
1. fix error messages of few experiment APIs
2. fix conclude api behaviour when all the keys are deleted for Delete_override type experiment

### DIFF
--- a/crates/frontend/src/components/context_card.rs
+++ b/crates/frontend/src/components/context_card.rs
@@ -12,23 +12,91 @@ use crate::{
 };
 
 #[component]
+fn option(
+    label: String,
+    icon: String,
+    #[prop(into)] on_click: Callback<(), ()>,
+    #[prop(default = "text-blue-500".to_string())] icon_color_class: String,
+) -> impl IntoView {
+    view! {
+        <li on:click=move |_| on_click.call(())>
+            <div class="flex gap-2">
+                <i class=format!("w-fit {icon} ri-lg {icon_color_class}") />
+                <span>{label}</span>
+            </div>
+        </li>
+    }
+}
+
+#[component]
+fn context_options(
+    #[prop(into)] handle_create_experiment: Callback<(), ()>,
+    #[prop(into)] handle_delete_experiment: Callback<(), ()>,
+    #[prop(into)] handle_clone: Callback<(), ()>,
+    #[prop(into)] handle_edit: Callback<(), ()>,
+    #[prop(into)] handle_delete: Callback<(), ()>,
+) -> impl IntoView {
+    let node_ref = create_node_ref::<html::Input>();
+
+    view! {
+        <div class="w-fit dropdown dropdown-left">
+            <label
+                tabindex="0"
+                class="btn btn-sm text-xs m-1 w-full"
+                on:click:undelegated=move |_| {
+                    if let Some(element) = node_ref.get() {
+                        let _ = element.focus();
+                    }
+                }
+            >
+                <i class="ri-more-2-fill" />
+            </label>
+            <ul
+                tabindex="0"
+                class="dropdown-content z-[999999] menu w-[350px] flex-nowrap p-2 shadow bg-base-100 rounded-box overflow-x-hidden"
+            >
+                <Option
+                    label="Update Overrides via Experiment".to_string()
+                    icon="ri-test-tube-line".to_string()
+                    on_click=handle_create_experiment
+                />
+                <Option
+                    label="Delete Overrides via Experiment".to_string()
+                    icon="ri-delete-row".to_string()
+                    on_click=handle_delete_experiment
+                />
+                <Option
+                    label="Update Overrides".to_string()
+                    icon="ri-pencil-line".to_string()
+                    on_click=handle_clone
+                />
+                <Option
+                    label="Clone Overrides".to_string()
+                    icon="ri-file-copy-line".to_string()
+                    on_click=handle_edit
+                />
+                <Option
+                    label="Delete Overrides".to_string()
+                    icon="ri-delete-bin-5-line".to_string()
+                    icon_color_class="text-red-500".to_string()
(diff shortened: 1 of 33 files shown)


crates/frontend/src/components/experiment.rs
```diff
@@ -143,6 +141,14 @@ where
                         }
                         ExperimentStatusType::INPROGRESS => {
```
mahatoankitkumar, line 142: This match statement is growing, i think it should be refactored a bit.
Maybe make a map keeping the on:click and label for different ExperimentStatusType.

crates/frontend/src/components/experiment_conclude_form.rs
```diff
@@ -1,138 +1,189 @@
...
-                value=change_reason_rs.get_untracked()
-                on_change=Callback::new(move |new_change_reason| {
-                    change_reason_ws.set(new_change_reason)
-                })
-            />
-
-            <For
-                each=move || {
-                    experiment_rc
+            {move || {
```
mahatoankitkumar, line 103: move needed here, if we use get_untracked?

mahatoankitkumar (changes requested):

crates/frontend/src/components/context_card.rs
```diff
@@ -110,32 +116,69 @@ pub fn context_card(
...
-                                }
-                            />
-                            <i
-                                class="ri-delete-bin-5-line ri-lg text-red-500 cursor-pointer"
-                                on:click=move |_| {
-                                    let context_id = context_id.get_value();
-                                    handle_delete.call(context_id);
-                                }
-                            />
+                            <div class="w-fit dropdown dropdown-left">
```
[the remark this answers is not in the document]
ayushjain17: passing individual callbacks makes it equivalent

crates/frontend/src/components/experiment_form.rs
```diff
@@ -29,13 +32,38 @@ fn get_init_state(variants: &[VariantFormT]) -> Vec<(String, VariantFormT)> {
...
+        }
+    }
+}
+
 #[component]
 pub fn experiment_form(
-    #[prop(default = false)] edit: bool,
-    #[prop(default = String::new())] id: String,
-    name: String,
+    #[prop(optional)] edit: Option<String>,
```
mahatoankitkumar, line 61: name specifically like edit_experiement_id or something similar

mahatoankitkumar (changes requested):

crates/frontend/src/components/modal.rs
```diff
@@ -28,3 +28,30 @@ where
...
+                    >
+                        <i class="ri-close-line"></i>
+                    </button>
+                    <h3 class="font-bold text-lg">{heading.clone()}</h3>
+                    {children()}
+                </dialog>
+            </div>
+        </Portal>
+    }
+}
```
sauraww, line 57: Nice

crates/frontend/src/components/experiment_conclude_form.rs
```diff
@@ -1,138 +1,189 @@
...
-                                        let handle_conclude_experiment_clone = handle_conclude_experiment_clone
-                                            .clone();
-                                        handle_conclude_experiment_clone(variant_id.to_string())
+                        .map(move |(idx, variant)| {
+                            let variant = variant.clone();
+                            let variant_type = variant.variant_type;
+                            let variant_id = variant.id;
+                            let disabled = req_inprogess_rws.get();
+                            match variant_type {
+                                VariantType::CONTROL => {
```
sauraww, line 148: Instead of matching and adding button two times , Can we match and add the values that are changing and add the button only one time ?
This type of pattern we can follow at other places also .

crates/frontend/src/components/variant_form.rs
```diff
@@ -338,3 +331,373 @@ where
...
+                    }}
+                </Show>
+            </div>
+        </div>
+    }
+            .into_view()
+}
+
+#[derive(Serialize, Deserialize, Clone, Default)]
+struct CombinedResouse {
```
sauraww, line 479: I think this is a typo 
You probably wanted to name it CombinedResource

crates/frontend/src/utils.rs
```diff
@@ -262,6 +265,18 @@ where
...
+pub async fn request<'a, T>(
+    url: String,
+    method: reqwest::Method,
+    body: Option<T>,
+    headers: HeaderMap,
+) -> Result<reqwest::Response, String>
+where
+    T: serde::Serialize,
+{
+    request_with_skip_error(url, method, body, headers, &[]).await
```
sauraww, line 277: why are we using it ?

sauraww (changes requested):

crates/frontend/src/components/experiment_conclude_form.rs
ayushjain17: not possible, leptos does not support that kind of param passing

crates/frontend/src/components/experiment_form.rs
```diff
@@ -199,13 +225,29 @@ pub fn experiment_form(
...
-                        edit=edit
-                        variants=variants
-                        default_config=default_config.get_value()
-                        handle_change=handle_variant_form_change
-                    />
+                match experiment_form_type.get_value() {
+                    ExperimentFormType::Default => {
+                        view! {
+                            <VariantForm
+                                edit=edit.get_value().is_some()
```
mahatoankitkumar, line 232: Change these too

ayushjain17: not relevant

crates/frontend/src/utils.rs
ayushjain17: need this to handle `412` separately

sauraww (approved):

PR #515, opened 2025-05-19
feat: introduce experiment groups data model

Datron:
## Problem
Isolate experiments from each other in experiment groups

## Solution
First PR for experiment groups

### DIFF
--- a/crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
+++ b/crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
@@ -1,0 +1,14 @@
+-- Your SQL goes here
+CREATE TABLE IF NOT EXISTS public.experiment_groups(
+    experiment_group_id TEXT PRIMARY KEY,
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
+    change_reason TEXT NOT NULL,
+    context JSON NOT NULL,
+    traffic_percentage integer NOT NULL CONSTRAINT traffic_percentage_range CHECK (traffic_percentage >= 0 AND traffic_percentage <= 100),
+    member_experiment_ids TEXT[] NOT NULL,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    created_by TEXT NOT NULL,
+    last_modified_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    last_modified_by TEXT NOT NULL
+);
--- a/crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/down.sql
+++ b/crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/down.sql
@@ -1,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+DROP TABLE IF EXISTS public.experiment_groups;
--- a/crates/superposition_types/src/database/models/experimentation.rs
+++ b/crates/superposition_types/src/database/models/experimentation.rs
@@ -279,3 +279,26 @@ pub struct EventLog {
     pub new_data: Option<Value>,
     pub query: String,
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(
+    feature = "diesel_derives",
+    derive(QueryableByName, Queryable, Selectable, Insertable)
+)]
+#[cfg_attr(feature = "diesel_derives", diesel(check_for_backend(diesel::pg::Pg)))]
+#[cfg_attr(feature = "diesel_derives", diesel(primary_key(experiment_group_id)))]
+pub struct ExperimentGroup {
+    pub experiment_group_id: String,
+    pub name: String,
+    pub description: Description,
+    pub change_reason: ChangeReason,
+    pub context: Condition,
+    pub traffic_percentage: TrafficPercentage,
+    pub member_experiment_ids: Vec<String>,
+    pub created_at: DateTime<Utc>,
+    pub created_by: String,
+    pub last_modified_at: DateTime<Utc>,
+    pub last_modified_by: String,
+}
+
+pub type ExperimentGroups = Vec<ExperimentGroup>;
--- a/crates/superposition_types/src/database/schema.rs
+++ b/crates/superposition_types/src/database/schema.rs
@@ -718,6 +718,22 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    experiment_groups (experiment_group_id) {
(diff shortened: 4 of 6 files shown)


crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
```diff
@@ -0,0 +1,13 @@
+-- Your SQL goes here
+CREATE TABLE IF NOT EXISTS public.experiment_groups(
+    hash TEXT PRIMARY KEY,
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
```
ayushjain17, line 5: change_reason ?

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
```diff
@@ -0,0 +1,13 @@
+-- Your SQL goes here
+CREATE TABLE IF NOT EXISTS public.experiment_groups(
+    hash TEXT PRIMARY KEY,
```
ayushjain17, line 3: curious, what is this `hash`?

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
```diff
@@ -0,0 +1,13 @@
...
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
+    context JSON NOT NULL,
+    traffic_percentage integer NOT NULL CONSTRAINT traffic_percentage_range CHECK (traffic_percentage >= 0 AND traffic_percentage <= 100),
+    experiments_inside_ids TEXT[] NOT NULL,
```
ayushjain17, line 8: `contained_experiment_ids` or `scoped_experiment_ids` or `included_experiment_ids` ?
```suggestion
    contained_experiment_ids TEXT[] NOT NULL,
```

ayushjain17 (changes requested):

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: Since it allows only create, no change reason is needed

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: The hash of the context, similar to context ID

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: I'm thinking `internal_experiment_ids`, does that work?

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
ayushjain17: then can we rename it to context id itself ?

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
ayushjain17: in that case `last_modified` and all should also not be needed right ?
but addition of new experiments to the group is also a change, even though it might be a system generated change, the system should provide a change reason like `adding xyz experiment to the group`

and there would be changes to traffic percentage, and probably context also ? to expand or contract the scope ?
PS: if we allow changing of context then we should ideally not use the context hash as the id

Datron: Last modified since you can add an experiment to the group, change reason is redundant here since it is not human provided. Discussed the other stuff with @knutties , and if you want to modify traffic percentage or context you close this group and create a new one

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: renamed to experiment_group_id since context_id as the primary key of experiment groups table seems wrong

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: went with member_experiment_ids

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: > same goes for experiment conclude, we modify an existing context for each we do send a change reason: Experiment concluded with xyz variant that is why this change is being made
even though this was a system change we needed to send the reason, to track changes better

FYI conclude now takes change reason from the user as input, it is not system generated. Its valid here, you would want to understand why someone concluded with a variant.

ayushjain17: >FYI conclude now takes change reason from the user as input
that is the reason for the experiment being concluded only
and not the reason which is passed to the CAC land, where a system generated message is sent - which actually makes sense

Datron: Okay, I still don't see how change reason is relevant here. The reason is always the same.

ayushjain17: > Sure, the question is how would change reason help with this? 

it can just simply say that this last modification which happened for which there is a difference in the data being shown for `last_modified` and `created_at` is for this particular reason and nothing else

it makes sure that I have the info in the system

Datron: > it makes sure that I have the info in the system

Why do I need this info? Seems like a case for an event_log, not really something change_reason solves

ayushjain17: change reason is meant to improve the event log tracking
that was the exact purpose of the introduction 😅

Datron: The diff will be in the event log, the change reason is implied, what is the purpose of recording that which is obvious?

Datron: The purpose of change_reason is to tell users why other users did what they did. Here it is implied.

ayushjain17 (approved):

PR #516, opened 2025-05-19
fix: Override update/clone/experiment after key filter

ayushjain17:
## Problem
On updating/cloning/experimenting on a override after a key filter has been applied, the operation is done only with the remaining set of keys

## Solution
use the context id to fetch the data of the context on which the operation is to be performed

### DIFF
--- a/crates/frontend/src/pages/context_override.rs
+++ b/crates/frontend/src/pages/context_override.rs
@@ -21,7 +21,6 @@ use superposition_types::{
     PaginatedResponse, SortBy,
 };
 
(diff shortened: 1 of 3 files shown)


mahatoankitkumar (approved):

crates/frontend/src/pages/context_override.rs
```diff
@@ -370,11 +370,132 @@ fn form(
     }
 }
 
+#[allow(clippy::type_complexity)]
+fn use_context_data(
```
Datron, line 374:
```suggestion
fn generate_context_resource(
```

sauraww (approved):

PR #522, opened 2025-05-23
feat: added cac-client, native-resolver and ffi layer

sauraww:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
```diff
@@ -0,0 +1,298 @@
+import { SuperpositionConfig, CACClientOptions, ConfigData, EvalOptions } from './types';
```
Datron: If you generate a new smithy client would this get deleted or overwritten?

clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
```diff
@@ -0,0 +1,298 @@
...
+            // This matches your benchmark test exactly
+            const result = this.resolver.resolveConfig(
+                configData.default_configs || {},
+                configData.contexts || [],
+                configData.overrides || {},
+                queryData,  // User's query data directly
+                'merge'
+            );
+
+            console.log('✅ Resolution result:', result);
```
Datron: Your AI is too cheerful. Lets not use emoji in our logs - we have enough parsing problems

clients/generated/smithy/typescript/src/resolver-client/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// Default client instance
+const config: SuperpositionConfig = {
+    endpoint: ENV.baseUrl,
+    token: {
+        token: "12345678",
+    },
+    org_id: ENV.org_id,
+    workspace_id: ENV.workspace_id,
+};
```
Datron: Should these be hardcoded?

clients/generated/smithy/typescript/src/resolver-client/native-resolver.ts
```diff
@@ -0,0 +1,150 @@
...
+        let filename;
+        if (process.platform === 'win32') {
+            filename = 'core.dll';
+        } else if (process.platform === 'darwin') {
+            filename = 'libcore.dylib';
+        } else {
+            filename = 'libcore.so';
+        }
+
+        return path.resolve(__dirname, `../../../../../../target/release/${filename}`);
```
Datron: Relative paths are risky, better not to use these

clients/generated/smithy/typescript/src/resolver-client/native-resolver.ts
```diff
@@ -0,0 +1,150 @@
...
+        }
+        if (!overrides) {
+            throw new Error('overrides parameter is required');
+        }
+        if (!queryData) {
+            throw new Error('queryData parameter is required');
+        }
+        if (!mergeStrategy) {
+            throw new Error('mergeStrategy parameter is required');
+        }
```
Datron: These checks are redundant since typescript doesn't allow undefined or null as an argument for this. Contexts could be typed better.

clients/generated/smithy/typescript/src/resolver-client/native-resolver.ts
```diff
@@ -0,0 +1,150 @@
...
+        const overridesJson = JSON.stringify(overrides);
+        const queryDataJson = JSON.stringify(queryData);
+
+        console.log('🔧 Calling FFI with parameters:');
+        // console.log('  defaultConfigs length:', defaultConfigsJson.length);
+        console.log('  defaultConfigs:', defaultConfigs);
+        console.log('  contexts length:', contextsJson.length);
+        console.log('  overrides length:', overridesJson.length);
+        console.log('  queryData length:', queryDataJson.length);
+        console.log('  mergeStrategy:', mergeStrategy);
```
Datron: simplify logging

clients/generated/smithy/typescript/src/resolver-client/superposition-provider.ts
```diff
@@ -0,0 +1,403 @@
...
+    };
+
+    events?: ProviderEventEmitter<AnyProviderEvent> | undefined = new OpenFeatureEventEmitter();
+    readonly hooks: Hook[] = [];
+
+    private client: ConfigurationClient;
+    status: ProviderStatus = ProviderStatus.NOT_READY;
+    private pollingInterval?: NodeJS.Timeout;
+    private config: SuperpositionProviderConfig;
+    private cachedConfig: Record<string, any> = {};
```
Datron: Can we avoid `any`?

clients/generated/smithy/typescript/src/resolver-client/types.ts
```diff
@@ -0,0 +1,33 @@
+export interface SuperpositionConfig {
+    endpoint: string;
+    token?: {
+        token: string;
+    };
```
Datron: can this be defined better? Inside token is another object of token

Datron: Support API keys as well

crates/core/src/config.rs
```diff
@@ -0,0 +1,268 @@
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Map, Value};
+use std::collections::HashMap;
+
+// Merge strategy for configuration resolution
+#[derive(Clone, Debug, PartialEq)]
+pub enum MergeStrategy {
+    MERGE,
+    REPLACE,
+}
```
Datron: We can redefine this in superposition_types and use it here, seems repeated across crates

crates/core/src/ffi.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn core_last_error_message() -> *const c_char {
+    LAST_ERROR.with(|prev| match prev.borrow().clone() {
+        Some(err) => match CString::new(err) {
+            Ok(c_str) => c_str.into_raw(),
+            Err(_) => ptr::null(),
```
Datron: we should log this error returned, also careful with null pointers

clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
sauraww: Sure ,I will remove it

clients/generated/smithy/typescript/src/resolver-client/index.ts
sauraww: I will remove the hardcoding and move it to ENV

clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
ayushjain17: more likely to get deleted, no ?

crates/core/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+itertools = { workspace = true }
+jsonlogic = { workspace = true }
+log = { workspace = true }
+mini-moka = { version = "0.10.3" }
+once_cell = { workspace = true }
+reqwest = { workspace = true }
+serde_json = { workspace = true }
+strum = { workspace = true }
+strum_macros = { workspace = true }
+superposition_types = { path = "../superposition_types" }
```
ayushjain17, line 19: should we consider moving the entire logic from here to maybe `superposition_types` maybe ?
and make that itself the core guy ?
we can maybe merge them and rename it to core ?

sauraww: What do we actually mean by core ? 
I think the intent here is core logic of superposition which is resolution logic and FFI interface. 
I don't know if superposition_types fit into this definition.

crates/core/src/config.rs
ShubhranshuSanjeev, line 1: filtering of keys in resolved config based on prefix is missing from these eval functions.

crates/core/src/core.h
ShubhranshuSanjeev, line 1: Why all functions are prefixed with `core_`, is it required?

clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        for (const part of keyParts) {
+            if (value === undefined || value === null || typeof value !== 'object') {
+                return undefined;
+            }
+            value = value[part];
+        }
+
+        return value;
+    }
```
ShubhranshuSanjeev: A similar filter by prefix logic was there in the core evaluation layer, and should continue to exist alongside the evaluation logic.

ShubhranshuSanjeev (changes requested):

crates/core/src/core.h
sauraww: The prefix signifies that it is pure, language-agnostic configuration resolution logic. 
Earlier we had a prefix of cac which didn't make sense to me . 
Also this is purely a naming convention choice.
The architecture is like that: 
Core layer : Pure configuration logic, no I/O, no caching, no HTTP
Client layer (TypeScript/Python/etc.): Business logic, caching, API calls, error handling

clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
ShubhranshuSanjeev: Since the existing logic is a filter, for open-feature implementation we can add a `find` operation, which can return a single value rather than a list of values.

crates/core/src/experiment.rs
```diff
@@ -0,0 +1,110 @@
...
+    for experiment in experiments {
+        if !filter_prefixes.is_empty() {
+            let matches_prefix = filter_prefixes.iter().any(|prefix| {
+                experiment.name.starts_with(prefix) || experiment.id.starts_with(prefix)
+            });
+            if !matches_prefix {
+                continue; // Skip this experiment if it doesn't match any prefix
+            }
+        }
+
```
ShubhranshuSanjeev: What's the point on filtering on experiment name?

clients/generated/smithy/typescript/src/resolver-client/configuration-client.ts
sauraww: I think you are conflating between two different concerns: 
Filtering (during resolution) vs Finding (after resolution)
Filter by prefix = "Give me only config keys that start with certain prefix like say database or cache "
Find operation = "From my already-resolved config, get me the value at  lets say 'database.connection.host'"

I think core_config_resolution should handle configuration resolution logic, not value extraction convenience methods.
Also it's a performance nightmare : 

Bad: Multiple FFI calls
const host = await core_find_value(configs, contexts, overrides, query, "database.host");
const port = await core_find_value(configs, contexts, overrides, query, "database.port");
const ttl = await core_find_value(configs, contexts, overrides, query, "cache.ttl");
It would result in 3 FFI calls = 3x the overhead

Current approach: 
const config = await client.eval(queryData);  
const host = getNestedValue(config, "database.host");    
const port = getNestedValue(config, "database.port");    
const ttl = getNestedValue(config, "cache.ttl");        
1 FFI call and rest of the calls are in-memory 


In nutshell , it's the client layer where the finding belongs

crates/core/src/experiment.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Variant {
+    pub id: String,
+    pub experiment_id: String,
+    pub context: Map<String, Value>,
+    pub variants: Vec<String>,
+    pub override_keys: Vec<String>,
+    pub status: String,
+}
```
ShubhranshuSanjeev: Can these types be re-used from superposition-types? Not mandatory, if they bring in extra dependencies from `superposition-types` crates.

clients/generated/smithy/typescript/src/resolver-client/superposition-provider.ts
ShubhranshuSanjeev, line 1: @sauraww , I was thinking to use experiments in the superposition provider, we can have a function which as usual finds applicable variants and then add the ids to the evaluation context. And this function can be added as a `before` hook to the superposition client evaluation calls.

ShubhranshuSanjeev (dismissed):

pratikmishra356 (approved):

ayushjain17 (approved):
reviewed config.rs
hopefully someday we will pick up all the refactoring and merge this and superposition_types crate into one and keep it as superposition_core

PR #525, opened 2025-05-23
feat: add experiment groups API

Datron:
## Problem
Continuation of discussion #508 

## Solution
- Add API support 
- Add Smithy model
- Generate Clients
- Add tests for experimentation group APIs
- Allow creation of multiple experiments with the same context (required a model change)


Cargo.toml
```diff
@@ -57,7 +57,7 @@ regex = "1.9.1"
 reqwest = { version = "0.11.18", features = ["json"] }
 rs-snowflake = "0.6.0"
 serde = { version = "^1", features = ["derive"] }
-serde_json = { version = "1.0" }
+serde_json = { version = "1.0.140" }
```
Datron, line 60: Earlier our cargo.lock had 1.0.113 which did not implement the trait Hash on Value. 1.0.140 solves this

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,141 @@
...
+    let experiment_group_hash = hash(&Value::Object(exp_context.clone().into()));
+    let new_experiment_group = ExperimentGroup {
+        experiment_group_id,
+        experiment_group_hash,
+        name: req.name,
+        description: req.description,
+        change_reason: req.change_reason,
+        created_by: user.email.clone(),
+        last_modified_by: user.email.clone(),
+        created_at: chrono::Utc::now(),
```
pratikmishra356, line 95: can we move it to a variable and use at both the places

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,170 @@
...
+    workspace_request: WorkspaceContext,
+    user: User,
+) -> superposition::Result<Json<ExperimentGroup>> {
+    let DbConnection(mut conn) = db_conn;
+    let req = req.into_inner();
+    log::trace!("Creating experiment group with request: {:?}", req);
+    helpers::create_experiment_group(state, req, &mut conn, workspace_request, user).await
+}
+
+#[put("/{exp_group_id}")]
```
pratikmishra356, line 48: should we use patch ?

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,170 @@
...
+        .returning(ExperimentGroup::as_returning())
+        .schema_name(&schema_name)
+        .get_result(&mut conn)?;
+    if !marked_group.member_experiment_ids.is_empty() {
+        return Err(bad_argument!(
+            "Cannot delete experiment group {} since it has members",
+            marked_group.name
+        ));
+    }
+    diesel::delete(experiment_groups::experiment_groups)
```
pratikmishra356, line 165: should we use soft delete ?

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,141 @@
+use std::collections::HashSet;
+
+use actix_web::web::{Data, Json};
+use diesel::SelectableHelper;
```
pratikmishra356, line 4: can we club imports from same crate at once

crates/superposition/src/main.rs
```diff
@@ -202,6 +202,11 @@ async fn main() -> Result<()> {
                             .wrap(AppExecutionScopeMiddlewareFactory::new(AppScope::SUPERPOSITION))
                             .service(webhooks::endpoints()),
                     )
+                    .service(
+                            experiment_groups::endpoints(scope("/experiment_groups"))
```
pratikmishra356, line 206:
```suggestion
                            experiment_groups::endpoints(scope("/experiment-groups"))
```

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
```diff
@@ -1,6 +1,7 @@
 -- Your SQL goes here
 CREATE TABLE IF NOT EXISTS public.experiment_groups(
-    experiment_group_id TEXT PRIMARY KEY,
+    experiment_group_id bigint PRIMARY KEY,
```
pratikmishra356, line 3: shouldnt this be only id ?

pratikmishra356 (changes requested):

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,141 @@
...
+        .filter(experiment_groups::experiment_group_id.eq(&id))
+        .schema_name(&workspace_request.schema_name)
+        .first::<Condition>(conn)?;
+    validate_members(
+        &req.member_experiment_ids[..],
+        &exp_context,
+        conn,
+        &workspace_request.schema_name,
+    )?;
+    let new_members: HashSet<String> = HashSet::from_iter(req.member_experiment_ids);
```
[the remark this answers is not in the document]
Datron: Will update the code

docker-compose/postgres/db_init.sql
```diff
@@ -1360,7 +1361,8 @@ CREATE TABLE IF NOT EXISTS localorg_dev.experiment_groups(
 );
 
 CREATE TABLE IF NOT EXISTS localorg_test.experiment_groups(
-    experiment_group_id TEXT PRIMARY KEY,
+    experiment_group_id bigint PRIMARY KEY,
+    experiment_group_hash TEXT NOT NULL,
```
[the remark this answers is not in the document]
Datron: Yes

crates/superposition/src/main.rs
Datron: Hmm experiments API is /experiments, same for workspaces and organisations but will update to:
```
/experiment-groups
```

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
Datron: either is fine, I wanted to be verbose.

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,141 @@
...
+            })
+        })
+        .collect::<superposition::Result<Vec<_>>>()?;
+    let mut accepted_members = Vec::new();
+    let experiment_contexts: Vec<(i64, Condition)> = experiments::experiments
+        .select((experiments::id, experiments::context))
+        .filter(
+            experiments::id
+                .eq_any(&members[..])
+                .and(experiments::status.eq(ExperimentStatusType::CREATED)),
```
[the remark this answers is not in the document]
Datron: Any suggestions on a fix?

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
mahatoankitkumar: i guess `id` is fine, `experiment_group_id` is too verbose
we are already following this for other tables and we should try to keep it consistent.

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,141 @@
...
+            accepted_members.push(*experiment_id);
+        } else {
+            return Err(bad_argument!(
+                "Experiment with id {} does not fit in with the experiment group. The contexts do not match.", experiment_id
+            ));
+        }
+    }
+    if members.len() != accepted_members.len() {
+        return Err(bad_argument!(
+            "Some experiment Ids do not exist/are invalid"
```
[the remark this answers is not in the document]
Datron: @mahatoankitkumar so the problem here is
lets say you run this query:
```sql
SELECT * from experiments where id IN ('id1','id2', 'id3');
```
and id3 doesn't exist, then the query still returns data but id3 is not present. Now if I loop over the results, its impossible to know id3 is missing since it never showed up in the query. The only way to check is to look back at the original members list and compare to decide if id3 is not even in the database. The error message can be better, so that can be fixed.

mahatoankitkumar (changes requested):

crates/experimentation_platform/src/api/experiment_groups.rs
mahatoankitkumar: for audit log i guess

ayushjain17 (changes requested):

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -318,26 +319,12 @@ pub fn create_ctx_from_put_req(
     let r_override = req.r#override.clone().into_inner();
     let ctx_override = Value::Object(r_override.clone().into());
 
-    let workspace_settings = get_workspace(schema_name, conn)?;
-
-    validate_condition_with_strict_mode(&ctx_condition, workspace_settings.strict_mode)?;
-
```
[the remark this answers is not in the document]
Datron: It was in move, made the change

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -431,3 +418,24 @@ pub fn update_override_of_existing_ctx(
...
 }
+
+pub fn validate_ctx(
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+    condition: Condition,
+) -> superposition::Result<HashMap<String, DimensionData>> {
+    let workspace_settings = get_workspace(schema_name, conn)?;
+    validate_condition_with_strict_mode(&condition, workspace_settings.strict_mode)?;
+    let context_map = extract_dimensions(&condition)?;
```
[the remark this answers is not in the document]
Datron: @ayushjain17 we can't use that yet here, I've only checked it for equals but this needs to handle IN and other operators. We can come back and refactor this

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -43,6 +43,12 @@ pub struct MoveReq {
     pub change_reason: String,
 }
 
+#[cfg_attr(test, derive(Debug, PartialEq))]
+#[derive(Deserialize, Clone)]
+pub struct ContextValidationRequest {
```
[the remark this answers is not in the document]
Datron: All the other request types are still in this file, keeping just this type elsewhere may make it hard for someone trying to understand our context APIs. This can be moved as part of a separate refactor

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,170 @@
...
+        .service(get_experiment_group)
+        .service(delete_experiment_group)
+}
+
+#[post("")]
+async fn create_experiment_group(
+    state: Data<AppState>,
+    req: Json<ExpGroupCreateRequest>,
+    db_conn: DbConnection,
+    workspace_request: WorkspaceContext,
```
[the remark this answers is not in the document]
Datron: Workspace request is needed by validate_members since it calls CAC to validate the context, so the API call needs org_id and workspace_id

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,170 @@
...
+    log::trace!("Creating experiment group with request: {:?}", req);
+    helpers::create_experiment_group(state, req, &mut conn, workspace_request, user).await
+}
+
+#[patch("/{exp_group_id}")]
+async fn update_experiment_group(
+    exp_group_id: web::Path<i64>,
+    req: Json<ExpGroupUpdateRequest>,
+    db_conn: DbConnection,
+    workspace_request: WorkspaceContext,
```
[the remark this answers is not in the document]
Datron: Workspace request is needed by validate_members since it calls CAC to validate the context, so the API call needs org_id and workspace_id

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,170 @@
...
+        ))
+        .returning(ExperimentGroup::as_returning())
+        .schema_name(&schema_name)
+        .get_result(&mut conn)?;
+    if !marked_group.member_experiment_ids.is_empty() {
+        return Err(bad_argument!(
+            "Cannot delete experiment group {} since it has members",
+            marked_group.name
+        ));
+    }
```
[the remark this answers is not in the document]
Datron: Great catch, moved this under a transaction

crates/experimentation_platform/src/api/experiment_groups.rs
Datron: Experiment groups can last as long as the user wants - different teams may use them differently. If you want the group to be there throughout, just don't delete it. 

If we don't give a delete option but the user doesn't want the experiment group for some reason, then its an inconvenience. 

"Better to have and not use, then to need and not have" is what I thought

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+    conn: &mut DBConnection,
+    schema_name: &String,
+) -> superposition::Result<Vec<i64>> {
+    if members.is_empty() {
+        return Ok(Vec::new());
+    }
+    let members = members
+        .iter()
+        .map(|member| {
+            member.parse::<i64>().map_err(|e| {
```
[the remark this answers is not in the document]
Datron: These members come from the frontend, as you know the precision for large integers in javascript is bad, so I took them as strings. 

They can be stored in the DB as integers though, that is valid

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+            })
+        })
+        .collect::<superposition::Result<Vec<_>>>()?;
+    let mut accepted_members = HashSet::new();
+    let experiment_contexts: Vec<(i64, Condition)> = experiments::experiments
+        .select((experiments::id, experiments::context))
+        .filter(
+            experiments::id
+                .eq_any(&members[..])
+                .and(experiments::status.eq(ExperimentStatusType::CREATED)),
```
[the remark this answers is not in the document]
Datron: Thats the rule. To add a member, they have to be in created state

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+    if !invalid_members.is_empty() {
+        return Err(bad_argument!(
+            "the following experiment IDs are not present in the database/are not in the created stage: {}",
+            invalid_members
+                .iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<_>>()
+                .join(", ")
+        ));
+    }
```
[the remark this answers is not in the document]
Datron: You can use validate_members itself, we can discuss how

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+    state: Data<AppState>,
+    req: ExpGroupCreateRequest,
+    conn: &mut DBConnection,
+    workspace_request: WorkspaceContext,
+    user: User,
+) -> superposition::Result<Json<ExperimentGroup>> {
+    let exp_context = req.context.clone();
+    validate_context(&state, &exp_context, &workspace_request, &user).await?;
+    validate_members(
+        &req.member_experiment_ids[..],
```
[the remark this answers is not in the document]
Datron: Removed

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+                .iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<_>>()
+                .join(", ")
+        ));
+    }
+    Ok(Vec::from_iter(accepted_members))
+}
+
+pub async fn create_experiment_group(
```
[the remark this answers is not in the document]
Datron: I did this so experiments is able to call this or update group. We're not clear on the flow yet, it can be changed later once things are more drawn out

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,149 @@
...
+    };
+    let new_experiment_group = diesel::insert_into(experiment_groups::experiment_groups)
+        .values(&new_experiment_group)
+        .returning(ExperimentGroup::as_returning())
+        .schema_name(&workspace_request.schema_name)
+        .get_result::<ExperimentGroup>(conn)?;
+    Ok(Json(new_experiment_group))
+}
+
+pub async fn update_experiment_group(
```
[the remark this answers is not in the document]
Datron: This definitely needs to be outside since experiments will call this function when adding an experiment to this group

crates/experimentation_platform/src/api/experiments/cac_api.rs
```diff
@@ -219,3 +219,67 @@ pub async fn get_context_override(
...
     Ok(resp_contexts)
 }
+
+pub async fn validate_context(
+    state: &Data<AppState>,
+    condition: &Condition,
+    workspace_request: &WorkspaceContext,
+    user: &User,
+) -> superposition::Result<()> {
+    let http_client = reqwest::Client::new();
```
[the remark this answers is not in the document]
Datron: Done

crates/experimentation_platform/src/api/experiments/cac_api.rs
```diff
@@ -219,3 +219,67 @@ pub async fn get_context_override(
...
+                .ok_or_else(|| unexpected_error!("failed to parse Context validate error. Please checks the system logs"))?;
+            log::error!(
+                "http call to context validate failed with error {}",
+                error_message
+            );
+            Err(bad_argument!(error_message))
+        }
+        Err(err) => {
+            log::error!("Context validation failed with the error: {err}");
+            Err(bad_argument!(err))
```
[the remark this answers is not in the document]
Datron: Its not really unexpected, because in that case we wouldn't have handled it. The context is wrong, so its a bad argument

crates/superposition/src/main.rs
```diff
@@ -202,6 +202,11 @@ async fn main() -> Result<()> {
                             .wrap(AppExecutionScopeMiddlewareFactory::new(AppScope::SUPERPOSITION))
                             .service(webhooks::endpoints()),
                     )
+                    .service(
```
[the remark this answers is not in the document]
Datron: Moved this after experiments

crates/superposition_types/migrations/2025-05-16-082925_experiment_groups/up.sql
```diff
@@ -1,6 +1,7 @@
 -- Your SQL goes here
 CREATE TABLE IF NOT EXISTS public.experiment_groups(
-    experiment_group_id TEXT PRIMARY KEY,
+    id bigint PRIMARY KEY,
+    context_hash TEXT NOT NULL,
```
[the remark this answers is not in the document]
Datron: Its not an id here, since multiple groups with the same context can be created. I've left this in so that we can do migrations in the future and validate data, it is not an identifier

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,86 @@
...
+use superposition_derives::IsEmpty;
+
+use crate::{database::models::experimentation::TrafficPercentage, Condition, SortBy};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExpGroupCreateRequest {
+    pub name: String,
+    pub description: Description,
+    pub change_reason: ChangeReason,
+    pub context: Condition,
```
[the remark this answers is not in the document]
Datron: Will use Cac<Condition>, it does not allow variantIds right?

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub traffic_percentage: TrafficPercentage,
+    #[serde(default)]
+    pub member_experiment_ids: Vec<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "diesel_derives", derive(AsChangeset))]
+#[cfg_attr(feature = "diesel_derives", diesel(table_name = experiment_groups))]
+pub struct ExpGroupUpdateRequest {
```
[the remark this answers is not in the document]
Datron: Done

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub change_reason: ChangeReason,
+    pub traffic_percentage: Option<TrafficPercentage>,
+    #[serde(default)]
+    pub member_experiment_ids: Vec<String>,
```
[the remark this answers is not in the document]
Datron: Got it, fixing

crates/superposition_types/src/config.rs
```diff
@@ -115,6 +115,16 @@ impl Condition {
         }
         Ok(Self(condition_map))
     }
+
+    pub fn contains(&self, other_condition: &Condition) -> Result<bool, String> {
+        let condition1 = json!(self.0);
```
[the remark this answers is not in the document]
Datron: Its one in the same, but sure

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -302,3 +304,25 @@ pub struct ExperimentGroup {
 }
 
 pub type ExperimentGroups = Vec<ExperimentGroup>;
+
+mod i64_formatter {
+    use serde::{self, Deserialize, Deserializer, Serializer};
+
+    // Serialize i64 to String
```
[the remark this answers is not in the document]
Datron: If I send big int back to javascript, it loses its precision so I convert it to string. I want the conversion to be done automatically on deserialization and serialization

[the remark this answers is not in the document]
Datron: So the user sends Strings, but be operate on ints. And we respond with ints in the handler, and the user gets strings. No loss of precision.

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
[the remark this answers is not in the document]
ayushjain17: @Datron I am saying something else
I am saying:
>  if any member has started, no new member can be added

this is what the logic seems to be doing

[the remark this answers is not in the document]
Datron: @ayushjain17 this function is used for new members that have to be added, not for existing members.

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::{
+    database::models::{experimentation::i64_vec_deserialize, ChangeReason, Description},
+    Cac, IsEmpty,
+};
+#[cfg(feature = "diesel_derives")]
+use diesel::query_builder::AsChangeset;
+use serde::{Deserialize, Serialize};
+use superposition_derives::IsEmpty;
+
+use crate::{database::models::experimentation::TrafficPercentage, Condition, SortBy};
```
pratikmishra356: can we combine this import with above ?

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub traffic_percentage: Option<TrafficPercentage>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[cfg_attr(feature = "diesel_derives", derive(AsChangeset))]
+#[cfg_attr(feature = "diesel_derives", diesel(table_name = experiment_groups))]
+pub struct ExpGroupMemberRequest {
+    pub change_reason: ChangeReason,
+    #[serde(with = "i64_vec_formatter")]
+    pub member_experiment_ids: Vec<i64>,
```
pratikmishra356, line 44: cant we use Hashset here ?
or maybe in the helper from starting lets se hashset and during db write convert it to vector

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,157 @@
...
+            .schema_name(schema_name)
+            .get_results::<(i64, Condition, TrafficPercentage)>(conn)?;
+    let existing_member_contexts: Vec<TrafficPercentage> = experiments::experiments
+        .select(experiments::traffic_percentage)
+        .filter(experiments::id.eq_any(existing_members))
+        .schema_name(schema_name)
+        .get_results::<TrafficPercentage>(conn)?;
+    let mut traffic_percentage_sum =
+        existing_member_contexts.iter().fold(0, |acc, x| acc + **x);
+    for (experiment_id, context, traffic) in experiment_contexts.iter() {
```
pratikmishra356, line 77: probably we are not checking exclusivity of existing and new members which might result in repeated traffic percentage calculation

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,254 @@
...
+            experiment_groups::last_modified_by.eq(user.email),
+            experiment_groups::last_modified_at.eq(chrono::Utc::now()),
+        ))
+        .returning(ExperimentGroup::as_returning())
+        .schema_name(&schema_name)
+        .get_result(&mut conn)?;
+    Ok(Json(updated_group))
+}
+
+#[post("/{exp_group_id}/add-members")]
```
pratikmishra356: we can use patch here also and for the next api remove-members

crates/superposition_types/src/api/experiment_groups.rs
Datron: The generated schema by diesel uses vector, and this type uses changeset to do the update so we can't directly use hashset here

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: Nice catch, this is done

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -288,17 +288,93 @@ pub struct EventLog {
...
+
+    // Deserialize String to i64
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<i64, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        s.parse::<i64>()
+            .map_err(|e| serde::de::Error::custom(format!("Failed to parse i64: {}", e)))
+    }
```
ayushjain17, line 328: this deserialisation exclusively works for string input only, this should have been an either or to achieve [this](https://github.com/juspay/superposition/pull/525/files#r2113409900)

```suggestion
    pub fn deserialize<'de, D>(deserializer: D) -> Result<i64, D::Error>
    where
        D: Deserializer<'de>,
    {
        let value = Value::deserialize(deserializer)?;
        match value {
        Value::String(s) => s.parse::<i64>()
                    .map_err(|e| serde::de::Error::custom(format!("Failed to parse to i64: {}", e)))
        }
        Value::Number(n) => {
            if let Some(id) = n.as_i64() {
                Ok(id)
            } else {
                log::error!("Expected a bigint as input");
                Err(serde::de::Error::custom(
                    "Expected a bigint as input",
                ))
            }
        }
      _ => {
                log::error!("Expected a bigint or stringified bigint as input");
                Err(serde::de::Error::custom(
                    "Expected a bigint or stringified bigint as input",
                ))
      }
    }
```

crates/experimentation_platform/src/api/experiments/cac_api.rs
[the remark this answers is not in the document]
ayushjain17: @Datron this error comes when there was a failure in making the API call, and not as an error in the response of the API, that section has already been handled, which is indeed `bad_argument` but this is indeed and `unexpected_error`

crates/experimentation_platform/src/api/experiments/cac_api.rs
```diff
@@ -219,3 +219,67 @@ pub async fn get_context_override(
...
+    })?;
+
+    let extra_headers = vec![("x-user", user_str)];
+
+    let headers_map = construct_header_map(
+        &workspace_request.workspace_id,
+        &workspace_request.organisation_id,
+        extra_headers,
+    )?;
+    let payload = json!({
```
ayushjain17: we could have used the type here now itself, if we move the type to `superposition_types`

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -288,17 +288,93 @@ pub struct EventLog {
...
+pub mod i64_formatter {
+    use serde::{self, Deserialize, Deserializer, Serializer};
+
+    // Serialize i64 to String
+    pub fn serialize<S>(value: &i64, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(&value.to_string())
+    }
```
ayushjain17, line 318: serialize is not needed, no ?
only custom deserialization should be enough

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -288,17 +288,93 @@ pub struct EventLog {
...
+        s.parse::<i64>()
+            .map_err(|e| serde::de::Error::custom(format!("Failed to parse i64: {}", e)))
+    }
+}
+
+pub mod i64_vec_formatter {
+    use serde::{self, Deserialize, Deserializer, Serialize, Serializer};
+
+    // Serialize Vec<i64> to Vec<String>
+    pub fn serialize<S>(value: &[i64], serializer: S) -> Result<S::Ok, S::Error>
```
ayushjain17, line 335: same for this serialize as well

crates/superposition_types/src/database/models/experimentation.rs
Datron: This a converter though, it takes a string and makes it an int. If you're passing int, why bother?

Datron: is the name confusing? Would `string_to_i64_vec_mapper` be better?

crates/superposition_types/src/database/models/experimentation.rs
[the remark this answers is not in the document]
ayushjain17: the DB type is `array of int`, the `deserialize` definition currently says to read only `array of strings` and convert to `array of int`, while reading from DB, you would get `array of int`, the `deserialize` would fail

you should support custom deserialise as suggested and not have serialise, just use default serialise

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -288,17 +288,93 @@ pub struct EventLog {
...
+    where
+        S: Serializer,
+    {
+        let string_values: Vec<String> = value.iter().map(|v| v.to_string()).collect();
+        string_values.serialize(serializer)
+    }
+
+    // Deserialize Vec<String> to Vec<i64>
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<i64>, D::Error>
```
ayushjain17, line 344: ideally, the vec deserialisers should also behave like the above suggested deserialise and accept both

crates/superposition_types/src/database/models/experimentation.rs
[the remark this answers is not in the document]
ayushjain17: your system might still be working, because your local might still have the old type `Array<Text>` which has been changed to `Array<Int>` now

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+    id: &i64,
+    mut req: ExpGroupMemberRequest,
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+    user: User,
+) -> superposition::Result<Json<ExperimentGroup>> {
+    let exp_context = experiment_groups::experiment_groups
+        .filter(experiment_groups::id.eq(&id))
+        .schema_name(schema_name)
+        .first::<ExperimentGroup>(conn)?;
```
ayushjain17, line 156: can we rather read the `member_experiment_ids` column only, instead of the entire row ?

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+    Ok(Json(updated_group))
+}
+
+pub fn remove_members(
+    id: &i64,
+    mut req: ExpGroupMemberRequest,
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+    user: User,
+) -> superposition::Result<Json<ExperimentGroup>> {
```
ayushjain17, line 152: just asking, should we add/do we need any validations before removing ?
like maybe the experiment should not have already started or something ?

I dont think we discussed much regarding the conditions for removal other than the conclusion of an experiment

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+    exp_group_id: &i64,
+    mut req: ExpGroupMemberRequest,
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+    user: User,
+) -> superposition::Result<Json<ExperimentGroup>> {
+    let exp_context = experiment_groups::experiment_groups
+        .filter(experiment_groups::id.eq(exp_group_id))
+        .schema_name(schema_name)
+        .first::<ExperimentGroup>(conn)?;
```
ayushjain17, line 125: here as well, should we read only the required 2 columns ?

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,254 @@
...
+        total_items,
+        data,
+    }))
+}
+
+#[get("/{exp_group_id}")]
+async fn get_experiment_group(
+    exp_group_id: web::Path<i64>,
+    db_conn: DbConnection,
+    workspace_request: WorkspaceContext,
```
ayushjain17:
```suggestion
    schema_name: SchemaName,
```

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,254 @@
...
+    let id = exp_group_id.into_inner();
+    helpers::remove_members(&id, req, &mut conn, &schema_name, user)
+}
+
+#[get("")]
+async fn list_experiment_groups(
+    pagination_params: Query<PaginationParams>,
+    filters: Query<ExpGroupFilters>,
+    db_conn: DbConnection,
+    workspace_request: WorkspaceContext,
```
ayushjain17:
```suggestion
    schema_name: SchemaName,
```

crates/experimentation_platform/src/api/experiment_groups.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    #[rustfmt::skip]
+    let base_query = match (sort_on, sort_by) {
+        (SortOn::LastModifiedAt, SortBy::Desc) => base_query.order(experiment_groups::last_modified_at.desc()),
+        (SortOn::LastModifiedAt, SortBy::Asc)  => base_query.order(experiment_groups::last_modified_at.asc()),
+        (SortOn::CreatedAt, SortBy::Desc)      => base_query.order(experiment_groups::created_at.desc()),
+        (SortOn::CreatedAt, SortBy::Asc)       => base_query.order(experiment_groups::created_at.asc()),
+        (SortOn::Name, SortBy::Desc)           => base_query.order(experiment_groups::name.desc()),
+        (SortOn::Name, SortBy::Asc)            => base_query.order(experiment_groups::name.asc()),
+    };
```
ayushjain17: can we please move this above the check for `all` param

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+/// - if they are in the created stage,
+/// - if their contexts contain the group context
+/// - if the sum of their traffic percentages does not exceed 100%
+/// - if there are no repeating members in the new members list
+pub fn validate_members(
+    new_members: &[i64],
+    existing_members: &[i64],
+    group_context: &Condition,
+    conn: &mut DBConnection,
+    schema_name: &String,
```
```suggestion
    schema_name: &SchemaName,
```

ayushjain17 (changes requested):

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: There no perf impact, the only reason I did that in validate_members is to make it easier to iterate

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: Removal is straight forward, we can see if any cases come up later

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: This can go either way, I'll leave it as is.

crates/experimentation_platform/src/api/experiments/cac_api.rs
Datron: Done

crates/superposition_types/src/database/models/experimentation.rs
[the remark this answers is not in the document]
Datron: No its changed and working

crates/superposition_types/src/database/models/experimentation.rs
Datron: Experiment group is returned as a response, if we respond with i64 the JS will truncate the number so both functions are required

crates/superposition_types/src/database/models/experimentation.rs
Datron: This is used ExperimentGroup to respond, and we don't want want loss of precision in those responses

crates/superposition_types/src/database/models/experimentation.rs
Datron: Why accept both? Let systems send a string, why support multiple ways to do this?

ayushjain17: I was of the opinion that, systems which support bigint, they should be allowed to send bigints 

Also, when someone sees the type of the API they would see an array of i64 but, instead, it takes an array of String
Which I feel is somewhat misleading, specifically when the type used here is not a custom type which could have indicated other way

crates/superposition_types/src/database/models/experimentation.rs
ayushjain17: For responding should we rather use a different Response type ?
Just like we have for Experiment and Workspace for this exact same reason

PR #526, opened 2025-05-26
feat: experiment group integration

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/superposition_types/migrations/2025-05-22-072420_experiment_group_integration/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TABLE localorg_test.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
```
ayushjain17:
```suggestion
ALTER TABLE public.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
```

crates/superposition_types/migrations/2025-05-22-072420_experiment_group_integration/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE localorg_test.experiments DROP COLUMN IF EXISTS experiment_group_id;
```
ayushjain17:
```suggestion
ALTER TABLE public.experiments DROP COLUMN IF EXISTS experiment_group_id;
```

crates/frontend/src/pages/experiment_list/utils.rs
```diff
@@ -156,6 +156,24 @@ pub fn experiment_table_columns(
...
+                    other => other.to_string(),
+                };
+
+                view! {
+                    <span>{label}</span>
+                }
+                .into_view()
+            },
+            ColumnSortable::No,
+            Expandable::Enabled(100),
```
ayushjain17: shouldn't this column be disabled ?
as it would be a static length string only which would not even reach 100 chars

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -374,9 +381,15 @@ pub fn experiment_list() -> impl IntoView {
                 current_tenant.to_string(),
                 org_id.clone(),
             );
+            let experiment_groups_future =
+                fetch_experiment_groups(current_tenant.to_string(), org_id.clone());
```
ayushjain17: why do we need to fetch this when we are in the list experiments page ?

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -526,6 +541,8 @@ pub fn experiment_list() -> impl IntoView {
                                     default_config=def_conf.clone()
                                     handle_submit=handle_submit_experiment_form
                                     metrics=workspace_settings.get_value().metrics
+                                    experiment_group_id=None
+                                    experiment_groups=exp_groups.clone()
```
ayushjain17: instead of sending this and fetching it in the list exp page, wont it be better to rather fetch it inside the component itself

crates/frontend/src/pages/experiment_list.rs
```diff
@@ -526,6 +541,8 @@ pub fn experiment_list() -> impl IntoView {
                                     default_config=def_conf.clone()
                                     handle_submit=handle_submit_experiment_form
                                     metrics=workspace_settings.get_value().metrics
+                                    experiment_group_id=None
```
ayushjain17: you can set the default value as `None` that way you wont need to send `None` from all the places

crates/frontend/src/pages/context_override.rs
```diff
@@ -440,7 +452,8 @@ pub fn context_override() -> impl IntoView {
                     &default_config_filters,
                     current_tenant.to_string(),
                     org_id.clone()
-                )
+                ),
+                fetch_experiment_groups(current_tenant.to_string(), org_id.clone(),),
```
ayushjain17: same for all the pages, this resource is local to the form, why block all the pages for this, which should be fetched on demand only

reduces most of the change log

crates/frontend/src/components/experiment_form/types.rs
ayushjain17, line 1: please revert

crates/frontend/src/api.rs
```diff
@@ -660,3 +661,25 @@ pub async fn execute_autocomplete_function(
...
     Ok(result)
 }
+
+pub async fn fetch_experiment_groups(
+    tenant: String,
+    org_id: String,
+) -> Result<PaginatedResponse<ExperimentGroup>, ServerFnError> {
+    let client = reqwest::Client::new();
+    let host = use_host_server();
+
```
ayushjain17: should we rather use this way for doing it ?
```
    let response = request(
        url,
        reqwest::Method::POST,
        Some(payload),
        construct_request_headers(&[("x-tenant", &tenant), ("x-org-id", &org_id)])?,
    )
    .await?;
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
         description,
         change_reason,
         metrics: req.metrics.clone().unwrap_or(workspace_settings.metrics),
+        experiment_group_id: req.experiment_group_id.unwrap_or(None),
```
ayushjain17:
```suggestion
        experiment_group_id: req.experiment_group_id,
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -23,12 +23,16 @@ use service_utils::{
         WorkspaceContext,
     },
 };
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{HashMap, HashSet},
+    vec,
+};
```
ayushjain17: can we move this import for `std` crate above all the imports and give a newline between these

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
...
-        .returning(Experiment::as_returning())
-        .schema_name(&workspace_request.schema_name)
-        .get_results(&mut conn)?;
+    let mut inserted_experiments =
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            let inserted_experiments = diesel::insert_into(experiments)
+                .values(&new_experiment)
+                .returning(Experiment::as_returning())
+                .schema_name(&workspace_request.schema_name)
+                .get_results(transaction_conn)?;
```
ayushjain17: can we fix this as well
```suggestion
                .get_result(transaction_conn)?;
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
...
+                    transaction_conn,
+                    &workspace_request.schema_name,
+                    user.clone(),
+                )?;
+            }
+
+            Ok(inserted_experiments)
+        })?;
 
     let inserted_experiment: Experiment = inserted_experiments.remove(0);
```
ayushjain17: this line would also not be needed then

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
...
         metrics: req.metrics.clone().unwrap_or(workspace_settings.metrics),
+        experiment_group_id: req.experiment_group_id.unwrap_or(None),
     };
 
-    let mut inserted_experiments = diesel::insert_into(experiments)
-        .values(&new_experiment)
-        .returning(Experiment::as_returning())
-        .schema_name(&workspace_request.schema_name)
-        .get_results(&mut conn)?;
+    let mut inserted_experiments =
```
ayushjain17: also, would need to update this as well
```suggestion
    let inserted_experiment =
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
...
+                .values(&new_experiment)
+                .returning(Experiment::as_returning())
+                .schema_name(&workspace_request.schema_name)
+                .get_results(transaction_conn)?;
+
+            if let Some(Some(experiment_group_id)) = &req.experiment_group_id {
+                add_members(
+                    experiment_group_id,
+                    ExpGroupMemberRequest {
+                        change_reason: req.change_reason.clone(),
```
ayushjain17: this change reason will not be same as the request change reason, this is a system action
```suggestion
                        change_reason: ChangeReason::try_from(format!("Adding experiment {experiment_id} to the group, while creating the experiment.")),
```

crates/superposition_types/src/api/experiments.rs
```diff
@@ -315,6 +318,28 @@ pub struct OverrideKeysUpdateRequest {
     pub description: Option<Description>,
     #[serde(default = "ChangeReason::default")]
     pub change_reason: ChangeReason,
+    #[serde(default, deserialize_with = "deserialize_experiment_group_id")]
+    pub experiment_group_id: Option<Option<i64>>,
```
Datron: Why not just Option<i64>?
```suggestion
    pub experiment_group_id: Option<i64>,
```

crates/superposition_types/src/api/experiments.rs
```diff
@@ -112,6 +114,7 @@ pub struct ExperimentCreateRequest {
     pub description: Description,
     #[serde(default = "ChangeReason::default")]
     pub change_reason: ChangeReason,
+    pub experiment_group_id: Option<Option<i64>>,
```
ayushjain17: for create you don't need it to be `Option<Option<i64>>`
also, add the deserialiser and serialiser which @Datron had added for passing the experiment id as i64
```suggestion
    pub experiment_group_id: Option<i64>,
```

crates/superposition_types/src/api/experiments.rs
```diff
@@ -315,6 +318,28 @@ pub struct OverrideKeysUpdateRequest {
     pub description: Option<Description>,
     #[serde(default = "ChangeReason::default")]
     pub change_reason: ChangeReason,
+    #[serde(default, deserialize_with = "deserialize_experiment_group_id")]
+    pub experiment_group_id: Option<Option<i64>>,
+}
+
+pub fn deserialize_experiment_group_id<'de, D>(
```
ayushjain17: can you verify once if this exact logic exists or not, if yes, then can we reuse that

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
...
-        .get_results(&mut conn)?;
+    let mut inserted_experiments =
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            let inserted_experiments = diesel::insert_into(experiments)
+                .values(&new_experiment)
+                .returning(Experiment::as_returning())
+                .schema_name(&workspace_request.schema_name)
+                .get_results(transaction_conn)?;
+
+            if let Some(Some(experiment_group_id)) = &req.experiment_group_id {
```
ayushjain17: why is this Option of Option ?

crates/superposition_types/src/api/experiments.rs
Datron: You'll have to handle the case where there is loss of precision in JS land (probably happens in other lands too). Have the user send string and convert it to i64. There is a function in experiment_groups that does this
```suggestion
    pub experiment_group_id: Option<i64>,
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,13 +302,32 @@ async fn create(
...
+                add_members(
+                    experiment_group_id,
+                    ExpGroupMemberRequest {
+                        change_reason: req.change_reason.clone(),
+                        member_experiment_ids: vec![experiment_id],
+                    },
+                    transaction_conn,
+                    &workspace_request.schema_name,
+                    user.clone(),
```
ayushjain17: can we please update the `add_member` function to accept the reference of `User` and borrow the data, instead of cloning

crates/frontend/src/pages/experiment_list/utils.rs
```diff
@@ -156,6 +156,24 @@ pub fn experiment_table_columns(
...
+            default_column_formatter,
+        ),
```
Datron: I think we already have too many columns. Lets show this in experiments page

crates/frontend/src/api.rs
```diff
@@ -660,3 +661,25 @@ pub async fn execute_autocomplete_function(
...
+        .header("x-tenant", &tenant)
+        .header("x-org-id", org_id)
+        .send()
+        .await
+        .map_err(|e| ServerFnError::new(e.to_string()))?
+        .json()
+        .await
+        .map_err(|e| ServerFnError::new(e.to_string()))?;
+
+    Ok(response)
```
Datron: If the experiment groups UI PR gets merged, you can use that function rather than this. You need to take pagination params probably to show all groups

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1289,18 +1316,58 @@ async fn update_overrides(
...
+                {
+                    remove_members(
+                        experiment_group_id,
+                        ExpGroupMemberRequest {
+                            change_reason: change_reason.clone(),
+                            member_experiment_ids: vec![experiment_id],
+                        },
+                        transaction_conn,
+                        &workspace_request.schema_name,
+                        user.clone(),
```
ayushjain17: same for this as well

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1289,18 +1316,58 @@ async fn update_overrides(
...
+                            &workspace_request.schema_name,
+                            user.clone(),
+                        )?;
+                    }
+                } else if let Some(experiment_group_id) = &experiment.experiment_group_id
+                {
+                    remove_members(
+                        experiment_group_id,
+                        ExpGroupMemberRequest {
+                            change_reason: change_reason.clone(),
```
ayushjain17: change reason in this section should also be different

crates/frontend/src/components/experiment_form.rs
```diff
@@ -210,6 +230,44 @@ pub fn experiment_form(
                 metrics=metrics_rws.get_untracked()
                 on_change=Callback::new(move |metrics| metrics_rws.set(metrics))
             />
+
+            {move || {
+                let mut experiment_options: Vec<Option<ExperimentGroup>> = experiment_groups
+                    .iter()
+                    .map(|group| Some(group.clone()))
+                    .collect();
+                experiment_options.insert(0, None);
```
ayushjain17: this way you wont need to derive PartialEq on ExperimentGroup, as you dont even need to store the entire experiment group, only the group id is needed
```suggestion
                let mut experiment_options: Vec<Option<i64>> = experiment_groups
                    .iter()
                    .map(|group| Some(group.id.clone()))
                    .collect();
                experiment_options.insert(0, None);
```

crates/frontend/src/types.rs
```diff
@@ -206,3 +206,28 @@ impl DropdownOption for PayloadVersion {
 pub type AutoCompleteCallback = leptos::Callback<(String, WriteSignal<Vec<String>>), ()>;
 
 pub type AutoCompleteCallbacks = std::collections::HashMap<String, AutoCompleteCallback>;
+
+impl DropdownOption for ExperimentGroup {
```
ayushjain17: not sure why both of these are needed

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -280,7 +283,7 @@ pub struct EventLog {
     pub query: String,
 }
 
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
ayushjain17: I guess this change is not needed, after resolving the above comments

ayushjain17 (changes requested):

crates/superposition_types/src/api/experiments.rs
ayushjain17: this is for update a data of optional type

crates/frontend/src/pages/experiment_list.rs
mahatoankitkumar: We are already fetching the dimensions and default_configs there.

crates/frontend/src/pages/context_override.rs
mahatoankitkumar: i get your point, but maybe this can be done in a cleanup (I'll do the honours for this), not in the scope of this PR though

crates/frontend/src/components/experiment_form/types.rs
mahatoankitkumar: cargo fmt did it though

crates/frontend/src/components/experiment_form.rs
mahatoankitkumar: need this as showing the group name instead of id

crates/superposition_types/src/api/experiments.rs
mahatoankitkumar: no its for function_name, not similar though

crates/frontend/src/api.rs
mahatoankitkumar: Let this be, i myself have added that in my bucketing PR, even if this goes will be overwritten by your PR, thats fine.

crates/frontend/src/pages/experiment_list/utils.rs
mahatoankitkumar: Am showing there too, i guess better to show it here too

crates/superposition_types/src/api/experiments.rs
mahatoankitkumar: yes

mahatoankitkumar:
> Mostly fine, but try to remove Option<Option> and simplify this. Also, this PR might have to wait for the UI PR to go since it has a few utilities that can be used here. Or a follow up PR where things can be refactored

Better not to wait, overwrite in your PR, anyways i have a PR lined up after this

crates/frontend/src/components/experiment_form.rs
ayushjain17: not a good idea, name is not unique, you will have collisions

crates/frontend/src/pages/context_override.rs
ayushjain17: this does not require for a separate cleanup right, you are adding a new resource, just fetch it in the correct way
fetch this in the form and others if needed can be moved to form later on, no ?

crates/frontend/src/components/experiment_form/types.rs
ayushjain17: no its not on cargo fmt, if you delete the separating newline, then cargo fmt changes it

crates/superposition_types/src/api/experiments.rs
ayushjain17: I meant few were added by @Datron for exp groups, most probably those can be reused here

crates/frontend/src/pages/experiment_list/utils.rs
Datron: If you're keen on showing this, show it upfront after traffic percentage or something. 

My suggestion is to remove it, since no user will at a glance make sense of experiments in a group in the listing page since its paginated. If they are interested in that experiment, they will click on it to reveal more information.

crates/superposition_types/src/api/experiments.rs
Datron: This is in the diesel docs:
```
If you wanted to assign NULL instead, you can either specify #[diesel(treat_none_as_null = true)] on the struct, or you can have the field be of type Option<Option<T>>. Diesel doesn’t currently provide a way to explicitly assign a field to its default value, though it may be provided in the future.
```
can we use 
```
#[diesel(treat_none_as_null = true)]
```

crates/superposition_types/src/api/experiments.rs
mahatoankitkumar: no its not there

crates/superposition_types/src/api/experiments.rs
ayushjain17: How do you manage the case when you don't want to update the field ?

We have discussed this long back, after that itself we started using this Option of Option thing
This is not something new, just to be clear

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1289,18 +1319,61 @@ async fn update_overrides(
...
+                        experiments::change_reason.eq(change_reason),
+                        experiments::description
+                            .eq(description.unwrap_or(experiment.description)),
+                        experiments::metrics.eq(updated_metrics),
+                        experiments::last_modified.eq(Utc::now()),
+                        experiments::last_modified_by.eq(user.get_email()),
+                        experiments::experiment_group_id
+                            .eq(experiment_group_id
+                                .unwrap_or(experiment.experiment_group_id)),
+                    ))
```
ayushjain17: are we not using the update request as change set here ?

mahatoankitkumar: no

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1289,18 +1319,61 @@ async fn update_overrides(
...
-        .get_result::<Experiment>(&mut conn)?;
+    let updated_experiment =
+        conn.transaction::<_, superposition::AppError, _>(|transaction_conn| {
+            if let Some(experiment_group_id) = &experiment_group_id {
+                if let Some(experiment_group_id) = experiment_group_id {
+                    if experiment.experiment_group_id.is_none() {
+                        add_members(
+                            experiment_group_id,
+                            ExpGroupMemberRequest {
+                                change_reason: change_reason.clone(),
```
ayushjain17: change reason not updated here ?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -295,15 +303,36 @@ async fn create(
...
+                .values(&new_experiment)
+                .returning(Experiment::as_returning())
+                .schema_name(&workspace_request.schema_name)
+                .get_result(transaction_conn)?;
+
+            if let Some(experiment_group_id) = &req.experiment_group_id {
+                add_members(
+                    experiment_group_id,
+                    ExpGroupMemberRequest {
+                        change_reason: ChangeReason::try_from(format!("Adding experiment {experiment_id} to the group, while creating the experiment.")).map_err(|err| {
```
ayushjain17, line 322: is this not failing the build for formatting  ?

smithy/models/experiments.smithy
```diff
@@ -23,6 +23,7 @@ resource Experiments {
         chosen_variant: String
         description: String
         change_reason: String
+        experiment_group_id: Integer
```
ayushjain17:
```suggestion
        experiment_group_id: String
```

smithy/models/experiments.smithy
ayushjain17, line 1: tests not updated

crates/superposition_types/src/api/experiments.rs
```diff
@@ -96,6 +97,7 @@ impl From<Experiment> for ExperimentResponse {
             change_reason: experiment.change_reason,
             metrics: experiment.metrics,
             metrics_url,
+            experiment_group_id: experiment.experiment_group_id,
```
ayushjain17:
```suggestion
            experiment_group_id: experiment.experiment_group_id.to_string(),
```

crates/superposition_types/src/api/experiments.rs
```diff
@@ -47,6 +47,7 @@ pub struct ExperimentResponse {
     pub change_reason: ChangeReason,
     pub metrics: Metrics,
     pub metrics_url: Option<String>,
+    pub experiment_group_id: Option<i64>,
```
ayushjain17:
```suggestion
    pub experiment_group_id: Option<String>,
```

crates/superposition_types/src/api/experiments.rs
```diff
@@ -112,6 +114,7 @@ pub struct ExperimentCreateRequest {
     pub description: Description,
     #[serde(default = "ChangeReason::default")]
     pub change_reason: ChangeReason,
+    pub experiment_group_id: Option<i64>,
```
ayushjain17, line 121: we need to convert this from String to i64 in deserialisation, can you please refer to what we are doing in experiment groups for experiment ids, these would fail in the JS world

crates/superposition_types/src/api/experiments.rs
```diff
@@ -315,6 +318,30 @@ pub struct OverrideKeysUpdateRequest {
     pub description: Option<Description>,
     #[serde(default = "ChangeReason::default")]
     pub change_reason: ChangeReason,
+    #[serde(default, deserialize_with = "deserialize_experiment_group_id")]
+    pub experiment_group_id: Option<Option<i64>>,
```
ayushjain17: we need to do the same string to i64 conversion here as well

crates/superposition_types/src/database/models/experimentation.rs
ayushjain17, line 1: do we need the `Debug` in `ExperimentGroup` and `PartialEq` in `TrafficPercentage` changes in this file ?

crates/frontend/src/components/experiment_form.rs
```diff
@@ -210,6 +231,39 @@ pub fn experiment_form(
                 metrics=metrics_rws.get_untracked()
                 on_change=Callback::new(move |metrics| metrics_rws.set(metrics))
             />
+
+            <Suspense>
```
ayushjain17: I think you need to add a fallback here of Block type, to constrain the Shimmer within the scope, the shimmer is missing altogether right now
something like this:
``` 
view! { <Skeleton variant=SkeletonVariant::Block style_class="h-10".to_string() /> }
```

crates/frontend/src/components/experiment_form.rs
```diff
@@ -90,6 +97,18 @@ pub fn experiment_form(
...
+    let (experiment_group_id_rs, experiment_group_id_ws) =
+        create_signal(experiment_group_id);
+
+    let experiment_groups_resource: Resource<(String, String), Vec<ExperimentGroup>> =
+        create_blocking_resource(
+            move || (tenant_rws.get().0, org_rws.get().0),
+            |(current_tenant, org)| async move {
+                fetch_experiment_groups(current_tenant.to_string(), org.clone())
+                    .await
+                    .map_or_else(|_| vec![], |data| data.data)
```
ayushjain17: no ?
```suggestion
                    .map(|data| data.data).unwrap_or_default()
```

crates/frontend/src/components/experiment_form.rs
ayushjain17: Just FYI, as per the current structure of this code, the section for the Dropdown will be in a shimmer till it loads, meanwhile other section of the form would still be available to use

ayushjain17 (changes requested):

crates/experimentation_platform/src/api/experiments/handlers.rs
mahatoankitkumar: no

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -378,3 +379,17 @@ where
         .collect::<Result<Vec<i64>, D::Error>>()?;
     Ok(Some(numbers))
 }
+
+pub fn i64_option_deserialize<'de, D>(deserializer: D) -> Result<Option<i64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let string_value: Option<String> = Option::deserialize(deserializer)?;
```
ayushjain17: needs to support Number as well

clients/tests/typescript/tests/experiments.test.ts
ayushjain17, line 1: we need to add tests to verify that whether the conditions in which add and remove should pass and fail are working or no

crates/frontend/src/components/experiment_form/utils.rs
```diff
@@ -66,14 +75,19 @@ pub async fn update_experiment(
...
+    let experiment_group_id = experiment_group_id
+        .map(|id| id.parse::<i64>())
+        .transpose()
+        .map_err(|_| "Invalid experiment group ID".to_string())?;
     let payload = OverrideKeysUpdateRequest {
         variants: Result::<Vec<VariantUpdateRequest>, String>::from_iter(variants)?,
         metrics,
         description: Some(Description::try_from(description)?),
         change_reason: ChangeReason::try_from(change_reason)?,
+        experiment_group_id: experiment_group_id.map(I64Update::Add).map(I64Update::from),
```
ayushjain17: this is wrong, over here one cannot delete the data
rather do how it was done for config_versions, pass the value of experiment_group_id as `serde_json::Value`
`Value::String(id)` when the data is present and `Value::Null` when the data is to be removed

and over here write this:
```suggestion
        experiment_group_id: Some(
            serde_json::from_value(experiment_group_id)
                .map_err(|e| format!("Invalid experiment group id: {e}"))?,
        ),
```

and changing the function parameter:
```
        experiment_group_id: Value,
```

also, this same above suggested changes would be needed to be done in the create function as well

ayushjain17 (changes requested):

crates/frontend/src/pages/experiment_list/utils.rs
```diff
@@ -156,6 +156,24 @@ pub fn experiment_table_columns(
             Expandable::Enabled(100),
             default_column_formatter,
         ),
+        Column::new(
+            "experiment_group_id".to_string(),
+            false,
+            |value: &str, _| {
+                let label = match value {
+                    "null" => "¯\\_(ツ)_/¯".to_string(),
```
Datron, line 164:
```suggestion
                    "null" => "¯\_(ツ)_/¯".to_string(),
```

crates/superposition_types/src/api.rs
```diff
@@ -8,3 +18,89 @@ pub mod experiments;
...
+                }
+            }
+        }
+        Value::Null => Ok(Some(I64Update::Remove)),
+        _ => {
+            log::error!("Expected a bigint, bigint string or null literal as the value.");
+            Err(serde::de::Error::custom(
+                "Expected a bigint, bigint string or null literal as the value.",
+            ))
+        }
```
Datron: @mahatoankitkumar we discussed this, just adding a comment here:
We don't need to parse for numbers, since different callers can end up sending us a wrong value due to loss of precision on their end. We don't need to handle the case of dealing with numbers if we only support strings in the adding experiment to group interface. Even languages with matching precision to rust will be able to send strings

mahatoankitkumar: @ayushjain17

ayushjain17: @Datron but its rather better to support extra than not supporting
for folks who support i64 they would unnecessarily convert first to string and then send

better to be make stuff easier for folks to use, no ?

mahatoankitkumar: @knutties can you please help resolve this conflict.

crates/superposition_types/src/api.rs
```diff
@@ -8,3 +18,89 @@ pub mod experiments;
 pub mod functions;
 pub mod webhook;
 pub mod workspace;
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Nullable<BigInt>))]
+pub enum I64Update {
```
knutties, line 25: What is the purpose of this enum ?

mahatoankitkumar: This was already in use for configVersionUpdate, just renamed it. 
Used this as it already had the serialization and deserialization implemented for it and served the same purpose.

crates/superposition_types/src/api.rs
Datron: Documenting the resolution here:
We will support only string for this interface

crates/superposition_types/src/api.rs
```diff
@@ -8,3 +18,126 @@ pub mod experiments;
...
+                Err("Expected a bigint or bigint string as the value.".to_string())
+            }
+        },
+        Value::Null => Ok(None),
+        _ => {
+            log::error!("Expected a bigint, bigint string or null literal as the value.");
+            Err(
+                "Expected a bigint, bigint string or null literal as the value."
+                    .to_string(),
+            )
```
ayushjain17:
```suggestion
            log::error!("Expected a bigint string or null literal as the value.");
            Err(
                "Expected a bigint string or null literal as the value."
                    .to_string(),
            )
```

crates/superposition_types/src/api.rs
```diff
@@ -8,3 +18,126 @@ pub mod experiments;
...
+    }
+}
+
+pub fn option_i64_from_value(value: Value) -> Result<Option<i64>, String> {
+    match value {
+        Value::String(val) => match val.parse::<i64>() {
+            Ok(config_version) => Ok(Some(config_version)),
+            Err(_) => {
+                log::error!("Expected a bigint or bigint string as the value.");
+                Err("Expected a bigint or bigint string as the value.".to_string())
```
ayushjain17:
```suggestion
                log::error!("Expected a bigint string as the value.");
                Err("Expected a bigint string as the value.".to_string())
```

crates/superposition_types/src/api.rs
```diff
@@ -8,3 +18,123 @@ pub mod experiments;
...
+                    }
+                }
+            }
+        }
+        Value::Null => Ok(Some(I64Update::Remove)),
+        _ => {
+            log::error!("Expected a bigint, bigint string or null literal as the value.");
+            Err(serde::de::Error::custom(
+                "Expected a bigint, bigint string or null literal as the value.",
+            ))
```
ayushjain17:
```suggestion
            log::error!("Expected a bigint string or null literal as the value.");
            Err(serde::de::Error::custom(
                "Expected a bigint string or null literal as the value.",
            ))
```

crates/superposition_types/src/api.rs
```diff
@@ -8,3 +18,123 @@ pub mod experiments;
...
+            if &val == "null" {
+                Ok(Some(I64Update::Remove))
+            } else {
+                match val.parse::<i64>() {
+                    Ok(config_version) => Ok(Some(I64Update::Add(config_version))),
+                    Err(_) => {
+                        log::error!("Expected a bigint or bigint string as the value.");
+                        Err(serde::de::Error::custom(
+                            "Expected a bigint or bigint string as the value.",
+                        ))
```
ayushjain17:
```suggestion
                        log::error!("Expected a bigint string as the value.");
                        Err(serde::de::Error::custom(
                            "Expected a bigint string as the value.",
                        ))
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+            if experiment_group_id == current_group_id =>
+        {
+            Ok(Some(*experiment_group_id))
+        }
+
+        // Case 5: No group specified in update, keep current state
+        (None, current_group_id) => Ok(*current_group_id),
+
+        // Case 6: All other cases (no change needed)
+        _ => Ok(*current_group_id),
```
ayushjain17:
```suggestion
        // Case 4: All other cases (no change needed)
        _ => Ok(*current_group_id),
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+    experiment: &Experiment,
+    new_group_id: &Option<I64Update>,
+    current_group_id: &Option<i64>,
+    transaction_conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    schema_name: &SchemaName,
+    user: &User,
+) -> superposition::Result<Option<i64>> {
+    let experiment_id = experiment.id;
+
+    fn create_change_reason(
```
ayushjain17: do we need this separately ?

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+        // Case 5: No group specified in update, keep current state
+        (None, current_group_id) => Ok(*current_group_id),
+
+        // Case 6: All other cases (no change needed)
+        _ => Ok(*current_group_id),
+    }
+}
+
+pub fn validate_and_add_experiment_group_id(
+    member_experiment_ids: &[Experiment],
```
ayushjain17:
```suggestion
    member_experiments: &[Experiment],
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+    for experiment in member_experiment_ids {
+        let experiment_id = experiment.id;
+
+        if let Some(existing_group_id) = experiment.experiment_group_id {
+            let error_msg = if existing_group_id == *new_experiment_group_id {
+                format!(
+                    "Experiment with id {} already has the same experiment group assigned: {}",
+                    experiment_id, existing_group_id
+                )
+            } else {
```
ayushjain17: this is not an error, this is for maintaining idempotency

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+                format!(
+                    "Experiment with id {} already has a different experiment group assigned: {}",
```
ayushjain17:
```suggestion
                    "Experiment {} is already a part of a different experiment group {}",
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+            log::error!("Failed to convert change reason: {}", err);
+            unexpected_error!("Failed to convert change reason")
+        })?;
+
+        // Update experiment
+        diesel::update(experiments::experiments.find(experiment_id))
+            .set((
+                experiments::experiment_group_id
+                    .eq(I64Update::Add(*new_experiment_group_id)),
+                experiments::last_modified_by.eq(&user.email),
```
ayushjain17:
```suggestion
                experiments::last_modified_by.eq(&user.get_email()),
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+    member_experiment_ids: &[i64],
+    experiment_group_id: &i64,
+    schema_name: &SchemaName,
+    transaction_conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    user: &User,
+) -> superposition::Result<()> {
+    let now = Utc::now();
+
+    for experiment_id in member_experiment_ids {
+        let experiment: Experiment = experiments::experiments
```
ayushjain17: can we use the function to fetch experiments

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -526,3 +533,212 @@ pub async fn fetch_webhook_by_event(
...
+        ))
+        .map_err(|err| {
+            log::error!("Failed to convert change reason: {}", err);
+            unexpected_error!("Failed to convert change reason")
+        })?;
+
+        diesel::update(experiments::experiments.find(experiment_id))
+            .set((
+                experiments::experiment_group_id.eq(I64Update::Remove),
+                experiments::last_modified_by.eq(&user.email),
```
ayushjain17:
```suggestion
                experiments::last_modified_by.eq(&user.get_email()),
```

clients/tests/typescript/tests/experiment_groups.test.ts
```diff
@@ -712,35 +746,6 @@ describe("Experiment Groups API Integration Tests", () => {
             );
             expect(response.member_experiment_ids).not.toContain(expValid2Id);
         });
-        
-        test("should successfully handle removing non-existent members", async () => {
```
ayushjain17, line 716: can we please add this back

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -51,21 +41,42 @@ pub fn validate_members(
                 .join(", ")
         ));
     }
+
+    let members: Vec<Experiment> = experiments::experiments
+        .filter(experiments::id.eq_any(&new_members))
+        .filter(experiments::status.eq(ExperimentStatusType::CREATED))
```
```suggestion
        .filter(experiments::id.eq_any(&new_members).and(experiments::status.eq(ExperimentStatusType::CREATED)))
```

crates/experimentation_platform/src/api/experiments/helpers.rs
ayushjain17: that has extra validation which is not valid for remove case

PR #527, opened 2025-05-27
fix: Resolving resolve page: Removing the All Contexts section

vraghunandhan:
Fixing resolve page - Removing some unused sections


mahatoankitkumar (changes requested):
Page doesn't open itself
Panics

PR #530, opened 2025-06-09
fix: latest support in config params

pratikmishra356:
## Problem
Currently we get_config and get_resolved_config api accepts version as query params , also api handler checks if 
version is being set at the workspace level , so if any version set at workspace level then there isn't a way to get the latest config .

## Solution
Query params for the above to api will now accept "latest" as version param , based on which the latest config will be served

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -67,12 +67,43 @@ pub fn endpoints() -> Scope {
     scope
 }
 
-fn validate_version_in_params(
+fn get_config_version_from_workspace(
+    workspace_context: &WorkspaceContext,
+    conn: &mut DBConnection,
+) -> Option<i64> {
+    match workspaces::dsl::workspaces
+        .select(workspaces::config_version)
+        .filter(
+            workspaces::organisation_id
+                .eq(&workspace_context.organisation_id.0)
+                .and(workspaces::workspace_name.eq(&workspace_context.workspace_id.0)),
+        )
+        .get_result::<Option<i64>>(conn)
+    {
+        Ok(version) => version,
+        Err(e) => {
+            log::error!(
+                "Failed to get config_version for org_id: {}, workspace_name: {} — {:?}",
+                workspace_context.organisation_id.0,
+                workspace_context.workspace_id.0,
+                e
+            );
+            None
+        }
+    }
+}
+fn get_config_version(
     query_params_map: &mut Map<String, Value>,
+    workspace_context: &WorkspaceContext,
+    conn: &mut DBConnection,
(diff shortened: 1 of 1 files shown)


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -767,6 +791,7 @@ async fn get_resolved_config(
     db_conn: DbConnection,
     query_map: superposition_query::Query<QueryMap>,
     schema_name: SchemaName,
+    workspace_context: WorkspaceContext,
```
ayushjain17: use only workspace_context, schema_name is redundant
```suggestion
    workspace_context: WorkspaceContext,
```

Datron:
Needs a proper description

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -781,9 +805,16 @@ async fn get_resolved_config(
         return Ok(HttpResponse::NotModified().finish());
     }
 
-    let mut config_version = validate_version_in_params(&mut query_params_map)?;
-    let mut config =
-        generate_config_from_version(&mut config_version, &mut conn, &schema_name)?;
+    let mut config_version = get_config_version_from_params(
```
ayushjain17:
```suggestion
    let mut config_version = get_config_version(
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -67,20 +67,52 @@ pub fn endpoints() -> Scope {
...
             version.as_i64().map_or_else(
                 || {
                     log::error!("failed to decode version as integer: {}", version);
                     Err(bad_argument!("version is not of type integer"))
                 },
                 |v| Ok(Some(v)),
             )
-        })
+        },
+    )
```
ayushjain17: this section can be simplified or refactored, but not a blocker of sorts

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -67,20 +67,52 @@ pub fn endpoints() -> Scope {
...
+        .filter(
+            workspaces::organisation_id
+                .eq(&workspace_context.organisation_id.0)
+                .and(workspaces::workspace_name.eq(&workspace_context.workspace_id.0)),
+        )
+        .get_result::<Option<i64>>(conn)
+    {
+        Ok(version) => version,
+        Err(e) => {
+            log::info!(
```
sauraww: Why is this log::info ? Shouldn't it be log::error ?

sauraww (changes requested):

PR #531, opened 2025-06-09
fix: Show context filters

ayushjain17:
Show context filters


crates/frontend/src/pages/context_override.rs
```diff
@@ -78,186 +83,6 @@ enum FormMode {
     Experiment(ExperimentType, String),
 }
 
-#[component]
-fn context_filter_drawer(
```
ayushjain17, line 82: moved this to filter.rs file

PR #532, opened 2025-06-09
feat: Add change log popup for override, default-config, dimension, type-template, webhook and experiment changes

ayushjain17:
Add change log popup for override, default-config, dimension, type-template, webhook and experiment changes

https://github.com/user-attachments/assets/4ce651c8-01af-441f-a644-bd2733342b7f

### DIFF
--- a/crates/frontend/src/api.rs
+++ b/crates/frontend/src/api.rs
@@ -106,7 +106,8 @@ pub async fn fetch_snapshots(
 
 pub async fn delete_context(
     context_id: String,
-    org_id: String,
+    tenant: &str,
+    org_id: &str,
 ) -> Result<(), ServerFnError> {
     let client = reqwest::Client::new();
     let host = use_host_server();
@@ -721,6 +722,12 @@ pub async fn get_webhook(
 }
 
 pub mod experiment_groups {
+    use superposition_types::{
+        database::models::experimentation::TrafficPercentage, Condition, Exp,
+    };
+
+    use crate::logic::Conditions;
+
     use super::*;
 
     pub async fn fetch_all(
@@ -770,11 +777,26 @@ pub mod experiment_groups {
             .map_err(ServerFnError::new)
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub async fn create(
-        payload: ExpGroupCreateRequest,
+        name: String,
+        description: String,
+        change_reason: String,
+        traffic_percentage: i32,
+        member_experiment_ids: Option<Vec<i64>>,
+        conditions: Conditions,
         tenant: &str,
         org_id: &str,
     ) -> Result<ExperimentGroup, String> {
+        let payload = ExpGroupCreateRequest {
+            name,
+            description: Description::try_from(description)?,
+            change_reason: ChangeReason::try_from(change_reason)?,
+            traffic_percentage: TrafficPercentage::try_from(traffic_percentage)?,
+            member_experiment_ids,
+            context: Exp::<Condition>::try_from(conditions.as_context_json())
+                .map_err(|e| e.to_string())?,
+        };
         let host = use_host_server();
         let url = format!("{host}/experiment-groups");
 
@@ -787,6 +809,18 @@ pub mod experiment_groups {
         .await?;
 
         parse_json_response(response).await
+    }
+
+    pub fn try_update_payload(
+        traffic_percentage: i32,
+        description: String,
+        change_reason: String,
+    ) -> Result<ExpGroupUpdateRequest, String> {
+        Ok(ExpGroupUpdateRequest {
+            traffic_percentage: Some(TrafficPercentage::try_from(traffic_percentage)?),
+            description: Some(Description::try_from(description)?),
+            change_reason: ChangeReason::try_from(change_reason)?,
+        })
     }
 
     pub async fn update(
--- a/crates/frontend/src/components/input.rs
+++ b/crates/frontend/src/components/input.rs
@@ -622,7 +622,7 @@ pub fn input(
             Some(ref v) => {
                 view! { <Toggle value=*v on_change class name disabled /> }.into_view()
             }
-            None => view! { <Toggle value=false on_change class name /> }.into_view(),
+            None => view! { <Toggle value=false on_change class name disabled /> }.into_view(),
         },
         InputType::Select(ref options) => view! { <Select id name class value on_change disabled options=options.0.clone() /> }
         .into_view(),
--- a/crates/frontend/src/components/experiment_group_form.rs
+++ b/crates/frontend/src/components/experiment_group_form.rs
@@ -1,32 +1,30 @@
+use std::ops::Deref;
+
 use leptos::*;
-use serde_json::{json, Value};
+use serde_json::{json, Map, Number, Value};
 use superposition_types::{
     api::{
-        experiment_groups::{
-            ExpGroupCreateRequest, ExpGroupMemberRequest, ExpGroupUpdateRequest,
-        },
+        experiment_groups::{ExpGroupMemberRequest, ExpGroupUpdateRequest},
         workspace::WorkspaceResponse,
     },
     database::{
-        models::{
-            experimentation::{ExperimentGroup, TrafficPercentage},
-            ChangeReason, Description,
-        },
+        models::{experimentation::ExperimentGroup, ChangeReason},
         types::DimensionWithMandatory,
     },
-    Condition, Exp,
 };
 use web_sys::MouseEvent;
 
 use crate::{
-    api::experiment_groups::{add_members, create, update},
+    api::experiment_groups::{add_members, create, fetch, try_update_payload, update},
     components::{
         alert::AlertType,
         button::Button,
         change_form::ChangeForm,
+        change_summary::{ChangeLogPopup, ChangeSummary, JsonChangeSummary},
         context_form::ContextForm,
         form::label::Label,
         input::{Input, InputType},
+        skeleton::{Skeleton, SkeletonVariant},
     },
     logic::Conditions,
     providers::{alert_provider::enqueue_alert, editor_provider::EditorProvider},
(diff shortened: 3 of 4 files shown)


crates/frontend/src/components/default_config_form.rs
```diff
@@ -450,12 +467,191 @@ where
...
+            "Confirm Update",
+            "Are you sure you want to update this config?",
+            "Yes, Update",
+        ),
+        ChangeType::Delete => (
+            "Confirm Delete",
+            "Are you sure you want to delete this config? Action is irreversible.",
+            "Yes, Delete",
+        ),
+    };
```
mahatoankitkumar, line 534: I guess better to move it either inside the component or somewhere common as we are repeating this everytime we call the component.

crates/frontend/src/components/webhook_form.rs
```diff
@@ -1,28 +1,63 @@
...
     types::{OrganisationId, Tenant},
 };
 
+enum ResponseType {
+    UpdatePrecheck,
+    Response,
+}
+
+#[allow(clippy::too_many_arguments)]
+fn try_update_payload(
```
mahatoankitkumar, line 37: A single-use function here just adds extra lines without benefit.

ayushjain17: this is not just single use it is for ergonomics and has purpose

crates/frontend/src/components/default_config_form.rs
ayushjain17: not same
everyone is different

mahatoankitkumar: Just the topic seems to be different i.e. config, type_templates,etc, that can be configured.

crates/frontend/src/components/webhook_form.rs
mahatoankitkumar: Still being used at a single place, better to write it there itself, will still cost you less line of code.

crates/frontend/src/api.rs
```diff
@@ -713,11 +777,26 @@ pub mod experiment_groups {
...
+        name: String,
+        description: String,
+        change_reason: String,
+        traffic_percentage: i32,
+        member_experiment_ids: Option<Vec<i64>>,
+        conditions: Conditions,
         tenant: &str,
         org_id: &str,
     ) -> Result<ExperimentGroup, String> {
+        let payload = ExpGroupCreateRequest {
```
Datron, line 791: @ayushjain17 In the webhook update API call you removed arguments and used the payload struct
https://github.com/juspay/superposition/pull/532/files#diff-1d862f8a80591534a887703db2d42bd96e188c1c740ad2873061a3ed6de5f210L432
Can we also use the payload struct here?

crates/frontend/src/components/change_summary.rs
```diff
@@ -0,0 +1,242 @@
...
+        if value == "##DELETED##" {
+            view! { <span class="text-red-500 bg-red-100 px-2 py-1 rounded">"Deleted"</span> }
+            .into_view()
+        } else if value == "##NOT_PRESENT##" {
+            view! { <span class="text-[#505050] bg-[#cdcdcd] px-2 py-1 rounded">"Not Present"</span> }
+            .into_view()
+        } else {
+            default_formatter(value, _row)
+        }
+    };
```
Datron, line 31: Nitpick:
We could use a Enum here and convert ##DELETED## into the enum with from_str. Later when we change the diff representation we don't need to change code

crates/frontend/src/components/default_config_form.rs
```diff
@@ -31,8 +35,13 @@ use crate::{
 
 use self::utils::{create_default_config, update_default_config};
 
+enum ResponseType {
+    UpdatePrecheck,
+    Response,
+}
```
Datron, line 41: What does ResponseType present?

crates/frontend/src/components/default_config_form.rs
```diff
@@ -450,12 +467,191 @@ where
...
+            }
+        }}
+    }
+}
+
+#[derive(Clone)]
+pub enum ChangeType {
+    Delete,
+    Update(DefaultConfigUpdateRequest),
+}
```
Datron, line 501: Might be good to use generics here and put this in change_log_summary

```suggestion
#[derive(Clone)]
pub enum ChangeType<T> {
    Delete,
    Update(T),
}
```
That way you can define some behaviours on the enum that apply to every page

crates/frontend/src/components/dimension_form.rs
```diff
@@ -37,8 +43,13 @@ use crate::{
     utils::set_function,
 };
 
+enum ResponseType {
+    UpdatePrecheck,
+    Response,
+}
```
Datron, line 49: Should this be defined in the common utils file?

crates/frontend/src/components/dimension_form/types.rs
```diff
@@ -17,9 +17,9 @@ pub struct DimensionCreateReq {
...
-    pub dependencies: Vec<String>,
-    pub function_name: Option<String>,
-    pub autocomplete_function_name: Option<String>,
-    pub description: String,
+    pub dependencies: Option<Vec<String>>,
+    pub function_name: Option<Option<String>>,
+    pub autocomplete_function_name: Option<Option<String>>,
+    pub description: Option<String>,
     pub change_reason: String,
 }
```
Datron, line 25: Can we use the request type from superposition_types?

crates/frontend/src/components/input.rs
```diff
@@ -620,7 +620,7 @@ pub fn input(
     match r#type {
         InputType::Toggle => match value.as_bool() {
             Some(ref v) => {
-                view! { <Toggle value=*v on_change class name /> }.into_view()
+                view! { <Toggle value=*v on_change class name disabled /> }.into_view()
             }
             None => view! { <Toggle value=false on_change class name /> }.into_view(),
```
Datron:
```suggestion
            None => view! { <Toggle value=false on_change class name disabled /> }.into_view(),
```
The disabled option should apply to both cases right?

crates/frontend/src/components/webhook_form.rs
Datron: I have the same opinion, these functions try_update_payload just construct the struct, they don't need to be a separate function call.

crates/frontend/src/components/dimension_form/types.rs
ayushjain17: This has not yet been moved to superposition_types

The type movement work is not yet completed

crates/frontend/src/api.rs
ayushjain17: this is for a pattern, please refer to the place where these are being used

crates/frontend/src/components/change_summary.rs
ayushjain17: how can I send enum here ?
this is the Table's cell formatter which gives me only String value

crates/frontend/src/components/default_config_form.rs
ayushjain17: there are no common behaviours between anywhere, have already generalised as much possible or makes sense

crates/frontend/src/components/dimension_form.rs
ayushjain17: everywhere does not have the same type, and it is over generalising, every page is not expected to have same flow, for intermediate steps which one might want to add, that would add a lot of friction, which is simply unnecessary

crates/frontend/src/components/default_config_form.rs
ayushjain17: that whether the current received response is the final response or not, or is some intermediary response

crates/frontend/src/components/webhook_form.rs
ayushjain17: I would rather suggest you to check the usage and what it is providing
and just to be verbose, this is not simply a function to to move a piece of code out of another function, this wraps the type creation errors as a single response, and I don't need to handle each field's error separately

this is exactly same as me implementing `UpdateWebhookRequest: TryFrom<(bool, String, HttpMethod, PayloadVersion, CustomHeaders, Vec<WebhookEvent>, String, String)>`
which is ugly and meaningless, as now no positional params carry any meaning along with it

crates/frontend/src/components/change_summary.rs
Datron: > We could use a Enum here and convert ##DELETED## into the enum with from_str

crates/frontend/src/components/dimension_form.rs
Datron: Can you give me an example of where this ResponseType is not the same? We can ignore those, the rest you can generalize

crates/frontend/src/components/default_config_form.rs
Datron: You can define a common behaviour among them with a trait

crates/frontend/src/components/default_config_form.rs
Datron: What is an intermediary response? Do you make an API call before the actual API call?

crates/frontend/src/components/default_config_form.rs
ayushjain17: there is no common behaviour among them, making any further changes its like if one word is common across components make another wrapper component to have that one word in the component

crates/frontend/src/components/change_summary.rs
ayushjain17: can you please rather give a code suggestion, not able to understand how this makes sense

crates/frontend/src/components/default_config_form.rs
ayushjain17: its not always about API calls, you want to have a mechanism which gives you an option to flow specific steps irrespective of whether its a simple UI step or an API call

and I would re-iterate that please have a look at the usage once

crates/frontend/src/components/dimension_form.rs
ayushjain17: >every page is not expected to have same flow, for intermediate steps which one might want to add, that would add a lot of friction, which is simply unnecessary 

https://github.com/juspay/superposition/pull/532#discussion_r2191530776

mahatoankitkumar (approved):

PR #534, opened 2025-06-11
Feat/experiment bucketing

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


clients/tests/typescript/tests/experiments.test.ts
```diff
@@ -544,7 +544,7 @@ describe("Experiments API", () => {
             expect(out.id).toBe(experimentId1);
             expect(out.traffic_percentage).toBe(rampPercentage);
             expect(out.status).toBe(ExperimentStatusType.INPROGRESS);
-            expect(out.experiment_group_id).toBeUndefined();
+            experimentGroupId = out.experiment_group_id;
```
ayushjain17:
```suggestion
            expect(out.experiment_group_id).toBeDefined();
            experimentGroupId = out.experiment_group_id;
```

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
             operations,
             types::{
                 BulkOperation, BulkOperationResponse, ContextAction, ContextBulkResponse,
-                MoveReq, PutReq, WeightRecomputeResponse,
+                ContextValidationRequest, MoveReq, PutReq, WeightRecomputeResponse,
```
ayushjain17: why is this type being moved out of `superposition_types` ?

crates/context_aware_config/src/api/context/types.rs
```diff
@@ -43,6 +43,12 @@ pub struct MoveReq {
     pub change_reason: String,
 }
 
+#[cfg_attr(test, derive(Debug, PartialEq))]
+#[derive(Deserialize, Clone)]
+pub struct ContextValidationRequest {
+    pub context: Cac<Condition>,
+}
```
ayushjain17: why is this being moved here ?

crates/experimentation_client/src/types.rs
```diff
@@ -12,3 +15,5 @@ pub struct Config {
 pub type Experiments = Vec<ExperimentResponse>;
 
 pub(crate) type ExperimentStore = HashMap<String, ExperimentResponse>;
```
ayushjain17, line 17: we shouldn't be needing this type anymore, right ?

workspace_template.sql
ayushjain17, line 1: migration files not added

workspace_template.sql
```diff
@@ -594,4 +594,21 @@ CREATE TABLE IF NOT EXISTS {replaceme}.experiment_groups(
...
 );
 
-ALTER TABLE {replaceme}.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
\ No newline at end of file
+ALTER TABLE {replaceme}.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
+
+ALTER TABLE {replaceme}.experiment_groups 
+ADD COLUMN buckets JSON[] 
+DEFAULT array_fill('{"experiment_id": null, "variant": ""}'::jsonb, ARRAY[100]) 
+NOT NULL;
```
ayushjain17: this has to work for already running experiments also, some migration or something would be needed

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -280,8 +280,37 @@ pub struct EventLog {
     pub new_data: Option<Value>,
     pub query: String,
 }
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    Eq,
```
ayushjain17, line 340: `Eq` not needed, `PartialEq` is enough

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
         Ok(Self(value))
     }
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Default)]
+pub struct Bucket {
```
ayushjain17: this does not seem to be a common type, it is needed in `models/experimentation.rs`, can we move it there only

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+    pub experiment_id: Option<i64>,
+    pub variant: Option<String>,
```
ayushjain17: you would need to add extra validation to make sure either both exist or neither exist, either of them cannot exist while the other does not exist

instead of doing all this, we simply not use `[Option<String>; 100]` as the type ?
and store only the `variant_id`, it is more than enough to convey all that is needed, isn't it ?

crates/superposition_types/src/database/models.rs
```diff
@@ -9,7 +9,14 @@ use chrono::{DateTime, Utc};
...
-    sql_types::{Json, Text},
+    deserialize::{self, FromSql},
+    pg::PgValue,
+    serialize::{self, Output, ToSql},
+};
+#[cfg(feature = "diesel_derives")]
+use diesel::{
+    pg::Pg,
+    sql_types::{Array, Json, Text},
     AsChangeset, AsExpression, FromSqlRow, Insertable, QueryId, Queryable, Selectable,
```
ayushjain17: just combine these two, why have them separately

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Deref, DerefMut, Default)]
+#[serde(try_from = "Vec<Bucket>")]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Array<Json>))]
+pub struct Buckets(Vec<Bucket>);
+
+impl Buckets {
+    pub fn from_array(arr: [Bucket; 100]) -> Self {
+        Buckets(arr.to_vec())
```
ayushjain17:
```suggestion
        Self(arr.to_vec())
```

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+    }
+
+    pub fn new() -> Self {
+        Buckets(vec![Bucket::default(); 100])
```
ayushjain17:
```suggestion
        Self(vec![Bucket::default(); 100])
```

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+    }
+}
+
+impl TryFrom<Vec<Bucket>> for Buckets {
+    type Error = String;
+
+    fn try_from(value: Vec<Bucket>) -> Result<Self, Self::Error> {
+        if value.len() != 100 {
```
ayushjain17: this check is insufficient, need to check for the keys also

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+        Ok(Buckets(value))
+    }
+}
+
+#[cfg(feature = "diesel_derives")]
+impl FromSql<Array<Json>, Pg> for Buckets {
+    fn from_sql(bytes: PgValue<'_>) -> deserialize::Result<Self> {
+        let json_array: Vec<serde_json::Value> =
+            FromSql::<Array<Json>, Pg>::from_sql(bytes)?;
+        let mut buckets = Vec::with_capacity(json_array.len());
```
ayushjain17: no ?
```suggestion
        let mut buckets = Vec::with_capacity(100);
```

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+
+        for json_value in json_array {
+            let bucket: Bucket = serde_json::from_value(json_value)?;
+            buckets.push(bucket);
+        }
+
+        Buckets::try_from(buckets).map_err(Into::into)
```
ayushjain17:
```suggestion
        Self::try_from(buckets).map_err(Into::into)
```

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+    type Error = String;
+
+    fn try_from(value: Vec<Bucket>) -> Result<Self, Self::Error> {
+        if value.len() != 100 {
+            return Err(format!(
+                "Buckets must contain exactly 100 elements, got {}",
+                value.len()
+            ));
+        }
+        Ok(Buckets(value))
```
ayushjain17:
```suggestion
        Ok(Self(value))
```

crates/superposition_types/src/database/models.rs
```diff
@@ -359,3 +366,69 @@ impl TryFrom<String> for NonEmptyString {
...
+impl ToSql<Array<Json>, Pg> for Buckets {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        let json_values: Result<Vec<serde_json::Value>, _> =
+            self.0.iter().map(serde_json::to_value).collect();
+
+        let json_array = json_values?;
+        <Vec<serde_json::Value> as ToSql<Array<Json>, Pg>>::to_sql(
+            &json_array,
+            &mut out.reborrow(),
+        )
```
ayushjain17:
```suggestion
        let json_array: Vec<serde_json::Value> =
            self.0.iter().map(serde_json::to_value).collect()?;

        <Vec<serde_json::Value> as ToSql<Array<Json>, Pg>>::to_sql(
            &json_array,
            &mut out.reborrow(),
        )
```

workspace_template.sql
```diff
@@ -594,4 +594,21 @@ CREATE TABLE IF NOT EXISTS {replaceme}.experiment_groups(
     last_modified_by TEXT NOT NULL
 );
 
-ALTER TABLE {replaceme}.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
\ No newline at end of file
+ALTER TABLE {replaceme}.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
+
+ALTER TABLE {replaceme}.experiment_groups 
+ADD COLUMN buckets JSON[] 
+DEFAULT array_fill('{"experiment_id": null, "variant": ""}'::jsonb, ARRAY[100])
```
ayushjain17:
```suggestion
DEFAULT array_fill('{}'::jsonb, ARRAY[100]) 
```

docker-compose/postgres/db_init.sql
```diff
@@ -1381,4 +1381,25 @@ ALTER TABLE localorg_dev.experiments ADD COLUMN IF NOT EXISTS experiment_group_i
...
+ADD COLUMN buckets JSON[] 
+DEFAULT array_fill('{"experiment_id": null, "variant": ""}'::jsonb, ARRAY[100]) 
+NOT NULL;
+
+ALTER TABLE localorg_dev.experiment_groups 
+ADD COLUMN group_type public.GROUP_TYPE DEFAULT 'USER_CREATED';
+
+ALTER TABLE localorg_test.experiment_groups 
+ADD COLUMN buckets JSON[] 
+DEFAULT array_fill('{"experiment_id": null, "variant": ""}'::jsonb, ARRAY[100])
```
ayushjain17:
```suggestion
DEFAULT array_fill('{}'::jsonb, ARRAY[100]) 

```

docker-compose/postgres/db_init.sql
```diff
@@ -1381,4 +1381,25 @@ ALTER TABLE localorg_dev.experiments ADD COLUMN IF NOT EXISTS experiment_group_i
...
 ALTER TABLE localorg_test.experiments ADD COLUMN IF NOT EXISTS experiment_group_id bigint;
 
+CREATE TYPE public.GROUP_TYPE AS ENUM (
+    'USER_CREATED',
+    'SYSTEM_GENERATED'
+);
+
+ALTER TABLE localorg_dev.experiment_groups 
+ADD COLUMN buckets JSON[] 
+DEFAULT array_fill('{"experiment_id": null, "variant": ""}'::jsonb, ARRAY[100])
```
ayushjain17:
```suggestion
DEFAULT array_fill('{}'::jsonb, ARRAY[100]) 

```

clients/tests/typescript/tests/experiments.test.ts
```diff
@@ -554,7 +554,7 @@ describe("Experiments API", () => {
             const updatedExp = await superpositionClient.send(getCmd);
             expect(updatedExp.traffic_percentage).toBe(rampPercentage);
             expect(updatedExp.status).toBe(ExperimentStatusType.INPROGRESS);
-            expect(updatedExp.experiment_group_id).toBeUndefined();
+            expect(updatedExp.experiment_group_id).toBe(experimentGroupId);
```
ayushjain17: need to add test to check that when the experiment is closed, the group should also get auto deleted

clients/tests/typescript/tests/experiment_groups.test.ts
```diff
@@ -457,6 +457,7 @@ describe("Experiment Groups API Integration Tests", () => {
                     expect.arrayContaining([expValid1Id, expValid2Id]),
                 );
                 expect(response.traffic_percentage).toBe(100);
+                expect(response.group_type).toBe("USER_CREATED");
```
ayushjain17:
```suggestion
                expect(response.group_type).toBe(GroupType.USER_CREATED);
```

clients/tests/typescript/tests/experiment_groups.test.ts
```diff
@@ -477,6 +478,7 @@ describe("Experiment Groups API Integration Tests", () => {
             expGroupId = response.id!;
             expect(response.name).toBe(groupName);
             expect(response.member_experiment_ids).toEqual([]);
+            expect(response.group_type).toBe("USER_CREATED");
```
ayushjain17:
```suggestion
            expect(response.group_type).toBe(GroupType.USER_CREATED);
```

clients/tests/typescript/tests/experiment_groups.test.ts
```diff
@@ -595,6 +597,7 @@ describe("Experiment Groups API Integration Tests", () => {
             expect(response.member_experiment_ids).toEqual(
                 currentGroup.member_experiment_ids || [],
             );
+            expect(response.group_type).toBe("USER_CREATED");
```
ayushjain17:
```suggestion
            expect(response.group_type).toBe(GroupType.USER_CREATED);
```

clients/tests/typescript/tests/experiments.test.ts
mahatoankitkumar: This is not a system generated group, so it should not be deleted, only system generated groups should get deleted after the experiment is closed

crates/superposition_types/migrations/2025-06-26-130933_experiment_buckets/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Your SQL goes here
+ALTER TABLE public.experiment_groups ADD COLUMN buckets TEXT[] DEFAULT array_fill(NULL::TEXT, ARRAY[100]) NOT NULL;
+
+CREATE TYPE public.GROUP_TYPE AS ENUM (
+    'USER_CREATED',
+    'SYSTEM_GENERATED'
+);
+
+ALTER TABLE public.experiment_groups
```
Datron: @mahatoankitkumar can you explain why we need this?

crates/superposition_types/src/database/models.rs
```diff
@@ -360,3 +364,52 @@ impl TryFrom<String> for NonEmptyString {
...
+pub struct Buckets(Vec<Option<String>>);
+
+impl Buckets {
+    pub fn from_array(arr: [Option<String>; 100]) -> Self {
+        Self(arr.to_vec())
+    }
+
+    pub fn new() -> Self {
+        Self(vec![None; 100])
+    }
```
Datron:
```suggestion
    pub fn new() -> Self {
        Self(Vec::with_capacity(100))
    }
```
But since vector is dynamic this limit doesn't really hold or help. You are better of with actual slices

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -415,31 +420,57 @@ pub fn extract_override_keys(overrides: &Map<String, Value>) -> HashSet<String>
...
-            if variant.variant_type == VariantType::EXPERIMENTAL {
-                return Ok(Some(variant.clone()));
-            }
-        }
-    }
-    let variant_count = applicable_variants.len() as u8;
-    let range = (traffic * variant_count) as i32;
-    if (toss as i32) >= range {
+    experiment_group: &ExperimentGroup,
+    toss: usize,
```
Datron: toss -1 represents CUG. usize doesn't take negative numbers

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -415,31 +420,57 @@ pub fn extract_override_keys(overrides: &Map<String, Value>) -> HashSet<String>
...
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+) -> superposition::Result<Option<Variant>> {
+    let bucket: Option<String> = experiment_group.buckets
+        .get(toss)
+        .ok_or_else(|| {
+            bad_argument!(
+                "Bucket index out of bounds. Ensure the bucket index is within the range of available buckets."
+            )
+        })?.clone();
```
Datron: So we are creating a bucket for every value of toss?

mahatoankitkumar: yes, 100 buckets, and we take in identifier instead of toss, and calculate the hash_toss on the go with the identifier.
And we create separate groups for experiments outside any group to keep the bucketing logic consistent, and those are marked as system_generated groups(which are deleted once the experiment is closed) and others as user_created.

crates/experimentation_client/src/lib.rs
```diff
@@ -67,20 +72,30 @@ impl Client {
...
-                .await;
-                match experiments {
-                    Ok(experiments) => {
-                        let mut exp_store = self.experiments.write().await;
-                        *exp_store = experiments;
-                        *start_date = Utc::now();
-                    }
-                    Err(e) => {
-                        log::error!(
-                            "Failed to fetch experiments from the server with error: {}",
```
pratikmishra356, line 79: keep this log

crates/superposition_types/src/database/models.rs
```diff
@@ -360,3 +364,52 @@ impl TryFrom<String> for NonEmptyString {
...
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Deref, DerefMut, Default)]
+#[serde(try_from = "Vec<Option<String>>")]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Array<Nullable<Text>>))]
+pub struct Buckets(Vec<Option<String>>);
+
+impl Buckets {
+    pub fn from_array(arr: [Option<String>; 100]) -> Self {
```
ayushjain17: can we implement `From` trait for this

crates/superposition_types/src/database/models.rs
```diff
@@ -360,3 +364,52 @@ impl TryFrom<String> for NonEmptyString {
...
+        Self(arr.to_vec())
+    }
+
+    pub fn new() -> Self {
```
ayushjain17: should this be as `Default` trait, instead of new ?

crates/superposition_types/src/api/experiments.rs
```diff
@@ -335,3 +309,46 @@ pub struct AuditQueryFilters {
     pub count: Option<i64>,
     pub page: Option<i64>,
 }
+
+pub fn option_i64_from_value(value: Value) -> Result<Option<i64>, String> {
```
ayushjain17: dont we already have this?

crates/superposition_types/src/api/experiments.rs
```diff
@@ -189,8 +178,7 @@ impl TryFrom<HashMap<String, String>> for ApplicableVariantsQuery {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ApplicableVariantsRequest {
     pub context: Map<String, Value>,
-    #[serde(deserialize_with = "deserialize_toss")]
-    pub toss: i8,
+    pub toss: String,
```
ayushjain17: this should be renamed now to `identifier` right ?
and to maintain backwards compatibility, if at all we want to, we can add serde alias

but ideally it should be a breaking change only, and we completely change it to `identifier` without keeping any traces of toss anywhere

crates/superposition_types/src/api/experiments.rs
```diff
@@ -164,23 +164,12 @@ impl TryFrom<HashMap<String, String>> for ApplicableVariantsQuery {
             .map(|(key, value)| (key, value.parse().unwrap_or(Value::String(value))))
             .collect::<Map<_, _>>();
 
-        let toss = value
+        let toss: Value = value
             .remove("toss")
```
ayushjain17:
```suggestion
        let identifier = value
            .remove("toss").and_then(|v| v.as_str().to_string())
```

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -63,13 +65,27 @@ pub enum SortOn {
...
+
+impl Default for ExpGroupFilters {
+    fn default() -> Self {
+        ExpGroupFilters {
+            name: None,
+            created_by: None,
+            last_modified_by: None,
+            sort_on: None,
+            sort_by: None,
+            group_type: Some(GroupType::UserCreated),
```
ayushjain17: just confused, why do we need this separately ?
it should have been default derivation only with all as `None`

crates/superposition_types/src/database/models.rs
```diff
@@ -360,3 +364,52 @@ impl TryFrom<String> for NonEmptyString {
         Ok(Self(value))
     }
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Deref, DerefMut, Default)]
+#[serde(try_from = "Vec<Option<String>>")]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Array<Nullable<Text>>))]
+pub struct Buckets(Vec<Option<String>>);
```
ayushjain17: this is not a common type and is very experiment specific file
can we move it to, `models/experimentation.rs`

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -219,7 +219,7 @@ pub enum VariantType {
 }
 
 #[repr(C)]
-#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record, PartialEq)]
```
ayushjain17: why is this needed ?

ayushjain17 (changes requested):
partial review
reviewed only types and surrounding stuff, core logic not yet reviewed

build failing, needs to be fixed and also rebased

clients/tests/typescript/tests/experiments.test.ts
ayushjain17: you need to add test to check whether the experiment group was deleted or not, currently the checks added just makes sure that whether it was unlinked from the experiment or not, but not regarding the existence of experiment group altogether

crates/experimentation_client/src/lib.rs
```diff
@@ -67,20 +72,30 @@ impl Client {
...
+                    hostname.clone(),
+                    self.http_client.clone(),
+                    self.client_config.tenant.to_string(),
+                )
+                .await
+                .unwrap_or(HashMap::new());
+
+                let mut exp_store = self.experiments.write().await;
+                for (exp_id, experiment) in experiments.into_iter() {
+                    if experiment.status.active() {
```
pratikmishra356: we have updated this logic

crates/experimentation_client/src/lib.rs
```diff
@@ -89,21 +104,41 @@ impl Client {
     pub async fn get_applicable_variant(
         &self,
         context: &Value,
-        toss: i8,
+        identifier: &str,
     ) -> Result<Vec<String>, String> {
-        let experiments: Experiments =
-            self.get_satisfied_experiments(context, None).await?;
+        let experiments: ExperimentGroups =
+            self.get_satisfied_experiment_groups(context).await?;
```
pratikmishra356: why we are only fetching satisfied_experiment_groups 
what about independent experiments ?

crates/experimentation_client/src/lib.rs
```diff
@@ -295,6 +342,47 @@ async fn get_experiments(
...
+async fn get_experiment_groups(
+    hostname: String,
+    http_client: reqwest::Client,
+    tenant: String,
+) -> Result<ExperimentGroupStore, String> {
+    let mut curr_exp_group_store: ExperimentGroupStore = HashMap::new();
+    let mut pagination_params = PaginationParams::default();
+    let mut page = 1;
+
+    loop {
```
pratikmishra356: lets not use pagination here , fetch all at once , we have changes the approach

crates/experimentation_client/src/lib.rs
```diff
@@ -213,37 +248,49 @@ impl Client {
...
+        let variant_id = match experiment_group
+            .buckets
+            .get(hashed_toss)
+            .and_then(|bucket| bucket.as_ref())
+        {
+            Some(id) => id,
+            None => return Ok(None),
+        };
+
+        let experiment_id = variant_id
```
pratikmishra356: would it be possible to get experiment_id from somewhere else , parsing from variant_id does not sounds good

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +203,223 @@ pub fn remove_members(
...
+            additional_needed,
+            unassigned_buckets.len()
+        ));
+    }
+
+    // Assign buckets in round-robin fashion across variants
+    unassigned_buckets.reverse();
+    let mut variant_index = 0;
+    for _ in 0..additional_needed {
+        if let Some(bucket) = unassigned_buckets.pop() {
```
pratikmishra356: Unassigned bucket will always have None element , so it will never enter the if loop right ?

pratikmishra356 (changes requested):

crates/experimentation_client/src/lib.rs
mahatoankitkumar: independent experiments get their own group i.e. System_generated group once started.

crates/experimentation_client/src/lib.rs
mahatoankitkumar: Either we keep both experiment_id and variant_id in the array as an object or we just keep this.
If we keep both then a lot of checks would be needed as both would be optional.

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
mahatoankitkumar: its does enter with none value.

crates/experimentation_client/src/lib.rs
mahatoankitkumar: where?

crates/superposition_types/src/api/experiment_groups.rs
mahatoankitkumar: I think we don't want to show the system generated groups anywhere thats why default to this for any case.

crates/superposition_types/migrations/2025-06-26-130933_experiment_buckets/up.sql
mahatoankitkumar: https://github.com/juspay/superposition/pull/534#discussion_r2176839825

crates/superposition_types/src/database/models.rs
mahatoankitkumar: vector size fixed.

crates/superposition_types/src/database/models.rs
mahatoankitkumar: this function not needed anymore.

crates/superposition_types/src/database/models.rs
mahatoankitkumar: method `default` can be confused for the standard trait method `std::default::Default::default`

clients/tests/typescript/tests/experiments.test.ts
ayushjain17: pending

examples/experimentation_client_integration_example/src/main.rs
```diff
@@ -31,18 +31,20 @@ async fn main() -> std::io::Result<()> {
...
 
-#[get("/variants/{client_id}/{platform}/{toss}")]
+#[get("/variants/{client_id}/{platform}/{identifier}")]
 async fn get_variants(
     state: Data<exp::Client>,
-    path: Path<(String, String, i8)>,
+    path: Path<(String, String, String, Option<Vec<String>>)>,
 ) -> HttpResponse {
-    let (client_id, platform, toss) = path.into_inner();
+    let (client_id, platform, identifier, variant_ids) = path.into_inner();
```
ayushjain17: from where is this variant_ids coming ?
path has only a rename change here:
#[get("/variants/{client_id}/{platform}/{toss}")]
#[get("/variants/{client_id}/{platform}/{identifier}")]

crates/superposition_types/src/database/models.rs
```diff
@@ -7,9 +7,10 @@ use std::str::FromStr;
 
 use chrono::{DateTime, Utc};
 use derive_more::{Deref, DerefMut};
+
 #[cfg(feature = "diesel_derives")]
 use diesel::{
-    sql_types::{Json, Text},
+    sql_types::{Json, Nullable, Text},
```
ayushjain17: unused import

crates/superposition_types/src/database/models.rs
```diff
@@ -7,9 +7,10 @@ use std::str::FromStr;
 
 use chrono::{DateTime, Utc};
 use derive_more::{Deref, DerefMut};
+
```
ayushjain17: wrongly added newline

crates/superposition_types/src/api/experiments.rs
```diff
@@ -164,58 +165,45 @@ impl TryFrom<HashMap<String, String>> for ApplicableVariantsQuery {
...
-                    None
-                }
-            })
-            .ok_or_else(|| {
-                log::error!("toss should be a an interger between -1 and 100 (included)");
-                String::from("toss should be a an interger between -1 and 100 (included)")
-            })?;
+        let identifier: Value = value
+            .remove("identifier")
+            .ok_or_else(|| "Missing 'toss' key in the request".to_string())?;
```
ayushjain17:
```suggestion
            .ok_or_else(|| "Missing 'identifier' key in the request".to_string())?;
```

crates/frontend/src/components/experiment_form.rs
ayushjain17, line 1: can we revert changes in this file ?

clients/tests/typescript/tests/experiments.test.ts
mahatoankitkumar: line 726

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -77,12 +80,14 @@ use super::{
...
         add_variant_dimension_to_ctx, check_variant_types,
-        check_variants_override_coverage, extract_override_keys, fetch_cac_config,
+        check_variants_override_coverage, decide_variant_with_variant_ids,
+        extract_override_keys, fetch_cac_config, get_experiment,
         handle_experiment_group_membership, hash, validate_experiment,
         validate_override_keys,
     },
     types::{ContextAction, ContextBulkResponse, ContextMoveReq, ContextPutReq},
 };
+use superposition_types::database::schema::experiment_groups::dsl as experiment_groups;
```
ayushjain17: can we club this with the other imports

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +203,223 @@ pub fn remove_members(
...
+    for (unassigned, i) in (0..current_buckets.len()).rev().enumerate() {
+        if unassigned >= excess_count {
+            break;
+        }
+
+        *current_buckets[i] = None;
+    }
+}
+
+pub fn detach_experiment_from_group(
```
ayushjain17, line 301: detaching logic needs to be run for remove members api as well

mainly the bucket de-allocation has to be there

crates/experimentation_client/src/lib.rs
ayushjain17: +1

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +203,223 @@ pub fn remove_members(
         .get_result(conn)?;
     Ok(Json(updated_group))
 }
+
+pub fn update_bucket_allocation(
```
ayushjain17, line 212: can we update `validate_experiment_group_constraints` to use available buckets to check whether new guys can be added or not

also, there is a check for this: `traffic_percentage_sum >= 100`, in that function it should rather become `traffic_percentage_sum > 100`, as we allow only created status to be added, so even if all slots are full, we can allow to add, just we cant allow to start or ramp, once all buckets all filled

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +203,223 @@ pub fn remove_members(
...
+    }
+
+    Ok(())
+}
+
+fn unassign_excess_buckets(
+    current_buckets: &mut [&mut Option<String>],
+    excess_count: usize,
+) {
+    for (unassigned, i) in (0..current_buckets.len()).rev().enumerate() {
```
ayushjain17: dont do assigning and un-assigning in this alternate fashion, in a situation of multiple allocations and de-allocation it will lead to inconsistent data as the assigning and un-assigning is supposed to be done per variant level, but you are not doing that, rather you are operating at a per experiment level

there are cases in this logic, where eventually your logic will assign more buckets to one variant than other
excess addition or removal calculation and operation should be at a variant level only as that is the atomic item here, and not the experiment

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1056,19 +1083,51 @@ async fn ramp(
...
+                    new_traffic_percentage,
+                    &state,
+                    transaction_conn,
+                    &workspace_request,
+                    &user,
+                )?;
+                experiment_group_id = Some(experiment_group.id);
+            }
+
+            if let Some(experiment_group_id) = experiment_group_id {
```
pratikmishra356: allocation of bucket , this logic is being done twice in case of system_generated_experiment_group
because in that function u are already assigning buckets and then here also

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -415,31 +420,128 @@ pub fn extract_override_keys(overrides: &Map<String, Value>) -> HashSet<String>
 }
 
 pub fn decide_variant(
```
pratikmishra356: clients also has this function , could we make it one

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -132,13 +140,21 @@ pub fn add_members(
         .filter(experiment_groups::id.eq(exp_group_id))
         .schema_name(schema_name)
         .first::<ExperimentGroup>(conn)?;
+
+    if exp_context.group_type == GroupType::SystemGenerated {
```
ayushjain17: need to add this check for update api and remove api as well

ayushjain17: also, while ramping an experiment which belongs to SYSTEM_GENERATED, for that group, when the experiment is ramped, the traffic to the group also should increase 
`experiment_group.traffic = experiment.traffic * experiment.variants.len()`

crates/experimentation_platform/src/api/experiments/helpers.rs
mahatoankitkumar: sure

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +198,246 @@ pub fn remove_members(
...
+    let new_traffic_percentage =
+        if experiment_group.group_type == GroupType::SystemGenerated {
+            TrafficPercentage::try_from(
+                experiment.variants.clone().into_inner().len() as i32
+                    * traffic_percentage.as_i32(),
+            )
+            .map_err(|e| unexpected_error!(e))?
+        } else {
+            experiment_group.traffic_percentage.clone()
+        };
```
ayushjain17: shouldn't this rather have been a part of ramp API ?
instead of doing it for all the times this is called ?

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
ayushjain17: this is not done yet

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -71,44 +79,26 @@ pub fn validate_experiment_group_constraints(
     member_experiments: &[Experiment],
     existing_members: &[i64],
     group_context: &Condition,
-    conn: &mut DBConnection,
-    schema_name: &SchemaName,
 ) -> superposition::Result<Vec<i64>> {
```
ayushjain17, line 82: this function is no longer any sort of validation

crates/experimentation_client/src/lib.rs
```diff
@@ -89,21 +122,53 @@ impl Client {
     pub async fn get_applicable_variant(
         &self,
         context: &Value,
-        toss: i8,
+        identifier: &str,
+        variant_ids: Option<Vec<String>>,
```
pratikmishra356: this function does not needs variant_ids as param,
i think you misunderstood , if variant_ids will be present in the context of openfeature client
then we will not call experiments functions , we will directly do cac eval

crates/experimentation_client/src/lib.rs
```diff
@@ -68,11 +73,37 @@ impl Client {
...
                 )
                 .await;
+
                 match experiments {
                     Ok(experiments) => {
                         let mut exp_store = self.experiments.write().await;
-                        *exp_store = experiments;
-                        *start_date = Utc::now();
+                        for (exp_id, experiment) in experiments.into_iter() {
+                            if experiment.status.active() {
```
pratikmishra356: can you check the main branch , we have removed this logic 
do not do any change in this , just add the experiment_group logic

crates/experimentation_client/src/lib.rs
```diff
@@ -213,37 +278,123 @@ impl Client {
...
+        };
+
+        Ok(experiment
+            .variants
+            .iter()
+            .find(|v| v.id == *variant_id)
+            .cloned())
+    }
+
+    pub async fn decide_variant_with_variant_ids(
```
pratikmishra356: we dont need this function right ?

crates/experimentation_client/src/lib.rs
```diff
@@ -89,21 +122,53 @@ impl Client {
     pub async fn get_applicable_variant(
         &self,
         context: &Value,
-        toss: i8,
+        identifier: &str,
+        variant_ids: Option<Vec<String>>,
     ) -> Result<Vec<String>, String> {
-        let experiments: Experiments =
-            self.get_satisfied_experiments(context, None).await?;
+        let experiments: ExperimentGroups = self
```
pratikmishra356:
```suggestion
        let exp_group: ExperimentGroups = self
```

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -415,31 +423,128 @@ pub fn extract_override_keys(overrides: &Map<String, Value>) -> HashSet<String>
...
-    applicable_variants: Vec<Variant>,
-    toss: i8,
-) -> Result<Option<Variant>, String> {
-    if toss < 0 {
-        for variant in applicable_variants.iter() {
-            if variant.variant_type == VariantType::EXPERIMENTAL {
-                return Ok(Some(variant.clone()));
-            }
+    experiment_group: &ExperimentGroup,
+    toss: usize,
```
ayushjain17: can we make sure we remove all occurrences of `toss` or is it something related to backwards compatibility ?
in that case we should rather update the deserialiser instead of having toss anywhere in the logic

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -412,3 +445,48 @@ where
         .collect::<Result<Vec<i64>, D::Error>>()?;
     Ok(Some(numbers))
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Deref, DerefMut, Default)]
+#[serde(try_from = "Vec<Option<String>>")]
+#[cfg_attr(feature = "diesel_derives", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "diesel_derives", diesel(sql_type = Array<Nullable<Text>>))]
+pub struct Buckets(Vec<Option<String>>);
```
ayushjain17:
```suggestion
pub struct Buckets([Option<String>; 100]);
```

crates/experimentation_client/src/lib.rs
ayushjain17: +1

crates/experimentation_client/src/lib.rs
```diff
@@ -81,6 +112,8 @@ impl Client {
                         );
                     }
                 };
+
+                *start_date = Utc::now();
```
ayushjain17, line 116: this also should not be needed

pratikmishra356:
@mahatoankitkumar When we are removing members from experiment_group
you first need to check if it is in created state do nothing , just remove the association
and if its in inprogress , you need to remove asscociation , create system generated experiment group
and associate to it , allocate appropriate buckets

crates/superposition_types/src/api/experiments.rs
```diff
@@ -159,7 +159,8 @@ pub struct ExperimentStateChangeRequest {
 #[serde(try_from = "HashMap<String,String>")]
 pub struct ApplicableVariantsQuery {
     pub context: Map<String, Value>,
-    pub toss: i8,
```
pratikmishra356, line 162: this changes the api contract , can we support both

crates/experimentation_client/src/interface.rs
```diff
@@ -168,16 +168,28 @@ pub extern "C" fn expt_get_client(tenant: *const c_char) -> *mut Arc<Client> {
 pub extern "C" fn expt_get_applicable_variant(
     client: *mut Arc<Client>,
     c_context: *const c_char,
-    toss: c_short,
+    identifier: *const c_char,
```
pratikmishra356, line 171: interface we are changing , but are we changing the corresponding clients implementation
and this is a breaking change

crates/experimentation_client/src/lib.rs
```diff
@@ -89,21 +122,53 @@ impl Client {
...
+        identifier: &str,
+    ) -> Result<ExperimentGroups, String> {
+        let running_experiment_groups = self
+            .experiment_groups
+            .read()
+            .await
+            .iter()
+            .filter(|(_, exp_group)| {
+                let hashed_toss = calculate_bucket_index(identifier, &exp_group.id);
+                exp_group.context.is_empty()
```
pratikmishra356: you should put the or combination within parenthesis

crates/experimentation_client/src/lib.rs
```diff
@@ -213,37 +278,123 @@ impl Client {
...
+            let variant_id = req_experiment_ids[&exp_id];
+
+            // Get experiment and check if active
+            let experiments = self.experiments.read().await;
+            let experiment = match experiments.get(&exp_id.to_string()) {
+                Some(exp) if exp.status.active() => exp,
+                _ => return Ok(None),
+            };
+
+            if experiment.status.active() {
```
pratikmishra356: this active status check is repetitive , also you are doing early return please verify it that is correct

crates/superposition_types/migrations/2025-06-26-130933_experiment_buckets/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Your SQL goes here
+ALTER TABLE public.experiment_groups ADD COLUMN buckets TEXT[] DEFAULT array_fill(NULL::TEXT, ARRAY[100]) NOT NULL;
```
ayushjain17: we should add this as well
```CHECK (cardinality(buckets) = 100);```

File: crates/experimentation_platform/src/api/experiment_groups/handlers.rs
Comment (ayushjain17) [inline crates/experimentation_platform/src/api/experiment_groups/handlers.rs:1]: we need to add a temporary backfill API

File: crates/experimentation_platform/src/api/experiment_groups/helpers.rs (hunk 53 above)
Comment (mahatoankitkumar) [reply crates/experimentation_platform/src/api/experiment_groups/helpers.rs]: when ever we do ramp, this function already updates the experiment_groups, better do it here only right?
Otherwise another function to update the traffic_percentage of the group.

File: crates/superposition_types/src/database/models/experimentation.rs [hunk 67]
```diff
@@ -313,6 +330,35 @@ pub struct EventLog {
     pub new_data: Option<Value>,
     pub query: String,
 }
+#[derive(
```
ayushjain17, line 336: can we give a new line here, before this derive

crates/superposition_types/src/api/experiments.rs
```diff
@@ -195,33 +184,18 @@ impl TryFrom<HashMap<String, String>> for ApplicableVariantsQuery {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ApplicableVariantsRequest {
     pub context: Map<String, Value>,
-    #[serde(deserialize_with = "deserialize_toss")]
```
ayushjain17, line 198: you will need custom deserialization to support toss (backwards compatible)
also, one question, by accepting toss as well, can we maintain backwards compatibility? or is it breaking enough in other places as well, that the user would need to make other changes mandatorily in order to be even able use experimentation ?

is this is the case, then there would be no point in maintaining backwards compatibility

crates/superposition_types/src/api/experiment_groups.rs
```diff
@@ -63,13 +66,14 @@ pub enum SortOn {
     LastModifiedAt,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, IsEmpty)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, IsEmpty)]
```
ayushjain17: weren't we going to keep default as `UserGenerated` ?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1056,19 +1075,65 @@ async fn ramp(
...
+                update_bucket_allocation(
+                    &experiment,
+                    &experiment_group,
+                    &req.traffic_percentage,
+                    transaction_conn,
+                    &workspace_request.schema_name,
+                    &user,
+                )?;
+                // update the traffic percentage of the system created experiment group
+                update_system_generated_experiment_group_traffic_percentage(
```
ayushjain17: shouldn't this be there in both the cases ?

crates/frontend/src/pages/experiment_group_listing/filter.rs
```diff
@@ -202,6 +220,42 @@ pub(super) fn experiment_group_filter_widget(
...
+                                                filters_buffer_rws
+                                                    .with(|f| f.group_type.clone())
+                                                    .is_some_and(|s| s.iter().any(|item| *item == status))
+                                            }
+                                            on:change=move |event| {
+                                                let checked = event_target_checked(&event);
+                                                status_filter_management(checked, status)
+                                            }
+                                        />
+                                        <label
```
ayushjain17, line 247: is this using the new UI for the pill ?
if yes can we make a component out of this and use it in both the places

crates/frontend/src/pages/experiment_group_listing/filter.rs
```diff
@@ -133,6 +135,22 @@ pub(super) fn experiment_group_filter_widget(
 ) -> impl IntoView {
     let filters_buffer_rws = RwSignal::new(filters_rws.get_untracked());
 
+    let status_filter_management = move |checked: bool, filter_status: GroupType| {
```
ayushjain17:
```suggestion
    let group_type_management = move |checked: bool, group_type: GroupType| {
```

crates/experimentation_platform/src/api/experiments/handlers.rs
mahatoankitkumar: User created group has predefined traffic percentage right.

crates/superposition_types/src/api/experiment_groups.rs
mahatoankitkumar: Will be easier for us to see it too
but we'll add predefined filter for usercreated to be shown as we do in experiments page for status

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +208,257 @@ pub fn remove_members(
...
+
+    let new_experiment_group = ExperimentGroup {
+        id,
+        context_hash,
+        name: experiment.name.clone(),
+        description: Description::try_from(format!(
+            "Experiment group for experiment {}",
+            experiment.name
+        ))
+        .map_err(|e| unexpected_error!(e))?,
```
ayushjain17, line 376: can we do the same as this, in other places and make the error handling simple, unexpected_error already adds a log for the error, we don't need to add it separately

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +208,257 @@ pub fn remove_members(
...
+    update_bucket_allocation(experiment, &mut buckets, &TrafficPercentage::default())?;
+
+    // Remove the experiment from the member_experiment_ids
+    let mut member_experiment_ids = experiment_group.member_experiment_ids;
+    member_experiment_ids.retain(|&id| id != experiment.id);
+
+    diesel::update(experiment_groups::experiment_groups)
+        .filter(experiment_groups::id.eq(&experiment_group_id))
+        .set((
+            experiment_groups::member_experiment_ids.eq(member_experiment_ids),
```
ayushjain17, line 328: change reason not added in all such queries in this file
can we check the other update queries as well for missing change_reason

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +208,257 @@ pub fn remove_members(
...
+    experiment: &Experiment,
+    excess_count: usize,
+) {
+    let variants = experiment.variants.clone().into_inner();
+    let variants_len = variants.len();
+    for variant in variants {
+        for _ in 0..excess_count / variants_len {
+            if let Some(bucket) = current_buckets
+                .iter_mut()
+                .rev()
```
ayushjain17, line 292: can we move this reverse operation outside the loop, we are revering every time in the loop

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +208,257 @@ pub fn remove_members(
...
+            "Not enough empty buckets to accommodate the updated traffic percentage. Required additional: {}, Available: {}",
+            additional_needed,
+            unassigned_buckets.len()
+        ));
+    }
+    let variants = experiment.variants.clone().into_inner();
+    let variants_len = variants.len();
+
+    // Reverse the unassigned_buckets to fill from the front
+    unassigned_buckets.reverse();
```
ayushjain17, line 265: since we are adding this reverse logic in these helper functions to assign and unassign, this becomes coupled to the way how `update_bucket_allocation` works, can we make sure that these functions are not available for others to use directly.
should we rather move this directly inside `update_bucket_allocation` itself ?

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -164,21 +157,49 @@ pub fn remove_members(
...
+
+    ensure_experiments_exist(
+        &HashSet::from_iter(members_to_remove),
+        &experiments_to_remove,
+        "The following experiment IDs are not present in the database",
+    )?;
+
+    let mut buckets = experiment_group.buckets;
+    for member_experiment in &experiments_to_remove {
+        // TODO:  remove redundant db calls to update the buckets for the same experiment group
```
ayushjain17: this is already done, we please remove the redundant or unnecessary comments

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +208,257 @@ pub fn remove_members(
...
+            }
+        } else {
+            unassigned_buckets.push(bucket);
+        }
+    }
+
+    let required_bucket_count =
+        **exp_traffic_percentage as usize * experiment.variants.len();
+    let current_bucket_count = current_exp_buckets.len();
+
```
ayushjain17, line 235: can we rather do this instead, the order of the operation would not matter then
```suggestion
   let bucket_diff = required_bucket_count.abs_diff(current_bucket_count);
```

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
mahatoankitkumar: I think its fine to be used outside the update_bucket_allocation as even if we pass the whole bucket array, we will have to rev as pop() provides the elements from the back.

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
mahatoankitkumar: This doesn't reverse the array, it just iterates in the reverse direction.

crates/experimentation_client/src/lib.rs
```diff
@@ -81,6 +85,28 @@ impl Client {
...
+                    }
+                    Err(e) => {
+                        log::error!(
+                            "Failed to fetch experiments from the server with error: {}",
+                            e
+                        );
+                    }
+                };
+
+                *start_date = Utc::now();
```
pratikmishra356: even on error , this start_date will get updates , that is not a good experience

crates/experimentation_client/src/lib.rs
```diff
@@ -212,36 +250,67 @@ impl Client {
...
+        })
+        .collect()
+}
+
+pub fn get_applicable_variants_from_group_response(
+    experiments: &[ExperimentResponse],
+    context: &Value,
+    bucket_response: &[(usize, Bucket)],
+) -> Vec<String> {
+    experiments
```
pratikmishra356: can we make it efficient ? currently its num_of_experiment * satisfied_buckets 
can we pass the experimentstore its already a may and then iter on satisfied_buckets and validate the context

crates/experimentation_client/src/lib.rs
```diff
@@ -89,19 +115,31 @@ impl Client {
...
+        let buckets =
+            get_applicable_buckets_from_group(&experiment_groups, context, identifier);
+
+        let experiments = self
+            .experiments
+            .read()
+            .await
+            .iter()
+            .map(|(_, exp)| exp.clone())
+            .collect::<Vec<_>>();
```
pratikmishra356, line 137: can we add the type here and above also

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
```diff
@@ -133,8 +140,17 @@ async fn update_experiment_group(
     schema_name: SchemaName,
     user: User,
 ) -> superposition::Result<Json<ExperimentGroup>> {
-    let req = req.into_inner();
     let DbConnection(mut conn) = db_conn;
+    let experiment_group =
```
pratikmishra356: instead of fetch and then check , we can directly write query to check if given id is of type systemgenerated

crates/experimentation_client/src/lib.rs
```diff
@@ -296,6 +353,48 @@ async fn get_experiments(
...
+        StatusCode::NOT_MODIFIED => {
+            return Err(format!(
+                "{} EXP: skipping update, remote not modified",
+                tenant
+            ));
+        }
+        StatusCode::OK => log::info!(
+            "{}",
+            format!("{} EXP: new experiment groups received, updating", tenant)
+        ),
```
ayushjain17:
```suggestion
        StatusCode::OK => log::info!(
            "{} EXP: new experiment groups received, updating", tenant
        ),
```

smithy/models/experiment_groups.smithy
```diff
@@ -262,6 +287,10 @@ operation ListExperimentGroups {
         @documentation("If true, returns all experiment groups, ignoring pagination parameters page and count.")
         @notProperty
         all: Boolean
+
+        @httpQuery("group_type")
+        @documentation("Filter by the type of group (USER_CREATED or SYSTEM_GENERATED).")
+        group_type: GroupType
```
ayushjain17, line 293: this CommaSeparatedQParam type
```suggestion
        group_type: String
```

this is wrong in Experiment List filters also, I dont if we change this, as group_type already exists as a property

mahatoankitkumar: Screenshot 2025-07-22 at 6 34 29 PM

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -187,3 +207,247 @@ pub fn remove_members(
...
+    match required_bucket_count.cmp(&current_bucket_count) {
+        std::cmp::Ordering::Greater => {
+            assign_additional_buckets(&mut unassigned_buckets, experiment, bucket_diff)?;
+        }
+        std::cmp::Ordering::Less => {
+            unassign_excess_buckets(&mut current_exp_buckets, experiment, bucket_diff);
+        }
+        std::cmp::Ordering::Equal => {
+            return Ok(());
+        }
```
```suggestion
        std::cmp::Ordering::Equal => ()
```

crates/experimentation_client/src/lib.rs
```diff
@@ -61,26 +66,44 @@ impl Client {
...
+                    *start_date,
+                    self.client_config.tenant.to_string(),
+                )
+                .await;
+
+                match (experiments_result, experiment_groups_result) {
+                    (Ok(experiments), Ok(experiment_groups)) => {
                         let mut exp_store = self.experiments.write().await;
                         *exp_store = experiments;
+                        let mut exp_group_store = self.experiment_groups.write().await;
```
ayushjain17, line 89: why not use separate start time for both and keep this separate for both instead of combining ?

mahatoankitkumar: I guess we should keep only one last_polled
Otherwise 2 last_polled would need to be maintained, that may become overwhelming.

pratikmishra356 (approved):

clients/python/sdk/superposition_python_sdk/_private/schemas.py
pratikmishra356, line 1: Can you delete this folder
superposition_python_sdk

mahatoankitkumar: Did it in this PR
https://github.com/juspay/superposition/pull/616

Datron (approved):

PR #540, opened 2025-06-17
feat: added a CRUD for experiment groups

Datron:
## Problem
continuation of #508 

## Solution
Add a UI for experiment groups

https://github.com/user-attachments/assets/ddb1a413-10fc-4742-a560-02dfd085a805

### DIFF
--- a/crates/experimentation_platform/src/api/experiment_groups/helpers.rs
+++ b/crates/experimentation_platform/src/api/experiment_groups/helpers.rs
@@ -95,7 +95,8 @@ pub fn validate_experiment_group_constraints(
             traffic_percentage_sum += *member_experiment.traffic_percentage;
         } else {
             return Err(bad_argument!(
-                "Experiment with id {} does not fit in with the experiment group. The contexts do not match.", member_experiment.id
+                "Experiment with id {} does not fit in with the experiment group. The contexts do not match.",
+                member_experiment.id
             ));
         }
     }
@@ -122,6 +123,11 @@ pub fn add_members(
     schema_name: &SchemaName,
     user: &User,
 ) -> superposition::Result<Json<ExperimentGroup>> {
+    if req.member_experiment_ids.is_empty() {
+        return Err(bad_argument!(
+            "Please provide at least one experiment ID to add to the group"
+        ));
+    }
     let exp_context = experiment_groups::experiment_groups
         .filter(experiment_groups::id.eq(exp_group_id))
         .schema_name(schema_name)
@@ -153,6 +159,11 @@ pub fn remove_members(
     schema_name: &SchemaName,
     user: &User,
 ) -> superposition::Result<Json<ExperimentGroup>> {
+    if req.member_experiment_ids.is_empty() {
+        return Err(bad_argument!(
+            "Please provide at least one experiment ID to remove from the group"
+        ));
+    }
     let exp_context = experiment_groups::experiment_groups
         .filter(experiment_groups::id.eq(&id))
         .schema_name(schema_name)
--- a/crates/frontend/src/api.rs
+++ b/crates/frontend/src/api.rs
@@ -7,6 +7,10 @@ use superposition_types::{
     api::{
         context::ContextListFilters,
         default_config::DefaultConfigFilters,
+        experiment_groups::{
+            ExpGroupCreateRequest, ExpGroupFilters, ExpGroupMemberRequest,
+            ExpGroupUpdateRequest,
+        },
         experiments::{
             ExperimentListFilters, ExperimentResponse, ExperimentStateChangeRequest,
         },
@@ -662,19 +666,151 @@ pub async fn execute_autocomplete_function(
     Ok(result)
 }
 
-pub async fn fetch_experiment_groups(
-    tenant: String,
-    org_id: String,
-) -> Result<PaginatedResponse<ExperimentGroup>, String> {
-    let host = use_host_server();
-    let url = format!("{}/experiment-groups", host);
+pub mod experiment_groups {
+    use super::*;
 
-    let response = request(
-        url,
-        reqwest::Method::GET,
-        None::<serde_json::Value>,
-        construct_request_headers(&[("x-tenant", &tenant), ("x-org-id", &org_id)])?,
-    )
-    .await?;
-    parse_json_response(response).await
+    pub async fn fetch_all(
+        filters: &ExpGroupFilters,
+        pagination: &PaginationParams,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<PaginatedResponse<ExperimentGroup>, ServerFnError> {
+        let host = use_host_server();
+        let pagination = pagination.to_string();
+
+        let url = format!("{}/experiment-groups?{}&{}", host, filters, pagination);
+        let response = request::<()>(
+            url,
+            reqwest::Method::GET,
+            None,
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])
+                .map_err(|e| ServerFnError::new(e.to_string()))?,
+        )
+        .await
+        .map_err(|e| ServerFnError::new(e.to_string()))?;
+        let response: PaginatedResponse<ExperimentGroup> = parse_json_response(response)
+            .await
+            .map_err(|e| ServerFnError::new(e.to_string()))?;
+
+        Ok(response)
+    }
+
+    pub async fn fetch(
+        group_id: &str,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<ExperimentGroup, ServerFnError> {
+        let host = use_host_server();
+        let url = format!("{}/experiment-groups/{}", host, group_id);
+
+        let headers =
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])
+                .map_err(ServerFnError::new)?;
+
+        let response = request(url, reqwest::Method::GET, None::<()>, headers)
+            .await
+            .map_err(ServerFnError::new)?;
+
+        parse_json_response(response)
+            .await
+            .map_err(ServerFnError::new)
+    }
+
+    pub async fn create(
+        payload: ExpGroupCreateRequest,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<ExperimentGroup, String> {
+        let host = use_host_server();
+        let url = format!("{host}/experiment-groups");
+
+        let response = request(
+            url,
+            reqwest::Method::POST,
+            Some(payload),
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
+
+    pub async fn update(
+        group_id: &str,
+        payload: ExpGroupUpdateRequest,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<ExperimentGroup, String> {
+        let host = use_host_server();
+        let url = format!("{}/experiment-groups/{}", host, group_id);
+
+        let response = request(
+            url,
+            reqwest::Method::PATCH,
+            Some(payload),
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
+
+    pub async fn delete(
+        group_id: &str,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<ExperimentGroup, String> {
+        let host = use_host_server();
+        let url = format!("{}/experiment-groups/{}", host, group_id);
+
+        let response = request(
+            url,
+            reqwest::Method::DELETE,
+            None::<Value>,
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
+
+    pub async fn add_members(
+        group_id: &str,
+        payload: ExpGroupMemberRequest,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<ExperimentGroup, String> {
+        let host = use_host_server();
+        let url = format!("{}/experiment-groups/{}/add-members", host, group_id);
+
+        let response = request(
+            url,
+            reqwest::Method::PATCH,
+            Some(payload),
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
+
+    pub async fn remove_members(
+        group_id: &str,
+        payload: &ExpGroupMemberRequest,
+        tenant: &str,
+        org_id: &str,
+    ) -> Result<ExperimentGroup, String> {
+        let host = use_host_server();
+        let url = format!("{}/experiment-groups/{}/remove-members", host, group_id);
+
+        let response = request(
+            url,
+            reqwest::Method::PATCH,
+            Some(payload),
+            construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
 }
--- a/crates/frontend/src/components/experiment_group_form.rs
+++ b/crates/frontend/src/components/experiment_group_form.rs
@@ -1,0 +1,441 @@
+use leptos::*;
+use serde_json::{json, Value};
+use superposition_types::{
+    api::{
+        experiment_groups::{
+            ExpGroupCreateRequest, ExpGroupMemberRequest, ExpGroupUpdateRequest,
+        },
+        workspace::WorkspaceResponse,
+    },
+    database::{
+        models::{
+            experimentation::{ExperimentGroup, TrafficPercentage},
+            ChangeReason, Description,
+        },
+        types::DimensionWithMandatory,
+    },
+    Condition, Exp,
+};
+use web_sys::MouseEvent;
+
+use crate::{
+    api::experiment_groups::{add_members, create, update},
+    components::{
+        alert::AlertType,
+        button::Button,
+        change_form::ChangeForm,
+        context_form::ContextForm,
+        input::{Input, InputType},
+    },
+    logic::Conditions,
+    providers::{alert_provider::enqueue_alert, editor_provider::EditorProvider},
+    schema::{JsonSchemaType, SchemaType},
+    types::{OrganisationId, Tenant},
+};
+
+#[component]
+pub fn add_experiment_to_group_form(
+    experiment_group: StoredValue<ExperimentGroup>,
+    handle_submit: Callback<()>,
+) -> impl IntoView {
+    let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
+    let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
+
+    let members_selected_rws = create_rw_signal(Vec::new());
+    let change_reason_rws = create_rw_signal(String::new());
+    let loading_rws = create_rw_signal(false);
+
+    let on_submit = move |event: MouseEvent| {
+        loading_rws.set(true);
+        event.prevent_default();
+
+        let tenant = tenant_rws.get().0;
+        let org_id = org_rws.get().0;
+
+        let change_reason =
+            match ChangeReason::try_from(change_reason_rws.get_untracked()) {
+                Ok(reason) => reason,
+                Err(err) => {
+                    logging::error!("{}", err);
+                    enqueue_alert(err, AlertType::Error, 5000);
+                    loading_rws.set(false);
+                    return;
+                }
+            };
+
+        spawn_local({
+            async move {
+                let payload = ExpGroupMemberRequest {
+                    change_reason,
+                    member_experiment_ids: members_selected_rws.get_untracked(),
+                };
+
+                let result = add_members(
+                    &experiment_group.get_value().id.to_string(),
+                    payload,
+                    &tenant,
+                    &org_id,
+                )
+                .await;
+
+                loading_rws.set(false);
+                match result {
+                    Ok(_) => {
+                        handle_submit.call(());
+                        let success_message =
+                            "Members added to Experiment Group successfully!";
+                        enqueue_alert(
+                            String::from(success_message),
+                            AlertType::Success,
+                            5000,
+                        );
+                    }
+                    Err(e) => {
+                        logging::error!("Error submitting member management form: {}", e);
+                        enqueue_alert(e, AlertType::Error, 5000);
+                    }
+                }
+            }
+        });
+    };
+    view! {
+        <EditorProvider>
+            <ChangeForm
+                title="Reason for Change".to_string()
+                placeholder="Enter a reason for this change".to_string()
+                value=change_reason_rws.get_untracked()
+                on_change=Callback::new(move |new_change_reason| change_reason_rws.set_untracked(new_change_reason))
+            />
+            <div class="form-control w-full">
+                <label class="label">
+                    <span class="label-text">Experiment Group Members</span>
+                </label>
+                <Input
+                    id="experiment_group_members"
+                    class="mt-5 rounded-md resize-y w-full max-w-md pt-3"
+                    schema_type=SchemaType::Single(JsonSchemaType::Array)
+                    value=Value::Array(
+                        members_selected_rws
+                            .get_untracked()
+                            .iter()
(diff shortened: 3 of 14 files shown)


crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                })
+            />
+            <div class="form-control w-full">
+                <label class="label">
+                    <span class="label-text">Experiment Group Members</span>
+                </label>
+                <Input
+                    id="experiment_group_members"
+                    class="mt-5 rounded-md resize-y w-full max-w-md pt-3"
+                    schema_type=SchemaType::Single(JsonSchemaType::Array)
```
mahatoankitkumar, line 116: Why Monaco, multiselect dropdown would be better here right, otherwise its too much of a manual task to go and copy the experiment_id and paste here.

crates/frontend/src/api.rs
```diff
@@ -660,3 +665,156 @@ pub async fn execute_autocomplete_function(
...
+        let host = use_host_server();
+        let pagination = pagination.to_string();
+
+        let url = if pagination.is_empty() {
+            format!("{}/experiment-groups?{}", host, filters)
+        } else {
+            format!("{}/experiment-groups?{}&{}", host, filters, pagination)
+        };
+
+        let response: PaginatedResponse<ExperimentGroup> = client
```
mahatoankitkumar: Use request function to do the api call

crates/frontend/src/pages/experiment_group_listing.rs
```diff
@@ -0,0 +1,661 @@
...
+    vec![
+        Column::new(
+            "name".to_string(),
+            false,
+            move |value: &str, row: &Map<String, Value>| {
+                let group_name = value.to_string();
+                let (copied_rs, copied_ws) = create_signal(false);
+                let group_id = row.get("id").map_or(String::from(""), |value| {
+                    value.as_str().unwrap_or("").to_string()
+                });
```
mahatoankitkumar:
```suggestion
                let group_id = row.get("id")
                    .and_then(|v| v.as_str())
                    .map(String::from)
                    .unwrap_or_default();
```

crates/frontend/src/pages/experiment_group_listing.rs
```diff
@@ -0,0 +1,661 @@
...
+                    .get("description")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or("")
+                    .to_string();
+
+                let change_reason = row
+                    .get("change_reason")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or("")
+                    .to_string();
```
mahatoankitkumar:
```suggestion
                let change_reason = row.get("change_reason")
                    .and_then(|v| v.as_str())
                    .map(String::from)
                    .unwrap_or_default();
```

crates/frontend/src/pages/experiment_group_listing.rs
```diff
@@ -0,0 +1,661 @@
...
+                currently_sorted: current_sort_on == SortOn::CreatedAt,
+            },
+        ),
+        Column::default("created_by".to_string()),
+        Column::new(
+            "actions".to_string(),
+            false,
+            move |_, row: &Map<String, Value>| {
+                let group = RowData {
+                    id: row
```
mahatoankitkumar: This repetition can be avoided
`fn get_row_string(row: &Map<String, Value>, key: &str) -> String {
    row.get(key)
        .and_then(|v| v.as_str())
        .map(String::from)
        .unwrap_or_default();
}`

`let group_id = get_row_string(row, "id");`

crates/frontend/src/pages/experiment_group_listing.rs
```diff
@@ -0,0 +1,661 @@
...
+                }),
+                sort_by: current_sort_by.clone(),
+                currently_sorted: current_sort_on == SortOn::LastModifiedAt,
+            },
+        ),
+        Column::default("last_modified_by".to_string()),
+        Column::default_with_sort(
+            "created_at".to_string(),
+            ColumnSortable::Yes {
+                sort_fn: Callback::new(move |_| {
```
mahatoankitkumar, line 206: Same here
`fn create_sort_callback(
    filters_rws: RwSignal<ExpGroupFilters>,
    sort_on: SortOn,
) -> Callback<(), ()> {
    Callback::new(move |_| {
        let filters = filters_rws.get();
        let sort_by = filters.sort_by.unwrap_or_default().flip();
        let new_filters = ExpGroupFilters {
            sort_on: Some(sort_on),
            sort_by: Some(sort_by),
            ..filters
        };
        filters_rws.set(new_filters);
    })
}`
Then call this for sortable columns

crates/frontend/src/components/experiment_group_form.rs
Datron: This was what was discussed with team as a first cut

crates/frontend/src/api.rs
Datron: @mahatoankitkumar For all list operations we use the client directly. Its the same for fetch_webhooks, fetch_workspaces, etc. I forgot why but request function isn't used for list operations

crates/frontend/src/pages/experiment_group_listing.rs
Datron: Yeah this needs to be improved at the table component level. This interface is bad. We can fix this once and for all with that change.

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -119,6 +119,11 @@ pub fn add_members(
     schema_name: &SchemaName,
     user: User,
 ) -> superposition::Result<Json<ExperimentGroup>> {
+    if req.member_experiment_ids.is_empty() {
```
ayushjain17, line 126: is this error needed ?
isn't it fine without this error also ? allows idempotency in a way
if we need this, then shouldn't we add this for remove as well ?

crates/frontend/src/pages/experiment_groups.rs
```diff
@@ -0,0 +1,345 @@
...
+
+                                <ExperimentGroupInfo group=resource_group />
+
+                                <div class="card rounded-xl w-full bg-base-100 shadow">
+                                    <div class="card-body">
+                                        <div class="flex justify-between">
+                                            <h2 class="card-title">Member Experiments</h2>
+                                            <DrawerBtn drawer_id="add_member_group_drawer"
+                                                .to_string()>
+                                                Add Member <i class="ri-add-large-fill ml-2"></i>
```
mahatoankitkumar:
```suggestion
                                                Add Members <i class="ri-add-large-fill ml-2"></i>
```
Can add multiple members in one go right?

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                            "Members added to Experiment Group successfully!";
+                        enqueue_alert(
+                            String::from(success_message),
+                            AlertType::Success,
+                            5000,
+                        );
+                    }
+                    Err(e) => {
+                        logging::error!(
+                            "Error submitting adding experiment group form: {}",
```
mahatoankitkumar:
```suggestion
                            "Error adding experiment group form: {}",
```

crates/frontend/src/pages/experiment_groups.rs
```diff
@@ -0,0 +1,345 @@
...
+                                </div>
+                                <Drawer
+                                    id="add_member_group_drawer".to_string()
+                                    header="Add a member to the group"
+                                    handle_close=move || { close_drawer("add_member_group_drawer") }
+                                >
+                                    <AddExperimentToGroupForm
+                                        experiment_group=resource_group
+                                        handle_submit=Callback::new(move |_| {
+                                            close_drawer("add_member_group_drawer");
```
mahatoankitkumar:
```suggestion
                                    id="add_members_group_drawer".to_string()
                                    header="Add members to the group"
                                    handle_close=move || { close_drawer("add_members_group_drawer") }
                                >
                                    <AddExperimentToGroupForm
                                        experiment_group=resource_group
                                        handle_submit=Callback::new(move |_| {
                                            close_drawer("add_members_group_drawer");
```

crates/frontend/src/api.rs
mahatoankitkumar: We can right, why not use it?

crates/frontend/src/api.rs
```diff
@@ -660,3 +665,156 @@ pub async fn execute_autocomplete_function(
...
+    ) -> Result<PaginatedResponse<ExperimentGroup>, ServerFnError> {
+        let client = reqwest::Client::new();
+        let host = use_host_server();
+        let pagination = pagination.to_string();
+
+        let url = if pagination.is_empty() {
+            format!("{}/experiment-groups?{}", host, filters)
+        } else {
+            format!("{}/experiment-groups?{}&{}", host, filters, pagination)
+        };
```
ayushjain17:
```suggestion
        let url = format!("{}/experiment-groups?{}&{}", host, filters, pagination);
```

crates/frontend/src/api.rs
```diff
@@ -660,3 +665,156 @@ pub async fn execute_autocomplete_function(
         .collect();
     Ok(result)
 }
+
+pub mod experiment_group {
```
ayushjain17: do we need this mod ?
even when the internal functions do repeat the phrase `experiment_group` in them
seems redundant

crates/frontend/src/api.rs
ayushjain17: @Datron ideally we should use the request function itself
at some point we would need to sit and convert all of them to request function
as, in frontend when you want to do something like a middleware you would be able to do it only when the client used is wrapped in some common functionality which you want across

so instead of adding the work later, we might do it while adding itself

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                        <Button
+                            class="pl-[70px] pr-[70px] w-48 h-12".to_string()
+                            text="Submit".to_string()
+                            on_click=on_submit
+                            loading
+                        />
+                    }
+                }}
+
+            </div>
```
ayushjain17:
```suggestion
          {move || {
              let loading = loading_rws.get();
              view! {
                  <Button
                      class="h-12 w-48 mt-8 px-[70px]".to_string()
                      text="Submit".to_string()
                      on_click=on_submit
                      loading
                  />
              }
          }}
```

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                            enqueue_alert(
+                                format!("Failed to parse change reason: {}", err),
+                                AlertType::Error,
+                                5000,
+                            );
+                            return;
+                        }
+                    };
+                    change_reason_rws.set_untracked(new_reason)
+                })
```
ayushjain17: can we move these validations to the submit call, just like how we have in all the places, everytime while someone is typing the validations is logic would be run and on all momentary errors alert will keep coming
not a very good practice or experience

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+#[component]
+pub fn add_experiment_to_group_form(
+    experiment_group: StoredValue<ExperimentGroup>,
+    handle_submit: Callback<()>,
+) -> impl IntoView {
+    let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
+    let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
+
+    let members_selected_rws = create_rw_signal(Vec::new());
+    let change_reason_rws = create_rw_signal(ChangeReason::default());
```
ayushjain17: using these default values will encourage people to not fill it in at all

crates/frontend/src/components/experiment_group_form.rs
ayushjain17: @Datron we already have multi-select component we can directly use it

and even if we dont want to make api call to fetch in the list of existing valid experiments, using monaco is here is really not a good experience at all

if I am not wrong, what was discussed was an input box where folks can past the value and that gets added as a pill (again component for this also already exists)
and this can be kept as the final cut as well

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                            <Button
+                                class="pl-[70px] pr-[70px] w-48 h-12".to_string()
+                                text="Submit".to_string()
+                                on_click=on_submit
+                                loading
+                            />
+                        }
+                    }}
+
+                </div>
```
ayushjain17:
```suggestion
          {move || {
              let loading = loading_rws.get();
              view! {
                  <Button
                      class="h-12 w-48 mt-8 px-[70px]".to_string()
                      text="Submit".to_string()
                      on_click=on_submit
                      loading
                  />
              }
          }}
```

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                        class="input input-bordered w-full max-w-md"
+                    />
+                </div>
+
+                <Show when=move || !is_edit>
+                    <div class="form-control w-full">
+                        <label class="label">
+                            <span class="label-text">Experiment Group Members (Optional)</span>
+                        </label>
+                        <Input
```
ayushjain17, line 387: same for this monaco input as well

@Datron we already have multi-select component we can directly use it

and even if we dont want to make api call to fetch in the list of existing valid experiments, using monaco is here is really not a good experience at all

if I am not wrong, what was discussed was an input box where folks can past the value and that gets added as a pill (again component for this also already exists)
and this can be kept as the final cut as well

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                                    enqueue_alert(
+                                        format!("Failed to parse traffic percentage: {}", err),
+                                        AlertType::Error,
+                                        5000,
+                                    );
+                                    return;
+                                }
+                            };
+                            traffic_percentage_rws.set_untracked(value);
+                        }
```
ayushjain17: same here for the validations
can we move these validations to the submit call, just like how we have in all the places, everytime while someone is typing the validations is logic would be run and on all momentary errors alert will keep coming
not a very good practice or experience

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                                logging::error!("Failed to parse description: {}", err);
+                                enqueue_alert(
+                                    format!("Failed to parse description: {}", err),
+                                    AlertType::Error,
+                                    5000,
+                                );
+                                return;
+                            }
+                        };
+                        change_reason_rws.set_untracked(new_reason)
```
ayushjain17: same
can we move these validations to the submit call, just like how we have in all the places, everytime while someone is typing the validations is logic would be run and on all momentary errors alert will keep coming
not a very good practice or experience

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                                logging::error!("Failed to parse description: {}", err);
+                                enqueue_alert(
+                                    format!("Failed to parse description: {}", err),
+                                    AlertType::Error,
+                                    5000,
+                                );
+                                return;
+                            }
+                        };
+                        group_description_rws.set_untracked(new_description)
```
ayushjain17: same
can we move these validations to the submit call, just like how we have in all the places, everytime while someone is typing the validations is logic would be run and on all momentary errors alert will keep coming
not a very good practice or experience

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                    <input
+                        disabled=is_edit
+                        value=group_name_rws.get_untracked()
+                        on:input=move |ev| group_name_rws.set_untracked(event_target_value(&ev))
+                        type="text"
+                        name="exp_group_name"
+                        id="exp_group_name"
+                        placeholder="Enter an experiment group name, eg: Bangalore Region"
+                        class="input input-bordered w-full max-w-md"
+                    />
```
ayushjain17, line 317: can we use the `Input` component here instead

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+                        };
+                        change_reason_rws.set_untracked(new_reason)
+                    })
+                />
+
+                <div class="form-control w-full">
+                    <label class="label">
+                        <span class="label-text">Traffic Percentage</span>
+                    </label>
+                    <input
```
ayushjain17, line 358: can we use the Input component here instead

crates/frontend/src/components/experiment_group_form.rs
```diff
@@ -0,0 +1,448 @@
...
+    let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
+    let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
+    let experiment_group_id_rws = create_rw_signal(group_id);
+    let workspace_settings = use_context::<StoredValue<WorkspaceResponse>>().unwrap();
+    let (context_rs, context_ws) = create_signal(context);
+    let group_name_rws = create_rw_signal(group_name);
+    let group_description_rws =
+        create_rw_signal(Description::try_from(group_description).unwrap_or_default());
+    let change_reason_rws = create_rw_signal(ChangeReason::default());
+    let traffic_percentage_rws = create_rw_signal(traffic_percentage);
```
ayushjain17: let's use String and i64 here simply
and then convert on API call

crates/frontend/src/pages/experiment_group_listing.rs
ayushjain17, line 1: review pending

crates/frontend/src/pages/experiment_groups.rs
ayushjain17, line 1: reviewing pending

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: I understand the argument for idem-potency but why do all the SQL queries, validation and operations just to return to the same state? I thought it better to exit early.

crates/frontend/src/api.rs
Datron: Yeah I'm thinking I'd use this to separate different module function calls. I want to refactor api.rs to include modules like this for each API set that we have. I can get rid of the `experiment_group` in the function names

crates/frontend/src/components/experiment_group_form.rs
Datron: @ayushjain17 if you remember, I had a component already in place that did this. that was rejected and a plain text box was recommended. The experience with monaco is better. We can come back to this, but its not a blocker.

crates/frontend/src/components/experiment_group_form.rs
Datron: @ayushjain17 the input component expects a serde_json::Value for the value field in input. It expects a more arguments that don't make a lot of sense for a simple string field. I believe @mahatoankitkumar pointed this out to me last time I asked him for this change.

```
#[component]
pub fn input(
    value: Value,
    schema_type: SchemaType,
    #[prop(into, default = String::new())] placeholder: String,
    #[prop(into)] on_change: Callback<Value, ()>,
    #[prop(into)] r#type: InputType,
    #[prop(default = false)] disabled: bool,
    #[prop(into, default = String::new())] id: String,
    #[prop(into, default = String::new())] class: String,
    #[prop(into, default = String::new())] name: String,
    #[prop(default = None)] operator: Option<Operator>,
    #[prop(default = None)] autocomplete_function: Option<AutoCompleteCallback>,
)
```

crates/frontend/src/api.rs
Datron: Sure trying this change.

crates/frontend/src/components/experiment_group_form.rs
Datron: https://github.com/juspay/superposition/pull/540/files#r2168425712

crates/frontend/src/components/experiment_group_form.rs
Datron: https://github.com/juspay/superposition/pull/540/files#r2168398742

crates/frontend/src/pages/experiment_groups.rs
Datron: Ah yes, singular vs plural

crates/frontend/src/components/experiment_group_form.rs
ayushjain17: but most of the fields are anyways optional here, right ?

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
ayushjain17: I am not against the change
all I was saying, if we need it then can we add it to remove function as well

crates/frontend/src/api.rs
ayushjain17: yeah that is still better to avoid the redundancy at least

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: Sure will add it

crates/frontend/src/components/experiment_group_form.rs
Datron: Yeah but converting string to value each time seems redundant right?

Datron: Also on every user input

ayushjain17: ideally we should exposed the proper raw components which was meant for the exact types

crates/frontend/src/pages/experiment_group_listing.rs
```diff
@@ -0,0 +1,637 @@
...
+                    .and_then(|v| v.as_object().cloned())
+                    .unwrap_or_default();
+                let id = get_row_string_fn(row, "id");
+                let conditions =
+                    Conditions::from_context_json(&context).unwrap_or_default();
+
+                view! {
+                    <div class="w-[300px]">
+                        <ConditionComponent conditions grouped_view=false id />
+                    </div>
```
ayushjain17: ConditionComponent already accepts class, we dont need to wrap it in a div for this
```suggestion
                <ConditionComponent conditions grouped_view=false id class="w-[300px] />
```

crates/frontend/src/pages/experiment_groups.rs
```diff
@@ -0,0 +1,341 @@
...
+#[component]
+fn experiment_group_info(group: StoredValue<ExperimentGroup>) -> impl IntoView {
+    let group = group.get_value();
+    let conditions = Conditions::from_context_json(&group.context).unwrap_or_default();
+    view! {
+        <div class="card bg-base-100 max-w-screen shadow">
+            <div class="card-body flex flex-row gap-2 flex-wrap">
+                <div class="h-fit w-[300px]">
+                    <ConditionComponent conditions id="experiment-group-context" />
+                </div>
```
ayushjain17:
```suggestion
                <ConditionComponent conditions id="experiment-group-context" class="h-fit w-[300px]" />
```

crates/frontend/src/pages/experiment_groups.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+#[component]
+pub fn experiment_groups() -> impl IntoView {
+    let group_params = use_params_map();
+    let tenant_rws = use_context::<RwSignal<Tenant>>().unwrap();
+    let org_rws = use_context::<RwSignal<OrganisationId>>().unwrap();
+
+    let (delete_modal_rs, delete_modal_ws) = create_signal(false);
+    let delete_group_rws = create_rw_signal(RemoveRequest::default());
```
ayushjain17, line 183: RemoveRequest has default ?!

ayushjain17 (changes requested):

crates/frontend/src/pages/experiment_groups.rs
Datron: Yeah both its fields are strings, so the default is empty strings. 😂

crates/frontend/src/components/experiment_group_form.rs
Datron: Yeah the component can change to better meet our requirements

PR #543, opened 2025-06-18
feat: AuthZ setup

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -429,8 +430,9 @@ async fn reduce_config_key(
     })
 }
 
+#[auth_action]
```
knutties: Let us call this macro `authorized`.

crates/frontend/src/api.rs
```diff
@@ -35,33 +32,34 @@ use superposition_types::{
...
+
 // #[server(GetDefaultConfig, "/fxn", "GetJson")]
 pub async fn fetch_default_config(
     pagination: &PaginationParams,
     filters: &DefaultConfigFilters,
-    tenant: String,
-    org_id: String,
-) -> Result<PaginatedResponse<DefaultConfig>, ServerFnError> {
-    let client = reqwest::Client::new();
+    tenant: &str,
```
knutties, line 43: let us change to `workspace` - we seem to be doing it in other places.

crates/frontend/src/api.rs
```diff
@@ -35,33 +32,34 @@ use superposition_types::{
...
-        .json()
-        .await
-        .map_err(|e| ServerFnError::new(e.to_string()))?;
-    Ok(response)
+
+    let response = request(
+        url,
+        reqwest::Method::GET,
+        None::<()>,
+        construct_request_headers(&[("x-tenant", tenant), ("x-org-id", org_id)])?,
```
knutties, line 58: Can we start sending `x-workspace` for all these calls now ?  Will it work in the internal deployment as well?

crates/service_utils/src/middlewares/auth_z.rs
```diff
@@ -0,0 +1,202 @@
...
+            // "CASBIN" => Arc::new(
+            //     CasbinPolicyEngine::new(kms_client, app_env, None)
+            //         .await
+            //         .unwrap(),
+            // ),
+            "DISABLED" => Arc::new(NoAuth),
+            _ => panic!("Missing/Unknown authorizer."),
+        };
+        Self(ap)
+    }
```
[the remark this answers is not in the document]
ayushjain17: expected

crates/frontend/src/api.rs
ayushjain17: will raise a separate PR for this cleanup

crates/frontend/src/components/workspace_form.rs
```diff
@@ -31,8 +31,8 @@ pub fn workspace_form(
     #[prop(default = Metrics::default())] metrics: Metrics,
     #[prop(default = false)] allow_experiment_self_approval: bool,
     #[prop(default = true)] auto_populate_control: bool,
-    #[prop(default = true)] enable_context_validation: bool,
-    #[prop(default = true)] enable_change_reason_validation: bool,
+    #[prop(default = false)] enable_context_validation: bool,
+    #[prop(default = false)] enable_change_reason_validation: bool,
```
knutties, line 35: change_reason_validation - we enable this by default?

ayushjain17: previously we were enabling this by default

knutties: Or maybe it is easier to keep it off for both and have them enable it if needed.

ayushjain17: yeah kept it that way only now

PR #546, opened 2025-06-20
Python openfeature client

pratikmishra356:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


clients/openfeature/python/superposition_client.py
```diff
@@ -0,0 +1,1210 @@
+
```
sauraww, line 1: Instead of calling this superposition client , Can we call it configuration client and have experimentation client as well seperately ?

sauraww (changes requested):

clients/openfeature/python/cac_client.py
```diff
@@ -0,0 +1,373 @@
...
+        self.evaluation_cache: Dict[str, Dict[str, Any]] = {}
+        self._polling_task = None
+
+    async def create_config(self) -> None:
+        async def poll_config(interval: int, timeout: int) -> None:
+            while True:
+                try:
+                    latest_config = await self._get_config(self.superposition_options)
+                    if latest_config is None and self.cached_config is None:
+                        print("⚠️ No config found, using fallback config.")
```
sauraww: Can we get rid of these emojis please ?

sauraww (changes requested):

PR #554, opened 2025-06-24
fix: changes to smithy model

Datron:
## Problem
Autocomplete function name is missing from smithy dimension model

## Solution
Add it


mahatoankitkumar (approved):

smithy/models/dimension.smithy
ayushjain17, line 1: can we please update tests as well for this

makefile
```diff
@@ -248,3 +249,6 @@ schema-file:
 	diesel print-schema --schema superposition > crates/superposition_types/src/database/superposition_schema.rs
 	git apply crates/superposition_types/src/database/superposition_schema.patch
 	git apply crates/superposition_types/src/database/schema-timestamp-migration.patch
+
+smithy: smithy-clients setup-clients
+	cd clients/tests/typescript && bun install
```
ayushjain17: can we add this to `smithy-clients` instead

makefile
```diff
@@ -42,7 +42,6 @@ export SMITHY_MAVEN_REPOS = https://repo.maven.apache.org/maven|https://sandbox.
 	validate-psql-connection
 	cac
 	schema-file
-	setup-clients
```
ayushjain17, line 45: this needs to be added back

makefile
```diff
@@ -209,6 +209,7 @@ smithy-clients: smithy-build
 		mkdir -p "$(SMITHY_CLIENT_DIR)/$$name"; \
 		cp -r "$$d"/* "$(SMITHY_CLIENT_DIR)/$$name"; \
 	done
+	make setup-clients
```
ayushjain17: @ShreyBana can you help here avoid recursive `make`

clients/tests/typescript/tests/default_config.test.ts
```diff
@@ -250,6 +279,29 @@ describe("Default Config API Integration Tests", () => {
...
+                    },
+                },
+                value: { name: "valid Value" },
+                description: "Test configuration",
+                autocomplete_function_name: "auto_fn",
+                change_reason: "Test function completion",
+            };
+
+            const cmd = new CreateDefaultConfigCommand(input);
+            expect(superpositionClient.send(cmd)).resolves.toBeDefined();
```
ayushjain17: can we add checks to see that the data returned is correct

docker-compose/postgres/db_init.sql
```diff
@@ -1378,7 +1378,6 @@ CREATE TABLE IF NOT EXISTS localorg_test.experiment_groups(
 ALTER TABLE superposition.workspaces add column if not exists config_version bigint;
 
 -- keeping self approval enabled by default for test and dev workspaces
-ALTER TABLE superposition.workspaces ADD COLUMN IF NOT EXISTS allow_experiment_self_approval boolean NOT NULL DEFAULT true;
-ALTER TABLE superposition.workspaces ALTER COLUMN allow_experiment_self_approval SET DEFAULT false;
+ALTER TABLE superposition.workspaces ADD COLUMN IF NOT EXISTS allow_experiment_self_approval boolean NOT NULL DEFAULT false;
```
ayushjain17, line 1381: why are we changing this ?

ayushjain17: we want to make it true in local setup, else with a single user only, you will not be able to approve any experiment requests

Datron: This is local setup right? I reduced the queries into one, so the default needs to be true?

makefile
Datron: What is the issue here?

ShreyBana: Don't know a way to avoid it as it is. People generally have different targets which do all the things, like: `clients: smithy-clients setup-clients`, then you just run `make clients`. This way you have a command which avoids the extra step as well.
Can we do it that way @Datron ?

ayushjain17: +1

Datron: Done

PR #555, opened 2025-06-24
fix: experiment client fetch fix

pratikmishra356:
## Problem
Fetch experiment from client using all

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/experimentation_client/src/lib.rs
```diff
@@ -249,13 +248,12 @@ async fn get_experiments(
...
-    let mut page = 1;
 
     let list_filters = ExperimentListFilters {
-        status: Some(CommaSeparatedQParams(
-            ExperimentStatusType::iter().collect(),
-        )),
+        status: Some(CommaSeparatedQParams(vec![
+            ExperimentStatusType::CREATED,
+            ExperimentStatusType::INPROGRESS,
+        ])),
```
ayushjain17: this needs to be a condition

crates/experimentation_client/src/lib.rs
```diff
@@ -265,34 +263,26 @@ async fn get_experiments(
         sort_on: None,
         sort_by: None,
     };
+    let pagination_params = PaginationParams {
+        page: None,
+        count: None,
+        all: Some(true), // Fetch all entries
+    };
```
ayushjain17:
```suggestion
    let pagination_params = PaginationParams::all();
```

crates/experimentation_client/src/lib.rs
```diff
@@ -265,34 +263,26 @@ async fn get_experiments(
...
-        if page < list_experiments_response.total_pages {
-            page += 1;
-        } else {
-            break;
-        }
+    for experiment in experiments.into_iter() {
+        curr_exp_store.insert(experiment.id.to_string(), experiment);
     }
 
     Ok(curr_exp_store)
```
ayushjain17: since using all no need to insert
```suggestion
    Ok(experiments.into_iter().map(|exp|| (experiment.id.to_string(), experiment)).collect())
```

crates/experimentation_client/src/lib.rs
pratikmishra356: it does need to be in condition

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -60,6 +60,10 @@ impl ExperimentStatusType {
         }
     }
 
+    pub fn get_active() -> Vec<ExperimentStatusType> {
```
ayushjain17:
```suggestion
    pub fn get_active() -> Vec<Self> {
```

ShreyBana (approved):

crates/experimentation_client/src/lib.rs
```diff
@@ -265,37 +260,23 @@ async fn get_experiments(
         sort_on: None,
         sort_by: None,
     };
+    let pagination_params = PaginationParams::all_entries();
+    let endpoint = format!("{hostname}/experiments?{pagination_params}&{list_filters}");
+    let list_experiments_response = http_client
+        .get(endpoint)
+        .header("x-tenant", tenant.to_string())
```
ayushjain17, line 268: don't know why but we are not passing `"If-Modified-Since"` header here, can we please add changes for that as well, it will reduce the amount of data transfer and the changes in the memory also

crates/experimentation_client/src/lib.rs
```diff
@@ -67,19 +67,21 @@ impl Client {
...
+                        let mut exp_store = self.experiments.write().await;
+                        *exp_store = experiments;
+                        *start_date = Utc::now();
+                    }
+                    Err(e) => {
+                        log::error!(
+                            "Failed to fetch experiments from the server with error: {}",
+                            e
+                        );
+                    }
```
Datron, line 82: Just logging the error isn't sufficient here. Either fail loudly by panicking or provide a default. The loop may also need to be exited

crates/experimentation_client/src/lib.rs
```diff
@@ -265,37 +261,39 @@ async fn get_experiments(
...
+            return Err(format!(
+                "{} EXP: skipping update, remote not modified",
+                tenant
+            ));
         }
-    }
+        StatusCode::OK => log::info!(
+            "{}",
+            format!("{} EXP: new config received, updating", tenant)
+        ),
```
Datron, line 284: Do this

```suggestion
        StatusCode::OK => log::trace!("{} EXP: new config received, updating", tenant),
```

Nitpick
Also the log line would be more useful if it contained the number of changes or something

crates/experimentation_client/src/lib.rs
ayushjain17: why should we fail ?
if one iteration fails, it can just go on safely to the next iteration, the previous data would anyways be there, right ?

pratikmishra356: yes @Datron its the polling strategy , we should not exit the loop

crates/experimentation_client/src/lib.rs
pratikmishra356: Let it be log::info , any reason for trace ?

PR #561, opened 2025-06-27
feat: add exp core

pratikmishra356:
## Problem
added experimentation logic and niffi

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/core/src/experiment.rs
+++ b/crates/core/src/experiment.rs
@@ -1,417 +1,139 @@
+use std::collections::HashSet;
+
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
-use std::collections::HashMap;
-use superposition_types::{
-    api::experiments::ExperimentResponse,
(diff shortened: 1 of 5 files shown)


crates/core/src/experiment.rs
```diff
@@ -1,417 +1,139 @@
...
-    _toss: i32,
-) -> Vec<&'a Variant> {
-    variants.to_vec()
 }
 
-// Bucketed variant selection: complex distribution logic from existing client
-fn select_variants_bucketed<'a>(
-    variants: &'a [&'a Variant],
+#[derive(Serialize, Deserialize, Clone, Debug, uniffi::Record)]
+pub struct Experiment {
```
ShreyBana: Can we call this `FfiExperiment`? It'll be clear in generated code that this is different from the usual `Experiment` object.

pratikmishra356: yes sure

PR #569, opened 2025-07-02
feat(java-openf): Experimentation support.

ShreyBana:
Experimentation support for Java Open-feature-provider.


.github/workflows/generate_binaries.yaml
```diff
@@ -99,22 +99,36 @@ jobs:
...
-                      superposition_cac_client-x86_64-apple-darwin.zip
-                      superposition_experimentation_client-x86_64-apple-darwin.zip
-                      superposition_core-x86_64-apple-darwin.zip
-                      superposition_cac_client-aarch64-apple-darwin.zip
-                      superposition_experimentation_client-aarch64-apple-darwin.zip
-                      superposition_core-aarch64-apple-darwin.zip
-                      superposition_cac_client-x86_64-pc-windows-msvc.zip
-                      superposition_experimentation_client-x86_64-pc-windows-msvc.zip
-                      superposition_core-x86_64-pc-windows-msvc.zip
+                      artifacts_dir/superposition_cac_client-x86_64-unknown-linux-gnu.zip/superposition_cac_client-x86_64-unknown-linux-gnu.zip
```
knutties: @Datron  - do we have separate FFI modules for cac/experimentation still?  I thought we have just one with 2 functions (`evaluate` and `getAvailableVariants`) ?

docs/docusaurus.config.ts
```diff
@@ -18,7 +18,7 @@ const config: Config = {
   url: 'https://your-docusaurus-site.example.com',
   // Set the /<baseUrl>/ pathname under which your site is served
   // For GitHub pages deployment, it is often '/<projectName>/'
-  baseUrl: '/juspay/superposition/',
+  baseUrl: '/superposition/',
```
knutties: @Datron - is this change correct ?  Did we need `juspay` in the basepath at all?

.github/workflows/generate_binaries.yaml
Datron: @knutties this is for legacy clients using haskell without nix

docs/docusaurus.config.ts
Datron: @knutties all these changes are already in main. I think we just need to rebase this PR

Datron (approved):

clients/java/open-feature-provider/src/main/java/io/superposition/openfeature/RefreshJob.java
```diff
@@ -1,35 +1,40 @@
...
 
-import java.sql.Ref;
-import java.util.Collection;
 import java.util.Optional;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
 interface RefreshJob<T> {
     Logger logger = LoggerFactory.getLogger(RefreshJob.class);
```
knutties, line 13: We might not need this yet as we do not have any default implementations in this interface.  Can be removed.

knutties (approved):

PR #574, opened 2025-07-03
feat(java): Publishing scripts

ShreyBana:
- Setup gradle publish for each of the required java modules.
  All modules are published to `(root)build/m2`.
- README for `open-feature-provider`.

### DIFF
--- a/clients/java/open-feature-provider/src/test/java/io/juspay/superposition/openfeature/DefaultConfigPopulator.java
+++ b/clients/java/open-feature-provider/src/test/java/io/juspay/superposition/openfeature/DefaultConfigPopulator.java
@@ -14,6 +14,7 @@ import java.net.http.HttpClient;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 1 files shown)


clients/java/open-feature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionProviderOptions.java
```diff
@@ -0,0 +1,51 @@
...
+     * Options for experimentation mode, allowing separate refresh and cache settings.
+     */
+    @Builder
+    public static class ExperimentationOptions {
+        /** Refresh strategy for experimentation data. */
+        @NonNull
+        RefreshStrategy refreshStrategy;
+        /** Evaluation cache options for experimentation data (optional). */
+        @Nullable
+        EvaluationCacheOptions evaluationCacheOptions;
```
knutties, line 49: Is this being used at all ?  Better to remove this I think as there is no evaluationCache implementation.

clients/java/open-feature-provider/src/main/java/io/juspay/superposition/openfeature/options/RefreshStrategy.java
knutties, line 1: @pratikmishra356 @sauraww - can you both compare your RefreshStrategy implementation and this one?  I see some differences compared to the Slack discussion we had.

knutties (approved):

Datron (approved):

PR #585, opened 2025-07-09
feat: simplify workspace migration with an API endpoint

Datron:
## Problem
It is hard to migrate existing workspaces to the latest db state, especially when a large number of workspaces are present

## Solution
Provide an API that can apply the current workspace_template.sql file to existing workspaces, bringing them up to the latest version


smithy/models/workspace.smithy
```diff
@@ -27,6 +27,7 @@ resource Workspace {
     put: UpdateWorkspace
     operations: [
         CreateWorkspace
+        UpdateWorkspaceDatabase
```
ayushjain17: should this be exposed via SDK ?

workspace_template.sql
```diff
@@ -335,42 +367,61 @@ CREATE TABLE IF NOT EXISTS {replaceme}.type_templates (
...
+        FROM information_schema.columns
+        WHERE table_name='type_templates' and table_schema='{replaceme}' and column_name='last_modified')
+  THEN
+    ALTER TABLE {replaceme}.type_templates rename column last_modified to last_modified_at;
+  END IF;
+END $$;
 
 ALTER TABLE {replaceme}.type_templates
-add column last_modified_by varchar(200) not null default('null');
+ADD COLUMN IF NOT EXISTS last_modified_by varchar(200) not null default('null');
```
ayushjain17, line 407: should we also update the migration files as well ?

docker-compose.yaml
```diff
@@ -7,6 +7,8 @@ services:
     environment:
       POSTGRES_PASSWORD: "docker"
       POSTGRES_DB: "config"
+    volumes:
+      - ./docker-compose/postgres/data:/var/lib/postgresql/data
```
ayushjain17, line 11: what is this change for ?

crates/frontend/src/api.rs
ayushjain17, line 1: why are there changes in the frontend crate ?

docker-compose.yaml
Datron: This helps persist data in the postgres container

smithy/models/workspace.smithy
Datron: It is right?

workspace_template.sql
Datron: No those can be straight forward - this file is not a representation of the migration files

crates/frontend/src/api.rs
Datron: Just wrapped the workspaces API calls into a module

smithy/models/workspace.smithy
Datron: Yes it needs to, that way folks can use the SDK to write scripts that can do workspace migrations. That is a use case

ayushjain17 (dismissed):
JS patch needs to be regenerated

crates/frontend/src/api.rs
```diff
@@ -743,6 +724,88 @@ pub async fn get_webhook(
     parse_json_response(response).await
 }
 
+pub mod workspaces {
```
ayushjain17: this should not be needed anymore

crates/frontend/src/api.rs
ayushjain17, line 1: changes in this PR are are already merged in control population PR, we can remove the changes from frontend crate

workspace_template.sql
ayushjain17, line 1: everything else seems fine
just going through all the queries once to cross verify

workspace_template.sql
```diff
@@ -384,34 +435,49 @@ ALTER TABLE {replaceme}.default_configs ADD COLUMN IF NOT EXISTS change_reason T
...
-        'default@superposition.io',
-        '{"type": "string","pattern": ".*"}'::json,
-        null,
-        'variantIds are used by experimentation module to manage and select variations',
-        'initial setup'
-);
+DO $$ BEGIN
+    IF NOT EXISTS(SELECT *
+        FROM {replaceme}.dimensions
+        WHERE dimension = 'variantIds')
```
ayushjain17, line 457: why are we not doing it the same way as its done for type_templates
`ON CONFLICT DO NOTHING` ?

if either works, can we keep things consistent

workspace_template.sql
```diff
@@ -617,13 +667,7 @@ ALTER TABLE {replaceme}.experiment_groups
 ADD COLUMN IF NOT EXISTS group_type public.group_type DEFAULT 'USER_CREATED';
 
 ALTER TABLE {replaceme}.functions
```
ayushjain17, line 669: why are these queries changed ?
they were already written in a way that there are no conflicts on re-running them

workspace_template.sql
```diff
@@ -335,42 +367,61 @@ CREATE TABLE IF NOT EXISTS {replaceme}.type_templates (
...
 CREATE INDEX IF NOT EXISTS type_templates_created_at_index ON {replaceme}.type_templates(created_at);
-CREATE INDEX IF NOT EXISTS type_templates_last_modifed_index ON {replaceme}.type_templates(last_modified);
+
+DO $$ BEGIN
+    IF EXISTS(SELECT *
+        FROM information_schema.columns
+        WHERE table_name='type_templates' and table_schema='{replaceme}' and column_name='last_modified')
+  THEN
+    CREATE INDEX IF NOT EXISTS type_templates_last_modifed_index ON {replaceme}.type_templates(last_modified);
+  END IF;
```
ayushjain17, line 377: shouldn't we add an else for this with the following query:
```CREATE INDEX IF NOT EXISTS type_templates_last_modifed_index ON {replaceme}.type_templates(last_modified_at);```

Comment (Datron) [reply workspace_template.sql:377]: Indexes are auto updated by postgres when the column is renamed

File: workspace_template.sql (hunk 5 above)
Comment (Datron) [reply workspace_template.sql:457]: dimensions table has a deferred constraint, which on conflict doesn't support
```ON CONFLICT does not support deferrable unique constraints/exclusion constraints as arbiters```

File: workspace_template.sql (hunk 6 above)
Comment (Datron) [reply workspace_template.sql:669]: I just merged them

Comment (ayushjain17) [reply workspace_template.sql:669]: its sort of incorrect, I had added them in that way because we need these all separately
there was a reason for it being separate steps as it was doing more, which has been removed now

can we please revert this change

File: workspace_template.sql (hunk 7 above)
Comment (ayushjain17) [reply workspace_template.sql:377]: fair, but if this `index` was never created for some reason (being left out in any manual update), would lead to a situation that it will never get created on the subsequent auto updations

just simply adding this `else` block would ensure both the cases

this is safe as well, as the suggested query is with `IF NOT EXISTS`

File: workspace_template.sql (hunk 6 above)
Comment (Datron) [reply workspace_template.sql:669]: These were the previous queries
```
ALTER TABLE {replaceme}.functions
ADD COLUMN IF NOT EXISTS created_at TIMESTAMP,
ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL DEFAULT('NOT AVAILABLE');

ALTER TABLE {replaceme}.functions ALTER COLUMN created_by DROP DEFAULT;

UPDATE {replaceme}.functions SET created_at = last_modified_at WHERE created_at IS NULL;

ALTER TABLE {replaceme}.functions
ALTER COLUMN created_at SET DEFAULT CURRENT_TIMESTAMP,
ALTER COLUMN created_at SET NOT NULL;
```
This is what I simplified it to

```
ALTER TABLE {replaceme}.functions
ADD COLUMN IF NOT EXISTS created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL;

UPDATE {replaceme}.functions SET created_at = last_modified_at WHERE created_at IS NULL;
```

Can you explain what is different?

Comment (ayushjain17) [reply workspace_template.sql:669]: `DEFAULT('NOT AVAILABLE')`

Comment (Datron) [reply workspace_template.sql:669]: You don't need that right, the default is dropped in the next query, and then this query is run
```
UPDATE {replaceme}.functions SET created_at = last_modified_at WHERE created_at IS NULL;
```
Isn't the above query enough?

File: workspace_template.sql (hunk 7 above)
Comment (Datron) [reply workspace_template.sql:377]: This PR doesn't solve cases where the original workspace was not setup correctly, it only migrates a workspace from previous valid state to the latest valid state.

File: workspace_template.sql (hunk 6 above)
Comment (Datron) [reply workspace_template.sql:669]: ~Oh I see, Would this work~
```
UPDATE {replaceme}.functions SET created_by = last_modified_by WHERE created_by IS NULL;
```
Will add this instead:
```
UPDATE {replaceme}.functions SET created_by = "NOT AVAILABLE" WHERE created_by IS NULL;
```

Comment (Datron) [reply workspace_template.sql:669]: Neither of these work, using DEFAULT itself.

File: workspace_template.sql (hunk 7 above)
Comment (ShreyBana) [reply workspace_template.sql:377]: @ayushjain17 - I think to provide that kind of stability, you must do database versioning as well.

File: workspace_template.sql [hunk 8]
```diff
@@ -617,13 +667,7 @@ ALTER TABLE {replaceme}.experiment_groups
 ADD COLUMN IF NOT EXISTS group_type public.group_type DEFAULT 'USER_CREATED';
 
 ALTER TABLE {replaceme}.functions
-ADD COLUMN IF NOT EXISTS created_at TIMESTAMP,
-ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL DEFAULT('NOT AVAILABLE');
-
-ALTER TABLE {replaceme}.functions ALTER COLUMN created_by DROP DEFAULT;
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
ayushjain17:
```suggestion
ADD COLUMN IF NOT EXISTS created_at TIMESTAMP,
```

workspace_template.sql
```diff
@@ -617,13 +667,7 @@ ALTER TABLE {replaceme}.experiment_groups
...
+ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL DEFAULT 'NOT AVAILABLE';
 
 UPDATE {replaceme}.functions SET created_at = last_modified_at WHERE created_at IS NULL;
-
-ALTER TABLE {replaceme}.functions
```
ayushjain17, line 627: this is needed

workspace_template.sql
ayushjain17: similarly other one is also needed

Datron: Done

ayushjain17 (approved):

crates/superposition/src/workspace/handlers.rs
```diff
@@ -258,3 +259,23 @@ fn validate_workspace_name(workspace_name: &String) -> superposition::Result<()>
         _ => Ok(()),
     }
 }
+
+#[post("/{workspace_name}/db/migrate")]
+async fn migrate_workspace_schema(
```
ayushjain17, line 264: should we not add tests for this ?
also, migrations for superposition schema is not yet handled

Datron: Yeah this was discussed earlier, sharing it here

> should we not add tests for this

The process of testing this is complex. I thought we could drop some columns and see if this function restores that, and can validate it by doing a dropped-column-specific command. The drop column operation however cannot happen from the tests, they have to be done via an API since the tests don't access the DB directly. I'll follow this PR up with a test, though we can discuss how this can be done

> migrations for superposition schema is not yet handled

That is outside the scope of this PR

PR #587, opened 2025-07-11
feat: release pipeline

Datron:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


.github/workflows/release.yaml
```diff
@@ -0,0 +1,407 @@
...
+            - name: install cocogitto
+              uses: baptiste0928/cargo-install@v2.2.0
+              with:
+                  crate: cocogitto
+                  version: "6.2.0"
+
+            - name: Semver release
+              shell: bash
+              run: |
+                  git config user.email "super_bot@juspay.in"
```
knutties, line 43: let us use `superposition@juspay.in`

knutties (approved):

knutties (approved):

PR #591, opened 2025-07-14
fix: restore setup.py and setup JS pipeline

Datron:


.github/workflows/release.yaml
```diff
@@ -332,7 +336,7 @@ jobs:
                           mkdir -p "$target_dir"
 
                           # Copy the file
-                          cp "$lib" "$target_dir/"
+                          cp "$lib" "$target_dir/libsuperposition_core.$extension"
```
knutties, line 339: @Datron  - won't the following work?  Why do we need to extract the extension and then add it back?

```
cp "$lib" "$target_dir/$filename"
```

Datron: @knutties windows build binary does not contain the word `lib` so everyone has to write an extra condition to handle that case. I rather rename the file consistently across platforms.

.github/workflows/release.yaml
```diff
@@ -465,7 +475,67 @@ jobs:
                   name: superposition_provider-${{needs.tag-release.outputs.version}}-py3-none-any.whl
                   path: python-wheels/superposition_provider-${{needs.tag-release.outputs.version}}-py3-none-any.whl
 
-    # generate-js-pacakages:
+    generate-js-pacakages:
```
knutties: typo: `generate-js-packages`

.github/workflows/release.yaml
knutties: oh ok - assuming we cannot generate the filename correctly at the time of generation - which might make this change moot?

.gitignore
```diff
@@ -61,4 +61,8 @@ clients/generated/smithy/ts
 
 # Intellij
 **/.idea
-docker-compose/postgres/data
\ No newline at end of file
+docker-compose/postgres/data
+clients/generated/smithy/typescript/dist-cjs/
```
knutties, line 65: Are there directories still relevant?

@sauraww can possibly confirm.

clients/python/bindings/setup.py
```diff
@@ -0,0 +1,44 @@
...
+
+setup(
+    name="superposition_bindings",
+    version="0.1.0",
+    packages=find_packages(),
+    package_data={"superposition_bindings": ["libsuperposition_core.*"]},
+    include_package_data=True,
+    cmdclass={"build_py": build_py},
+    python_requires=">=3.8",
+)
```
knutties: minor nit: newline at end of file.

.gitignore
Datron: This gets removed in the workspace PR, it needed a smithy and makefile change that is not here

.github/workflows/release.yaml
Datron: This decision is made by the Rust compiler for windows, I'm not sure why they do this

sauraww (approved):

PR #594, opened 2025-07-15
feat(audit-log): add UI for Audit Log API

knutties:
## Problem
The Audit Log feature that lists all changes on the core tables of Superposition is not available for view in the UI.  Teams using Superposition need it for easy review - particularly when they want to port the changes to other instances.

## Solution
This PR adds a UI over the Audit Log API.  It uses the same patterns for tables/filters as the existing pages (particularly the experiment list page).

N/A

## Pre-deployment activity
N/A

## Post-deployment activity

## API changes
N/A

N/A


crates/frontend/src/pages/audit_log/filter.rs
```diff
@@ -0,0 +1,452 @@
...
+use crate::components::{
+    badge::{GrayPill, ListPills},
+    button::Button,
+    drawer::{close_drawer, Drawer, DrawerBtn, DrawerButtonStyle},
+    form::label::Label,
+    input::DateInput,
+};
+
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+pub struct AuditLogFilters {
```
knutties: Move this struct and associated impls to `superposition_types`.

cc: @ayushjain17

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -39,7 +39,7 @@ use superposition_types::{
         default_config::DefaultConfigUpdateRequest,
         experiment_groups::ExpGroupMemberRequest,
         experiments::{
-            ApplicableVariantsQuery, ApplicableVariantsRequest, AuditQueryFilters,
```
knutties, line 43: This file's change are removing the audit log that was present in the experimentation schema earlier.

crates/frontend/src/pages/audit_log.rs
```diff
@@ -0,0 +1,201 @@
...
+            "timestamp".to_string(),
+            false,
+            move |value: &str, _row: &Map<String, Value>| {
+                let timestamp = value.to_string();
+                view! { <span class="text-gray-900">{timestamp}</span> }.into_view()
+            },
+            ColumnSortable::No,
+            Expandable::Disabled,
+            default_column_formatter,
+        ),
```
Datron: Can this be the last column, this column is getting pinned. It should be table_name

crates/frontend/src/pages/audit_log.rs
```diff
@@ -0,0 +1,201 @@
...
+            "user_name".to_string(),
+            false,
+            move |value: &str, _row: &Map<String, Value>| {
+                let user_name = value.to_string();
+                view! { <span class="text-gray-900">{user_name}</span> }.into_view()
+            },
+            ColumnSortable::No,
+            Expandable::Disabled,
+            default_column_formatter,
+        ),
```
Datron: We can remove the user_name, since it just shows the DB username

crates/frontend/src/pages/audit_log.rs
```diff
@@ -0,0 +1,201 @@
...
+fn audit_log_table_columns() -> Vec<Column> {
+    vec![
+        Column::new(
+            "timestamp".to_string(),
+            false,
+            move |value: &str, _row: &Map<String, Value>| {
+                let timestamp = value.to_string();
+                view! { <span class="text-gray-900">{timestamp}</span> }.into_view()
+            },
+            ColumnSortable::No,
```
Datron:
```suggestion
            ColumnSortable::Yes,
```

crates/frontend/src/pages/audit_log.rs
```diff
@@ -0,0 +1,201 @@
...
+            Expandable::Enabled(100),
+            default_column_formatter,
+        ),
+    ]
+}
+
+#[component]
+pub fn AuditLog() -> impl IntoView {
+    let _workspace_settings = use_context::<StoredValue<WorkspaceResponse>>().unwrap();
```
Datron: this can be removed since it's not used

crates/frontend/src/pages/audit_log.rs
```diff
@@ -0,0 +1,201 @@
...
+                let value = audit_log_resource.get().unwrap_or_default();
+                let table_rows = value
+                    .data
+                    .iter()
+                    .map(|ele| {
+                        let mut ele_map = json!(ele).as_object().unwrap().clone();
+                        ele_map
+                            .insert(
+                                "timestamp".to_string(),
+                                Value::String(ele.timestamp.format("%v %T").to_string()),
```
Datron, line 206: Can this be formatted to IST for now since all our customers are in India

crates/context_aware_config/src/api/audit_log/handlers.rs
```diff
@@ -40,16 +41,17 @@ async fn get_audit_logs(
         builder
             .filter(
                 event_log::timestamp
-                    .ge(filters.from_date.unwrap_or(now - Duration::hours(24))),
+                    .ge(filters.from_date.unwrap_or(now - Duration::days(30))),
```
Datron: This is a huge set, could it be 7 days?

crates/frontend/src/pages/audit_log/filter.rs
```diff
@@ -0,0 +1,318 @@
+use std::{fmt::Display, str::FromStr};
```
Datron, line 1: This should be in the folder crates/frontend/src/components/audit_log_filters.rs

Datron (changes requested):

crates/frontend/src/pages/audit_log.rs
ayushjain17: We dont need to do this right now
We need to achieve that throughout the framework

Which can be done later

crates/frontend/src/pages/audit_log/filter.rs
ayushjain17: Its fine
Its following as per other already existing ones

crates/frontend/src/pages/audit_log.rs
ayushjain17: Api does not support sort

crates/frontend/src/pages/audit_log.rs
ayushjain17: +1

crates/frontend/src/pages/audit_log.rs
Datron: Can we add this? Would be useful to sort it

crates/frontend/src/pages/audit_log.rs
Datron: It can be hardcoded for now, the UTC time is annoying to convert in my head

crates/frontend/src/pages/audit_log/filter.rs
Datron: Its a component right? Lets put it in the component folder. If others are in the wrong place, we can move those to it another PR. Seems weird otherwise

crates/frontend/src/pages/audit_log.rs
ayushjain17: lets not do it  now, I have a solution which is more dynamic and robust based on user's Locale

ayushjain17: I am not saying not to do it at all

crates/context_aware_config/src/api/audit_log/handlers.rs
mahatoankitkumar: +1

crates/frontend/src/pages/audit_log.rs
```diff
@@ -0,0 +1,201 @@
...
+                    _ => "badge-info",
+                };
+                view! { <span class=format!("badge {}", badge_class)>{action}</span> }
+                    .into_view()
+            },
+            ColumnSortable::No,
+            Expandable::Disabled,
+            default_column_formatter,
+        ),
+        Column::new(
```
mahatoankitkumar, line 91: Should we add a modal showing the diff between old and new data? Currently, it's difficult to spot the differences between them.

crates/superposition_types/src/api/config.rs
```diff
@@ -27,3 +33,56 @@ pub struct ResolveConfigQuery {
     #[query_param(skip_if_empty)]
     pub prefix: Option<CommaSeparatedStringQParams>,
 }
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, QueryParam, IsEmpty)]
+pub struct AuditQueryFilters {
```
mahatoankitkumar: Shouldn't we add a check to prevent "To Date" from being before "From Date"?

crates/frontend/src/pages/audit_log.rs
ayushjain17: This is just an introduction PR

crates/superposition_types/src/api/config.rs
ayushjain17: Postgres will handle that, no data will match such a query

mahatoankitkumar: Yeah it does handle it, shouldn't we throw an error was my question.

crates/frontend/src/pages/audit_log.rs
mahatoankitkumar: we already have that component right?

clients/haskell/sdk/Io/Superposition/Model/ListAuditLogsInput.hs
Datron, line 1: https://github.com/juspay/superposition/blob/5e061aa8a2a1c589990637a34c44ada6de6f66bf/crates/frontend/src/pages/audit_log/mod.rs#L33-L43

timestamp is still present as the key column. Is this okay?

PR #597, opened 2025-07-16
feat: rust open-feature implementation

pratikmishra356:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/superposition_provider/src/provider.rs
```diff
@@ -0,0 +1,340 @@
...
+            }
+        }
+
+        let mut status = self.status.write().await;
+        *status = ProviderStatus::Ready;
+
+        info!("SuperpositionProvider initialized successfully");
+    }
+
+    // async fn shutdown(&self) -> Result<()> {
```
sauraww: Can we remove these lines ?

crates/superposition_provider/src/provider.rs
```diff
@@ -0,0 +1,340 @@
...
+    provider::{ProviderMetadata, ProviderStatus, ResolutionDetails},
+    EvaluationContext, EvaluationError, EvaluationErrorCode, EvaluationResult,
+    StructValue,
+};
+
+#[derive(Debug)]
+pub struct SuperpositionProvider {
+    metadata: ProviderMetadata,
+    status: RwLock<ProviderStatus>,
+    hooks: Vec<String>,
```
sauraww:
```
#[warn(dead_code)]
hooks: Vec<String>,
```

sauraww (changes requested):

crates/superposition_provider/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+### With Experimentation
+
+Enable A/B testing and experimentation features:
+
+```rust
+use superposition_provider::ExperimentationOptions;
+
+let options = SuperpositionProviderOptions {
+    superposition_options: SuperpositionOptions {
```
knutties: Is this grouping of the configs the same as in other clients?  I see it a bit different in Java - should we keep it the same ?

Eg: https://github.com/juspay/superposition/blob/main/clients/java/open-feature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionProviderOptions.java#L19

makefile
```diff
@@ -6,7 +6,7 @@ TENANT ?= dev
 SHELL := /usr/bin/env bash
 FEATURES ?= ssr
 FMT_FLAGS := --all
```
knutties: Do we need the `FMT_FLAGS` at all?  Should use the `--package` to do the inclusions?  little sad that `fmt` command does not behave the same way as the linter.  You can run a negative grep to get everything else other than smithy - so that any new package that is added is automatically covered by formatting.

crates/superposition_provider/README.md
pratikmishra356: yes made the same

makefile
pratikmishra356: FMT_FLAGS is not being used , 
I am ignoring the smithy/rust sdk in the rustfmt command

knutties: @pratikmishra356 - you can use the following makefile snippet to make the `fmt` targetted.  You can use the same vars for generating the LINT_FLAGS as well.

```makefile
# Packages to exclude from formatting
EXCLUDE_PACKAGES := superposition_core cac_toml experimentation_client_integration_example
FMT_EXCLUDE_PACKAGES_REGEX := $(shell echo "$(EXCLUDE_PACKAGES)" | sed "s/ /|/g")
LINT_FLAGS := --workspace --all-targets $(addprefix --exclude ,$(EXCLUDE_PACKAGES))

# Get all workspace members and filter out excluded ones
WORKSPACE_MEMBERS := $(shell cargo metadata --format-version 1 --no-deps | \
	jq -r '.workspace_members[]' | \
    awk -F '#' '{print $$1;}' | awk -F '/' '{print $$NF;}' | \
    grep -v -E '$(FMT_EXCLUDE_PACKAGES_REGEX)')

# Create the package flags for cargo fmt
FMT_PACKAGE_FLAGS := $(addprefix -p ,$(WORKSPACE_MEMBERS))

# Target to run cargo fmt on filtered packages
fmt-workspace:
	@echo "Running cargo fmt on packages: $(WORKSPACE_MEMBERS)"
	cargo fmt $(FMT_PACKAGE_FLAGS)

# Debug target to see what packages would be formatted
fmt-debug:
	@echo "Excluded packages: $(EXCLUDE_PACKAGES)"
	@echo "Workspace members: $(WORKSPACE_MEMBERS)"
	@echo "Format flags: $(FMT_PACKAGE_FLAGS)"

lint:
	cargo clippy $(LINT_FLAGS)
```

Cargo.toml
```diff
@@ -16,7 +16,8 @@ members = [
     "examples/superposition-demo-app",
     "crates/superposition_macros",
     "crates/superposition_derives",
-    "crates/superposition_core"
+    "crates/superposition_core",
+    "crates/superposition_provider"
 ]
```
knutties, line 22: Why no `superposition-sdk` here ?

crates/superposition_provider/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+thiserror = "1.0"
+tokio = { version = "1.0", features = ["full"] }
+async-trait = "0.1"
+log = "0.4"
+reqwest = { version = "0.11", features = ["json"] }
+chrono = { version = "0.4", features = ["serde"] }
```
sauraww: Don't we have these dependencies already in the workspace ?
Can we do  { workspace = true } instead of specifying the versions ?

sauraww (changes requested):

crates/superposition_provider/Cargo.toml
pratikmishra356: yes will do that

makefile
```diff
@@ -238,6 +247,22 @@ check: LEPTOS_FMT_FLAGS += --check
...
 check: fmt lint
 
+
+# Target to run cargo fmt on filtered packages
+fmt:
+	@echo "Running cargo fmt on packages: $(WORKSPACE_MEMBERS)"
+	cargo fmt $(FMT_PACKAGE_FLAGS)
+
+# Debug target to see what packages would be formatted
+fmt-debug:
```
knutties: We can remove this debug target `fmt-debug` @pratikmishra356  - was only needed for debugging.

knutties (commented):
Looks good to me except for the one target removal in the makefile.

crates/superposition_provider/src/provider.rs
```diff
@@ -0,0 +1,321 @@
...
+        &self,
+        evaluation_context: &EvaluationContext,
+    ) -> Result<serde_json::Map<String, Value>> {
+        // Get cached config from CAC
+        let (mut context, targeting_key) =
+            self.get_context_from_evaluation_context(evaluation_context);
+        let variant_ids = if let Some(exp_config) = &self.exp_config {
+            log::info!(
+                "Evaluating config with context: {:?}, targeting_key: {:?}, variant_ids: {:?}",
+                5.clone(), 5.clone(), 5
```
sauraww: Can we remove this ?

sauraww (changes requested):

PR #600, opened 2025-07-18
fix: smithy build

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/makefile
+++ b/makefile
@@ -208,11 +208,11 @@ smithy-clients: smithy-build
 				clients/java/sdk/src/main/java
 	cp -r $(SMITHY_BUILD_SRC)/python-client-codegen/*\
 				clients/python/sdk
-	cp -r $(SMITHY_BUILD_SRC)/javascript-client-codegen/*\
+	cp -r $(SMITHY_BUILD_SRC)/typescript-client-codegen/*\
 				clients/javascript/sdk 			
 	@for d in $(SMITHY_BUILD_SRC)/*-client-codegen; do \
 		[ -d "$$d" ] || continue; \
-		[[ "$$d" =~ "java" || "$$d" =~ "python" || "$$d" =~ "javascript" ]] && continue; \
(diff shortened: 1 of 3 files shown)


smithy/patches/java.patch
```diff
@@ -66,7 +66,7 @@ index 518d71c9..68775fc5 100644
 +        Map.entry("change_reason", PreludeSchemas.STRING),
 +        Map.entry("function_name", PreludeSchemas.STRING),
 +        Map.entry("dependents", SharedSchemas.DEPENDENTS),
-+        Map.entry("position", PreludeSchemas.INTEGER)); 
++        Map.entry("position", PreludeSchemas.INTEGER));
```
ayushjain17, line 69: What's the change here ?

mahatoankitkumar: This gives a warning at the end
`smithy/patches/java.patch:69: trailing whitespace.
        Map.entry("position", PreludeSchemas.INTEGER)); 
warning: 1 line adds whitespace errors.`

makefile
```diff
@@ -208,11 +208,11 @@ smithy-clients: smithy-build
 				clients/java/sdk/src/main/java
 	cp -r $(SMITHY_BUILD_SRC)/python-client-codegen/*\
 				clients/python/sdk
-	cp -r $(SMITHY_BUILD_SRC)/javascript-client-codegen/*\
```
knutties, line 211: is this Smithy generating as `typescript` and we cannot change ?

Datron: The plugin is typescript-codegen or something so it uses the same name

knutties: we can fix line 38 in `smithy/smithy-build.json` ?

smithy/smithy-build.json
```diff
@@ -36,7 +36,7 @@
         "moduleAuthors": ["protocoltest@example.com"]
     },
     "typescript-client-codegen": {
```
knutties, line 38: is this line causing the output directory to have typescript ?

PR #603, opened 2025-07-18
feat: push packages to AWS codeArtifact

Datron:
## Problem
Working with github package registries have a lot of friction

## Solution
Push to AWS code artifact


clients/java/bindings/build.gradle.kts
```diff
@@ -26,11 +26,11 @@ publishing {
     }
     repositories {
         maven {
-            name = "GitHubPackages"
-            url = uri("https://maven.pkg.github.com/juspay/superposition")
+            name = "CodeArtifact"
```
knutties: @Datron - can we update in the outside `build.grade.kts` (https://github.com/juspay/superposition/blob/main/clients/java/build.gradle.kts)  so that it applies to all sub-modules?  Would that work?

Datron: @knutties since we're publishing individual packages, not the whole java package - this doesn't seem to work or I don't know enough gradle to make it work

knutties (approved):

PR #605, opened 2025-07-18
build: fix npm issue

knutties:
## Problem
The auth for private npm registry needs to be setup before the setup-node action.

## Solution
This moves the aws auth stuff before the setup-node action

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
NA

NA

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -540,7 +540,7 @@ jobs:
             AWS_REGION: ${{ vars.AWS_REGION }}
             CODEARTIFACT_DOMAIN: ${{ vars.CODEARTIFACT_DOMAIN }}
             CODEARTIFACT_REPOSITORY: ${{ vars.CODEARTIFACT_REPOSITORY }}
-            NODE_AUTH_TOKEN: ${{ env.CODEARTIFACT_AUTH_TOKEN }}
+
         steps:
             - name: Checkout repository
               uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -555,6 +555,27 @@ jobs:
...
+                      --format npm --query repositoryEndpoint --output text)
+
+                  echo "CODEARTIFACT_AUTH_TOKEN=$CODEARTIFACT_AUTH_TOKEN" >> $GITHUB_ENV
+                  echo "CODEARTIFACT_REPOSITORY_ENDPOINT=$CODEARTIFACT_REPOSITORY_ENDPOINT" >> $GITHUB_ENV
+                  echo "NODE_AUTH_TOKEN=$CODEARTIFACT_REPOSITORY_ENDPOINT" >> $GITHUB_ENV
+
+                  # Configure npm to use CodeArtifact
+                  npm config set registry $CODEARTIFACT_REPOSITORY_ENDPOINT
+                  npm config set //$CODEARTIFACT_REPOSITORY_ENDPOINT:_authToken=$CODEARTIFACT_AUTH_TOKEN
+
```
sauraww, line 578: Is this syntax correct ?
Shouldn't it be `npm config set //$CODEARTIFACT_REPOSITORY_ENDPOINT:_authToken$CODEARTIFACT_AUTH_TOKEN`

knutties: good call @sauraww  - this is npm set - so the `=` is wrong.  It shows as `=` in `.npmrc` .  Fixing this.

There is stil some problem where the workflow itself does not start.  No errors as well.

PR #608, opened 2025-07-21
fix: support complete bundling of python and JS packages

Datron:


clients/javascript/open-feature-provider/package.json
```diff
@@ -1,38 +1,57 @@
...
     "dependencies": {
-        "@openfeature/server-sdk": "^1.13.0",
-        "superposition-bindings": "file:../bindings",
-        "superposition-sdk": "file:../sdk"
+      "koffi": "^2.11.0",
+      "tslib": "^2.6.2"
     },
     "devDependencies": {
+        "@openfeature/server-sdk": "^1.13.0",
+        "superposition-bindings": "file:../bindings",
```
sauraww, line 30: These are still the local path ? Can we use the package name with the version directly here ?

Datron: These are in dev dependencies

pratikmishra356 (approved):

sauraww (approved):

PR #609, opened 2025-07-21
feat: Dimension api types unification

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -506,13 +512,13 @@ fn fetch_dimensions_map(
 fn update_dimensions_in_db(
     dimensions_map: &mut HashMap<String, (Dimension, bool)>,
     user_email: &str,
-    reason: &str,
+    reason: &ChangeReason,
```
mahatoankitkumar: reason being str and converting and handling in the function was better i guess.

mahatoankitkumar (approved):

PR #612, opened 2025-07-22
fix: JS builds to include sdk and bindings

Datron:


sauraww (approved):

clients/javascript/open-feature-provider/configuration-client.ts
```diff
@@ -1,6 +1,6 @@
 import { SuperpositionOptions, ConfigOptions, ConfigData, PollingStrategy, ExperimentationArgs } from './types';
-import { NativeResolver } from 'superposition-bindings';
-import { SuperpositionClient, GetConfigCommand, GetConfigCommandInput } from 'superposition-sdk';
+import { NativeResolver } from '@juspay/superposition-bindings';
```
knutties, line 2: Do we need the `juspay` namespace qualifier?  Based on what @JamesGeorg said I thought we will not keep it right ?

It is ok for now.

PR #614, opened 2025-07-23
feat: Standalone page for dimensions, default-configs, type-templates, webhooks

ayushjain17:
https://github.com/user-attachments/assets/f3939918-75e7-444c-96c5-7c4a25d11482

### DIFF
--- a/crates/frontend/src/pages/dimension.rs
+++ b/crates/frontend/src/pages/dimension.rs
@@ -7,6 +7,7 @@ use superposition_types::{
 };
 
 use crate::api::{delete_dimension, get_dimension};
+use crate::components::badge::Badge;
 use crate::components::{
     alert::AlertType,
     button::Button,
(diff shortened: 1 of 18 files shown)


crates/frontend/src/pages/dimension.rs
```diff
@@ -0,0 +1,354 @@
...
+            prefix.push_str(if *is_last { "    " } else { "│   " });
+        }
+    }
+
+    let children = data.get(&name).cloned().unwrap_or_default();
+
+    view! {
+        <div class="whitespace-pre">
+            <span class="text-gray-500 font-mono">{prefix}</span>
+            <A href=format!("../{name}") class="underline text-blue-600 hover:text-blue-800">
```
mahatoankitkumar: Can we add a check here for the root node not being a hyperlink here,

mahatoankitkumar (changes requested):

crates/frontend/src/pages/dimension.rs
ayushjain17: why do you want that though ?
keeping root as hyperlink is not an issue though

mahatoankitkumar: But it does nothing, And i am already in the root, so making it a hyperlink doesn't make sense
But just keeping it a hyperlink gives an assumption that something will happen but nothing happens
May lead to bad user experience.

mahatoankitkumar:
Just was thinking if we are showing the dependency graph, should we also show the dependents of a particular dimension?

crates/frontend/src/pages/dimension.rs
ayushjain17: added

mahatoankitkumar (approved):

Datron (approved):

PR #616, opened 2025-07-23
fix: provider changes for identifier

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


clients/javascript/open-feature-provider/configuration-client.ts
```diff
@@ -146,10 +146,10 @@ export class ConfigurationClient {
 
             if (this.experimentationClient && targetingKey) {
                 const experiments = await this.experimentationClient.getExperiments();
-                const toss = this.experimentationOptions?.defaultToss ||
-                    (parseInt(targetingKey) || Math.floor(Math.random() * 100)) % 100;
+                const identifier = this.experimentationOptions?.defaultIdentifier ||
+                    (targetingKey ? targetingKey : 'default');
```
mahatoankitkumar, line 150: Not sure of this though

sauraww: Should we even have a default ?

clients/javascript/open-feature-provider/experimentation-client.ts
```diff
@@ -190,6 +240,34 @@ export class ExperimentationClient {
         return this.cachedExperiments || [];
     }
 
+    async getExperimentGroups(): Promise<ExperimentGroup[]> {
```
sauraww, line 247: Should we expose this in provider?

sauraww (changes requested):

clients/javascript/open-feature-provider/experimentation-client.ts
mahatoankitkumar: i guess we might have to.

sauraww:
@mahatoankitkumar , Can you update the native-resolver inside bindings as well , getApplicableVariants is still expecting a number rather it should be a string ?

clients/javascript/open-feature-provider/experimentation-client.ts
```diff
@@ -132,6 +147,41 @@ export class ExperimentationClient {
...
+
+            if (!response.data) {
+                return null;
+            }
+
+            // Convert response to internal format with proper type checking
+            const experimentGroups: ExperimentGroup[] = [];
+
+            for (const exp_group of response.data) {
+                experimentGroups.push({
```
sauraww, line 173:
```
                if (!exp_group.id) {
                    console.warn('Skipping experiment group without ID');
                    continue;
                }
```
Can we add this check ? We should skip if there is no id

clients/javascript/open-feature-provider/experimentation-client.ts
sauraww: Can you do it ?

clients/javascript/open-feature-provider/experimentation-client.ts
```diff
@@ -9,17 +9,28 @@ export interface Variant {
     overrides: Record<string, string>;
 }
 
+export type GroupType = 'USER_CREATED' | 'SYSTEM_GENERATED';
```
sauraww: Can we make it an enum: 
```
export enum GroupType {
    USER_CREATED = 'USER_CREATED',
    SYSTEM_GENERATED = 'SYSTEM_GENERATED'
}
```

pratikmishra356:
@mahatoankitkumar where is the superpositon_Core changes and python provider changes ?

pratikmishra356 (approved):

PR #619, opened 2025-07-24
feat: Context API types unification

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


clients/javascript/sdk/src/models/models_0.ts
```diff
@@ -359,23 +359,18 @@ export interface BulkOperationInput {
...
-  description?: string | undefined;
-  change_reason?: string | undefined;
-}
-
-/**
- * @public
- */
-export interface ContextPutOut {
-  context_id?: string | undefined;
+export interface ContextFull {
```
mahatoankitkumar: Why ContextFull, Just Context or ContextResponse?

ayushjain17: There is no such type, and I am not adding this type the diff is coming up as wrong, I deleted the unused types 

Also this is the smithy generated file which you are seeing

mahatoankitkumar: Then if we are unifying the types to one type then better rename this.

crates/superposition_types/src/config.rs
```diff
@@ -185,6 +213,38 @@ impl Condition {
...
+
+impl TryFrom<Condition> for Cac<Condition> {
+    type Error = String;
+
+    fn try_from(value: Condition) -> Result<Self, Self::Error> {
+        let map: Map<String, Value> = value.into();
+        Self::try_from(map)
+    }
+}
+
```
sauraww, line 247: This is a critical bug: https://github.com/juspay/superposition/pull/619/files#diff-61d7d423a4c1cc86aa26179f4586efff51e9d8dc8a0d414868d820073e622de6L221. In filter_by_prefix also something similar is done . 
Both locations access override_with_keys[0] without checking if the vector is empty, which will cause a panic. Either we can do it here or in a seperate PR.

crates/superposition_types/src/database/models.rs
```diff
@@ -144,6 +144,19 @@ impl TryFrom<String> for Description {
     }
 }
 
+impl From<ChangeReason> for Description {
+    fn from(value: ChangeReason) -> Self {
+        Self(value.0)
+    }
+}
+
+impl FromIterator<ChangeReason> for Description {
```
sauraww: Why are doing this ? What is the use case ?

crates/superposition_types/src/database/models.rs
```diff
@@ -144,6 +144,19 @@ impl TryFrom<String> for Description {
...
+    fn from_iter<T: IntoIterator<Item = ChangeReason>>(iter: T) -> Self {
+        let reasons: Vec<String> = iter.into_iter().map(|r| r.0).collect();
+        Self(reasons.join(","))
```
sauraww: Also this is buggy , it can easily break the description char limit since we are doing no validation for length here.

sauraww: Multiple 255-character reasons could easily exceed it

sauraww (changes requested):

crates/superposition_types/src/database/models.rs
ayushjain17: These issues were actually already there
Introducing types is highlighting them

crates/superposition_types/src/config.rs
ayushjain17: where is [] being used over here ?
please comment on the line of usage or on the section of lines

mahatoankitkumar (approved):

PR #620, opened 2025-07-24
fix: Change log related UX issues

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/frontend/src/components/change_summary.rs
+++ b/crates/frontend/src/components/change_summary.rs
@@ -3,7 +3,7 @@ use std::collections::HashSet;
 use leptos::*;
 use serde_json::{Map, Value};
 
-use crate::components::table::Table;
+use crate::components::{button::Button, table::Table};
 
 use super::{
     form::label::Label,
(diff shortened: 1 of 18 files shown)


crates/frontend/src/components/change_summary.rs
```diff
@@ -207,27 +207,33 @@ pub fn change_log_popup(
...
                         <p class="flex-0 text-sm text-gray-600">{description.clone()}</p>
                         <div class="flex-1 flex flex-col gap-8 overflow-auto">{children()}</div>
                         <div class="flex-0 flex justify-end gap-4">
+                            {move || {
+                                view! {
+                                    <Button
+                                        loading=disabled.get() || inprogress.get()
+                                        on_click=move |_| on_confirm.call(())
+                                        text=confirm_text.get_value()
+                                        icon_class=""
```
sauraww, line 231: Can we remove this line?
icon_class=""
Or make it optional and not pass it ?

ayushjain17: it has default value, and I dont want any icon to be there, so setting it to "" to not use any icon

sauraww (approved):

PR #625, opened 2025-07-28
build: use make functions instead of inline commands for wider compat…

knutties:
## Problem
Current makefile execution fails because we use features not supported by GNU Make 3.81 which is typically available in our Mac books.

## Solution
Use a primitive that works across both versions - moving them to a Makefile function solves the problem.

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
NA

NA

### DIFF
--- a/makefile
+++ b/makefile
@@ -11,10 +11,15 @@ FMT_EXCLUDE_PACKAGES_REGEX := $(shell echo "$(EXCLUDE_PACKAGES)" | sed "s/ /|/g"
 LINT_FLAGS := --workspace --all-targets $(addprefix --exclude ,$(EXCLUDE_PACKAGES)) --no-deps
 
 # Get all workspace members and filter out excluded ones
-WORKSPACE_MEMBERS := $(shell cargo metadata --format-version 1 --no-deps | \
-	jq -r '.workspace_members[]' | \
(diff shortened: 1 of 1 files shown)


pratikmishra356 (approved):

makefile
```diff
@@ -11,10 +11,15 @@ FMT_EXCLUDE_PACKAGES_REGEX := $(shell echo "$(EXCLUDE_PACKAGES)" | sed "s/ /|/g"
...
-WORKSPACE_MEMBERS := $(shell cargo metadata --format-version 1 --no-deps | \
-	jq -r '.workspace_members[]' | \
-    awk -F '#' '{print $$1;}' | awk -F '/' '{print $$NF;}' | \
-    grep -v -E '$(FMT_EXCLUDE_PACKAGES_REGEX)')
+#
+define get_workspace_members
+cargo metadata --format-version 1 --no-deps | \
+jq -r '.workspace_members[]' | \
+awk -F '#' '{print $$1;}' | awk -F '/' '{print $$NF;}' | \
+grep -v -E '$(FMT_EXCLUDE_PACKAGES_REGEX)'
```
ayushjain17, line 19: can we remove `superposition_core` and `cac_toml` from `EXCLUDE_PACKAGES`

knutties: done

ayushjain17 (approved):

PR #626, opened 2025-07-28
fix: Add created_by and created_at column for functions table

ayushjain17:
## Problem
Function creation data not being stored

## Solution
Introduce column for created_by and created_at

### DIFF
--- a/crates/superposition_types/migrations/2025-07-28-120301_function_create_data/up.sql
+++ b/crates/superposition_types/migrations/2025-07-28-120301_function_create_data/up.sql
@@ -1,0 +1,12 @@
+-- Your SQL goes here
+ALTER TABLE public.functions
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP,
+ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL DEFAULT('NOT AVAILABLE');
+
+ALTER TABLE public.functions ALTER COLUMN created_by DROP DEFAULT;
+
+UPDATE public.functions SET created_at = last_modified_at WHERE created_at IS NULL;
+
+ALTER TABLE public.functions
+ALTER COLUMN created_at SET DEFAULT CURRENT_TIMESTAMP,
+ALTER COLUMN created_at SET NOT NULL;
--- a/workspace_template.sql
+++ b/workspace_template.sql
@@ -615,3 +615,15 @@ END $$;
 
 ALTER TABLE {replaceme}.experiment_groups 
 ADD COLUMN IF NOT EXISTS group_type public.group_type DEFAULT 'USER_CREATED';
+
+ALTER TABLE {replaceme}.functions
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP,
+ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL DEFAULT('NOT AVAILABLE');
+
+ALTER TABLE {replaceme}.functions ALTER COLUMN created_by DROP DEFAULT;
+
+UPDATE {replaceme}.functions SET created_at = last_modified_at WHERE created_at IS NULL;
+
+ALTER TABLE {replaceme}.functions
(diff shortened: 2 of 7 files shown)


workspace_template.sql
```diff
@@ -601,3 +601,9 @@ DO $$ BEGIN
 EXCEPTION
     WHEN duplicate_object THEN null;
 END $$;
+
+ALTER TABLE {replaceme}.functions
+ADD COLUMN created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+ADD COLUMN created_by CHARACTER VARYING NOT NULL DEFAULT('NOT AVAILABLE');
```
Datron:
```suggestion
ADD COLUMN IF NOT EXISTS created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
ADD COLUMN IF NOT EXISTS created_by CHARACTER VARYING NOT NULL DEFAULT('NOT AVAILABLE');
```

workspace_template.sql
```diff
@@ -601,3 +601,9 @@ DO $$ BEGIN
...
+
+ALTER TABLE {replaceme}.functions ALTER COLUMN created_by DROP DEFAULT;
```
Datron: Wrap this in an exception handler, so that we can run it on tenants that already have this column present and it won't terminate the transaction

crates/superposition_types/migrations/2025-07-28-120301_function_create_data/up.sql
```diff
@@ -0,0 +1,8 @@
+-- Your SQL goes here
+ALTER TABLE public.functions
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+ADD COLUMN IF NOT EXISTS created_by CHARACTER VARYING NOT NULL DEFAULT('NOT AVAILABLE');
```
Datron:
```suggestion
ADD COLUMN IF NOT EXISTS created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
ADD COLUMN IF NOT EXISTS created_by TEXT NOT NULL DEFAULT('NOT AVAILABLE');
```

workspace_template.sql
```diff
@@ -615,3 +615,11 @@ END $$;
...
 ALTER TABLE {replaceme}.experiment_groups 
 ADD COLUMN IF NOT EXISTS group_type public.group_type DEFAULT 'USER_CREATED';
+
+ALTER TABLE {replaceme}.functions
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+ADD COLUMN IF NOT EXISTS created_by CHARACTER VARYING NOT NULL DEFAULT('NOT AVAILABLE');
+
+ALTER TABLE {replaceme}.functions ALTER COLUMN created_by DROP DEFAULT;
+
+UPDATE {replaceme}.functions SET created_at = last_modified_at;
```
Datron: So in PR #585 we are re-applying workspace.sql on a workspace to update it to the current state. If we ran this query on a workspace that was already updated it would overwrite the created_by field. Is there a way we can properly handle this?

Datron (changes requested):

crates/superposition_types/migrations/2025-07-28-120301_function_create_data/up.sql
ayushjain17: had not made this change thinking to keep it same throughout for the time being
and if we want we can just change the type for all at once, later some time ?

Datron: Right but we've been doing this since experiment groups, its better to reduce our work later on right?

Datron (approved):

PR #627, opened 2025-07-29
feat: publish rust libraries through crates.io

Datron:


Cargo.toml
```diff
@@ -40,7 +45,7 @@ bigdecimal = { version = "0.3.1", features = ["serde"] }
 cfg-if = "1.0.0"
 chrono = { version = "0.4.26", features = ["serde"] }
 derive_more = "^0.99"
-diesel = { git = "https://github.com/juspay/diesel.git", branch = "dynamic-schema", features = [
+diesel = { git = "https://github.com/juspay/diesel.git", version = "2.2.4", branch = "dynamic-schema", features = [
```
knutties: does version work for git dependencies at all?  

ref: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#the-role-of-the-version-key

.github/workflows/release.yaml
```diff
@@ -588,14 +588,66 @@ jobs:
...
+            - name: Set cargo dependencies to crates.io
+              run: |
+                  sed -i '' 's/path = "\.\.\/superposition_derives"/version = "${{ steps.metadata.outputs.derives_version }}"/' crates/superposition_types/Cargo.toml
+                  sed -i '' 's/path = "\.\.\/superposition_types"/version = "${{ steps.metadata.outputs.types_version }}"/' crates/superposition_core/Cargo.toml
+                  sed -i '' 's/path = "\.\.\/superposition_core"/version = "${{ steps.metadata.outputs.core_version }}"/' crates/superposition_provider/Cargo.toml
+                  
+            - name: Build and package superposition
+              run: |
+                  cargo login ${{ secrets.CRATES_IO_TOKEN }}
+                  cargo publish --package superposition_derives
```
knutties: this is interesting now - so we will have separate versions for some of the packages right ?  This is the first case where we have had to push some base packages to repository.

Should we just use one version all along to keep things simple ?

Datron: I thought of this, saw some problems. `superposition_derives` is rarely every changed, but changing it every time any other package changes would be weird - since it has no changes. I feel this is simpler

Datron: Similar example:
Changing superposition_derives would mean I need to bump the SDK version. This would indicate to users that some change occurred in the SDK, when actually nothing changed.

Cargo.toml
Datron: @knutties thanks for the link - yeah it seems like this would result in unintended behaviour. I guess we may need to publish our diesel fork as well

knutties: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

Looks like it works - but will we really work well with 2.2.4 - we will run into connection problems is it without branch ?

.github/workflows/release.yaml
```diff
@@ -38,32 +38,37 @@ jobs:
...
+              run: |
+                  git push origin main
+                  git push origin --tags
 
     docker-build:
         needs: tag-release
         permissions:
             contents: read
             packages: write
+        if: github.ref == 'refs/heads/main'
```
knutties, line 80: is this check to ensure this runs only on main?  Will we restrict it at workflow level as well?  Would be good to do so.  Not necessary for this PR.

.github/workflows/release.yaml
```diff
@@ -588,17 +600,56 @@ jobs:
                   ls -la clients/javascript/bindings/native-lib
 
             - run: |
-                cd clients/javascript
-                npm ci
-                npm publish --access public --scope @juspay --workspace bindings --workspace sdk --workspace open-feature-provider
+                  cd clients/javascript
+                  npm ci
+                  npm publish --access public --scope @juspay --workspace bindings --workspace sdk --workspace open-feature-provider
```
knutties, line 614: We are ditching the @juspay namespace for npm right for (or is this still code-artifact).

Datron: This is still for code artifact

.github/workflows/release.yaml
Datron: This is just for my testing, once we've got a concrete pipeline I'll limit this only to main at the action level

.github/workflows/release.yaml
```diff
@@ -588,17 +600,56 @@ jobs:
...
+                  toolchain: 1.86.0
+                  targets: wasm32-unknown-unknown
+                  components: rustfmt, clippy, cargo
+
+            - name: Set cargo dependencies to crates.io
+              run: |
+                  sed -i '' 's/path = "\.\.\/superposition_derives"/version = "${{ env.VERSION }}"/' crates/superposition_types/Cargo.toml
+                  sed -i '' 's/path = "\.\.\/superposition_types"/version = "${{ env.VERSION  }}"/' crates/superposition_core/Cargo.toml
+                  sed -i '' 's/path = "\.\.\/superposition_core"/version = "${{ env.VERSION  }}"/' crates/superposition_provider/Cargo.toml
+
```
knutties: These changes need to be committed right.  Should we do this along with the set-version itself ?

knutties: And that commit will have the actual tag:

https://github.com/serde-rs/serde/commit/49d098debdf8b5c38bfb6868f455c6ce542c422c

See this commit by serde - the version edit change is the chore commit that has the actual version tag.

Datron: @knutties We can't commit these - since rust considers types from different sources (but same crate) as different. This breaks our generics in certain middleware and frontend

.github/workflows/release.yaml
```diff
@@ -38,32 +38,45 @@ jobs:
...
               id: git_tag
               shell: bash
               run: |
                   version=`git tag -l --sort=-creatordate | grep "^v" | head -n 1 | sed 's/^v//'`
                   echo "version=$version" >> $GITHUB_OUTPUT
+            
+            - name: Set the tag version to all crates
+              run: |
+                  cargo set-version --workspace ${{ steps.git_tag.outputs.version }}
+                  cargo upgrade --package superposition_types@${{ steps.git_tag.outputs.version }}
```
knutties, line 60: since we set everyone's version in Cargo.toml via 

```
version.workspace = true
```
do we need to upgrade individual versions ?  What does the upgrade command for us in this release file ?

.github/workflows/release.yaml
```diff
@@ -588,17 +608,53 @@ jobs:
...
+                  components: rustfmt, clippy, cargo
+
+            - name: Build and package superposition
+              run: |
+                  cargo login ${{ secrets.CRATES_IO_TOKEN }}
+                  cargo publish --package superposition_derives 
+                  cargo publish --package superposition_sdk
+                  sed -i 's/path = "\.\.\/superposition_derives"/version = "${{ env.VERSION }}"/' crates/superposition_types/Cargo.toml
+                  cargo publish --allow-dirty --package superposition_types
+                  sed -i 's/path = "\.\.\/superposition_types"/version = "${{ env.VERSION  }}"/' crates/superposition_core/Cargo.toml
```
knutties: Do we need these sed commands anymore?  The references to `path` based dependency is removed with `workspace = true` I guess.

.github/workflows/release.yaml
ayushjain17: should we also define a SOP for how development needs to be done after this change
I am a bit confused here, does the packages internally refer to the crate or local directory ?

if its crates then it makes multi crate change for a single change a bit difficult

knutties: @ayushjain17  - the internal packages should refer to the local directory.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

knutties (approved):

.github/workflows/release.yaml
ayushjain17: @knutties but I don't see that being used
instead we changed it follow the workspace versions of the crate

I dont think the dev experience is going to be any good

PR #634, opened 2025-08-04
fix: UX Improvements

ayushjain17:
Primarily changed snapshot page, rest are just refactors

https://github.com/user-attachments/assets/d777d761-8486-41d2-b04d-025f35c70858


crates/superposition_types/src/api/config.rs
```diff
@@ -1,7 +1,33 @@
...
+        Self {
+            id: config_version.id.to_string(),
+            config: config_version.config,
+            config_hash: config_version.config_hash,
+            tags: config_version.tags,
+            created_at: config_version.created_at,
+            description: config_version.description,
+        }
+    }
+}
```
Datron: You don't need this response type just to stringify the ID. The ID is already stringified on serialization by i64_formatter

crates/frontend/src/pages/config_version_list.rs
```diff
@@ -73,7 +75,10 @@ pub fn config_version_list() -> impl IntoView {
                                         .data
                                         .into_iter()
                                         .map(|config_version| {
-                                            let mut map = Map::new();
+                                            let mut map = json!(config_version)
+                                                .as_object()
+                                                .unwrap()
```
Datron: nit: can we do unwrap or throw up a snackbar expection with an empty jsonviewer object on screen?

Datron (approved):

PR #638, opened 2025-08-05
feat(java): Publishing to central-sonatype.

ShreyBana:
Publishing for Maven Central via Sonatype.


clients/java/build.gradle.kts
```diff
@@ -1,22 +1,83 @@
...
+
+plugins {
+    id("org.jreleaser") version "1.19.0"
+    id("base")
+}
+
+fun readVersion(): String {
+    val cargoTomlFile = file("../../Cargo.toml")
+    val content = cargoTomlFile.readText()
+
```
Datron: Read the version from ENV, since in github actions the checkout event happens for the commit that the workflow was triggered for not the commit that gets committed in tag-release job. 

If you read from Cargo.toml, you will end up reading the previously tagged version of superposition.

.github/workflows/java-checks.yaml
```diff
@@ -0,0 +1,46 @@
+name: Java Checks
```
Datron: better name would be helpful

ShreyBana: Any suggestions?

knutties: @ShreyBana - pardon my ignorance here -  what does this workflow do?  or is it only to detect failure and upload things in case of failure?

That might help in getting to a right name.

knutties: do we have a main build workflow that happens on main changes (which is what is supposed to power the build passing badge in the main README.md) - and this should be part of that workflow ?

clients/java/build.gradle.kts
```diff
@@ -1,22 +1,83 @@
+import org.jreleaser.model.Active
+
+plugins {
+    id("org.jreleaser") version "1.19.0"
+    id("base")
+}
+
+fun readVersion(): String {
```
knutties: Saw @Datron's comment - @ShreyBana  - are we leaving reading from Cargo.toml here for fallback - should we just fail if the VERSION env is not defined?

ShreyBana: It'll be a little painful to test some things out locally. One would have to either remember to set the env or hardcode it here. 
We could set it back to the snapshot version tho, that deploy is garunteed to fail and will work locally as well.

.github/workflows/java-checks.yaml
ShreyBana: @knutties - It just compiles the java code atm. Testing requires more setup. 
As for your second question, I am not aware of one, don't know if @Datron knows better.

Datron: @ShreyBana does `Java sdk pre-publish checks` work?

clients/java/sdk/build.gradle.kts
```diff
@@ -1,11 +1,18 @@
+
 plugins {
     `java-library`
-    `maven-publish`
+    `publishing-conventions`
 }
 
+extra["displayName"] = "Superposition SDK"
+description = "Java SDK for Superposition."
```
Datron, line 8: @ShreyBana can we read a markdown file here or should we share a link to our Java SDK documentation website/javadoc? Same for the provider as well

ShreyBana: All javadocs published in maven-central are available at `javadoc.io`. We can just link that in our main readme.

.github/workflows/java-checks.yaml
ShreyBana: @Datron - This has been moved to our CI checks workflow.

knutties (approved):

PR #639, opened 2025-08-05
feat: Move jsonlogic behind compile time flag

ayushjain17:
Crate wise summary of changes for removing `jsonlogic`:
1. cac_client:
a. use custom logic for evaluation
2. context_aware_config
a. context validation logic - removal of strict mode checks
b. changes in places requiring the list of dimensions used in the condition of the override context
c. change in logic to validate the dimensions used in the condition of the override context
d. dimension schema restricted to primitive types only (rejecting `array` and `object` types)
3. experimentation_client
a. use of custom `apply` and `partial_apply` logic for evaluation and filtering
4. experimentation_platform
a. changes in places requiring the list of dimensions used in the context of the experiment
b. change in logic of adding `variantIds` while creating override contexts for experiments
5. frontend
a. `strict_mode` option and details totally made hidden
b. relies on the fact that backend is returning `strict_mode` as `true` for all workspaces, in case, strict mode is forcibly made true, other operator options are also removed from UI
c. `variantIds` in context even though is a simple string, retained `has` syntactic sugar in the UI
d. change in context parsing logic (not to expect `jsonlogic`, but just a simple map)
6. service_utils
a. disabled `jsonlogic` related functions
7. superposition
a. `strict_mode` option is not unavailable to the APIs and is forced to `true` while creating workspaces
8. superposition_core
a. use of custom `apply` and `partial_apply` logic for evaluation and filtering
9. superposition_types
a. addition of custom `apply` and `partial_apply` logics
i. `variantIds` has special evaluation logic, it checks if the given value of `variantIds` in the context condition is part of the `array` provided in context for evaluation, for rest of the dimensions of the context condition it does a direct `!=` check for failure
b. Changes in logic related to:
i. check if a condition is a superset of another or not
ii. validating experiment context - empty object allowed, `variantIds` not allowed
iii. validating CAC context - empty object not allowed

Note:
The context condition has been made into a simple key value pair, where even the value for `variantIds` is expected to be simple string value
key (dimension) is validated if it is a valid dimension or not
value (value for the dimension) is validated if it follows the `jsonschema` provided for the dimension

### DIFF
--- a/crates/cac_client/src/eval.rs
+++ b/crates/cac_client/src/eval.rs
@@ -55,13 +55,22 @@ fn get_overrides(
         }
     };
 
+    #[cfg(feature = "jsonlogic")]
     let query_data = Value::Object(query_data.clone());
 
     for context in contexts {
-        if let Ok(Value::Bool(true)) = jsonlogic::apply(
-            &Value::Object(context.condition.clone().into()),
-            &query_data,
-        ) {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "jsonlogic")] {
+                let valid_context = jsonlogic::apply(
+                    &Value::Object(context.condition.clone().into()),
+                    &query_data,
+                ) == Ok(Value::Bool(true));
+            } else {
+                let valid_context = superposition_types::apply(&context.condition, query_data);
+            }
+        }
+
+        if valid_context {
             let override_key = context.override_with_keys.get_key();
             if let Some(overriden_value) = overrides.get(override_key) {
                 match merge_strategy {
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -19,11 +19,12 @@ use diesel::{dsl::max, BoolExpressionMethods, ExpressionMethods, QueryDsl, RunQu
 use fred::interfaces::KeysInterface;
 use itertools::Itertools;
 use serde_json::{json, Map, Value};
+#[cfg(feature = "jsonlogic")]
+use service_utils::helpers::extract_dimensions;
 #[cfg(feature = "high-performance-mode")]
 use service_utils::service::types::AppState;
-use service_utils::{
-    helpers::extract_dimensions,
-    service::types::{AppHeader, DbConnection, SchemaName, WorkspaceContext},
+use service_utils::service::types::{
+    AppHeader, DbConnection, SchemaName, WorkspaceContext,
 };
 #[cfg(feature = "high-performance-mode")]
 use superposition_macros::response_error;
@@ -276,7 +277,14 @@ fn reduce(
 ) -> superposition::Result<Vec<Map<String, Value>>> {
     let mut dimensions: Vec<Map<String, Value>> = Vec::new();
     for (context, overrides, key_val, override_id) in contexts_overrides_values {
-        let mut ct_dimensions = extract_dimensions(&context.condition)?;
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "jsonlogic")] {
+                let mut ct_dimensions = extract_dimensions(&context.condition)?;
+            } else {
+                let mut ct_dimensions: Map<String, Value> = context.condition.clone().into();
+            }
+        }
+
         ct_dimensions.insert("key_val".to_string(), key_val);
         let request_payload = json!({
             "override": overrides,
@@ -367,7 +375,14 @@ fn get_contextids_from_overrideid(
 ) -> superposition::Result<Vec<(Context, Map<String, Value>, Value, String)>> {
     let mut res: Vec<(Context, Map<String, Value>, Value, String)> = Vec::new();
     for ct in contexts {
-        let ct_dimensions = extract_dimensions(&ct.condition)?;
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "jsonlogic")] {
+                let ct_dimensions = extract_dimensions(&ct.condition)?;
+            } else {
+                let ct_dimensions = &ct.condition;
+            }
+        }
+
         if ct_dimensions.contains_key("variantIds") {
             continue;
         }
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -119,12 +119,13 @@ async fn put_handler(
         AppHeader::XConfigVersion.to_string(),
         version_id.to_string(),
     ));
-    cfg_if::cfg_if! {
-        if #[cfg(feature = "high-performance-mode")] {
-            let DbConnection(mut conn) = db_conn;
-            put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
-        }
+
+    #[cfg(feature = "high-performance-mode")]
+    {
+        let DbConnection(mut conn) = db_conn;
+        put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
     }
+
     Ok(http_resp.json(put_response))
 }
 
@@ -166,12 +167,13 @@ async fn update_override_handler(
         AppHeader::XConfigVersion.to_string(),
         version_id.to_string(),
     ));
-    cfg_if::cfg_if! {
-        if #[cfg(feature = "high-performance-mode")] {
-            let DbConnection(mut conn) = db_conn;
-            put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
-        }
+
+    #[cfg(feature = "high-performance-mode")]
+    {
+        let DbConnection(mut conn) = db_conn;
+        put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
     }
+
     Ok(http_resp.json(override_resp))
 }
 
@@ -228,12 +230,13 @@ async fn move_handler(
         AppHeader::XConfigVersion.to_string(),
         version_id.to_string(),
     ));
-    cfg_if::cfg_if! {
-        if #[cfg(feature = "high-performance-mode")] {
-            let DbConnection(mut conn) = db_conn;
-            put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
-        }
+
+    #[cfg(feature = "high-performance-mode")]
+    {
+        let DbConnection(mut conn) = db_conn;
+        put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
     }
+
     Ok(http_resp.json(move_response))
 }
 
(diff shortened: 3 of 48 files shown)


crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -415,17 +417,33 @@ pub fn validate_ctx(
...
+    #[cfg(feature = "jsonlogic")]
+    let context_map = &extract_dimensions(&condition)?;
+    #[cfg(feature = "jsonlogic")]
     let condition_val = Value::Object(condition.clone().into());
+
+    #[cfg(not(feature = "jsonlogic"))]
+    let context_map = &condition;
+    #[cfg(not(feature = "jsonlogic"))]
+    let condition_val = condition.clone();
+
```
mahatoankitkumar: Kinda difficult to go through the code like this
Wouldn't it be better to extract only the feature-specific parts into separate functions throughout the entire codebase?

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/context/helpers.rs
ayushjain17: that will increase complexity unnecessarily
those functions will return tuples
which I will have to destruct 
then these functions don't carry any meaning also

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -367,7 +371,11 @@ fn get_contextids_from_overrideid(
 ) -> superposition::Result<Vec<(Context, Map<String, Value>, Value, String)>> {
     let mut res: Vec<(Context, Map<String, Value>, Value, String)> = Vec::new();
     for ct in contexts {
+        #[cfg(feature = "jsonlogic")]
         let ct_dimensions = extract_dimensions(&ct.condition)?;
+        #[cfg(not(feature = "jsonlogic"))]
+        let ct_dimensions = &ct.condition;
+
```
Datron: We could use cfg_if! here, makes this more readable

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -5,11 +5,12 @@ use cac_client::utils::json_to_sorted_string;
...
-    helpers::{extract_dimensions, get_variable_name_and_value},
-    service::types::SchemaName,
-};
-use superposition_macros::{bad_argument, unexpected_error, validation_error};
+#[cfg(feature = "jsonlogic")]
+use service_utils::helpers::{extract_dimensions, get_variable_name_and_value};
+use service_utils::service::types::SchemaName;
+#[cfg(feature = "jsonlogic")]
+use superposition_macros::bad_argument;
+use superposition_macros::{unexpected_error, validation_error};
```
Datron, line 13: group all consecutive lines of code that use feature jsonlogic under a cfg block

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -415,17 +417,33 @@ pub fn validate_ctx(
...
+
+    #[cfg(feature = "jsonlogic")]
+    let context_map = &extract_dimensions(&condition)?;
+    #[cfg(feature = "jsonlogic")]
     let condition_val = Value::Object(condition.clone().into());
+
+    #[cfg(not(feature = "jsonlogic"))]
+    let context_map = &condition;
+    #[cfg(not(feature = "jsonlogic"))]
+    let condition_val = condition.clone();
```
Datron: group all consecutive lines of code that use feature jsonlogic under a cfg block and use cfg_if where possible

crates/cac_client/src/eval.rs
```diff
@@ -55,13 +55,20 @@ fn get_overrides(
...
-            &query_data,
-        ) {
+        #[cfg(feature = "jsonlogic")]
+        let valid_context = Ok(Value::Bool(true))
+            == jsonlogic::apply(
+                &Value::Object(context.condition.clone().into()),
+                &query_data,
+            );
+        #[cfg(not(feature = "jsonlogic"))]
+        let valid_context = superposition_types::apply(&context.condition, query_data);
```
Datron: We could use cfg_if! here, makes this more readable

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -129,3 +132,17 @@ pub fn validate_dimensions(
         _ => Ok(()),
     }
 }
+
+#[cfg(not(feature = "jsonlogic"))]
```
Datron, line 136: We should define two feature sets - `jsonlogic-contexts` and `simple-contexts`. Would be easier to document, the default feature can be `simple-contexts`. Then you don't need to check not cases like this.

Datron (changes requested):
I think using two feature flags:

- jsonlogic-contexts
- simple-contexts

Will simplify this a lot along with cfg blocks and use of cfg_if where possible.

makefile
```diff
@@ -205,6 +214,21 @@ test: setup frontend superposition
 	cd tests && bun test
 	-@pkill -f target/debug/superposition &
 
+test_jsonlogic: WASM_PACK_MODE=--profiling
+test_jsonlogic: FE_FEATURES += jsonlogic
+test_jsonlogic: setup frontend superposition_jsonlogic
+	cargo test --features=jsonlogic
```
knutties, line 220: can re-use the variable we set in 218 ?

makefile
```diff
@@ -205,6 +214,21 @@ test: setup frontend superposition
 	cd tests && bun test
 	-@pkill -f target/debug/superposition &
 
+test_jsonlogic: WASM_PACK_MODE=--profiling
+test_jsonlogic: FE_FEATURES += jsonlogic
```
knutties, line 218: Why `FE_FEATURES` ?

ayushjain17:
@Datron 
> I think using two feature flags:
> 
> * jsonlogic-contexts
> * simple-contexts
> 
> Will simplify this a lot along with cfg blocks and use of cfg_if where possible.

using 2 separate cfg is not a compile safe code anymore

crates/context_aware_config/src/api/context/validations.rs
ayushjain17: using 2 separate feature flags is not a compile safe code anymore

makefile
ayushjain17: for compiling the wasm, we need different set of features [line 177](https://github.com/juspay/superposition/pull/639/files/87b291150c330fe4b8106e9fc15f0979e03635d1#diff-beda42571c095172ab63437d050612a571d0d9ddd3ad4f2aecbce907a9b7e3d0R177)

makefile
ayushjain17: that variable has other options in it as well, which are needed for frontend crate (frontend crate specific features)

crates/cac_client/src/eval.rs
ayushjain17: does this even help ?

```
                cfg_if! {
                    if #[cfg(feature = "jsonlogic")] {
                        let valid_context = exp.context.is_empty()
                            || jsonlogic::apply(
                                &Value::Object(exp.context.clone().into()),
                                context,
                            ) == Ok(Value::Bool(true));
                    } else {
                        let valid_context = superposition_types::apply(
                            &exp.context,
                            &context.as_object().cloned().unwrap_or_default(),
                        );
                    }
                }
```

Datron: It helps me, unsure of anyone else. This is way more clearer, I missed the not(feature = ...) in earlier reviews

ayushjain17: Missed in earlier reviews as in ?

Datron: When I was reviewing the changes the first time, I read this quickly and assumed all lines were enabled under `#[cfg(feature = "jsonlogic")]`

ayushjain17: sure
for comparision, I have raised another PR with this style of code using `cfg_if`
you guys take a look and decide https://github.com/juspay/superposition/pull/664
cc: @knutties @ShreyBana 

meanwhile, I am trying to compare the compile time perf difference between the two implementations

crates/cac_client/src/eval.rs
```diff
@@ -55,13 +55,22 @@ fn get_overrides(
         }
     };
 
+    #[cfg(feature = "jsonlogic")]
```
knutties, line 58: wondering should we just use `cfg_if` everywhere - so that there is no scope of a miss in blocks.  Subsequent changes and reviews would be easy to handle as well.

@ShreyBana - what do you think?

ShreyBana: @ayushjain17 claimed that most of this code is never touched & if we are planning to do the move quickly, I think it's fine as it is.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -228,12 +230,13 @@ async fn move_handler(
         AppHeader::XConfigVersion.to_string(),
         version_id.to_string(),
     ));
-    cfg_if::cfg_if! {
-        if #[cfg(feature = "high-performance-mode")] {
-            let DbConnection(mut conn) = db_conn;
-            put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
-        }
+
```
knutties, line 233: Somehow I feel comfortable with the cfg_if style even if it feels a bit redundant.  Maybe that is my limited hands on stuff with Rust.  Will go with what your think @ShreyBana @ayushjain17.

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -3,7 +3,11 @@ use std::collections::HashMap;
 use chrono::Utc;
 use diesel::{query_dsl::methods::SchemaNameDsl, ExpressionMethods, RunQueryDsl};
 use jsonschema::{Draft, JSONSchema};
-use service_utils::{helpers::extract_dimensions, service::types::SchemaName};
+#[cfg(not(feature = "jsonlogic"))]
```
knutties, line 6: `cfg_if` possible here ?

ShreyBana (approved):

crates/context_aware_config/src/api/context/handlers.rs
ShreyBana: I don't mind it here and there, certainly better for most cases.

crates/context_aware_config/src/api/dimension/utils.rs
ayushjain17: its for the imports, formatter will not be able format this section correctly if `cfg_if` is used here

knutties (approved):

PR #647, opened 2025-08-07
fix: experiment group backfilling user details

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


ayushjain17 (approved):

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
```diff
@@ -347,10 +347,10 @@ async fn backfill_experiment_groups(
     state: Data<AppState>,
     db_conn: DbConnection,
     schema_name: SchemaName,
-    user: User,
 ) -> superposition::Result<Json<Vec<ExperimentGroup>>> {
     log::info!("Backfilling experiment groups");
     let DbConnection(mut conn) = db_conn;
+    let user = User::default();
```
ayushjain17: instead of doing this, I think its better to rather use `system@superposition.io`

crates/superposition_provider/src/lib.rs
```diff
@@ -7,7 +7,6 @@ pub use client::*;
 pub use provider::*;
 pub use types::*;
 
-
```
mahatoankitkumar, line 10: make fmt

PR #651, opened 2025-08-09
chore: boot-time feature support for request/response logging in trace mode

knutties:
## Problem
As a developer I would like to be able to dump request/response information as part of logs to facilitate debugging.

## Solution
This PR adds a middleware that can optionally attached by passing in the following compile-time flag to the build/run make targets

```sh
make run FEATURES=request-response-logging
```
GET request/response sample log
```log
[2025-08-11T06:57:42Z INFO  service_utils::middlewares::request_response_logging] REQUEST: GET /types?all=true query=all=true headers=[sec-fetch-site: same-origin, connection: keep-alive, user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36, accept-language: en-GB,en-US;q=0.9,en;q=0.8, x-tenant: dev, x-org-id: localorg, host: localhost:8080, sec-ch-ua-mobile: ?0, sec-fetch-dest: empty, sec-ch-ua-platform: "macOS", accept: */*, sec-fetch-mode: cors, accept-encoding: gzip, deflate, br, zstd, referer: http://localhost:8080/admin/localorg/dev/default-config/g, sec-ch-ua: "Not)A;Brand";v="8", "Chromium";v="138", "Google Chrome";v="138"]
[2025-08-11T06:57:42Z INFO  service_utils::middlewares::request_response_logging] RESPONSE: 200 OK headers=[content-type: application/json] body={"total_pages":1,"total_items":0,"data":[]}
```

PUT request/response sample log
```log
[2025-08-11T06:58:26Z INFO  service_utils::middlewares::request_response_logging] REQUEST: PUT /default-config/g query=(none) headers=[content-type: application/json, accept: */*, x-org-id: localorg, sec-fetch-site: same-origin, referer: http://localhost:8080/admin/localorg/dev/default-config/g, accept-encoding: gzip, deflate, br, zstd, origin: http://localhost:8080, sec-fetch-dest: empty, host: localhost:8080, content-length: 135, sec-fetch-mode: cors, sec-ch-ua-platform: "macOS", x-tenant: dev, user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36, accept-language: en-GB,en-US;q=0.9,en;q=0.8, connection: keep-alive, sec-ch-ua-mobile: ?0, sec-ch-ua: "Not)A;Brand";v="8", "Chromium";v="138", "Google Chrome";v="138"] body={"value":true,"schema":{"type":"boolean"},"function_name":null,"autocomplete_function_name":null,"description":"g","change_reason":"t"}
[2025-08-11T06:58:26Z INFO  service_utils::middlewares::request_response_logging] RESPONSE: 200 OK headers=[content-type: application/json, x-config-version: 7360565243979894784] body={"key":"g","value":true,"created_at":"2025-08-10T16:27:50.378669Z","created_by":"user@superposition.io","schema":{"type":"boolean"},"function_name":null,"last_modified_at":"2025-08-11T06:58:26.815662Z","last_modified_by":"user@superposition.io","description":"g","change_reason":"t","autocomplete_function_name":null}
```

No env changes - but a new compile-time feature flag is added - `request-response-logging`

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
Applicable for all APIs

It is to be ensured that the production builds are not compiled with this flag.

### DIFF
--- a/crates/service_utils/src/middlewares/request_response_logging.rs
+++ b/crates/service_utils/src/middlewares/request_response_logging.rs
@@ -12,8 +12,6 @@ use std::pin::Pin;
 use std::rc::Rc;
 use std::task::{Context, Poll};
 
-pub struct RequestResponseLogger;
-
 #[derive(Default)]
 pub struct RequestResponseLogger;
 
@@ -185,19 +183,13 @@ where
                 );
 
                 // Reconstruct the request with the body
+                #[rustfmt::skip]
                 let new_payload = if body_bytes.is_empty() {
                     Payload::None
                 } else {
                     let bytes = Bytes::from(body_bytes);
                     let stream = once(async move { Ok::<Bytes, PayloadError>(bytes) });
-                    Payload::from(Box::pin(stream)
-                        as Pin<
-                            Box<
-                                dyn futures_util::Stream<
-                                    Item = Result<Bytes, PayloadError>,
-                                >,
-                            >,
-                        >)
+                    Payload::from(Box::pin(stream) as Pin< Box< dyn futures_util::Stream< Item = Result<Bytes, PayloadError> > > >)
                 };
                 let new_req = ServiceRequest::from_parts(http_req, new_payload);
 


crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -0,0 +1,255 @@
...
+                    if let Ok(value_str) = value.to_str() {
+                        response_headers.push(format!("{}: {}", name, value_str));
+                    }
+                }
+
+                // Wrap the response body with our logging wrapper
+                let logged_res = res
+                    .map_body(|_, body| LoggingBody::new(body, status, response_headers));
+
+                Ok(logged_res)
```
ayushjain17: these lines seem repeated in both the blocks

knutties: @ayushjain17  - addressed.

The blame goes to claude :)

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -0,0 +1,243 @@
...
+                    Payload::from(Box::pin(stream)
+                        as Pin<
+                            Box<
+                                dyn futures_util::Stream<
+                                    Item = Result<Bytes, PayloadError>,
+                                >,
+                            >,
+                        >)
+                };
+                let new_req = ServiceRequest::from_parts(http_req, new_payload);
```
ayushjain17: are we re-constructing the request again because the body is a stream and it has been consumed here, so a new stream needs to be set for the request ?

knutties: Yes - the request needs to be available for the actual handler to consume.  Since this is only dev stuff - should be ok.  Strict no no in prod however.

ayushjain17: if that is the case should we add this with a condition like this:
```.wrap(Condition::new(matches!(app_env, AppEnv::PROD | AppEnv::SANDBOX), Compress::default()))```
instead of it being behind a feature flag ?

Comment (knutties) [reply crates/service_utils/src/middlewares/request_response_logging.rs]: @ayushjain17 - I did not want this feature to be available by mistake as well in prod setups - so felt it should be an explicit build flag not a runtime code check or a runtime flag.

File: makefile [hunk 3]
```diff
@@ -193,6 +193,9 @@ build: frontend backend
 run: kill db localstack frontend superposition
 	@./target/debug/superposition
 
+%_run: kill db localstack frontend superposition
+	@RUST_LOG=$* ./target/debug/superposition
```
knutties, line 197: @Datron @ayushjain17  - This allows 

```shell

$ make trace_run

# or

$ make debug_run
```

We need more refactors in the makefile - maybe I will do it someday.

crates/superposition/src/main.rs
```diff
@@ -112,17 +116,24 @@ async fn main() -> Result<()> {
...
         App::new()
+
             .wrap(Condition::new(matches!(app_env, AppEnv::PROD | AppEnv::SANDBOX), Compress::default()))
             .wrap(Logger::default())
-            .app_data(app_state.clone())
-            .app_data(PathConfig::default().error_handler(|err, _| {
-                actix_web::error::ErrorBadRequest(err)
-            }))
+            // Conditionally add request/response logging middleware for development
+            .wrap(Condition::new(log_enabled!(Level::Trace), RequestResponseLogger::default()))
```
knutties: Moved all the wraps and the app_data together for better readability.

crates/service_utils/src/middlewares.rs
```diff
@@ -1,2 +1,4 @@
 pub mod auth_n;
 pub mod tenant;
+
```
ayushjain17: this new should not be there
due to this new, cargofmt is not able to sort this alphabetically

crates/superposition/src/main.rs
```diff
@@ -208,8 +219,6 @@ async fn main() -> Result<()> {
                 "/health",
                 get().to(|| async { HttpResponse::Ok().body("Health is good :D") }),
             )
-            .app_data(Data::new(leptos_options.to_owned()))
```
ayushjain17: why did we reorder this ?
i don't remember exactly but I think there was some exact case for which auth_n was kept last, don't remember it exactly though

crates/superposition/src/main.rs
```diff
@@ -26,6 +26,10 @@ use service_utils::{
     service::types::AppEnv,
 };
 
+use log::{log_enabled, Level};
+
+use service_utils::middlewares::request_response_logging::RequestResponseLogger;
```
ayushjain17: same here as well, the newline should not be there

makefile
ayushjain17: nice, 
by the way any reason to add this command separately though ? we could achieve the same by updating my env as well

crates/superposition/src/main.rs
```diff
@@ -112,17 +116,24 @@ async fn main() -> Result<()> {
         let site_root = &leptos_options.site_root;
         let leptos_envs = ui_envs.clone();
         App::new()
+
```
ayushjain17: this newline also seems wrong, dont what we are doing wrong, but weirdly cargofmt fails without any valid known reason to format our main.rs file specifically

makefile
knutties: Yes - explicit env will work.  Just added as convenience for development.

crates/superposition/src/main.rs
knutties: I changed to keep all wraps together and app_data setup together (so that it is clear).

I did not know the ordering had a significance.  Will restore it back - might need to understand why though.

ayushjain17 (approved):

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -0,0 +1,243 @@
...
+use std::rc::Rc;
+use std::task::{Context, Poll};
+
+pub struct RequestResponseLogger;
+
+impl Default for RequestResponseLogger {
+    fn default() -> Self {
+        RequestResponseLogger
+    }
+}
```
Datron: nit: your could just do:

```suggestion
#[derive(Default)]
pub struct RequestResponseLogger;
```

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -0,0 +1,243 @@
...
+                        this.headers.join(", "),
+                        response_body
+                    );
+                }
+                Poll::Ready(None)
+            }
+            Poll::Pending => Poll::Pending,
+        }
+    }
+}
```
Datron, line 86: This struct and code seems superfluous. We could just add the response body processing into the middleware handler and use the generic <B> as usual.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -0,0 +1,243 @@
...
+                    let bytes = Bytes::from(body_bytes);
+                    let stream = once(async move { Ok::<Bytes, PayloadError>(bytes) });
+                    Payload::from(Box::pin(stream)
+                        as Pin<
+                            Box<
+                                dyn futures_util::Stream<
+                                    Item = Result<Bytes, PayloadError>,
+                                >,
+                            >,
+                        >)
```
Datron: nit: Looks weird. Should be just ignore rustfmt and inline this

crates/service_utils/src/middlewares/request_response_logging.rs
Datron: You could generalize the body processing with a function.

knutties: @Datron - [ServiceResponse](https://docs.rs/actix-web/latest/actix_web/dev/struct.ServiceResponse.html#method.map_body) does not have a reconstruction setup like [ServiceRequest::from_parts](https://docs.rs/actix-web/latest/actix_web/dev/struct.ServiceRequest.html#method.from_parts).  The logging is setup as part of the consumption of the response by the framework and hence this implementation.  I am not sure if we can manually construct the response back again.

Datron (approved):

PR #654, opened 2025-08-11
fix: bucketing logic

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/experimentation_client/src/lib.rs
```diff
@@ -256,9 +256,8 @@ pub fn get_applicable_buckets_from_group(
             let res = (exp_context.is_empty()
                 || jsonlogic::apply(&Value::Object(exp_context), context)
                     == Ok(Value::Bool(true)))
+                && *exp_group.traffic_percentage > 0
```
ayushjain17: The change is fine, but can we move it inside then_some so that this check makes more sense that is to avoid divide by 0 issue

PR #656, opened 2025-08-12
feat: Add QueryString proc_macro_derive

ayushjain17:
## Problem
Similar `Display` trait being written over and over for different struct for converting the struct to query param format. 
This often leads to confusion as to why a field is missed as it might not have been added in the implementation of the `Display` trait

## Solution
Added `QueryParam` proc_macro_derive which derives the QueryParam trait for the struct to get the desired query param. 
It being auto generated makes sure that newly added fields get auto added to the `QueryParam` trait's implementation

### DIFF
--- a/crates/superposition_derives/src/lib.rs
+++ b/crates/superposition_derives/src/lib.rs
@@ -196,3 +196,95 @@ pub fn derive_is_empty(input: TokenStream) -> TokenStream {
 
     TokenStream::from(expanded)
 }
+
+/// Implements `QueryParam` trait for the struct, allowing it to be used as a query string
+///
+#[proc_macro_derive(QueryParam, attributes(query_param))]
+pub fn derive_query_param(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let struct_name = input.ident;
+
+    let fields = if let Data::Struct(data) = input.data {
+        match data.fields {
+            Fields::Named(fields) => fields.named,
+            Fields::Unnamed(_) | Fields::Unit => {
+                return syn::Error::new_spanned(
+                    struct_name,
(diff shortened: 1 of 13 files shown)


crates/superposition_derives/src/lib.rs
```diff
@@ -196,3 +196,95 @@ pub fn derive_is_empty(input: TokenStream) -> TokenStream {
 
     TokenStream::from(expanded)
 }
+
+/// Implements `Display` trait for the struct, allowing it to be used as a query string
```
ayushjain17: `QueryString`

crates/superposition_derives/src/lib.rs
```diff
@@ -196,3 +196,95 @@ pub fn derive_is_empty(input: TokenStream) -> TokenStream {
...
+///
+#[proc_macro_derive(QueryString, attributes(query_string))]
+pub fn derive_display_query(input: TokenStream) -> TokenStream {
```
ayushjain17: `derive_query_string`

crates/superposition_types/src/custom_query.rs
```diff
@@ -10,12 +9,16 @@ use serde::{
...
 use strum::IntoEnumIterator;
-use superposition_derives::IsEmpty;
+use superposition_derives::{IsEmpty, QueryString};
 
 #[cfg(feature = "experimentation")]
 use crate::database::models::experimentation::ExperimentStatusType;
 use crate::IsEmpty;
 
+pub trait QueryString {
+    fn to_query(&self) -> String;
```
Datron: Can we call this 
```suggestion
    fn to_query_params(&self) -> String;
```

mahatoankitkumar (approved):

PR #667, opened 2025-08-14
fix: Make host related ENVs non mandatory

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,5 @@
-DATABASE_URL=postgres://postgres:docker@localhost:5432/config?sslmode=disable
+DATABASE_URL=postgres://postgres:docker@localhost:5432/config?sslmode=disable # Sample format (unused ENV variable)
+PORT=8080
 RUST_LOG=debug
 AWS_ACCESS_KEY_ID=test
 AWS_SECRET_ACCESS_KEY=test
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -1,4 +1,5 @@
-DATABASE_URL=postgres://postgres:docker@localhost:5432/config?sslmode=disable
+DATABASE_URL=postgres://postgres:docker@localhost:5432/config?sslmode=disable # Sample format, not needed
```
knutties: not needed or is it NOT MANDATORY.  The comment isn't fully explanatory IMHO.

.env.example
```diff
@@ -7,13 +8,14 @@ AWS_REGION=ap-south-1
 DB_USER=postgres
 DB_HOST=localhost:5432
 DB_NAME=config
+# DB_PASSWORD=KMS_SECRET
```
knutties: Let us move the ones from KMS to the bottom with a block comment header - so that it is clear.  We can add non-AWS instructions as well there.

```shell
################################################
## Following values are to be set in KMS and not directly in ENV
# 
# DB_PASSWORD
#################################################
```

.env.example
ayushjain17: Its actually not needed

Datron (approved):

PR #675, opened 2025-08-20
feat: introduce cohort dimensions

Datron:
## Problem
Removal of operators and making strict mode the default mode of operations increases verbosity and limits use cases since operators like in, >, <, != are not allowed

## Solution
Introduce cohorts, which aim to solve this while making sure there are no ambiguous overrides during resolution

https://github.com/juspay/superposition/discussions/589

A look at the UI: https://drive.google.com/file/d/1Y0XEBe-c61wQ3VoGzPBuonAd7dM_9Xkb/view?usp=sharing


crates/superposition_types/migrations/2023-10-16-133815_context-aware-config-init/up.sql
```diff
@@ -4,17 +4,7 @@ CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
 -- Name: public; Type: SCHEMA; Schema: -; Owner: -
 --
 CREATE SCHEMA IF NOT EXISTS public;
---
```
mahatoankitkumar, line 7: Shouldn't you add a drop command instead?

crates/frontend/src/api.rs
```diff
@@ -310,24 +285,115 @@ pub async fn delete_default_config(
...
+        let payload = CreateRequest {
+            dimension: DimensionName::try_from(dimension)?,
+            position: Position::from(position),
+            schema,
+            dependencies: Some(dependencies),
+            function_name: validation_fn_name,
+            autocomplete_function_name: autocomplete_fn_name,
+            description: Description::try_from(description)?,
+            change_reason: ChangeReason::try_from(change_reason)?,
+            // TODO: change this
```
mahatoankitkumar: I guess you can remove this comment now.

crates/context_aware_config/src/helpers.rs
```diff
@@ -176,6 +177,40 @@ pub fn validate_jsonschema(
     }
 }
 
+pub fn validate_cohort_schema(cohort_schema: &Value) -> superposition::Result<()> {
```
mahatoankitkumar: Should cohort dimension depend on single or any number of dimensions?

crates/superposition_types/migrations/2023-10-16-133815_context-aware-config-init/up.sql
Datron: This isn't used anywhere, unsure if it was created in production or sandbox

crates/frontend/src/api.rs
Datron: No still need this for frontend changes

crates/context_aware_config/src/helpers.rs
Datron: Single, I have added validation now

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -88,8 +95,9 @@ async fn create(
         change_reason: create_req.change_reason,
         dependency_graph: DependencyGraph::default(),
```
mahatoankitkumar, line 96: Are we sure in keeping this `dependency_graph` now, its useless as we only have one level of dependency
@Datron @knutties

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -201,9 +249,21 @@ async fn update(
     let update_req = req.into_inner();
 
     if let Some(schema_value) = update_req.schema.clone() {
-        #[cfg(not(feature = "jsonlogic"))]
-        allow_primitive_types(&schema_value)?;
-        validate_jsonschema(&state.meta_schema, &schema_value)?;
+        match dimension_data.dimension_type {
```
mahatoankitkumar, line 252: Why checking dimension_data.dimension_type?
Type can be changed in the req also right?

Datron: Yes it can, I'm checking the dimension type since  the schema will change and require different validations for different schema.

mahatoankitkumar: But you are comparing the new schema provided with the old dimension_type.

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -64,6 +64,41 @@ impl Overridden<Cac<Overrides>> for Context {
...
+    uniffi::Enum,
+    Default,
+)]
+#[serde(rename_all = "UPPERCASE")]
+#[strum(serialize_all = "UPPERCASE")]
+#[cfg_attr(
+    feature = "diesel_derives",
+    derive(diesel_derive_enum::DbEnum, QueryId)
+)]
+#[cfg_attr(feature = "diesel_derives", DbValueStyle = "UPPERCASE")]
```
ayushjain17:
```suggestion
#[cfg_attr(feature = "diesel_derives", DbValueStyle = "SCREAMING_SNAKE_CASE")]
```

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -64,6 +64,41 @@ impl Overridden<Cac<Overrides>> for Context {
...
+#[cfg_attr(
+    feature = "diesel_derives",
+    ExistingTypePath = "crate::database::schema::sql_types::DimensionType"
+)]
+pub enum DimensionType {
+    #[default]
+    Regular,
+    #[cfg_attr(feature = "diesel_derives", db_rename = "LOCAL_COHORT")]
```
ayushjain17: after making the above suggested change, this is not more needed

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -64,6 +64,41 @@ impl Overridden<Cac<Overrides>> for Context {
...
+    Deserialize,
+    Serialize,
+    strum_macros::Display,
+    strum_macros::EnumIter,
+    strum_macros::EnumString,
+    uniffi::Enum,
+    Default,
+)]
+#[serde(rename_all = "UPPERCASE")]
+#[strum(serialize_all = "UPPERCASE")]
```
ayushjain17:
```suggestion
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]
```

crates/frontend/src/api.rs
```diff
@@ -360,24 +336,115 @@ pub async fn delete_default_config(
...
+            .map_err(|e| ServerFnError::new(e.to_string()))?
+            .json()
+            .await
+            .map_err(|e| ServerFnError::new(e.to_string()))?;
+
+        Ok(response)
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub async fn create_dimension(
```
ayushjain17:
```suggestion
    pub async fn create(
```

crates/frontend/src/api.rs
```diff
@@ -360,24 +336,115 @@ pub async fn delete_default_config(
...
+            reqwest::Method::POST,
+            Some(payload),
+            construct_request_headers(&[("x-tenant", &tenant), ("x-org-id", &org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
+
+    pub async fn update_dimension(
```
ayushjain17:
```suggestion
    pub async fn update(
```

crates/frontend/src/api.rs
```diff
@@ -360,24 +336,115 @@ pub async fn delete_default_config(
...
+            reqwest::Method::PUT,
+            Some(payload),
+            construct_request_headers(&[("x-tenant", &tenant), ("x-org-id", &org_id)])?,
+        )
+        .await?;
+
+        parse_json_response(response).await
+    }
+
+    pub async fn delete_dimension(
```
ayushjain17:
```suggestion
    pub async fn delete(
```

crates/frontend/src/pages/home.rs
```diff
@@ -14,7 +14,7 @@ use crate::logic::Conditions;
 use crate::providers::condition_collapse_provider::ConditionCollapseProvider;
 use crate::types::{OrganisationId, Tenant};
 use crate::{
-    api::{fetch_config, fetch_dimensions, resolve_config},
+    api::{dimensions::fetch_dimensions, fetch_config, resolve_config},
```
ayushjain17: ideally we should get the function renamed to simply `fetch` inside `dimensions` module and import here like
```suggestion
    api::{dimensions, fetch_config, resolve_config},
```

and use it like
```
dimensions::fetch(...params)
```

crates/frontend/src/components/dimension_form/utils.rs
```diff
@@ -1,85 +1,26 @@
...
 pub fn try_update_payload(
     position: u32,
     schema: Value,
-    dependencies: Vec<String>,
     validation_fn_name: Option<String>,
     autocomplete_fn_name: Option<String>,
     description: String,
     change_reason: String,
+    _dependencies: Vec<String>,
+    _cohort_based_on: Option<String>,
```
ayushjain17: why not removed ?
neither are used

crates/frontend/src/api.rs
```diff
@@ -360,24 +336,115 @@ pub async fn delete_default_config(
...
+    use super::*;
 
-    request(
-        url,
-        reqwest::Method::DELETE,
-        None::<serde_json::Value>,
-        construct_request_headers(&[("x-tenant", &tenant), ("x-org-id", &org_id)])?,
-    )
-    .await?;
+    pub async fn fetch_dimensions(
```
ayushjain17:
```suggestion
    pub async fn fetch(
```

ayushjain17 (changes requested):
server code review pending

crates/context_aware_config/src/helpers.rs
```diff
@@ -176,6 +181,116 @@ pub fn validate_jsonschema(
...
+    if dimensions_used.len() > 1 {
+        log::error!(
+            "Multiple dimensions used in cohort schema: {:?}",
+            dimensions_used
+        );
+        return Err(validation_error!(
+            "Multiple dimensions used in cohort schema and that is not allowed: {:?}",
+            dimensions_used
+        ));
+    }
```
mahatoankitkumar:
```
match all_dimensions.len() {
        0 => {
            log::error!("No dimensions used in cohort schema");
            Err(validation_error!("No dimensions used in cohort schema, one dimension is required"))
        },
        1 => Ok(all_dimensions.into_iter().next().unwrap()),
        _ => {
            log::error!("Multiple dimensions used in cohort schema: {:?}", all_dimensions);
            Err(validation_error!("Multiple dimensions used in cohort schema and that is not allowed: {:?}", all_dimensions))
        }
    }
```
Much cleaner this way

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: Its useful to the normal dimension which multiple cohorts can derive from. We can make a decision to drop it at a later point

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: The old dimension_type was never used in the dimensions table, never defined in code. It was just a dead SQL that was written in our migration. The dimension_type that I've introduced is a new column from the perspective of our code

crates/frontend/src/components/dimension_form/utils.rs
Datron: removed in the frontend PR

crates/context_aware_config/src/helpers.rs
Datron: Agreed, thanks for the suggestion

crates/superposition_types/src/api/dimension.rs
```diff
@@ -78,7 +85,6 @@ pub struct UpdateRequest {
     pub function_name: Option<Option<String>>,
     #[serde(default, deserialize_with = "deserialize_function_name")]
     pub autocomplete_function_name: Option<Option<String>>,
-    pub dependencies: Option<Vec<String>>,
```
ayushjain17, line 81: quick question, can the `cohort_based_on ` never be updated then ?

crates/superposition_types/migrations/2023-10-16-133815_context-aware-config-init/up.sql
```diff
@@ -4,17 +4,7 @@ CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
...
--- Name: dimension_type; Type: TYPE; Schema: public; Owner: -
---
-CREATE TYPE public.dimension_type AS ENUM (
```
ayushjain17, line 10: this type would have already existed in all the current deployments, be it local or some prod deployment
needs to be taken care of

crates/frontend/src/components/dimension_form.rs
```diff
@@ -559,17 +560,7 @@ pub fn change_log_summary(
                                                 "Position".to_string(),
                                                 Value::Number((*position).into()),
                                             )),
-                                            Some((
```
ayushjain17, line 562: why is this removed ?
this is for showing the change log, please revert this

crates/frontend/src/components/dimension_form/utils.rs
ayushjain17: can we not do it like this, and rather remove it here itself, its difficult to keep track of such changes, more so because we are even suppressing the warnings also

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -104,20 +112,60 @@ async fn create(
...
+                &user.get_email(),
+                &schema_name,
+                transaction_conn,
+            )?;
+
+            diesel::update(dimensions::dsl::dimensions)
+                .filter(dimensions::dsl::dimension.eq(&cohorted_dimension.dimension))
+                .set((
+                    dimensions::dsl::change_reason.eq(format!(
+                        "System Added new dependent cohort dimension: {}",
```
ayushjain17:
```suggestion
                        "System added new dependent cohort dimension: {}",
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -104,20 +112,60 @@ async fn create(
...
+                .filter(dimensions::dimension.eq(&cohort_based_on))
+                .schema_name(&schema_name)
+                .get_result::<Dimension>(transaction_conn)?;
+
+            if cohorted_dimension.dimension_type != DimensionType::Regular {
+                return Err(bad_argument!(
+                    "Cohort dimension cannot be based on another cohort dimension {}",
+                    cohorted_dimension.dimension
+                ));
+            }
```
ayushjain17: shouldn't the condition be:
```suggestion
            if cohorted_dimension.dimension_type == DimensionType::LocalCohort {
                return Err(bad_argument!(
                    "Cohort dimension cannot be based on another local cohort dimension {}",
                    cohorted_dimension.dimension
                ));
            }
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -104,20 +112,60 @@ async fn create(
...
         let insert_resp = diesel::insert_into(dimensions::table)
             .values(&dimension_data)
             .returning(Dimension::as_returning())
             .schema_name(&schema_name)
             .get_result(transaction_conn);
 
+        if dimension_data.dimension_type != DimensionType::Regular {
+            let Some(cohort_based_on) = create_req.cohort_based_on else {
+                return Err(bad_argument!(
+                    "Cohort based on argument is required for cohort dimensions"
```
ayushjain17:
```suggestion
                    "`cohort_based_on` argument is required for cohort dimensions"
```

crates/frontend/src/components/dimension_form.rs
Datron: The user cannot add dependencies, this not a change log

crates/superposition_types/src/api/dimension.rs
```diff
@@ -62,10 +67,12 @@ pub struct CreateRequest {
     pub position: Position,
     pub schema: Value,
     pub function_name: Option<String>,
-    pub dependencies: Option<Vec<String>>,
     pub description: Description,
     pub change_reason: ChangeReason,
     pub autocomplete_function_name: Option<String>,
+    #[serde(default)]
+    pub dimension_type: DimensionType,
```
ShreyBana, line 74: Why not something like:

```rust
type ElementryDimension = String;
pub enum DimensionVariant {
    Elementry,
    Cohort(ElementryDimension),
    RemoteCohort(ElementryDimension),
}
```
That way you can avoid invalid combinations. Such as  when `DimensionType` is `Regular` but `cohort_based_on` has been set, or vice-versa. Though, I believe you'll have to write up custom to-and-from sql implementations. Maybe something like: `cohort:<elementry-dimension>`, `elementry` & `remote-cohort:<elementry-dimension>`. Or store it as a json, though that might be even worse as you sacrifice some searchability.
What do you think?
CC: @mahatoankitkumar @ayushjain17

crates/context_aware_config/src/api/dimension/handlers.rs
mahatoankitkumar: You didn't get my question
But old dimension_type i mean, the type which was there before calling the update api
so the user might have changed the type in the updateReq as well right?
So we need to validate the schema according to that i guess

Datron: You can't change dimension types in update API, because the rules around local cohorts,remote cohorts and regular dimensions makes this conversion super complicated. Its better to just delete the dimension and create a new one if you want to change the dimension type

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -354,23 +404,66 @@ async fn delete_dimension(
...
-            validate_dimension_deletability(
-                &name,
-                &dimension_data,
-                &user.get_email(),
-                transaction_conn,
-                &schema_name,
-            )?;
+            if dimension_data.dimension_type != DimensionType::Regular {
+                let Some(cohort_based_on) = dimension_data.dependents.first() else {
+                    return Err(bad_argument!(
```
ShreyBana: Shouldn't this be an internal server error?

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -354,23 +404,66 @@ async fn delete_dimension(
...
+                    .filter(dimensions::dimension.eq(&cohort_based_on))
+                    .schema_name(&schema_name)
+                    .get_result::<Dimension>(transaction_conn)?;
+
+                let deps = cohorted_dimension.dependencies
+                    .iter()
+                    .filter(|dep| **dep != dimension_data.dimension)
+                    .cloned()
+                    .collect::<Vec<_>>();
+                // create dependency between the cohort and the dimension it is based on
```
ShreyBana: Isn't this removing the dependency?

crates/superposition_types/src/api/dimension.rs
ayushjain17: not having an option to update this is like, if I make a mistake in choosing the based on dimension I will have to delete and re-create 😞

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17: @Datron fait point, but I was thinking we should consider providing this option

if I already have a dimension in the old setup migration becomes impossible
plus, code here seems entirely seems to be on the assumption there is no data already present for dependency

also, in future, I am not sure how valid this use case is, if I have a Regular dimension which I later realise that it is actually a RemoteCohort, it becomes too difficult to migrate

maybe deleting and recreating is a fair ask, but maybe if we could do something here make this kind of process smooth

cc: @knutties

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17: marking this as resolved, as we have requirement for this, as the graph can in fact be multi-leveled

crates/superposition_types/src/api/dimension.rs
ShreyBana: Seems simple that way. Create a new one & delete this one.

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -354,23 +404,66 @@ async fn delete_dimension(
...
+                // Direction of this is Dimension depends on cohort
+                // Dimension -> Cohort
+
+                validate_and_update_dimension_hierarchy(
```
ayushjain17, line 444: this function already updates the dimensions in the db, following update query is not needed, and the message by this this function is not correct based on the current operation

can we please revert back the `validate_dimension_deletability` function and rather add the required modifications in that function itself

it will help reduce the change log, and avoid confusions

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -354,23 +404,66 @@ async fn delete_dimension(
...
 
-            validate_dimension_deletability(
-                &name,
-                &dimension_data,
-                &user.get_email(),
-                transaction_conn,
-                &schema_name,
-            )?;
+            if dimension_data.dimension_type != DimensionType::Regular {
+                let Some(cohort_based_on) = dimension_data.dependents.first() else {
```
ayushjain17, line 425: am not very confident on this usage of `first` here, would cause issues in migration to this setup from the previous setup

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17: I would prefer to bring back this `validate_and_initialize_dimension_hierarchy` as well with necessary modifications

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -104,20 +112,60 @@ async fn create(
...
+                ));
+            }
+
+            let mut deps = cohorted_dimension.dependencies.clone();
+            deps.push(dimension_data.dimension.clone());
+            // create dependency between the cohort and the dimension it is based on
+            // Direction of this is Dimension depends on cohort
+            // Dimension -> Cohort
+
+            validate_and_update_dimension_hierarchy(
```
ayushjain17, line 146: this function already updates the dimensions in the db, following update query is not needed

crates/superposition_types/src/api/dimension.rs
ayushjain17: I also wanted something like this, only for API type though:
```rust
type ElementryDimension = String;
pub enum DimensionVariant {
    Regular,
    LocalCohort(ElementryDimension),
    RemoteCohort(ElementryDimension),
}
```

in the API, take 2 fields, ` pub dimension_type: DimensionType,` and `pub cohort_based_on: Option<String>,`
but while deserialization convert them to a single key of this suggested type

this would remove a lot checks from the code as well

crates/context_aware_config/src/helpers.rs
```diff
@@ -176,6 +181,119 @@ pub fn validate_jsonschema(
...
+        0 => {
+            // no dimensions? not allowed
+            log::error!("No dimensions used in cohort schema");
+            Err(validation_error!(
+                "No dimensions used in cohort schema, one dimension is required"
+            ))
+        }
+        1 => {
+            // check if the single dimension used exists in the dimensions table
+            for dim in dimensions_used.iter() {
```
ayushjain17, line 259: can we avoid doing this and rather do something like:

```rust
match dimensions_used.into_iter().collect::<Vec<_>>().as_slice() {
[] => // handle error,
[dim] =>  // logic for the single dimension using the `dim` variable,
_ => // handle error,
}
```

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: It sadly does not, this was tested. The issue comes up when you try to delete the cohorted dimension and it is still shown in the dependency graph of the cohorted dimension since it was not inserted in dependencies.

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: Would need to be handled in the migration script then, this behaviour is what cohorts expects

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: Let me update the comment

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: We can do it on a need basis?

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: It sadly does not, this was tested. The issue comes up when you try to delete the cohorted dimension and it is still shown in the dependency graph of the cohorted dimension since it was not inserted in dependencies.

crates/context_aware_config/src/helpers.rs
Datron: They seem similar enough that this doesn't warrant the change. The code is still readable and understandable

crates/superposition_types/migrations/2023-10-16-133815_context-aware-config-init/up.sql
Datron: Yeah we can take care of it

crates/superposition_types/src/api/dimension.rs
Datron: This seems unnecessary to me - the cohort_based_on value is available in the dimensions table, it mostly used the first time at creation to validate the cohort definition and create dependency graphs. Everywhere else I'm just looking for the type. I don't see the benefit of this yet, lets keep it simple for now

crates/context_aware_config/src/helpers.rs
ayushjain17: I meant lets avoid going through the loop when we already know that the number or elements is actually 1 only

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17: we need to something for this
the current set of code would anyways not work correctly if there exists any data wrt to dependencies prior to this migration, probably it would be better to actually drop all data from that column or just drop the column and move it to a new type of `Option<String>`

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17: its very much there no

https://github.com/juspay/superposition/pull/675/files/104655659818be79fa226db5fa6abdc67ac8c7e2#diff-233cc1e0caa13c0a3d1760902c7b412fef6d58908ae29ccde37143793484fcd6R174

Datron:
closing for #709

PR #682, opened 2025-08-28
Cohort dimensions smithyandtests

Datron:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


LICENSE
```diff
@@ -179,15 +178,15 @@
...
+      boilerplate notice, with the fields enclosed by brackets "{}"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2018-2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ayushjain17, line 189: please take a pull and regenerate the smithy changes, also make sure to remove the haskell files from the old path (which are added in this PR)

clients/generated/smithy/haskell/Io/Superposition/Command/CreateDimension.hs
ayushjain17, line 1: these are to be deleted

clients/python/sdk/README.md
ayushjain17, line 1: this also needs to be reverted after the latest pull

crates/context_aware_config/Cargo.toml
ayushjain17, line 1: wasn't this PR supposed to be pointed to the other branch instead of main ?

PR #688, opened 2025-09-02
feat: Add map support for js-browser client

ayushjain17:
1. index.js / index.ts  - map version of the client
2. indexJsonlogic.js / indexJsonlogic.ts - json-logic version of the client

## Note
Support for `ExperimentGroups` is not there in this client yet


clients/javascript-browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "cac-client",
```
knutties: @ayushjain17  - let us change to superposition_browser_client ?  So that we can push this to npm as well separately if needed.

In hindsight - we need an openfeature compatible browser version as well.  That should replace this one ideally.  That would probably be named superposition_provider_browser ??

ayushjain17: Yeah that might be better, for npm push, I think we need slight more change as to be able to support both es6 and commonJs correctly as well

clients/javascript-browser/src/index.ts
```diff
@@ -1,42 +1,40 @@
...
         this.defaultConfig = completeConfig.default_configs;
     }
 
     public evaluateConfig(data: IObject): IObject {
-
         const requiredOverrides: Array<IObject> = [];
         for (let i = 0; i < this.contexts.length; i++) {
             try {
-                if (jsonLogic.apply(this.contexts[i].condition, data)) {
+                if (apply(this.contexts[i].condition, data)) {
```
knutties: Let us use a different function name (evaluateContext or evaluationCondition) - I got confused with the [Javascript apply](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) and spent 10 mins reading if we can actually call it globally like this :) before I looked up the import.

ayushjain17: Yeah can change it
I had just kept the names same across
maybe i can rather do `logic.apply` and have the file imported as `logic`

clients/javascript-browser/test-logic.js
```diff
@@ -0,0 +1,89 @@
+/**
```
knutties: Should we use something like `jest` or `chai` here and run these in the build ?

clients/javascript-browser/package.json
```diff
@@ -1,6 +1,6 @@
 {
-    "name": "cac-client",
-    "version": "0.0.1",
+    "name": "superposition_browser_client",
```
knutties, line 2:
```suggestion
    "name": "superposition_browser_provider",
```

ayushjain17: Should we rename it to provider already ?
Its not the provider yet, and might create confusion later on, no ?

clients/javascript-browser/src/indexJsonlogic.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export class ExperimentReader {
+    experiments: Experiments;
+
+    constructor(experiments: Experiments) {
+        this.experiments = experiments;
+    }
+
+    public getApplicableVariant(data: IObject, toss: number): Array<String> {
```
Datron:
```suggestion
    public getApplicableVariants(data: IObject, toss: number): Array<String> {
```
nitpick

PR #690, opened 2025-09-03
docs: added api doc generation inside docusaurus via smithy and openapi

sauraww:
Added api docs inside docusaurus using openapi and smithy 

image
image


smithy/smithy-build.json
```diff
@@ -55,6 +55,21 @@
...
+      "protocol": "aws.protocols#restJson1",
+      "version": "3.0.2",
+      "tags": true,
+      "supportedTags": ["deprecated"],
+      "defaultBlobFormat": "byte",
+      "keepUnusedComponents": false,
+      "substitutions": {
+        "io.superposition#Superposition": {
+          "title": "Superposition API",
+          "description": "API documentation for Superposition - Context Aware Configuration and Experimentation Platform"
```
ayushjain17: ??
```suggestion
          "description": "API documentation for Superposition"
```

makefile
```diff
@@ -304,6 +304,16 @@ grafana-local:
...
+# API Documentation targets
+api-docs-generate: smithy-build
+	cd docs && npx docusaurus gen-api-docs superposition
+
+docs-build: api-docs-generate
+	cd docs && npm run build
+
+docs-dev: api-docs-generate
+	cd docs && npm start
+
```
ayushjain17: lets not add separate command for this
please add this in `smithy-clients` command itself, adding new commands adds more chances for folks to miss it

ideally `dev` and `build` options should not be needed also

docs/sidebars.ts
```diff
@@ -14,7 +14,19 @@ import type {SidebarsConfig} from '@docusaurus/plugin-content-docs';
...
-  superpositionSidebar: [{type: 'autogenerated', dirName: '.'}],
+  superpositionSidebar: [
+    { type: 'autogenerated', dirName: '.' },
+    {
+      type: 'category',
+      label: 'API Reference',
+      link: {
+        type: 'generated-index',
+        title: 'Superposition API Reference',
+        description: 'Complete API documentation generated from Smithy models.',
```
ayushjain17: do we need to specify smithy models here ?
this description will come up in docusaurus as well right ?

ayushjain17 (changes requested):

makefile
```diff
@@ -304,6 +304,16 @@ grafana-local:
...
 	cd docs && npm ci && npm start
 
+# API Documentation targets
+api-docs-generate: smithy-build
+	cd docs && npx docusaurus gen-api-docs superposition
+
+docs-build: api-docs-generate
+	cd docs && npm run build
+
+docs-dev: api-docs-generate
```
knutties: @sauraww  - this is already done in `local-docs-view` target - do we need one more - we can make that depend on api-docs-generate itself.

makefile
```diff
@@ -301,9 +302,13 @@ amend-no-edit: amend
 grafana-local:
 	cd grafana && $(COMPOSE) up
 
-local-docs-view:
+local-docs-view: api-docs-generate
```
ayushjain17: what is this command for ?
do we really need to update this ?

ideally shouldn't it that the person who needs to modify first updates the smithy and then view the docs ?

or am I missing something over here ?

knutties: @ayushjain17  - are you asking about the `local-docs-view` target?  It is needed to generate and load the docs locally.  For that docs to be complete - we need the api-docs to be available as well - so that all of the docs can be previewed locally.  Hence the `local-docs-view` target depends on the `api-docs-generate`.

makefile
knutties: @ayushjain17 @sauraww - have a more fundamental question - why should the api-docs generate depend on the smithy-clients generation?  Aren't they independent?  The docs are generated from the source smithy files (via open-api) just like the clients.

Feels like this should just be in the generation of docs.  I will raise a PR with my suggestions.

We should have @Datron review this as well tomorrow before we merge.

ayushjain17: I am more of the opinion that any change in smithy should ensure that the corresponding change is made available across all generated output

Having separate command here, calls for a mandatory situation where when the doc is changed is smithy file, only the docusaurus change would be pushed, whereas all the clients should also be updated with the doc

From my perspective, having separate commands leads to uncalled for gaps.

knutties: `docs-build` and `docs-dev` commands are gone now.  `api-docs-generate` generates the API docs and is used in the `local-docs-view` and in the Github workflow.

@ayushjain17 @sauraww @Datron - please take a look.

docs/docusaurus.config.ts
```diff
@@ -33,13 +34,35 @@ const config: Config = {
...
+        ],
+    ],
+
     presets: [
         [
             "classic",
             {
                 docs: {
-                    routeBasePath: '/docs',
+                    routeBasePath: '',
```
knutties, line 63: @Datron - this was causing the 404 on the initial loading of the docs on the laptop.

Datron: @knutties we'll need this for the prod deployment, since the redirection is for /docs

Datron (approved):

docs/docusaurus.config.ts
Datron: nevermind, saw that you updated the root url

knutties: @Datron - I think the base url and the routeBasePath are not in alignment in the current setup

Why do think it will not work in prod (where the /docs will be given by the base path - and all routes will be relative to it)?  We can deploy in a test place and see.

docs/package.json
```diff
@@ -6,19 +6,23 @@
     "docusaurus": "docusaurus",
     "start": "docusaurus start",
     "build": "docusaurus build",
+    "build-all": "npm run build && npm run gen-api-docs",
```
Datron:
```suggestion
    "build-all": "npm run gen-api-docs && npm run build",
```

This should be inverted

Datron (approved):

PR #691, opened 2025-09-03
feat: Add option to fetch global experiments only

ayushjain17:
Contains changes for backend, frontend and smithy


crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1016,29 +1009,71 @@ async fn list_experiments(
...
+            let data = experiments
+                .get(start..end)
+                .map(|slice| slice.to_vec())
+                .unwrap_or_default();
+
+            PaginatedResponse {
+                total_pages: (total_items as f64 / limit as f64).ceil() as i64,
+                total_items: total_items as i64,
+                data,
+            }
```
mahatoankitkumar, line 1055: Can we please get the if and else block code into a proper function?
This if-else chain is hard to follow.

mahatoankitkumar (changes requested):

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: it is using a lot of variables, will have to pass all of them

also, I would want to know in what way you are suggesting to write it which would improve it

mahatoankitkumar: I think passing those parameters to a function is an acceptable trade-off that improves code readability.
```
if (condition1) {
    func1(...)
} else if (conditon2) {
    func2(...)
} else {
    func3(...)
}
```

crates/superposition_types/src/api/experiments.rs
```diff
@@ -241,6 +241,7 @@ pub struct ExperimentListFilters {
     pub created_by: Option<CommaSeparatedStringQParams>,
     pub sort_on: Option<ExperimentSortOn>,
     pub sort_by: Option<SortBy>,
+    pub global_experiments_only: Option<bool>,
```
Datron, line 244: Instead of Option<bool> can we do:
```suggestion
    #[serde(default = false)]
    pub global_experiments_only: bool,
```

crates/frontend/src/pages/experiment_list/filter.rs
```diff
@@ -280,29 +291,54 @@ pub(super) fn experiment_table_filter_widget(
...
+                            disabled=filters_buffer_rws
+                                .with(|f| f.global_experiments_only.unwrap_or_default())
+                        />
                     }
-                    heading_sub_text="Search By Context"
-                    resolve_mode=true
-                    fn_environment
-                />
-                <div class="w-full flex flex-row justify-start items-end gap-10">
+                }} <div class="w-fit flex items-center gap-2">
```
Datron, line 321:
```suggestion
                }} 
                <div class="w-fit flex items-center gap-2">
```
Has this been formatted this way by leptosfmt?

ayushjain17: yes, this got moved by leptosfmt

crates/superposition_types/src/api/experiments.rs
ayushjain17: why this ?
we keep all the filters as optional only no

quite a few logic in frontend is based on that, few behaviors would break with this

Datron: I usually don't like Option<bool> because it gives 3 possible states to a simple boolean: 
Some(true) -> true
Some(false) -> false
None -> false

You get false when None is specified, so using the serde default to handle this case simplifies your code from handling optional bools. 

> quite a few logic in frontend is based on that, few behaviors would break with this

Since this is a related to a filter only, would it break?

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: not feasible to pass query builder param, it has a very complex type and the type would change every time a new filter param is added, which is not good

crates/superposition_types/src/api/experiments.rs
ayushjain17: yes, default behaviors would break on the pages

mahatoankitkumar (approved):

PR #693, opened 2025-09-08
docs: commit smithy generated openapi spec

knutties:
## Problem
The openapi spec generated from Smithy is not committed and doc generation fails.

https://github.com/juspay/superposition/actions/runs/17551148409/job/49843680518

## Solution
Commit it.

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
NA

NA

### DIFF
--- a/makefile
+++ b/makefile
@@ -311,7 +311,7 @@ grafana-local:
 	cd grafana && $(COMPOSE) up
 
 local-docs-view: smithy-api-docs
-	cd docs && npm ci && npm start
+	cd docs && npm start
 
 default: dev-build frontend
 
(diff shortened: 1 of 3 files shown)


makefile
```diff
@@ -269,6 +269,7 @@ smithy-clients: smithy-build
 
 # API Documentation targets
 smithy-api-docs: smithy-build
+	cp $(SMITHY_BUILD_SRC)/openapi/Superposition.openapi.json docs/docs/api/
 	cd docs && npm ci && npm run openapi-docs
```
ayushjain17, line 273: Can you remove the `npm ci ` from local-docs-view
That us redundant and running that triggers npm install twice

PR #697, opened 2025-09-09
fix: add all query param to list contexts filter

Datron:
## Problem
The all pagination filter is missing for list contexts

## Solution
Add it


ayushjain17 (approved):

clients/python/sdk/README.md
ayushjain17, line 1: this needs to be reverted

clients/javascript/sdk/src/runtimeConfig.browser.ts
ayushjain17, line 1: needs to be reverted

clients/javascript/sdk/src/runtimeConfig.ts
ayushjain17, line 1: needs to be reverted

LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2018-2025 Juspay Technologies Private Limited or its affiliates. All Rights Reserved.
```
ayushjain17: why are we changing this ?

ayushjain17 (changes requested):

LICENSE
Datron: This should be the proper version so the license applies

Datron: I had set it previously, someone reverted it

PR #700, opened 2025-09-11
feat: Exact match option for context and experiment filter

ayushjain17:
Includes changes for backend, frontend as well smithy - supporting both map and jsonlogic versions

## Changelog
Added `dimension_match_strategy` for context and experiment filtering
Options for the strategy - `exact` and `subset`
`subset` is the default strategy

https://github.com/user-attachments/assets/dacfd53b-0b17-4c64-a7d2-8cf8b5809795


smithy/models/context.smithy
```diff
@@ -239,6 +239,10 @@ operation ListContexts {
         @httpQuery("plaintext")
         @notProperty
         plaintext: String
+        
+        @httpQuery("filter_exact_match")
+        @notProperty
+        filter_exact_match: Boolean
```
knutties: just `exact_match` suffices ?  It conveys the meaning right ?

smithy/models/experiments.smithy
```diff
@@ -377,6 +377,10 @@ operation ListExperiment {
         @httpQuery("global_experiments_only")
         @notProperty
         global_experiments_only: Boolean
+
+        @httpQuery("filter_exact_match")
+        @notProperty
+        filter_exact_match: Boolean
```
knutties: same comment as previously.

smithy/models/context.smithy
knutties: or

```
dimension_match_strategy = exact | subset 
```

might be more appropriate.

ayushjain17: 👍

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -351,10 +354,15 @@ async fn list_contexts(
...
         }
         let dimension_keys = dimension_params.keys().cloned().collect::<Vec<_>>();
-        let dimension_filter_contexts =
+        let dimension_filtered_contexts =
             Context::filter_by_dimension(all_contexts, &dimension_keys);
+
+        let filter_fn = match filter_params.dimension_match_strategy.unwrap_or_default() {
+            DimensionMatchStrategy::Exact => Context::filter_exact_match,
+            DimensionMatchStrategy::Subset => Context::filter_by_eval,
+        };
```
Datron, line 363: awesome use of functions as values

knutties (approved):

PR #702, opened 2025-09-12
chore: make smithy generation consistent, with deadcode removal

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


makefile
```diff
@@ -245,21 +245,39 @@ tailwind:
...
+	mkdir -p clients/python/sdk
 	cp -r $(SMITHY_BUILD_SRC)/python-client-codegen/*\
 				clients/python/sdk
+
+	rm -rf clients/haskell/sdk
+	mkdir -p clients/haskell/sdk
 	cp -r $(SMITHY_BUILD_SRC)/haskell-client-codegen/*\
 				clients/haskell/sdk
+
+## not removing the root folder here, deleting the src folder only
```
knutties, line 275: why so ?  are those parts not generated by Smithy ?

makefile
```diff
@@ -245,21 +245,39 @@ tailwind:
 	cd crates/frontend && npx tailwindcss -i ./styles/tailwind.css -o ./pkg/style.css --watch
 
 smithy-build:
+	rm -rf smithy/output
```
knutties, line 250: It is better to have a separate target to do the `rm -rf` - say something like `smithy-clean` and invoke it separately.

```make
smithy-clean:
   ## all rm -rf

smithy-build:
  ## existing step

smithy-clean-build:
   make smithy-clean 
   make smithy-build
```


You don't want to make the build in-efficient for non-removal changes.

makefile
Datron: Yes we have some custom code in here, for deployments, registry LICENSES symlinks, etc

knutties (approved):

makefile
knutties: @ayushjain17 - thanks for taking care of this.

PR #705, opened 2025-09-15
docs: fix tests/doc for python openfeature provider usage

knutties:
## Problem
The wrong usage of provider functions with `get_{type}_details` pattern was tripping LLMs and was also pure incorrect code and needs to be fixed.

## Solution
Use the correct client methods and include test execution in CI.

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
NA

NA

### DIFF
--- a/clients/python/provider/README.md
+++ b/clients/python/provider/README.md
@@ -59,7 +59,7 @@ ctx = EvaluationContext(
   attributes={'d1': 'd1'}
 )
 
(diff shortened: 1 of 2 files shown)


knutties:
Invalid issue.  `get_{type}_details` is the format for detailed evaluation - supported by openfeature.

https://openfeature.dev/docs/reference/concepts/evaluation-api/#detailed-evaluation

PR #706, opened 2025-09-16
build: setup msrv and edition consistently in all Cargo.toml

knutties:
## Problem
@SanchithHegde pointed out the missing `rust-version` and `edition` values in our crates.

## Solution
Set the value in the top-level Cargo.toml and reference it in all the crates.


crates/superposition_sdk/Cargo.toml
```diff
@@ -1,10 +1,11 @@
 # Code generated by software.amazon.smithy.rust.codegen.smithy-rs. DO NOT EDIT.
 [package]
 name = "superposition_sdk"
-version.workspace = true
+version = { workspace = true }
 authors = ["superposition@juspay.in"]
 description = "Rust SDK to work with Superposition"
-edition = "2021"
+edition = { workspace = true }
+rust-version = { workspace = true }
```
ayushjain17: if I am not wrong, this would get reverted in smithy generation
and I suppose we should use a different value just for the provider, so that we can have a higher inclusion percent

the server can be at a high MSRV like 1.82.0 but the sdk and the provider should not be dependent on this high of a number

knutties: This was based on @SanchithHegde's comment using the cargo msrv command (I think uniffi forces it).  Running it currently.  Will add the output to the PR.

Can we configure smithy to generate this the way we want it? cc: @ShubhranshuSanjeev @ShreyBana

ayushjain17: @knutties 
smithy-build.json takes `minimumSupportedRustVersion` as config

knutties:
```shell
➜  superposition git:(fix-msrv) ✗ cargo msrv verify --rust-version 1.81
  [Meta]   cargo-msrv 0.18.4  

Compatibility Check #1: Rust 1.81.0
◝ Rust 1.81.0      Elapsed 2s                                                                                                                                                                                                                                 ◞ Rust 1.81.0      Elapsed 3s                                                                                                                                                                                                                                 ◡ Rust 1.81.0      Elapsed 3s                                                                                                   ◜ Rust 1.81.0      Elapsed 4s                                                                                                         [FAIL]   Is incompatible 

...

uniffi errors
```
One option is to skip uniffi in rust builds via feature flag (which is actually a good idea by @SanchithHegde).  I will do this as well in this PR.

ayushjain17: @knutties you would need to check the MSRV for the crate individually I guess
also, uniffi would be needed for the provider only (from the import point of view)

I was thinking something:
sdk and provider to have their respective MSRVs
and all other crates can use a common one from workspace (if at all a common MSRV is needed)

knutties: Yes - that is the plan.  For crates that wider people have to use - we can lower the msrv - but we will need to remove uniffi dependency in Rust SDKs as it is not needed.  uniffi is not needed for the Rust provider - it is needed only in the other language bindings.

We might need some additional CI checks to ensure we compile with some lower msrv as well.  I will do that in this PR.

ayushjain17: But the rust sdk does not include uniffi

crates/frontend/src/components/experiment_form.rs
```diff
@@ -72,7 +72,7 @@ impl From<ExperimentType> for ExperimentFormType {
 
 enum ResponseType {
     UpdatePrecheck,
-    Response(ExperimentResponse),
+    Response(Box<ExperimentResponse>),
```
ayushjain17: Why is this change needed ?

crates/cac_client/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "cac_client"
 description = "CAC Client provides interfaces for applications to interact with the context-aware-config module in superposition. Note: this is deprecated, please use the provider"
 version.workspace = true
 edition = "2021"
+rust-version.workspace = true
```
ayushjain17: For client also same version ?

Datron: resolved

crates/frontend/src/components/experiment_form.rs
Datron: This becomes a clippy failure with edition set, the enum variants have different sizes and therefore Box is needed for dynamic allocation

smithy/smithy-build.json
```diff
@@ -34,6 +34,7 @@
       "service": "io.superposition#Superposition",
       "module": "superposition_sdk",
       "moduleVersion": "0.1.0",
+      "minimumSupportedRustVersion": "1.82.0",
```
ayushjain17, line 37: are you sure this is correct ?
I ran `cargo msrv find` and I found it to be `1.80.1`

knutties: @ayushjain17  @Datron - we were able to validate the lower bound ?  Looks like it is easy for verify.  Worst case we can take this forward and lower it later as well.

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.86.0"
+channel = "1.90.0"
```
ayushjain17, line 2: can we add `cargo-msrv` in `flake.nix`'s list of packages

ayushjain17 (approved):

PR #709, opened 2025-09-22
feat: introduce cohort dimensions

Datron:
## Problem
discussion #589 

## Solution
Support cohort dimensions in superposition

https://drive.google.com/file/d/1BfXRCQLds73cAXt92bKTVFiRmEAykUXZ/view?usp=sharing


crates/superposition_types/src/database/models/cac.rs
```diff
@@ -368,43 +425,17 @@ where
...
-    Deserialize, Serialize, Clone, Deref, Debug, PartialEq, Into, AsRef, Default,
+    Deserialize, Serialize, Clone, Deref, DerefMut, Debug, PartialEq, Into, AsRef, Default,
 )]
 #[cfg_attr(
     feature = "diesel_derives",
     derive(AsExpression, FromSqlRow, JsonFromSql, JsonToSql)
 )]
 #[cfg_attr(feature = "diesel_derives", diesel(sql_type = Json))]
-pub struct DependencyGraph(HashMap<String, Vec<String>>);
+pub struct DependencyGraph(pub HashMap<String, Vec<String>>);
```
ayushjain17, line 440: we already have `new` function can we avoid making the internal data public

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -13,12 +13,14 @@ use diesel::{
...
     pg::{Pg, PgValue},
     serialize::{Output, Result as SResult, ToSql},
-    sql_types::{Integer, Json},
+    sql_types::{Integer, Json, Text},
     AsChangeset, Insertable, QueryId, Queryable, Selectable,
 };
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
 #[cfg(feature = "diesel_derives")]
+use std::str::FromStr;
```
ayushjain17: we have this imported at line 2 already

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -64,6 +66,62 @@ impl Overridden<Cac<Overrides>> for Context {
...
+    }
+}
+
+#[cfg(feature = "diesel_derives")]
+impl ToSql<Text, Pg> for DimensionType {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> SResult {
+        let value = self.to_string();
+        <String as ToSql<Text, Pg>>::to_sql(&value, &mut out.reborrow())
+    }
+}
```
ayushjain17, line 128: you dont need to implement this manually, you can simple derive `TextFromSql` and `TextToSql` on the type

and you only need to implement `From<&DimensionType> for String` and `TryFrom<String> for DimensionType`

rest are taken care of internally

crates/superposition_types/src/database/models/cac.rs
```diff
@@ -64,6 +66,62 @@ impl Overridden<Cac<Overrides>> for Context {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
```
ayushjain17, line 68: just to highlight, this would just be generating a externally tagged value with the external tag (key) being in `SCREAMING_SNAKE_CASE `

crates/context_aware_config/src/helpers.rs
```diff
@@ -82,6 +85,32 @@ pub fn get_meta_schema() -> JSONSchema {
...
+            }
+        },
+        "required": ["type", "enum", "definitions"],
+    });
+
+    JSONSchema::options()
+        .with_draft(Draft::Draft7)
+        .compile(&my_schema)
+        .expect("Error encountered: Failed to compile 'context_dimension_schema_value'. Ensure it adheres to the correct format and data type.")
+}
```
ayushjain17, line 115: nice

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -71,11 +72,31 @@ async fn create(
...
+            validate_jsonschema(&state.meta_schema, &schema_value)?;
+            if let DimensionType::RemoteCohort(ref cohort_based_on) =
+                create_req.dimension_type
+            {
+                does_dimension_exist_for_cohorting(
+                    cohort_based_on,
+                    &schema_name,
+                    &mut conn,
+                )?;
+            }
```
ayushjain17, line 89: won't it be better to have a separate arm itself for this, instead of doing an if check,

also, just to verify, `does_dimension_exist_for_cohorting` check would also be needed for `LocalCohort` as well right ?

crates/frontend/src/types.rs
```diff
@@ -219,3 +219,51 @@ impl DropdownOption for Option<String> {
...
+    Clone,
+    Default,
+    Debug,
+    PartialEq,
+    Eq,
+    strum_macros::Display,
+    strum_macros::EnumString,
+    strum_macros::EnumIter,
+)]
+pub enum DimensionTypeOptions {
```
ayushjain17, line 235: why do we need to add an extra type for this ?
original type could only be used directly, no ?

separate strum should also not be needed, there is a function in FE which converts `snake_case` to `Title Case`, you can use that

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -358,20 +396,31 @@ async fn delete_dimension(
...
 
-            validate_dimension_deletability(
-                &name,
-                &dimension_data,
-                &user.get_email(),
-                transaction_conn,
-                &schema_name,
-            )?;
+            if !dimension_data.dependency_graph.is_empty() {
+                return Err(bad_argument!("The dimension {} currently has other dimensions that are using it in their cohort definitions. To delete this dimension, you need to delete these cohorts", &dimension_data.dimension))
```
ayushjain17, line 405:
```suggestion
                return Err(bad_argument!("The dimension {} currently has other dimensions that are using it in their cohort definitions. To delete this dimension, you need to delete those cohorts", &dimension_data.dimension))
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -20,16 +18,21 @@ use service_utils::{
     helpers::{generate_snowflake_id, validation_err_to_str},
     service::types::{AppState, SchemaName},
 };
+use std::collections::{HashMap, HashSet};
```
ayushjain17: std imports are kept before all, just like the one which got deleted

crates/frontend/src/pages/dimension.rs
```diff
@@ -275,62 +282,68 @@ pub fn dimension_page() -> impl IntoView {
...
+                        {if dimension.dimension_type != DimensionType::Regular {
+                            view! {
+                                <div class="card bg-base-100 max-w-screen shadow">
+                                    <div class="card-body">
+                                        <h2 class="card-title">"Cohort Dimension"</h2>
+                                        <div class="flex flex-col gap-4">
+                                            <div class="flex flex-row gap-6 flex-wrap">
+                                                <div class="h-fit flex flex-col gap-1">
+                                                    <div class="stat-title">"Cohort Dimension"</div>
+                                                    <Badge
```
ayushjain17: curse of developing UI, unnecessary div nesting

crates/frontend/src/components/context_form.rs
```diff
@@ -338,7 +343,7 @@ pub fn context_form(
                 dimension_map
                     .get_value()
                     .get(dimension)
-                    .map(|d| d.dependencies.clone())
+                    .map(|_| vec![])
```
ayushjain17: empty ?
then basically this entire piece of code is not needed, if at all this is correct

crates/frontend/src/components/context_form.rs
```diff
@@ -391,10 +396,8 @@ pub fn context_form(
             return "Mandatory Dimension".to_string();
         }
         if context_dependencies.get().contains(&variable) {
-            if let Some(parents) = dimension_map
-                .get_value()
-                .get(&variable)
-                .map(|d| d.dependents.clone())
+            if let Some(parents) =
+                dimension_map.get_value().get(&variable).map(|_| Vec::new())
```
ayushjain17: parents is always an empty array here

ayushjain17:
can we directly upload the view here, uploading it in drive makes it un-accessible to the open world

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: I would have to repeat
```
#[cfg(not(feature = "jsonlogic"))]
allow_primitive_types(&schema_value)?;
validate_jsonschema(&state.meta_schema, &schema_value)?;
```

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: Both are grammatically correct, since this sentence precedes this one:
`has other dimensions that are using it in their cohort definitions`

crates/context_aware_config/src/helpers.rs
Datron: This has been auto formatted

crates/frontend/src/components/context_form.rs
Datron: Yeah probably not needed, let me take a look. Some of this was done to make code compile

crates/frontend/src/components/context_form.rs
Datron: Will update

crates/frontend/src/types.rs
Datron: This is to display the type in the UI, other wise the dependency gets serialized and displayed as {"local_cohort: "name"}

crates/superposition_types/src/database/models/cac.rs
Datron: I need this, I use the hashmap functions which can't be used without this being public. Why not make this public anyway?

crates/frontend/src/pages/dimension.rs
Datron: Just collapse it in the IDE 😅

crates/context_aware_config/src/api/dimension/handlers.rs
ayushjain17: but it is just 2 lines right, better to repeat no ?
makes it cleaner I think ?

crates/frontend/src/pages/dimension.rs
ayushjain17: I meant this is not correct, you do not require so many divs to make it look however it is looking over here

will share the code in sometime

crates/context_aware_config/src/helpers.rs
ayushjain17: this happens if the line separator between them is deleted which the formatter introduces, then it ignore the separations and sorts it irrespective

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,200 @@ pub fn validate_jsonschema(
...
+) -> superposition::Result<()> {
+    if cohort_based_on.is_empty() {
+        return Err(validation_error!(
+            "Please specify a valid dimension that this cohort can derive from. Refer our API docs for examples",
+        ));
+    }
+
+    let enum_options = validate_cohort_jsonschema(cohort_schema)?;
+
+    let cohort_schema = cohort_schema.get("definitions").ok_or(validation_error!(
```
Datron, line 283: The function `validate_cohort_jsonschema` does this

crates/superposition_types/src/database/models/cac.rs
ayushjain17: should we consider adding wrapper fucntions on top ?

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -121,418 +121,84 @@ pub fn validate_dimension_position(
...
-    }
-
-    // Add the dependencies to the new dependency graph
-    new_dependency_graph.insert(
-        dimension_name.to_string(),
-        dependent_dimensions
-            .iter()
-            .map(String::from)
-            .collect::<Vec<_>>(),
+    let mut dimensions_vector = get_dimension_data(conn, schema_name)?;
```
ayushjain17: can we make this plural - function name

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -121,418 +121,84 @@ pub fn validate_dimension_position(
...
-}
-
-pub fn validate_and_update_dimension_hierarchy(
-    dimension_data: &Dimension,
-    dependent_dimensions: &[String],
+/// Update the dependency graph of the cohorted dimension
+/// Follow its parents and update their graphs as well
+pub fn create_connections_with_dependents(
+    cohorted_dimension: &str,
+    cohort_dimension: &str,
```
ayushjain17:
```suggestion
    dimension_name: &str,
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,210 @@ pub fn validate_jsonschema(
...
+    conn: &mut DBConnection,
+) -> superposition::Result<()> {
+    if let Some(dim) = dimensions::dsl::dimensions
+        .filter(dimensions::dsl::dimension.eq(dim))
+        .schema_name(schema_name)
+        .get_result::<Dimension>(conn)
+        .optional()?
+    {
+        match dim.dimension_type {
+            DimensionType::LocalCohort(_) => {
```
ayushjain17, line 246: we should allow for LocalCohort to depend on LocalCohort, as discussed earlier

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,210 @@ pub fn validate_jsonschema(
...
+    cohort_schema: &Value,
+    cohort_based_on: &String,
+    schema_name: &SchemaName,
+    conn: &mut DBConnection,
+) -> superposition::Result<()> {
+    if cohort_based_on.is_empty() {
+        return Err(validation_error!(
+            "Please specify a valid dimension that this cohort can derive from. Refer our API docs for examples",
+        ));
+    }
```
ayushjain17, line 279: not needed DB will catch it

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,210 @@ pub fn validate_jsonschema(
...
+
+    let logic = match cohort_schema {
+        Value::Object(logic) if logic.is_empty() => {
+            log::error!("Empty JSON Logic object is not allowed");
+            return Err(validation_error!(
+                "Empty object is not allowed as a schema, mention at least one cohort"
+            ));
+        }
+        Value::Object(logic) => {
+            let cohorts = logic.keys();
```
ayushjain17:
```suggestion
            let cohort_options = logic.keys();
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,210 @@ pub fn validate_jsonschema(
...
+                "Invalid JSON Logic schema: expected an object, found: {}",
+                cohort_schema
+            ));
+        }
+    };
+
+    // check if only one dimension is used across all cohort enums
+    let mut dimensions_used = HashSet::new();
+
+    for (cohort, expression) in logic.iter() {
```
ayushjain17:
```suggestion
    for (cohort_option, expression) in logic.iter() {
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,210 @@ pub fn validate_jsonschema(
...
+                e
+            )
+        })?;
+        dimensions_used.extend(dims);
+
+        // validate the jsonlogic itself by running it against an empty object
+        jsonlogic::apply(cohort_schema, &json!("{}")).map_err(|e| {
+            log::error!("jsonlogic validation error for cohort {cohort}: {e:?}");
+            validation_error!("Invalid jsonlogic in cohort {} : {:?}", cohort, e)
+        })?;
```
ayushjain17: expression parsing is enough for this
this just has resoltuion on top of expression parsing, so we can drop this

crates/frontend/src/components/dimension_form.rs
```diff
@@ -255,50 +250,167 @@ pub fn dimension_form(
...
+                                view! {
+                                    <div class="form-control">
+                                        <Label title="Cohort Based on" />
+                                        <Dropdown
+                                            disabled=edit
+                                            dropdown_text=cohort_based_on_rs.get()
+                                            dropdown_direction=DropdownDirection::Down
+                                            dropdown_btn_type=DropdownBtnType::Select
+                                            dropdown_options
+                                            on_select=Callback::new(move |selected_item: String| {
```
ayushjain17: you can avoid wrapping it in Callback, it is handled internally
```suggestion
                                            on_select=move |selected_item: String| {
```

crates/frontend/src/types.rs
```diff
@@ -219,3 +219,51 @@ impl DropdownOption for Option<String> {
...
+            DimensionTypeOptions::RemoteCohort => DimensionType::RemoteCohort(cohort),
+        }
+    }
+    pub fn from_dimension_type(dim_type: &DimensionType) -> Self {
+        match dim_type {
+            DimensionType::Regular{} => DimensionTypeOptions::Regular,
+            DimensionType::LocalCohort(_) => DimensionTypeOptions::LocalCohort,
+            DimensionType::RemoteCohort(_) => DimensionTypeOptions::RemoteCohort,
+        }
+    }
```
ayushjain17:
```suggestion
impl From<DimensionType> for DimensionTypeOptions {
    fn from(dim_type: DimensionType) -> Self {
        match dim_type {
            DimensionType::Regular{} => Self::Regular,
            DimensionType::LocalCohort(_) => Self::LocalCohort,
            DimensionType::RemoteCohort(_) => Self::RemoteCohort,
        }
    }
}
```

crates/frontend/src/pages/dimension.rs
```diff
@@ -275,62 +282,67 @@ pub fn dimension_page() -> impl IntoView {
...
+                                                </div>
+                                            </div>
+                                        </div>
+                                    </div>
+                                </div>
+                            }
+                                .into_view()
+                        } else {
+                            ().into_view()
+                        }}
```
ayushjain17, line 340: can we simply use this here, reduces a lot of change log:
(starting from the line right after this line: `<DimensionInfo dimension=dimension.clone() />`)
```rust
                        {if dimension.dependency_graph.len() < 2
                            && (DimensionType::Regular {} == dimension.dimension_type)
                        {
                            ().into_view()
                        } else {
                            view! {
                                <div class="card bg-base-100 max-w-screen shadow">
                                    <div class="card-body">
                                        <h2 class="card-title">"Dependency Data"</h2>
                                        <div class="flex flex-col gap-4">
                                            {if dimension.dependency_graph.len() < 2 {
                                                ().into_view()
                                            } else {
                                                view! {
                                                    <div class="flex flex-row gap-6 flex-wrap">
                                                        <div class="h-fit flex flex-col gap-1 overflow-x-scroll">
                                                            <div class="stat-title">
                                                                "Dimensions on which this dimension depends on"
                                                            </div>
                                                            <div class="w-[inherit] pl-5 whitespace-pre overflow-x-auto">
                                                                <TreeNode
                                                                    name=dimension.dimension.clone()
                                                                    data=dimension.dependency_graph.clone()
                                                                    root=true
                                                                />
                                                            </div>
                                                        </div>
                                                    </div>
                                                }
                                                    .into_view()
                                            }}
                                            {match dimension.dimension_type {
                                                DimensionType::Regular {} => ().into_view(),
                                                DimensionType::LocalCohort(cohort_based_on)
                                                | DimensionType::RemoteCohort(cohort_based_on) => {
                                                    view! {
                                                        <div class="flex flex-row gap-6 flex-wrap">
                                                            <div class="h-fit flex flex-col gap-1">
                                                                <div class="stat-title">"Cohort is based on"</div>
                                                                <Badge
                                                                    href_fn=|d| format!("../{d}")
                                                                    options=Signal::derive(move || {
                                                                        vec![cohort_based_on.clone()]
                                                                    })
                                                                />
                                                            </div>
                                                        </div>
                                                    }
                                                        .into_view()
                                                }
                                            }}
                                        </div>
                                    </div>
                                </div>
                            }
                                .into_view()
                        }}
```

crates/frontend/src/components/dimension_form.rs
```diff
@@ -65,24 +69,30 @@ pub fn dimension_form(
...
         create_signal(validation_function_name);
     let (autocomplete_fn_name_rs, autocomplete_fn_name_ws) =
         create_signal(autocomplete_function_name);
     let (description_rs, description_ws) = create_signal(description);
     let (change_reason_rs, change_reason_ws) = create_signal(String::new());
+    let (cohort_based_on_rs, cohort_based_on_ws) = match dimension_type {
+        DimensionType::Regular{} => create_signal(String::new()),
+        DimensionType::LocalCohort(cohort_based_on)
+        | DimensionType::RemoteCohort(cohort_based_on) => create_signal(cohort_based_on),
+    };
```
ayushjain17: dont do it like this, this can have implications, safer to do it this way
```suggestion
    let (cohort_based_on_rs, cohort_based_on_ws) = create_signal(match dimension_type {
        DimensionType::Regular{} => String::new(),
        DimensionType::LocalCohort(cohort_based_on)
        | DimensionType::RemoteCohort(cohort_based_on) => cohort_based_on,
    });
```

crates/frontend/src/components/dimension_form.rs
```diff
@@ -255,50 +250,167 @@ pub fn dimension_form(
...
+                        disabled=edit
+                        dropdown_width="w-100"
+                        dropdown_icon="".to_string()
+                        dropdown_text=dimension_type_rs.get().to_string()
+                        dropdown_direction=DropdownDirection::Down
+                        dropdown_btn_type=DropdownBtnType::Select
+                        dropdown_options=DimensionTypeOptions::iter().collect()
+                        on_select=Callback::new(move |selected_item: DimensionTypeOptions| {
+                            logging::log!("selected item {}", selected_item);
+                            if !edit && selected_item == DimensionTypeOptions::LocalCohort {
```
ayushjain17:
```suggestion
                            if selected_item == DimensionTypeOptions::LocalCohort {
```

crates/frontend/src/components/dimension_form.rs
```diff
@@ -255,50 +250,167 @@ pub fn dimension_form(
...
+                                                }
+                                                r#type=InputType::Monaco(vec![])
+                                            />
+                                        </EditorProvider>
+                                    </div>
+                                }
+                            }
                         }
                     }}
                 </Suspense>
```
ayushjain17, line 417: this entire `Suspense` section could have simply been this much, way less code:
```rust
                <Suspense>
                    {move || {
                        let dimension_type_schema = SchemaType::Single(
                            JsonSchemaType::from(&dimension_schema_rs.get()),
                        );
                        let schema_label = match dimension_type_rs.get() {
                            DimensionTypeOptions::Regular | DimensionTypeOptions::RemoteCohort => {
                                "Set Schema"
                            }
                            DimensionTypeOptions::LocalCohort => "Set Cohort Definition",
                        };
                        view! {
                            <div class="form-control">
                                <Show when=move || {
                                    dimension_type_rs.get() == DimensionTypeOptions::RemoteCohort
                                }>
                                    {
                                        let dimension_name = dimension_name_rs.get();
                                        let dimension_options = combined_resources
                                            .with(|c| c.as_ref().map(|c| c.dimensions.clone()))
                                            .unwrap_or_default()
                                            .iter()
                                            .filter_map(|d| {
                                                (d.dimension != dimension_name
                                                    && !matches!(
                                                        d.dimension_type,
                                                        DimensionType::LocalCohort(_)
                                                    ))
                                                    .then_some(d.dimension.clone())
                                            })
                                            .collect::<Vec<_>>();
                                        let dimension_options = StoredValue::new(dimension_options);

                                        view! {
                                            <Label title="Cohort Based on" />
                                            <Dropdown
                                                disabled=edit
                                                dropdown_text=cohort_based_on_rs.get()
                                                dropdown_direction=DropdownDirection::Down
                                                dropdown_btn_type=DropdownBtnType::Select
                                                dropdown_options=dimension_options.get_value()
                                                on_select=Callback::new(move |selected_item: String| {
                                                    logging::log!("selected item {:?}", selected_item);
                                                    cohort_based_on_ws.set(selected_item);
                                                })
                                            />
                                        }
                                    }
                                </Show>

                                <Label title=schema_label />
                                <Show when=move || {
                                    matches!(
                                        dimension_type_rs.get(),
                                        DimensionTypeOptions::RemoteCohort
                                    )
                                }>
                                    {
                                        let options = combined_resources
                                            .with(|c| c.as_ref().map(|c| c.type_templates.clone()))
                                            .unwrap_or_default();
                                        let dimension_t = if dimension_type_template_rs
                                            .get()
                                            .is_empty() && edit
                                        {
                                            "change current type template".into()
                                        } else if dimension_type_template_rs.get().is_empty()
                                            && !edit
                                        {
                                            "choose a type template".into()
                                        } else {
                                            dimension_type_template_rs.get()
                                        };

                                        view! {
                                            <Dropdown
                                                dropdown_width="w-100"
                                                dropdown_icon="".to_string()
                                                dropdown_text=dimension_t
                                                dropdown_direction=DropdownDirection::Down
                                                dropdown_btn_type=DropdownBtnType::Select
                                                dropdown_options=options
                                                on_select=Callback::new(move |selected_item: TypeTemplate| {
                                                    logging::log!("selected item {:?}", selected_item);
                                                    dimension_type_template_ws.set(selected_item.type_name);
                                                    dimension_schema_ws.set(selected_item.type_schema);
                                                })
                                            />
                                        }
                                    }
                                </Show>
                                <EditorProvider>
                                    <Input
                                        id="type-schema"
                                        class="mt-5 rounded-md resize-y w-full max-w-md pt-3"
                                        schema_type=dimension_type_schema
                                        value=dimension_schema_rs.get_untracked()
                                        on_change=move |new_type_schema| {
                                            dimension_schema_ws.set(new_type_schema)
                                        }
                                        r#type=InputType::Monaco(vec![])
                                    />
                                </EditorProvider>
                            </div>
                        }
                    }}
                </Suspense>
```

crates/context_aware_config/src/helpers.rs
Datron: Will keep this for early exit, the DB check is way down

crates/frontend/src/types.rs
Datron: I don't want to take ownership while converting

crates/frontend/src/pages/dimension.rs
Datron: This shows an empty Tree node graph for dimensions that do not have any cohorts on them yet. The change was made to avoid displaying anything for these cases

ayushjain17: @Datron there was one condition which I forgot to update
I have updated in the above suggestion itself

can you please try again once

crates/frontend/src/types.rs
ayushjain17: enum anyways does not take the ownership no, and even if it does you can still do:
```suggestion
impl From<&DimensionType> for DimensionTypeOptions {
    fn from(dim_type: &DimensionType) -> Self {
        match dim_type {
            DimensionType::Regular{} => Self::Regular,
            DimensionType::LocalCohort(_) => Self::LocalCohort,
            DimensionType::RemoteCohort(_) => Self::RemoteCohort,
        }
    }
}
```

crates/context_aware_config/src/helpers.rs
ayushjain17: pending

crates/context_aware_config/src/helpers.rs
```diff
@@ -161,16 +190,210 @@ pub fn validate_jsonschema(
...
+                log::error!(
+                    "The definition of the cohort and the enum options do not match. Some enum options do not have a definition, found {} cohorts and {} enum options (not including otherwise)",
+                    cohorts.len(),
+                    enum_options.len() - 1
+                );
+                return Err(validation_error!(
+                    "The definition of the cohort and the enum options do not match. Some enum options do not have a definition, found {} cohorts and {} enum options (not including otherwise)",
+                    cohorts.len(),
+                    enum_options.len() - 1
+                ));
```
ayushjain17, line 306:
```suggestion
                log::error!(
                    "The definition of the enums and the enum options do not match. Some enum options do not have a definition, found {} enum definitions and {} enum options (not including otherwise)",
                    cohorts.len(),
                    enum_options.len() - 1
                );
                return Err(validation_error!(
                    "The definition of the enums and the enum options do not match. Some enum options do not have a definition, found {} enum definitions and {} enum options (not including otherwise)",
                    cohorts.len(),
                    enum_options.len() - 1
                ));
```

Datron: Lets use the word cohort here, it pertains to the type. Enum is the jsonschema field while cohort is the keys defined in definition. We should make that distinction

crates/frontend/src/components/dimension_form.rs
Datron: Cohort based on field is required for the local cohorts as well

crates/context_aware_config/src/helpers.rs
ayushjain17: honestly, keeping the message as what it is right now, is very confusing for me

crates/frontend/src/types.rs
ayushjain17: we could have done something better to do this, but maybe sometime later

crates/context_aware_config/src/helpers.rs
ayushjain17: pending

crates/context_aware_config/src/helpers.rs
Datron: I would like to make the distinction - but if users complain about it we can change it

crates/context_aware_config/src/helpers.rs
```diff
@@ -82,6 +86,32 @@ pub fn get_meta_schema() -> JSONSchema {
...
+        "properties": {
+            "type": {
+                "type": "string"
+            },
+            "enum": {
+                "type": "array",
+                "items": {
+                    "type": "string"
+                },
+            },
```
ayushjain17, line 100:
```suggestion
            "enum": {
                "type": "array",
                "items": {
                    "type": "string"
                },
                "contains": {
                    "const": "otherwise"
                },
                "minContains": 1
            },
```

crates/superposition_types/src/database/models/cac.rs
Datron: Why? Just make the hashmap public and deref it when using

tests/src/dimension.test.ts
```diff
@@ -574,6 +627,432 @@ describe("Dimension API", () => {
...
+        };
+
+        expect(
+            superpositionClient.send(new CreateDimensionCommand(input)),
+        ).rejects.toThrow(
+            "The definition of the cohort and the enum options do not match. Some enum options do not have a definition, found 2 cohorts and 1 enum options (not including otherwise)",
+        );
+    });
+    
+    test("fail2create a local cohort dimension without enum otherwise", async () => {
```
ayushjain17, line 860: now we dont have a test anymore to check if multiple dimensions are used or not

ayushjain17 (approved):
all fine, just need to see when we push the change for allowing LocalCohort on LocalCohort

tests/src/dimension.test.ts
Datron: Added

PR #711, opened 2025-09-22
feat: cohort dimensions SDK, test and smithy changes

Datron:
tests for #709


tests/src/dimension.test.ts
```diff
@@ -574,6 +627,432 @@ describe("Dimension API", () => {
...
+        const input = {
+            workspace_id: ENV.workspace_id,
+            org_id: ENV.org_id,
+            ...wrongCohort,
+        };
+
+        expect(
+            superpositionClient.send(new CreateDimensionCommand(input)),
+        ).rejects.toThrow(
+            "",
```
ayushjain17, line 856: message not needed here ?

tests/src/dimension.test.ts
```diff
@@ -574,6 +627,432 @@ describe("Dimension API", () => {
...
+        const input = {
+            workspace_id: ENV.workspace_id,
+            org_id: ENV.org_id,
+            ...wrongCohort,
+        };
+
+        expect(
+            superpositionClient.send(new CreateDimensionCommand(input)),
+        ).rejects.toThrow(
+            /The definition of the cohort and the enum options do not match. Some enum options do not have a definition, found [0-9]+ cohorts and [0-9]+ enum options \(not including otherwise\)/,
```
ayushjain17, line 910: expected error message is wrong here
this test was for checking what happens when more than 1 dimension is used

but since the enum list does not have otherwise in it, it is testing something else altogether and not the intended check which was to be tested

PR #712, opened 2025-09-22
feat: Cohort resolution

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,130 @@ fn merge_overrides_on_default_config(
...
+    log::error!("{}", error_msg);
+    error_msg
+}
+
+fn evaluates_to_true(
+    expression: &Value,
+    cohort_based_on: &str,
+    cohort_based_on_value: &Value,
+) -> Result<bool, String> {
+    let evaluation_data = json!({cohort_based_on: cohort_based_on_value});
```
ayushjain17: the entire context data so far could have been blindly sent, no ?

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,130 @@ fn merge_overrides_on_default_config(
...
 
+fn evaluate_cohort(
+    dimensions: &HashMap<String, DimensionInfo>,
+    query_data: &Map<String, Value>,
+) -> Result<Map<String, Value>, String> {
+    let mut modified_context = Map::new();
+
+    populate_regular_dimensions(dimensions, query_data, &mut modified_context);
+    populate_cohort_dimensions(dimensions, &mut modified_context)?;
+
```
ayushjain17: shouldn't we try and perform this is in one go itself, to keep the number of iterations minimum

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,130 @@ fn merge_overrides_on_default_config(
...
+fn populate_cohort_dimensions(
+    dimensions: &HashMap<String, DimensionInfo>,
+    context: &mut Map<String, Value>,
+) -> Result<(), String> {
+    for (dimension_key, dimension_info) in dimensions {
+        if let DimensionType::LocalCohort(ref cohort_based_on) =
+            dimension_info.dimension_type
+        {
+            evaluate_local_cohort_dimension(
```
ayushjain17: the function name says evaluate, but this populates as well
can we stick to single responsibily here, and rather have it only evaluate and return the value

and perform the insertion here itself
just like the way it is done in `populate_regular_dimensions`

crates/superposition_types/src/config.rs
```diff
@@ -344,6 +347,7 @@ pub struct Config {
     pub contexts: Vec<Context>,
     pub overrides: HashMap<String, Overrides>,
     pub default_configs: Map<String, Value>,
+    pub dimensions: HashMap<String, DimensionInfo>,
```
ayushjain17, line 353: it needs to have a default to avoid backwards compatibility issues
```suggestion
    #[serde(default)]
    pub dimensions: HashMap<String, DimensionInfo>,
```

ayushjain17: need to add test for that case as well, where the dimensions in not present in the data

crates/superposition_provider/src/client.rs
```diff
@@ -221,6 +221,7 @@ impl CacConfig {
                     cached_config.default_configs.clone(),
                     &cached_config.contexts,
                     &cached_config.overrides,
+                    &cached_config.dimensions,
```
ayushjain17, line 224: need to check if proper changes have been made for fast mode as well or not, there is no compilation step as yet to verify that

crates/cac_client/src/lib.rs
```diff
@@ -227,13 +227,7 @@ impl Client {
...
                 config = config.filter_by_prefix(&HashSet::from_iter(keys));
             }
-            let evaled_cac = eval::eval_cac(
-                config.default_configs.to_owned(),
-                &config.contexts,
-                &config.overrides,
-                &query_data,
-                merge_strategy,
-            )?;
+            let evaled_cac = eval::eval_cac(&config, &query_data, merge_strategy)?;
```
ayushjain17, line 230: need to update the legacy client as well

mahatoankitkumar: Its already a part of the config
what to be updated?

crates/cac_client/src/eval.rs
mahatoankitkumar: This is much simpler
doing in one iteration would have a lot of checks and complexities.
Anyways dimensions data wouldn't be that much, so this won't be an issue.

crates/cac_client/src/eval.rs
mahatoankitkumar: Yeah but this is more specific of whats needed

crates/cac_client/src/eval.rs
ayushjain17: This function would be called heavily on every single resolve call, this flow should be the most optimised flow

crates/cac_client/src/lib.rs
ayushjain17: I meant the interface
that also needs to be updated I guess

just take a look if that is handled already or not

crates/superposition_core/src/ffi.rs
```diff
@@ -51,6 +53,14 @@ fn ffi_eval_logic(
         .map_err(|err| OperationError::Unexpected(err.to_string()))?;
     let mut _q = json_from_map(query_data)
         .map_err(|err| OperationError::Unexpected(err.to_string()))?;
+    let _dimensions: HashMap<String, DimensionInfo> = dimensions
+        .iter()
```
ayushjain17: to avoid cloning
```suggestion
        .into_iter()
```

crates/superposition_core/src/ffi.rs
```diff
@@ -51,6 +53,14 @@ fn ffi_eval_logic(
         .map_err(|err| OperationError::Unexpected(err.to_string()))?;
     let mut _q = json_from_map(query_data)
         .map_err(|err| OperationError::Unexpected(err.to_string()))?;
+    let _dimensions: HashMap<String, DimensionInfo> = dimensions
```
ayushjain17: this does not need `_` right ?
```suggestion
    let dimensions: HashMap<String, DimensionInfo> = dimensions
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -617,10 +617,24 @@ async fn reduce_config_key(
         }
     }
 
+    let dimensions: HashMap<String, DimensionInfo> = dimension_schema_map
+        .iter()
```
ayushjain17: to avoid cloning
```suggestion
        .into_iter()
```

crates/superposition_types/src/config.rs
```diff
@@ -344,6 +347,8 @@ pub struct Config {
     pub contexts: Vec<Context>,
     pub overrides: HashMap<String, Overrides>,
     pub default_configs: Map<String, Value>,
+    #[serde(default)]
+    pub dimensions: HashMap<String, DimensionInfo>,
 }
 
 impl Config {
```
Datron, line 356: filter_by_eval should also include the pre-resolution step, where cohorts are resolved

crates/superposition_types/src/config.rs
```diff
@@ -405,10 +411,19 @@ impl Config {
...
             default_configs: filtered_default_config,
+            dimensions: self.dimensions.clone(),
         }
     }
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct DimensionInfo {
+    // Todo: Add dimension name
```
Datron: Pending TODO, should this be done?

crates/superposition_core/src/core.h
```diff
@@ -11,6 +11,7 @@ char* core_get_resolved_config(
     const char* default_config_json,
     const char* contexts_json,
     const char* overrides_json,
+    const char* cohort_dimensions,
```
Datron: nitpick: define if its json or str, following the pattern used by other variables
```suggestion
    const char* cohort_dimensions_json,
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -536,10 +547,39 @@ pub fn generate_cac(
...
 
+    let dimensions: HashMap<String, DimensionInfo> = dimensions::dsl::dimensions
+        .select((
+            dimensions::dsl::dimension,
+            dimensions::dsl::schema,
+            dimensions::dsl::dimension_type,
+        ))
+        .order_by(dimensions::dsl::position.asc())
+        .schema_name(schema_name)
+        .load::<(String, Value, DimensionType)>(conn)
```
Datron: nitpick
```suggestion
        .get_results::<Vec<DimensionInfo>>(conn)
```
Would this not work? May need to implement From<(String, Value, DimensionType)> for DimensionInfo

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,168 @@ fn merge_overrides_on_default_config(
...
+                .get(dimension_key)
+                .cloned()
+                .unwrap_or(Value::Null);
+            modified_context.insert(dimension_key.clone(), value);
+        } else if let DimensionType::LocalCohort(ref cohort_based_on) = dimension_info.dimension_type
+        {
+            let matched_cohort = evaluate_local_cohort_dimension(
+                dimension_key,
+                cohort_based_on,
+                dimension_info,
```
Datron: nitpick: just pass the schema here

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,168 @@ fn merge_overrides_on_default_config(
...
+        }
+        DimensionType::LocalCohort(ref dependency) => {
+            return resolve_dimension_value(dimensions, dependency, query_data, modified_context);
+        }
+        _ => {
+            return Err(create_schema_error(
+                cohort_based_on,
+                "dimension is a Remote_Cohort (not supported for cohort resolution)",
+            ));
+        }
```
Datron: The remote dimension and regular dimension behaviour should be the same. Since you can't resolve it, just return the dimension value from cohort_based_on.

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,168 @@ fn merge_overrides_on_default_config(
...
+            Some(cohort) => {
+                let schema = definitions_object.get(cohort).ok_or_else(|| {
+                    create_schema_error(
+                        local_cohort,
+                        &format!("missing definition for cohort '{cohort}'"),
+                    )
+                })?;
+
+                // Find the first matching cohort definition
+                if evaluates_to_true(schema, cohort_based_on, cohort_based_on_value)? {
```
Datron: `evaluates_to_true` needs a better name, something like `check_cohort_membership`

Datron: We should avoid the function jump here, just evaluate JSON logic in place.

crates/superposition_types/src/config.rs
ayushjain17: +1

crates/superposition_types/src/config.rs
```diff
@@ -565,6 +580,30 @@ pub(crate) mod tests {
                 "test.test1": 12,
                 "test2.key": false,
                 "test2.test": "def_val"
+            },
+            "dimensions" : {
+                "test3": {
+                    "schema": {
+                        "type": "boolean"
+                    },
+                    "dependencies": [],
```
ayushjain17: there is no such field

crates/superposition_types/src/config.rs
```diff
@@ -565,6 +580,30 @@ pub(crate) mod tests {
...
+                    "dimension_type": "REGULAR"
+                },
+                "test2": {
+                    "schema": {
+                        "type": "integer"
+                    },
+                    "dependencies": [],
+                    "dimension_type": "REGULAR"
```
ayushjain17: dimension_type's ser-de has changed

crates/superposition_provider/src/utils.rs
```diff
@@ -99,10 +99,31 @@ impl ConversionUtils {
...
+                                SuperpositionError::SerializationError(format!(
+                                    "Invalid dimension info for '{}': {}",
+                                    key, e
+                                ))
+                            })?;
+                        Ok((key.clone(), dim_info))
+                    })
+                    .collect::<Result<HashMap<String, DimensionInfo>>>()
+            })
+            .unwrap_or_else(|| Ok(HashMap::new()))?;
```
ayushjain17, line 120: ??
```suggestion
            .unwrap_or_default();
```

crates/superposition_provider/src/utils.rs
```diff
@@ -256,10 +277,32 @@ impl ConversionUtils {
...
             .clone();
 
+        let dimensions = map
+            .get("dimensions")
+            .and_then(|v| v.as_object())
+            .ok_or_else(|| {
+                SuperpositionError::ConfigError(
+                    "Missing or invalid 'dimensions' field".to_string(),
+                )
+            })?
```
ayushjain17: why is this being treated as an error, when in the above case it was not considered as an error
and this field is supposed to be optional only

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -301,11 +314,20 @@ pub extern "C" fn core_get_resolved_config_with_reasoning(
         }
     }
 
+    let dimensions = match parse_json::<HashMap<String, DimensionInfo>>(dimensions) {
```
ayushjain17, line 317: just to verify, is this parse function correct ?
it handles the internal conversions right ?

crates/superposition_core/src/config.rs
ayushjain17, line 1: not reviewing this file, assuming it has same code as eval.rs

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -411,6 +415,97 @@ pub fn update_override_of_existing_ctx(
...
+
+fn is_remote_cohort_dimension(
+    dimensions: &HashMap<String, DimensionInfo>, 
+    dimension_key: &str
+) -> bool {
+    dimensions
+        .get(dimension_key)
+        .map(|info| matches!(&info.dimension_type, DimensionType::RemoteCohort(_)))
+        .unwrap_or(false)
+}
```
Datron: Why not inline this function?

crates/superposition_types/src/config.rs
```diff
@@ -344,10 +347,14 @@ pub struct Config {
...
     pub overrides: HashMap<String, Overrides>,
     pub default_configs: Map<String, Value>,
+    #[serde(default)]
+    pub dimensions: HashMap<String, DimensionInfo>,
 }
 
 impl Config {
     pub fn filter_by_dimensions(&self, dimension_data: &Map<String, Value>) -> Self {
+        // Need to copy the evaluate_cohort function here too?, because cyclic dependency
+        // let modified_context = evaluate_cohort(&self.dimensions, dimension_data)?;
```
ayushjain17: pending

crates/superposition_sdk/src/operation/get_config/_get_config_output.rs
ayushjain17, line 1: how is this file getting changed without updating smithy ?
also, smithy file change not added yet

crates/superposition_core/src/ffi.rs
```diff
@@ -51,6 +53,14 @@ fn ffi_eval_logic(
...
     let mut _q = json_from_map(query_data)
         .map_err(|err| OperationError::Unexpected(err.to_string()))?;
+    let dimensions: HashMap<String, DimensionInfo> = dimensions
+        .into_iter()
+        .map(|(k, v)| {
+            let dim_info: DimensionInfo = serde_json::from_str(&v)
+                .map_err(|err| OperationError::Unexpected(err.to_string()))?;
+            Ok((k, dim_info))
+        })
+        .collect::<Result<HashMap<String, DimensionInfo>, OperationError>>()?;
```
ayushjain17:
```suggestion
        .map(|(k, v)| {
            serde_json::from_str(&v)
                .map(|dim_info| (k, dim_info))
                .map_err(|err| OperationError::Unexpected(err.to_string()))
        })
        .collect::<Result<HashMap<_, _>, _>>()?;
```

crates/context_aware_config/src/helpers.rs
```diff
@@ -115,8 +118,18 @@ pub fn get_cohort_meta_schema() -> JSONSchema {
...
 ) -> superposition::Result<()> {
+    let dimension_schema = JSONSchema::options()
+        .with_draft(Draft::Draft7)
+        .compile(dimension_schema)
+        .map_err(|e| {
+            log::error!(
+                "Failed to compile as a Draft-7 JSON schema: {}",
+                e.to_string()
+            );
+            bad_argument!("Error encountered: Failed to compile 'context_dimension_schema_value'. Ensure it adheres to the correct format and data type.")
```
ayushjain17:
```suggestion
            bad_argument!("Error encountered: invalid jsonschema for dimension.")
```

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -202,3 +197,35 @@ fn update_dimensions_in_db(
         .execute(conn)?;
     Ok(())
 }
+
+pub fn fetch_dimensions_map(
```
ayushjain17: files are not formatted, lint checks will fail

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -202,3 +197,35 @@ fn update_dimensions_in_db(
...
+                acc.insert(
+                    key,
+                    DimensionInfo {
+                        schema: schema_value,
+                        dimension_type: dim_type,
+                    },
+                );
+                acc
+            },
+        );
```
ayushjain17:
```suggestion
        .map(|(key, schema_value, dim_type)| {
            (
                key,
                DimensionInfo {
                    schema: schema_value,
                    dimension_type: dim_type,
                },
            )
        })
        .collect();
```

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -38,16 +37,12 @@ pub fn get_dimension_data_map(
     let dimension_schema_map = dimensions_vec
```
ayushjain17: `get_dimensions_data` defined right above this fetches all the fields of all the dimensions, and is never used alone, is always used along with this function `get_dimension_data_map` which removes the extra fields

and the combined behaviour of these functions is exactly same as the newly created function `fetch_dimensions_map` which is infact better in the sense that it reduces the amount of data being transferred from DB

so can we remove these 2 functions and use the new function itself by unifying the 2 types `DimensionData` and `DimensionInfo` as well, as they barely have any difference

and call the new function as `fetch_dimensions_info_map` if type `DimensionInfo` is kept as the name
or `fetch_dimensions_data_map` if type `DimensionData` is kept as the name of the type

also, can we keep the type of `schema` as `Map<String, Value>`, inside the type, instead of `Value`, it is always going to be a map only

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -38,16 +37,12 @@ pub fn get_dimension_data_map(
...
         .iter()
         .filter_map(|item| {
-            let compiled_schema = JSONSchema::options()
-                .with_draft(Draft::Draft7)
-                .compile(&item.schema)
-                .ok()?;
-
             Some((
```
ayushjain17: `filter_map` and `Some` are not needed anymore here
```suggestion
        .map(|item| {
            (
```

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -313,7 +340,17 @@ async fn update(
...
+                tags,
+                dimension_data.change_reason.into(),
+                transaction_conn,
+                &schema_name,
+            )?;
+
+            Ok((result, version_id))
         })?;
 
     let workspace_settings = get_workspace(&schema_name, &mut conn)?;
```
ayushjain17: can we move this line above the DB updates, if this fails, even though the DB was updated successfully, but for some other reasons, a successful response wont be sent, and the end user would not be able to receive the new version id, leading to inconsistent behaviour on that system's side

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -444,13 +495,38 @@ async fn delete_dimension(
...
                 Err(e) => {
                     log::error!("dimension delete query failed with error: {e}");
                     Err(unexpected_error!("Something went wrong."))
                 }
             }
-        })
+        });
+        #[cfg(feature = "high-performance-mode")]
+        put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;
+        resp
```
ayushjain17: why are we updating redis in failure case also ?

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::str;
 
+use cac_client::evaluate_cohort;
```
ayushjain17: can we combine the imports

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -11,6 +12,8 @@ use service_utils::service::types::SchemaName;
 #[cfg(feature = "jsonlogic")]
 use superposition_macros::bad_argument;
 use superposition_macros::{unexpected_error, validation_error};
+use superposition_types::database::models::cac::DimensionType;
+use superposition_types::DimensionInfo;
```
ayushjain17: same here as well, combine the imports with the ones which are already present

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -411,6 +415,91 @@ pub fn update_override_of_existing_ctx(
...
+            get_dimension_root(cohort_based_on, dimensions, dimension_root_map)?
+        },
+        
+        DimensionType::RemoteCohort(_) => {
+            return Err(validation_error!(
+                "Remote cohort dimensions are not supported in context validation: {}",
+                dimension_key
+            ));
+        },
+    };
```
ayushjain17: for this case, `RemoteCohort` should be treated same as `Regular` 
so in either case the dimension should be returned as root
```suggestion
    let dimension_root = match &dimension_info.dimension_type {
        DimensionType::Regular {} | DimensionType::RemoteCohort(_) => dimension_key.to_string(),
        DimensionType::LocalCohort(cohort_based_on) => {
            get_dimension_root(cohort_based_on, dimensions, dimension_root_map)?
        }
    };
```

or if you want, you can actually still travel upwards and find the actual root the correct for `RemoteCohort` as well, to be able to achieve the correct and consistent behviour w.r.t caching also
but either case, you should not fail for `RemoteCohort`, it is very much a valid case here

```rust
    let dimension_root = match &dimension_info.dimension_type {
        DimensionType::Regular {} => dimension_key.to_string(),
        DimensionType::RemoteCohort(cohort_based_on) | DimensionType::LocalCohort(cohort_based_on) => {
            get_dimension_root(cohort_based_on, dimensions, dimension_root_map)?
        }
    };
```

ayushjain17: and then this function does not need to return a `Result` type, a simple `String` type is enough

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
 }
 
+pub fn evaluate_cohort(
+    dimensions: &HashMap<String, DimensionInfo>,
+    query_data: &Map<String, Value>,
+) -> Result<Map<String, Value>, String> {
+    let mut modified_context = Map::new();
+
+    for (dimension_key, dimension_info) in dimensions {
+        if (dimension_info.dimension_type == DimensionType::Regular {} && !modified_context.contains_key(dimension_key)) {
```
ayushjain17: from evaluation perspective, it has to be taken care of that both `Regular` and `RemoteCohort` needs to be treated the same way, that is, both would have same behaviour and both their values would already be avialable in the `context`/`query_data`

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+            let value = query_data
+                .get(dimension_key)
+                .cloned()
+                .unwrap_or(Value::Null);
+            modified_context.insert(dimension_key.clone(), value);
```
ayushjain17: instead of doing this, what should have done is, simply make the incoming `query_data` parameter as `mut` (not mutable reference, but rather a local mutable copy) and go over the list of dimensions, and for the `LocalCohort` simply perform the evaluation and insert the value, for `Regular` and `RemoteCohort` nothing needs to be done as their values would already be present in the `query_data`

also, setting `Value::Null` as the default value, when not found, is wrong here, as a case of `undefined` and `null` would give different evaluation result
in such cases, we simply need to skip that dimension and move ahead and not set any value

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -411,6 +415,91 @@ pub fn update_override_of_existing_ctx(
...
+
+    // Cache and return the result
+    dimension_root_map.insert(dimension_key.to_string(), dimension_root.clone());
+    Ok(dimension_root)
+}
+
+fn validate_cohorts(
```
ayushjain17: can we rename this to something like `validate_cohort_dimension_values`
with the current name, I was somewhat getting the impression that this function is meant to perform JSONSchema validations and all

but rather that is happening in a different function `validate_dimensions`

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -435,6 +524,7 @@ pub fn validate_ctx(
         &workspace_settings.mandatory_dimensions.unwrap_or_default(),
     )?;
     validate_condition_with_dependent_dimensions(conn, context_map, schema_name)?;
+    validate_cohorts(conn, context_map, schema_name)?;
```
ayushjain17: can we move this call to after `validate_dimensions` call, as in that the JSONSchema is validated
and technically JSONSchema validation should run before any other validations on the value is done

and I think for that matter even `validate_condition_with_functions` function should be moved as well

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -411,6 +415,91 @@ pub fn update_override_of_existing_ctx(
...
+
+    for (dimension_key, value) in context_map {
+        // Skip remote cohort dimensions as they cannot be validated locally
+        if dimensions
+            .get(dimension_key)
+            .map(|info| matches!(&info.dimension_type, DimensionType::RemoteCohort(_)))
+            .unwrap_or(false) 
+        {
+            continue;
+        }
```
ayushjain17: this is correct, this should stay

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -411,6 +415,91 @@ pub fn update_override_of_existing_ctx(
...
+        })?;
+
+    let mut dimension_root_map = HashMap::new();
+
+    for (dimension_key, value) in context_map {
+        // Skip remote cohort dimensions as they cannot be validated locally
+        if dimensions
+            .get(dimension_key)
+            .map(|info| matches!(&info.dimension_type, DimensionType::RemoteCohort(_)))
+            .unwrap_or(false)
```
ayushjain17:
```suggestion
            .unwrap_or_default()
```

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -411,6 +415,91 @@ pub fn update_override_of_existing_ctx(
...
+    let evaluated_context = evaluate_cohort(&dimensions, context_map)
+        .map_err(|err| {
+            log::error!("Failed to evaluate cohort: {}", err);
+            unexpected_error!(err)
+        })?;
+
+    let mut dimension_root_map = HashMap::new();
+
+    for (dimension_key, value) in context_map {
+        // Skip remote cohort dimensions as they cannot be validated locally
```
ayushjain17: we should skip for `Regular` also, no ?

crates/context_aware_config/src/api/context/helpers.rs
ayushjain17: also, this function would never get called for `RemoteCohort` as the prior check would pass and would skip the iteration, but from the perspective of a standalone utility function its behviour would become correct and consistent

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+    }
+
+    Ok(modified_context)
+}
+
+fn resolve_dimension_value(
+    dimensions: &HashMap<String, DimensionInfo>,
+    cohort_based_on: &str,
+    query_data: &Map<String, Value>,
+    modified_context: &mut Map<String, Value>,
```
ayushjain17: after making changes for the above solution, 2 different context wont be required to passed along, only one of them would be enough
and all operations would be performed on that one single `mutable query_data`, making a lot the things quite simple as well

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+    let mut modified_context = Map::new();
+
+    for (dimension_key, dimension_info) in dimensions {
+        if (dimension_info.dimension_type == DimensionType::Regular {} && !modified_context.contains_key(dimension_key)) {
+            let value = query_data
+                .get(dimension_key)
+                .cloned()
+                .unwrap_or(Value::Null);
+            modified_context.insert(dimension_key.clone(), value);
+        } else if let DimensionType::LocalCohort(ref cohort_based_on) = dimension_info.dimension_type
```
ayushjain17: after making the above change, we could also add a check that if the current cohort is not already present in `query_data` only then add it, as while doing DFS it might have already been populated

crates/superposition_types/src/config.rs
```diff
@@ -405,10 +413,43 @@ impl Config {
...
             default_configs: filtered_default_config,
+            dimensions: self.dimensions.clone(),
         }
     }
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct DimensionInfo {
+    pub schema: Value,
```
ayushjain17, line 426:
```suggestion
    pub schema: Map<String, Value>,
```

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+        .as_object()
+        .ok_or_else(|| create_schema_error(local_cohort, "failed to parse schema"))?;
+
+    let definitions_object = cohort_schema
+        .get("definitions")
+        .ok_or_else(|| create_schema_error(local_cohort, "missing 'definitions'"))?
+        .as_object()
+        .ok_or_else(|| {
+            create_schema_error(local_cohort, "'definitions' is not an object")
+        })?;
```
ayushjain17:
```suggestion
    let definitions_object = cohort_schema
        .get("definitions")
        .and_then(|v| v.as_object())
        .ok_or_else(|| {
            create_schema_error(local_cohort, "'definitions' is missing/not an object")
        })?;
```

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+
+    // Get the array of cohort names from the "enum" field and remove "otherwise"
+    let cohort_enums = cohort_schema
+        .get("enum")
+        .and_then(|v| v.as_array())
+        .ok_or_else(|| create_schema_error(local_cohort, "missing or invalid 'enum'"))?
+        .iter()
+        .filter(|v| v.as_str() != Some("otherwise"))
+        .collect::<Vec<&Value>>();
```
ayushjain17:
```suggestion
        .filter_map(|v| v.as_str())
        .filter(|s| *s != "otherwise")
        .collect::<Vec<&str>>();
```

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+
+    for cohort in cohort_enums {
+        match cohort.as_str() {
+            Some(cohort) => {
+                let schema = definitions_object.get(cohort).ok_or_else(|| {
```
ayushjain17:
```suggestion
                let jsonlogic = definitions_object.get(cohort).ok_or_else(|| {
```

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+    // Get the array of cohort names from the "enum" field and remove "otherwise"
+    let cohort_enums = cohort_schema
+        .get("enum")
+        .and_then(|v| v.as_array())
+        .ok_or_else(|| create_schema_error(local_cohort, "missing or invalid 'enum'"))?
+        .iter()
+        .filter(|v| v.as_str() != Some("otherwise"))
+        .collect::<Vec<&Value>>();
+
+    for cohort in cohort_enums {
```
ayushjain17:
```suggestion
    for cohort_option in cohort_enums {
```

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
...
+                let evaluation_data = json!({cohort_based_on: cohort_based_on_value});
+                let check_cohort_membership = jsonlogic::apply(schema, &evaluation_data)
+                    .map(|result| matches!(result, Value::Bool(true)))
+                    .map_err(|e| {
+                        log::error!("Failed to evaluate JSON logic expression: {}", e);
+                        format!("Failed to evaluate cohort expression: {}", e)
+                    })?;
+                if check_cohort_membership {
+                    return Ok(cohort.to_string());
+                }
```
ayushjain17:
```suggestion
                if jsonlogic::apply(schema, &evaluation_data) == Ok(Value::Bool(true)) {
                    return Ok(cohort.to_string());
                }
```

crates/cac_client/src/eval.rs
```diff
@@ -114,18 +116,152 @@ fn merge_overrides_on_default_config(
     })
 }
 
+pub fn evaluate_cohort(
+    dimensions: &HashMap<String, DimensionInfo>,
+    query_data: &Map<String, Value>,
+) -> Result<Map<String, Value>, String> {
```
ayushjain17: also, we need to understand that this logic cannot return errors
it is in the core resolution logic where Errors if at all are to be handled correctly and returned outside

just like how it is being done in core `eval_cac` function
although the type says it has a response of type `Result` but that is just a tech-debt which was never corrected, if you will check the internal workings, there are no places where the function generates an error, and we can safely change the type of it and remove the `Result` from it

the same needs to be achieved here as well

crates/superposition_sdk/src/operation/get_config/_get_config_output.rs
mahatoankitkumar: Yeah after the whole review completes, will push

crates/superposition_core/src/config.rs
mahatoankitkumar: yeah

crates/context_aware_config/src/api/dimension/utils.rs
mahatoankitkumar: DimensionData has position which is being used in recompute and reduce
No need to store position unnecessarily in client so we would need 2 types
can just reuse the `get_dimensions_data` in `fetch_dimensions_map`

ayushjain17: Thats why I was suggesting to combine
I dont find any issue in storing an extra field which barely takes any space

crates/context_aware_config/src/api/context/helpers.rs
mahatoankitkumar: > also, this function would never get called for RemoteCohort as the prior check would pass and would skip the iteration

This was the reason i failed for remote

> but from the perspective of a standalone utility function its behviour would become correct and consistent

Fair point

crates/context_aware_config/src/api/context/helpers.rs
mahatoankitkumar: Din't get you here

crates/cac_client/src/eval.rs
mahatoankitkumar: will discuss

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -617,10 +611,25 @@ async fn reduce_config_key(
...
+            (
+                dim_name.clone(),
+                DimensionInfo {
+                    schema: dim_data.schema.clone(),
+                    position: dim_data.position.clone(),
+                    dimension_type: dim_data.dimension_type.clone(),
+                },
+            )
+        })
+        .collect();
```
ayushjain17: this is not needed, the type is already same
redundant operation

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -202,3 +175,28 @@ fn update_dimensions_in_db(
         .execute(conn)?;
     Ok(())
 }
+
+pub fn fetch_dimensions_map(
```
ayushjain17: ??
```suggestion
pub fn fetch_dimensions_info_map(
```

crates/superposition_core/src/config.rs
```diff
@@ -21,10 +23,154 @@ impl From<&str> for MergeStrategy {
     }
 }
 
+pub fn evaluate_cohort(
```
ayushjain17: this not modified with the latest changes

crates/superposition_types/src/config.rs
ayushjain17: pending

crates/superposition_types/src/config.rs
```diff
@@ -344,12 +347,16 @@ pub struct Config {
     pub contexts: Vec<Context>,
     pub overrides: HashMap<String, Overrides>,
     pub default_configs: Map<String, Value>,
+    #[serde(default)]
+    pub dimensions: HashMap<String, DimensionInfo>,
 }
 
 impl Config {
     pub fn filter_by_dimensions(&self, dimension_data: &Map<String, Value>) -> Self {
+        let modified_context = evaluate_cohort(&self.dimensions, dimension_data);
```
ayushjain17, line 358: did we add a test for this change ?

crates/superposition_types/src/logic.rs
```diff
@@ -37,3 +41,122 @@ pub fn partial_apply(
...
+    query_data: &Map<String, Value>,
+    dimensions: &HashMap<String, DimensionInfo>,
+) -> String {
+    let cohort_based_on_value = &resolve_dimension_value(
+        dimensions,
+        cohort_based_on,
+        query_data,
+        modified_context,
+    );
+    match schema.as_object() {
```
ayushjain17: you dont need so many match  cases no
it can be simply chained
and one last default value as `"otherwise"`

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -383,22 +429,28 @@ async fn list(
...
+    let tags = parse_config_tags(custom_headers.config_tags)?;
+
     let dimension_data: Dimension = dimensions::dsl::dimensions
         .filter(dimensions::dimension.eq(&name))
         .select(Dimension::as_select())
         .schema_name(&schema_name)
         .get_result(&mut conn)?;
 
     let context_ids = get_dimension_usage_context_ids(&name, &mut conn, &schema_name)?;
+    let mut version_id = 0;
```
ayushjain17: this is not needed

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -313,16 +340,35 @@ async fn update(
...
                 .get_result::<Dimension>(transaction_conn)
-                .map_err(|err| db_error!(err))
+                .map_err(|err| db_error!(err))?;
+
+            let workspace_settings = get_workspace(&schema_name, transaction_conn)?;
+
+            let is_mandatory = workspace_settings
+                .mandatory_dimensions
+                .unwrap_or_default()
+                .contains(&result.dimension);
```
ayushjain17, line 350: I was suggesting to move this pre-transaction

crates/context_aware_config/src/api/dimension/utils.rs
```diff
@@ -13,16 +12,14 @@ use superposition_types::{
...
         },
         schema::{contexts::dsl::contexts, dimensions::dsl::*},
     },
-    result as superposition, Cac, Condition, DBConnection,
+    result as superposition, Cac, Condition, DBConnection, DimensionInfo,
 };
 
-use crate::helpers::DimensionData;
-
 pub fn get_dimensions_data(
```
ayushjain17, line 25: this is not used anymore

mahatoankitkumar: being used at 2 places

crates/context_aware_config/src/api/dimension/handlers.rs
mahatoankitkumar: better to have all db queries in transaction

crates/context_aware_config/src/api/dimension/handlers.rs
mahatoankitkumar: needed for the `put_config_in_redis(version_id, state, &schema_name, &mut conn).await?;` outside the transaction

crates/superposition_types/src/config.rs
mahatoankitkumar: no

crates/superposition_core/src/config.rs
mahatoankitkumar: yeah will update

crates/superposition_types/src/config.rs
```diff
@@ -344,12 +349,16 @@ pub struct Config {
...
```
ayushjain17, line 358: have slight doubts here, might have to be tested

crates/superposition_core/src/config.rs
ayushjain17, line 1: this file has not been updated

crates/superposition_types/src/config.rs
ayushjain17: we should add tests for this

ayushjain17: as well as `evaluate_cohorts` in general

crates/superposition_core/src/config.rs
```diff
@@ -21,10 +24,149 @@ impl From<&str> for MergeStrategy {
...
+                &dimension_info.dependency_graph,
+                dimensions,
+                modified_context,
+            );
+            modified_context.insert(cohort_dimension, cohort_value);
+        }
+    }
+}
+
+pub fn evaluate_cohort(
```
ayushjain17: these functions dont need to be copied
`evaluate_cohort` can be imported from `superposition_types` directly

ayushjain17 (approved):

ayushjain17 (approved):

Datron (approved):

PR #713, opened 2025-09-22
fix: use dimension prefix for context in get and resolve config

ayushjain17:
## Problem
In `get` and `resolve` config api dimension data was being taken as plain query params, leading to collisions with filtering params like `prefix` and `version`

## Solution
1. use `dimension[]` prefix to avoid collision
2. also, changed logic to get `max_created_at` time to use `config_versions` instead of `event_log` table

smithy changes not applicable -  as this is change only in the GET method (smithy uses POST method for these calls)


crates/superposition_types/src/api/config.rs
```diff
@@ -1,7 +1,29 @@
...
+
+#[derive(Deserialize, Serialize, IsEmpty, QueryParam, Default)]
+pub struct ConfigQuery {
+    pub version: Option<String>,
+    #[query_param(skip_if_empty)]
+    pub prefix: Option<CommaSeparatedStringQParams>,
+}
+
+#[derive(Deserialize, Serialize, IsEmpty, QueryParam, Default)]
+pub struct ResolveQuery {
```
Datron: nitpick: 
```suggestion
pub struct ResolveConfigQuery {
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -100,21 +102,24 @@ fn get_config_version_from_workspace(
...
     workspace_context: &WorkspaceContext,
     conn: &mut DBConnection,
 ) -> superposition::Result<Option<i64>> {
-    query_params_map.remove("version").map_or_else(
+    version.as_ref().map_or_else(
         || Ok(get_config_version_from_workspace(workspace_context, conn)),
         |version| {
-            if version == Value::String("latest".to_string()) {
+            if *version == *"latest" {
                 log::info!("latest config request");
```
Datron:
```suggestion
                log::trace!("latest config request");
```
Since we don't want this to log for every request and only need it for debug or tracing cases

PR #714, opened 2025-09-24
build: move to new self hosted github runner for CI checks

Datron:
config.test.ts was renamed to zconfig.test.ts so that it is executed last, with data from all other tests present. Github runners used to run this correctly somehow, but the custom runner is running the tests alphabetically.


knutties:
@Datron - minor nit: commit should be `build:` instead `feat:` - don't want this change to bump up version (might not be a big issue - but a detail to keep track of in general).

.github/workflows/ci_check_pr.yaml
```diff
@@ -41,31 +41,32 @@ env:
...
-
-            # - name: Check git settings
-            #   shell: bash
-            #   run: |
-            #       echo "${{ github.event.pull_request.head.ref }}"
-            #       git log --pretty=oneline --abbrev-commit
-            #       echo "----------------"
-            #       git tag "abc_tag" ${{github.event.pull_request.head.sha}}
-            #       git tag
+            - uses: actions/cache/restore@v4
```
knutties: Needs a name here? `Cache Restore` ?

.github/workflows/ci_check_pr.yaml
```diff
@@ -101,25 +102,37 @@ jobs:
...
         steps:
             - name: Checkout repository
               uses: actions/checkout@v4
               with:
                   fetch-depth: 0
-
-            - uses: Swatinem/rust-cache@v2
+                  
+            - uses: actions/cache/restore@v4
+              id: cache
```
knutties: `cache-restore` is a better choice for `id` ?

.github/workflows/ci_check_pr.yaml
```diff
@@ -180,7 +214,17 @@ jobs:
...
-
+            
+            - uses: actions/cache/save@v4
+              if: github.ref == 'refs/heads/main'
+              with:
+                path: |
+                  ~/.gradle/caches
+                  ~/.gradle/wrapper
+                key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+                restore-keys: |
```
knutties: Why use `restore-keys` in a save action ?

.github/workflows/ci_check_pr.yaml
Datron: @knutties its not mandatory here, it adds some context already
Screenshot 2025-10-06 at 11 28 13 AM

.github/workflows/ci_check_pr.yaml
Datron: This is not used, I can remove it.

.github/workflows/ci_check_pr.yaml
Datron: If the key is not found, restore-keys is used to find the closest match (or restore point) cache to use

mahatoankitkumar (approved):

PR #715, opened 2025-09-26
fix(cohorts): eval logic and update providers

ayushjain17:
# Changes
1. bug fix in local cohort evaluation logic
2. added make command for `uniffi` binding generation automation
3. updated smithy models to use better and verbose types for `schema` - (JSONSchema)
4. added `ExtendedMap` type in rust, to be used across as a newtype wrapper for `Map<String, Value>`
    - primarily used to extended features and behaviors on `Map<String, Value>` type
    - used this type as the type for `JSONSchema` in DB types as well as API types in `default-config`, `dimension` and `type-template`, as JSONSchema is always a map and previously used `Value` type is way too generic
    - usage of `ExtendedMap` type, for schema's type provides better operability with `uniffi`
    - type name open for discussion
5. `frontend` and `context-aware-config` crate has changes to accommodate this type change
6. Providers updated:
    - Java (tested)
    - Python (not-tested)
    - Javascript (not-tested)
7. Restructured config unit tests for better readability


clients/java/openfeature-provider/src/main/kotlin/io/juspay/superposition/openfeature/EvaluationArgs.kt
```diff
@@ -18,6 +20,8 @@ internal class EvaluationArgs {
     val contexts: List<Context>
     // Values in the 2nd Map are serialized json values.
     val overrides: Map<String, Map<String, String>>
+    // Values in the 2nd Map are serialized json values.
```
Datron:
```suggestion

```

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -118,11 +118,8 @@ pub fn validate_dimensions(
                     expected_dimension_name
                 ))?;
 
-                validate_context_jsonschema(
-                    object_key,
-                    &dimension_value,
-                    &dimension_data.schema,
-                )?;
+                let schema_value = Value::Object(dimension_data.schema.deref().clone());
```
Datron, line 121: Isn't the schema already an object? Why objectify it more?

Datron (commented):
Are we replacing the AWS type `Document` with `Object`? Is this to simplify parsing on the provider/SDK end? I would recommend not making the change on the APIs but rather write utils for parsing the `Document` for its shape and extract the underlying JSON. This reduces the number of changes in this PR and mostly importantly doesn't introduce smithy specific code into the repo that will need to be documented on why it is present

crates/context_aware_config/src/api/context/validations.rs
ayushjain17: Map and Value are 2 different types
here `Map` is getting converted to `Value`

Datron: > here Map is getting converted to Value

dimension.schema is already a Value

ayushjain17: dimension.schema is `Map`

PR #717, opened 2025-09-28
fix(cohorts): eval logic and update providers

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


crates/context_aware_config/src/api/type_templates/handlers.rs
Datron, line 1: https://github.com/juspay/superposition/blob/e9a2ec99c498318cb5676a2d18f9f0d5db3bb073/crates/context_aware_config/src/api/type_templates/handlers.rs#L92-L92

Should we set a date after which this is removed, or do we support both ways forever?

PR #719, opened 2025-10-03
fix: cohort resolution in get applicable variants

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/crates/superposition_core/src/ffi.rs
+++ b/crates/superposition_core/src/ffi.rs
@@ -63,6 +63,7 @@ fn ffi_eval_logic(
             .unwrap_or(rand::rng().random_range(0..=99))
             % 100;
         let variants = get_applicable_variants(
+            &dimensions,
             &e_args.experiments,
             &_q,
             toss,
@@ -139,6 +140,7 @@ fn ffi_eval_config_with_reasoning(
 #[uniffi::export]
 fn ffi_get_applicable_variants(
     eargs: ExperimentationArgs,
+    dimensions_info: HashMap<String, DimensionInfo>,
     query_data: HashMap<String, String>,
     prefix: Option<Vec<String>>,
 ) -> Result<Vec<String>, OperationError> {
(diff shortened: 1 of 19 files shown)


crates/superposition_core/src/ffi.rs
```diff
@@ -151,8 +153,14 @@ fn ffi_get_applicable_variants(
         .parse::<i8>()
         .unwrap_or(rand::rng().random_range(0..=99))
         % 100;
-    let r = get_applicable_variants(&eargs.experiments, &_query_data, toss, prefix)
-        .map_err(OperationError::Unexpected)?;
+    let r = get_applicable_variants(
```
mahatoankitkumar, line 156: r?
Better to rename it

examples/experimentation_client_integration_example/src/main.rs
```diff
@@ -42,7 +44,9 @@ async fn get_variants(
         "clientId": client_id,
         "os": platform
     });
-    let variant = state.get_applicable_variant(&contexts, &identifier).await;
+    let variant = state
+        .get_applicable_variant(&HashMap::new(), &contexts, &identifier)
```
mahatoankitkumar, line 48: Better to send a proper dimensions_map

ayushjain17: Not required
They still work

Datron: Its a pass by reference value, rather than create an empty hashmap you could pass the dimensions_map

Datron (approved):

examples/experimentation_client_integration_example/src/main.rs
ayushjain17: fair point to add it, but this is some example file
and I dont even know what all dimensions are used here

if you guys can help with that then it would be much appreciated

mahatoankitkumar (approved):

PR #723, opened 2025-10-10
feat: add provider checks

Datron:
## Problem
Missing checks for provider made it hard to make changes to our FFI, sdk and provider layers to support multiple languages

## Solution
Add provider checks that run in a PR to validate any FFI and provider changes

### DIFF
--- a/clients/javascript/open-feature-provider/configuration-client.ts
+++ b/clients/javascript/open-feature-provider/configuration-client.ts
@@ -165,7 +165,7 @@ export class ConfigurationClient {
             throw new Error(`Failed to fetch configuration: ${errorMessage}`);
         }
     }
-
+    // TODO: defaultValue is taken but not used. Should it be used as a fallback?
     async getAllConfigValue(
         defaultValue: Record<string, any>,
         context: Record<string, any>,
@@ -186,6 +186,7 @@ export class ConfigurationClient {
 
                 const variantIds = await this.getApplicableVariants(
                     experiments,
+                    this.currentConfigData?.dimensions || {},
                     queryData,
                     identifier
                 );
@@ -222,6 +223,7 @@ export class ConfigurationClient {
     // Add method to get applicable variants
     private async getApplicableVariants(
         experiments: Experiment[],
+        dimensions: Record<string, Record<string, any>>,
         queryData: Record<string, any>,
         identifier: string,
         filterPrefixes?: string[]
@@ -229,6 +231,7 @@ export class ConfigurationClient {
         // This would use the native resolver's getApplicableVariants method
         return this.resolver.getApplicableVariants(
             experiments,
+            dimensions,
             queryData,
             identifier,
             filterPrefixes || []
--- a/clients/python/sdk/README.md
+++ b/clients/python/sdk/README.md
@@ -1,3 +1,36 @@
-## Superposition Client
+# Superposition SDK
 
-superposition_sdk client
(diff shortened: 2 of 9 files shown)


clients/javascript/open-feature-provider/configuration-client.ts
```diff
@@ -186,6 +186,7 @@ export class ConfigurationClient {
 
                 const variantIds = await this.getApplicableVariants(
                     experiments,
+                    this.currentConfigData?.dimensions || {},
```
sauraww, line 189: Should we change the dimensions argument type to : 
 dimensions?: Record<string, Record<string, any>>
inside getApplicableVariants ?
Also , should we also move getApplicableVariants to experiments client ?

clients/python/provider-sdk-tests/pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+license = { text = "Apache-2.0" }
+
+dependencies = [
+    "superposition-provider",
+    "superposition-sdk",
+    "openfeature-sdk==0.8.3"
+]
+
+[tool.uv.sources]
+superposition-provider = { path = "../provider", editable = true }
```
sauraww, line 19: Can we use uv workspaces ? 
If the relative path changes , this might break.

clients/javascript/open-feature-provider/configuration-client.ts
Datron: > Should we change the dimensions argument type to : 
   dimensions?: Record<string, Record<string, any>>
inside getApplicableVariants ?

At some point you'll need to handle the empty case since the resolver expects this field. Better to keep the interface to applicable variants clean and let the calling function decide what it wants to pass

Datron: > Also , should we also move getApplicableVariants to experiments client ?

Yeah that makes sense, good call. I'll move it

clients/python/provider-sdk-tests/pyproject.toml
Datron: A number of changes will be required, this can come a bit later. We are not changing the relative path right now and provider-sdk-tests is run only in our CI

clients/javascript/open-feature-provider/configuration-client.ts
Datron: > Yeah that makes sense, good call. I'll move it

looked into it some more, and the native resolver is not present in the experimentation client, so this needs a lot more changes before being moved. Lets move this once tests are present for experimentation so that the change can be validated to not break.

sauraww: Sure @Datron , yeah , we can do it later on. Works!!

clients/python/provider-sdk-tests/pyproject.toml
sauraww: Can you rebase ?

clients/python/sdk/README.md
ayushjain17, line 1: How did this file reverted last time ?
there is a restore step in the make command to prevent this file from getting overwritten

clients/python/provider-sdk-tests/pyproject.toml
ayushjain17: what is `uv` ?

Datron: uv is a python build system written in rust, its like cargo for python

clients/python/sdk/README.md
Datron: Not sure how, but saw that this was gone and reverted the file

PR #724, opened 2025-10-14
feat(frontend): better UI for cohort schema

ayushjain17:


crates/frontend/src/bindings/sortable.rs
```diff
@@ -0,0 +1,208 @@
...
+
+        if let Some(ref chosen_class) = self.chosen_class {
+            js_sys::Reflect::set(&obj, &"chosenClass".into(), &chosen_class.into())?;
+        }
+
+        Ok(obj.into())
+    }
+}
+
+///  ```
```
sauraww: Can we remove this ?

crates/frontend/src/bindings/sortable.rs
```diff
@@ -0,0 +1,208 @@
...
+        self.ghost_class = Some(class.into());
+        self
+    }
+
+    pub fn chosen_class(mut self, class: impl Into<String>) -> Self {
+        self.chosen_class = Some(class.into());
+        self
+    }
+
+    /// Convert to JavaScript object for SortableJS
```
sauraww, line 82: Instead of manual serialization , Can we use serde with field renaming attributes: 
#[derive(Default, Clone, Debug, Serialize)]
pub struct SortableConfig {
    #[serde(skip_serializing_if = "Option::is_none")]
    pub animation: Option<u32>,
    
    #[serde(skip_serializing_if = "Option::is_none")]
    pub handle: Option<String>,
    
    #[serde(skip_serializing_if = "Option::is_none")]
    pub disabled: Option<bool>,
    
    #[serde(skip_serializing_if = "Option::is_none")]
    pub group: Option<String>,
    
    #[serde(rename = "ghostClass", skip_serializing_if = "Option::is_none")]
    pub ghost_class: Option<String>,
    
    #[serde(rename = "chosenClass", skip_serializing_if = "Option::is_none")]
    pub chosen_class: Option<String>,
}

crates/frontend/src/bindings/sortable.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn to_js_value(&self) -> Result<JsValue, JsValue> {
```
sauraww, line 83: you can then just do : 
pub fn to_js_value(&self) -> Result<JsValue, JsValue> {
    serde_wasm_bindgen::to_value(self)
        .map_err(|e| JsValue::from_str(&e.to_string()))
}

crates/frontend/src/bindings/sortable.rs
```diff
@@ -0,0 +1,208 @@
...
+            &options,
+            &"onEnd".into(),
+            callback.as_ref().unchecked_ref(),
+        )?;
+
+        // Create the Sortable instance using bindings
+        let sortable = Sortable::new(element, &options);
+
+        // Keep callback alive
+        callback.forget();
```
sauraww, line 182: This is a problem. This keeps callback alive. It doesn't cleanup when it goes out of scope. The callback becomes permanently allocated in memory. 
The memory will never be freed - it's a deliberate memory leak. 

Each creation leaks memory permanently.

In an SPA, when you navigate between pages:
1. Leptos unmounts the old component
2. Leptos mounts the new component
3. Rust variables go out of scope and get cleaned up
BUT the leaked callbacks remain in memory!

After 1 hour of usage:
User might have navigated 50+ times
50+ callbacks leaked in memory
Each callback holds references to closures, event handlers, etc.
Memory usage keeps growing and never goes down
In a long running SPA , it can leak 100+ callbacks .

crates/frontend/src/bindings/sortable.rs
ayushjain17: this is not serialisation over here
this is setting values inside an object which exists in JS land

and all of these operations are also happening in JS and not in rust

sauraww: I know that @ayushjain17 , it's not traditional serialization but that's the problem I am trying to solve : 
**The current approach**: 

let obj = Object::new();
if let Some(animation) = self.animation {
    js_sys::Reflect::set(&obj, &"animation".into(), &animation.into())?;
}
if let Some(ref ghost_class) = self.ghost_class {
    js_sys::Reflect::set(&obj, &"ghostClass".into(), &ghost_class.into())?;
}
// ... 15 more manual property sets

**Serde Approach**: 
#[derive(Serialize)]
pub struct SortableConfig {
    #[serde(rename = "ghostClass", skip_serializing_if = "Option::is_none")]
    pub ghost_class: Option<String>,
}

pub fn to_js_value(&self) -> Result<JsValue, JsValue> {
    serde_wasm_bindgen::to_value(self)
        .map_err(|e| JsValue::from_str(&e.to_string()))
}

What serde_wasm_bindgen::to_value() does under the hood:

Serializes the Rust struct to an intermediate representation
Converts that representation to a JavaScript object
Applies the serde attributes (rename, skip_serializing_if)
Returns a JsValue that is a JavaScript object
So yes, the end result is still 'setting values inside an object in JS land,' but:

**Benefits of the Serde Approach**:
Type Safety: Field names checked at compile time
Maintainability: Adding new fields is just adding serde attributes

ayushjain17: the thing which you are saying is same only in terms of end result
performance wise it is very bad as you are involving serialization and deserialization, and this process is very costly and not something which should be favoured in general

sauraww: That is correct. I missed the performance part , I was mostly focussed on maintainability . serde would be much more expensive.

crates/frontend/src/bindings/sortable.rs
ayushjain17: it needs to be leaked for it to work, else it will not work

on top of it, this would only happen when one opens to create the create/edit dimensions page
someone opening the page for 1 hour and not reloading at all is something which is entirely unlikely

that too the navigation should specifically involve selecting the dimension type as LocalCohort and then choose a based on dependent dimension or in edit screen only for LocalCohorts the component would be mounted

now this memory leak is an issue only in platforms which do not support weak references, otherwise this memory would get reclaimed when the JS gc would run

ayushjain17: I did a memory utilisation anaysis as well,  if at all, I dont think the leak is significant

PR #726, opened 2025-10-16
fix: experiment groups provider changes

mahatoankitkumar:
## Problem
This PR implements experiment groups support across the codebase. The primary goal is to add a new bucketing mechanism using experiment groups instead of the legacy random toss-based variant selection. The changes include updating FFI interfaces, adding experiment group data structures, modifying variant calculation logic, and updating client implementations across multiple languages.

Key Changes:

Replaced toss (random percentage) with deterministic identifier-based bucketing using hash functions
Added ExperimentGroups data structures and API endpoints for fetching/caching experiment groups
Updated get_applicable_variants function signature to accept experiment_groups and identifier parameters
Modified all client bindings (Python, JavaScript, Java/Kotlin, Haskell) to support the new experiment groups functionality

### DIFF
--- a/clients/haskell/hs-exp-client/src/Main.hs
+++ b/clients/haskell/hs-exp-client/src/Main.hs
@@ -25,8 +25,8 @@ main = do
         loopNTimes n client = do
             runningExperiments   <- getRunningExperiments client
             satisfiedExperiments <- getSatisfiedExperiments client "{\"os\": \"android\", \"client\": \"1mg\"}" Nothing
-            filteredExperiments <- getFilteredSatisfiedExperiments client (Just "{\"os\": \"android\"}") (Just "hyperpay")
-            variants             <- getApplicableVariants client "{}" "{\"os\": \"android\", \"client\": \"1mg\"}" "1mg-android"
+            filteredExperiments  <- getFilteredSatisfiedExperiments client (Just "{\"os\": \"android\"}") (Just "hyperpay")
+            variants             <- getApplicableVariants client "{}" "{\"os\": \"android\", \"client\": \"1mg\"}" "1mg-android" Nothing
             print "Running experiments"
             print runningExperiments
             print "experiments that satisfy context"
--- a/clients/java/openfeature-provider/src/main/kotlin/io/juspay/superposition/openfeature/EvaluationArgs.kt
+++ b/clients/java/openfeature-provider/src/main/kotlin/io/juspay/superposition/openfeature/EvaluationArgs.kt
@@ -5,12 +5,15 @@ import dev.openfeature.sdk.EvaluationContext
 import dev.openfeature.sdk.Value
 import io.juspay.superposition.model.ContextPartial
 import io.juspay.superposition.model.ExperimentResponse
+import io.juspay.superposition.model.ExperimentGroupResponse
 import io.juspay.superposition.model.GetConfigOutput
 import software.amazon.smithy.java.core.serde.document.Document
 import uniffi.superposition_client.*
+import uniffi.superposition_types.Buckets
 import uniffi.superposition_types.Context
 import uniffi.superposition_types.DimensionInfo
 import uniffi.superposition_types.DimensionType
+import uniffi.superposition_types.GroupType
 import uniffi.superposition_types.Variant
 import uniffi.superposition_types.VariantType
 
@@ -82,6 +85,25 @@ internal class EvaluationArgs {
                 er.trafficPercentage().toUByte(),
                 variants,
                 serializeDocumentValues(er.context())
+            )
+        }
+
+        private fun toFfiGroupType(gt: io.juspay.superposition.model.GroupType): GroupType {
+            return when (gt) {
+                io.juspay.superposition.model.GroupType.USER_CREATED -> GroupType.USER_CREATED
+                else -> GroupType.SYSTEM_GENERATED
+            }
+        }
+
+        @JvmStatic
+        fun toFfiExperimentGroup(er: ExperimentGroupResponse): FfiExperimentGroup {
+            return FfiExperimentGroup(
+                er.id(),
+                serializeDocumentValues(er.context()),
+                er.trafficPercentage().toUByte(),
+                er.memberExperimentIds(),
+                toFfiGroupType(er.groupType()),
+                er.buckets() as Buckets
             )
         }
     }
--- a/clients/javascript/bindings/native-resolver.ts
+++ b/clients/javascript/bindings/native-resolver.ts
@@ -239,6 +239,7 @@ export class NativeResolver {
 
     getApplicableVariants(
         experiments: any[],
+        experiment_groups: any[],
         dimensions: Record<string, Record<string, any>>,
         userContext: Record<string, any>,
         identifier: string,
@@ -258,19 +259,22 @@ export class NativeResolver {
         }
 
         const experimentsJson = JSON.stringify(experiments);
+        const experimentGroupsJson = JSON.stringify(experiment_groups);
         const userContextJson = JSON.stringify(userContext);
         const dimensionsJson = JSON.stringify(dimensions);
         const filterPrefixesJson =
             filterPrefixes.length > 0 ? JSON.stringify(filterPrefixes) : null;
 
         console.log("Calling FFI getApplicableVariants with parameters:");
-        console.log("  experiments:", experiments);
+        console.log("  experiments:", experiments.length);
+        console.log("  experimentGroups:", experiment_groups.length);
         console.log("  userContext:", userContext);
         console.log("  identifier:", identifier);
         console.log("  filterPrefixes:", filterPrefixes);
 
         const result = this.lib.core_get_applicable_variants(
             experimentsJson,
+            experimentGroupsJson,
             dimensionsJson,
             userContextJson,
             identifier,
--- a/clients/python/provider/superposition_provider/exp_config.py
+++ b/clients/python/provider/superposition_provider/exp_config.py
@@ -2,9 +2,9 @@ import json
 import logging
 from decimal import Decimal
 from typing import Any, Dict, Optional, TypeVar, Generic
-from superposition_bindings.superposition_client import FfiExperiment
+from superposition_bindings.superposition_client import FfiExperiment, FfiExperimentGroup
 from .types import OnDemandStrategy, PollingStrategy, SuperpositionOptions, ConfigurationOptions, ExperimentationOptions
-from superposition_sdk.client import Superposition, Config, ListExperimentInput
+from superposition_sdk.client import Superposition, Config, ListExperimentInput, ListExperimentGroupsInput
 import asyncio
 from datetime import datetime, timedelta
 from superposition_bindings.superposition_types import Variant, VariantType
@@ -73,6 +73,7 @@ class ExperimentationConfig():
         self.superposition_options = superposition_options
         self.options = experiment_options
         self.cached_experiments = None
+        self.cached_experiment_groups = None
         self.last_updated = None
         self.evaluation_cache: Dict[str, Dict[str, Any]] = {}
         self._polling_task = None
@@ -82,10 +83,12 @@ class ExperimentationConfig():
             while True:
                 try:
                     latest_exp_list = await self._get_experiments(self.superposition_options)
-                    if latest_exp_list is not None:
+                    latest_exp_grp_list = await self._get_experiment_groups(self.superposition_options)
+                    if latest_exp_list is not None and latest_exp_grp_list is not None:
                         self.cached_experiments = latest_exp_list
+                        self.cached_experiment_groups = latest_exp_grp_list
                         self.last_updated = datetime.utcnow()
-                        logger.info("Experiment List fetched successfully.")
+                        logger.info("Experiment List and Experiment Group List updated successfully.")
 
                 except Exception as e:
                     logger.error(f"Polling error: {e}")
@@ -93,11 +96,14 @@ class ExperimentationConfig():
                 await asyncio.sleep(interval)
 
         latest_exp_list = await self._get_experiments(self.superposition_options)
+        latest_exp_grp_list = await self._get_experiment_groups(self.superposition_options)
         logger.info("response from experimentation: " + str(latest_exp_list))
-        if latest_exp_list is not None:
+        logger.info("response from experimentation groups: " + str(latest_exp_grp_list))
+        if latest_exp_list is not None and latest_exp_grp_list is not None:
             self.cached_experiments = latest_exp_list
+            self.cached_experiment_groups = latest_exp_grp_list
             self.last_updated = datetime.utcnow()
-            logger.info("Experiment List fetched successfully.")
+            logger.info("Experiment List and Experiment Group List fetched successfully.")
         
         match self.options.refresh_strategy:
             case PollingStrategy(interval=interval, timeout=timeout):
@@ -185,6 +191,67 @@ class ExperimentationConfig():
             # Log the error and return empty config as fallback
             logger.error(f"Error fetching config from Superposition: {e}")
             return None
+        
+    @staticmethod
+    async def _get_experiment_groups(superposition_options: SuperpositionOptions) -> Optional[Dict[str, Any]]:
+        """
+        Fetch configuration from Superposition service using the generated Python SDK.
+        
+        Args:
+            superposition_options: Options containing endpoint, token, org_id, workspace_id
+            
+        Returns:
+            Dict containing the configuration data
+        """
+        try:
+            # Create SDK config
+            sdk_config = Config(
+                endpoint_uri=superposition_options.endpoint
+            )
+            
+            # Create Superposition client
+            client = Superposition(config=sdk_config)
+            
+           
+
+            list_exp_grp_input = ListExperimentGroupsInput(
+                workspace_id=superposition_options.workspace_id,
+                org_id=superposition_options.org_id,
+                all=True
+            )
+            
+
+            response = await client.list_experiment_groups(list_exp_grp_input)
+           
+            exp_grp_list = response.data
+            logger.info(f"Fetched {len(exp_grp_list)} experiment groups from Superposition")
+            trimmed_exp_grp_list = []
+            for exp_gr in exp_grp_list:
+                condition = {}
+                for key, value in exp_gr.context.items():
+                    condition[key] = json.dumps(document_to_python_value(value))
+                 
+                    
+                trimmed_exp_grp = FfiExperimentGroup(
+                    id=exp_gr.id,
+                    context=condition,
+                    member_experiment_ids=exp_gr.member_experiment_ids,
+                    traffic_percentage=exp_gr.traffic_percentage,
+                    group_type=exp_gr.group_type,
+                    buckets=exp_gr.buckets
+                )
+
+
+                trimmed_exp_grp_list.append(trimmed_exp_grp)
+                
+            
+            return trimmed_exp_grp_list
+            
+        except Exception as e:
+            # Log the error and return empty config as fallback
+            logger.error(f"Error fetching config from Superposition: {e}")
+            return None
+    
     
     async def on_demand_config(self, ttl, use_stale) -> dict:
         """Get config on-demand based on TTL or fall back to stale if needed."""
@@ -198,10 +265,12 @@ class ExperimentationConfig():
             try:
                 logger.debug("TTL expired. Fetching config on-demand.")
                 latest_exp_list = await self._get_experiments(self.superposition_options)
+                latest_exp_grp_list = await self._get_experiment_groups(self.superposition_options)
 
-                if latest_exp_list is not None:
-                    logger.info("Experiment List fetched successfully.")
+                if latest_exp_list is not None and latest_exp_grp_list is not None:
+                    logger.info("Experiment List and Experiment Group List updated successfully on-demand.")
                     self.cached_experiments = latest_exp_list
+                    self.cached_experiment_groups = latest_exp_grp_list
                     self.last_updated = datetime.utcnow()
                 
             except Exception as e:
--- a/crates/superposition_core/src/experiment.rs
+++ b/crates/superposition_core/src/experiment.rs
@@ -1,9 +1,10 @@
 use std::collections::{HashMap, HashSet};
+use std::hash::{DefaultHasher, Hash, Hasher};
 
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
 use superposition_types::database::models::experimentation::{
-    Variant, VariantType, Variants,
+    Bucket, Buckets, GroupType, Variant, Variants,
 };
 use superposition_types::{logic::evaluate_cohort, Condition, DimensionInfo, Overridden};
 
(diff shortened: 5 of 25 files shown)


crates/superposition_core/src/experiment.rs
```diff
@@ -30,32 +31,148 @@ pub struct FfiExperiment {
...
+                            || jsonlogic::apply(
+                                &Value::Object(exp.context.clone().into()),
+                                context,
+                            ) == Ok(Value::Bool(true));
+                    } else {
+                        let valid_context = superposition_types::apply(
+                            &exp.context,
+                            &context.as_object().cloned().unwrap_or_default(),
+                        );
+                    }
```
Datron, line 165: Do we need this cfg_if provision in the provider? I think we can discard this, since we're moving forward with Map anyway

crates/superposition_core/src/experiment.rs
```diff
@@ -30,32 +31,148 @@ pub struct FfiExperiment {
...
+                    let valid_context = exp_context.is_empty()
+                        || jsonlogic::apply(&Value::Object(exp_context.clone().into()), context)
+                            == Ok(Value::Bool(true));
+                } else {
+                    let valid_context = superposition_types::apply(
+                        exp_context,
+                        &context.as_object().cloned().unwrap_or_default(),
+                    );
+                }
+            }
```
Datron, line 116: Do we need cfg_if here to check for jsonlogic? Provider is mostly used by new guys who will be on the simple map setup

crates/superposition_provider/src/utils.rs
```diff
@@ -425,6 +426,66 @@ impl ConversionUtils {
         Ok(trimmed_exp_list)
     }
 
+    pub fn convert_experiment_groups_response(
+        response: &superposition_sdk::operation::list_experiment_groups::ListExperimentGroupsOutput,
```
Datron: Can we just import ListExperimentGroupsOutput

crates/superposition_provider/src/utils.rs
```diff
@@ -425,6 +426,66 @@ impl ConversionUtils {
...
+            let context = Exp::<Condition>::try_from(condition_map)
+                .map_err(|e| {
+                    SuperpositionError::SerializationError(format!(
+                        "Invalid condition: {}",
+                        e
+                    ))
+                })?
+                .into_inner();
+            let group_type = match exp_group.group_type {
+                superposition_sdk::types::GroupType::SystemGenerated => {
```
Datron: nitpick: Lets import SystemGenerated

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -503,3 +505,10 @@ impl ToSql<Array<Nullable<Json>>, Pg> for Buckets {
         )
     }
 }
+
+uniffi::custom_type!(Buckets, Vec<Option<Bucket>>, {
+    try_lift: |val: Vec<Option<Bucket>>| {
+        Buckets::try_from(val).map_err(|e| anyhow::anyhow!(e))
+    },
+    lower: |obj: Buckets| obj.0.to_vec(),
+});
```
Datron, line 514: What does this do?

clients/python/provider/superposition_provider/exp_config.py
```diff
@@ -185,6 +192,67 @@ async def _get_experiments(superposition_options: SuperpositionOptions) -> Optio
...
+
+            response = await client.list_experiment_groups(list_exp_grp_input)
+           
+            exp_grp_list = response.data
+            logger.info(f"Fetched {len(exp_grp_list)} experiment groups from Superposition")
+            trimmed_exp_grp_list = []
+            for exp_gr in exp_grp_list:
+                condition = {}
+                for key, value in exp_gr.context.items():
+                    condition[key] = json.dumps(document_to_python_value(value))
```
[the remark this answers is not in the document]
Datron: Do we want to stringify the response of document_to_python_value?

clients/haskell/hs-exp-client/src/Main.hs
```diff
@@ -25,8 +25,8 @@ main = do
         loopNTimes n client = do
             runningExperiments   <- getRunningExperiments client
             satisfiedExperiments <- getSatisfiedExperiments client "{\"os\": \"android\", \"client\": \"1mg\"}" Nothing
-            filteredExperiments <- getFilteredSatisfiedExperiments client (Just "{\"os\": \"android\"}") (Just "hyperpay")
-            variants             <- getApplicableVariants client "{}" "{\"os\": \"android\", \"client\": \"1mg\"}" "1mg-android"
+            filteredExperiments  <- getFilteredSatisfiedExperiments client (Just "{\"os\": \"android\"}") (Just "hyperpay")
```
Datron, line 28: nitpick: remove the additional space

crates/superposition_provider/src/client.rs
```diff
@@ -472,28 +551,35 @@ impl ExperimentationConfig {
...
+            cached_experiment_groups.as_ref(),
+        ) {
+            (Some(experiments), Some(experiment_groups)) => {
+                // Use get_applicable_variants from superposition_core
+                get_applicable_variants(
+                    dimensions_info,
+                    experiments,
+                    experiment_groups,
+                    contexts,
+                    &identifier.unwrap_or_default(),
```
ayushjain17, line 586: should we use a default value here ?
it would be `""` every time, not sure if this is going to have any implications over her

crates/superposition_provider/src/client.rs
```diff
@@ -447,11 +481,56 @@ impl ExperimentationConfig {
...
+    ) -> Result<Option<ExperimentGroups>> {
+        use superposition_sdk::{Client, Config as SdkConfig};
+
+        info!("Fetching experiment groups from Superposition service using SDK");
+
+        // Create SDK config
+        let sdk_config = SdkConfig::builder()
+            .endpoint_url(&options.endpoint)
+            .bearer_token(options.token.clone().into())
+            .build();
```
ayushjain17, line 511: based on @Datron currently opened PR for provider fixes, I think you need to add `.behaviour_version_latest()` over here
please verify once

crates/superposition_provider/src/client.rs
```diff
@@ -447,11 +481,56 @@ impl ExperimentationConfig {
...
+
+        // Create Superposition client
+        let client = Client::from_conf(sdk_config);
+
+        let response = client
+            .list_experiment_groups()
```
ayushjain17, line 517: @mahatoankitkumar 
should we consider making changes for [this issue](https://github.com/juspay/superposition/issues/640) as well in this PR ? or a separate for this as well ?

crates/superposition_provider/src/client.rs
```diff
@@ -331,24 +349,33 @@ impl ExperimentationConfig {
...
 
         tokio::spawn(async move {
             loop {
-                match Self::get_experiments_static(&superposition_options).await {
-                    Ok(Some(experiments)) => {
+                match (
+                    Self::get_experiments_static(&superposition_options).await,
+                    Self::get_experiment_groups_static(&superposition_options).await,
+                ) {
+                    (Ok(Some(experiments)), Ok(Some(experiment_groups))) => {
```
ayushjain17: why are we clubbing them together ?
let them have their own separate blocks

why store error for the one which has passed ?

ayushjain17: can we, re-write this to run both in parallel ?

crates/superposition_core/src/experiment.rs
```diff
@@ -30,32 +31,148 @@ pub struct FfiExperiment {
...
+            log::info!(
+                "Identifier: {}, Experiment Group ID: {}, Hashed Percentage: {}",
+                identifier,
+                exp_group.id,
+                hashed_percentage
+            );
+            let exp_context = &exp_group.context;
+
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "jsonlogic")] {
```
ayushjain17, line 106: Cargo.toml needs to be updated for this

crates/superposition_core/src/experiment.rs
ayushjain17: @Datron its the same code which is present in experimentation_client
lets keep the code consistent everywhere, then we can remove this from all the places at once when all the current blockers are resolved

Datron: @ayushjain17 the context is different here. The providers are not currently used by anyone and hence don't need this code. We can remove it.

crates/superposition_provider/src/utils.rs
```diff
@@ -425,6 +426,66 @@ impl ConversionUtils {
...
+                member_experiment_ids: exp_group.member_experiment_ids().to_vec(),
+                group_type,
+                buckets: Buckets::from(std::array::from_fn(|i| {
+                    exp_group.buckets.get(i).and_then(|b| {
+                        b.as_ref().map(|bucket| Bucket {
+                            variant_id: bucket.variant_id.clone(),
+                            experiment_id: bucket.experiment_id.clone(),
+                        })
+                    })
+                })),
```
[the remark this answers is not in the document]
ayushjain17: should we do something like this ?
```suggestion
                buckets: Buckets::try_from(
                    exp_group
                        .buckets
                        .iter()
                        .map(|b| {
                            b.as_ref().map(|bucket| Bucket {
                                variant_id: bucket.variant_id.clone(),
                                experiment_id: bucket.experiment_id.clone(),
                            })
                        })
                        .collect::<Vec<_>>(),
                )
                .map_err(SuperpositionError::SerializationError)?,
```

crates/superposition_provider/src/utils.rs
```diff
@@ -425,6 +426,66 @@ impl ConversionUtils {
...
+        let group_list = response.data();
+        let mut trimmed_group_list: ExperimentGroups = Vec::new();
+
+        for exp_group in group_list {
+            // Convert experiment context (condition)
+            let mut condition_map = Map::new();
+            for (key, doc) in exp_group.context() {
+                let value = Self::document_to_value(doc)?;
+                condition_map.insert(key.clone(), value);
+            }
```
ayushjain17: this format is repeated multiple times
can we make a function for this

which is converting `HashMap<String, Document>` to `Map<String, Value>`

clients/python/provider/superposition_provider/exp_config.py
```diff
@@ -82,22 +84,27 @@ async def poll_config(interval: int, timeout: int) -> None:
...
                 try:
                     latest_exp_list = await self._get_experiments(self.superposition_options)
-                    if latest_exp_list is not None:
+                    latest_exp_grp_list = await self._get_experiment_groups(self.superposition_options)
+                    if latest_exp_list is not None and latest_exp_grp_list is not None:
                         self.cached_experiments = latest_exp_list
+                        self.cached_experiment_groups = latest_exp_grp_list
                         self.last_updated = datetime.utcnow()
-                        logger.info("Experiment List fetched successfully.")
+                        logger.info("Experiment List and Experiment Group List updated successfully.")
```
ayushjain17, line 91: should we break them into individual if checks ?

clients/python/provider/superposition_provider/exp_config.py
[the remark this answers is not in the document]
ayushjain17: @Datron yeah we need to do this, on rust side the type is defined as `HashMap<String, String>`

had raised this issue earlier, we need to solve this at some point

crates/superposition_core/src/experiment.rs
```diff
@@ -30,32 +31,148 @@ pub struct FfiExperiment {
...
+            )
+            .flatten()
+            .and_then(|b| {
+                if exp_group.group_type == GroupType::SystemGenerated {
+                    Some((hashed_percentage, b))
+                } else {
+                    (exp_group.traffic_percentage > 0).then_some((
+                        (hashed_percentage * 100) / exp_group.traffic_percentage as usize,
+                        b,
+                    ))
```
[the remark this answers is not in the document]
ayushjain17: @mahatoankitkumar this is a valid point raised by @Copilot, can we update this in both the places

clients/javascript/bindings/native-resolver.ts
```diff
@@ -258,19 +259,22 @@ export class NativeResolver {
...
         const dimensionsJson = JSON.stringify(dimensions);
         const filterPrefixesJson =
             filterPrefixes.length > 0 ? JSON.stringify(filterPrefixes) : null;
 
         console.log("🔧 Calling FFI getApplicableVariants with parameters:");
         console.log("  experiments:", experiments);
+        console.log("  experimentGroups:", experiment_groups);
         console.log("  userContext:", userContext);
         console.log("  identifier:", identifier);
         console.log("  filterPrefixes:", filterPrefixes);
```
ayushjain17: can we update these logs, to the length of the arrays being passed instead of the entire data

ayushjain17 (commented):
overall review is done, need do just re-check once from the correct working point of view

crates/superposition_provider/src/client.rs
mahatoankitkumar: separate should be better.

crates/superposition_provider/src/client.rs
mahatoankitkumar: I think it should be clubbed together, as failure of one will impact the other
Both should pass in order to work properly.

clients/python/provider/superposition_provider/exp_config.py
mahatoankitkumar: https://github.com/juspay/superposition/pull/726#discussion_r2458920110

clients/javascript/bindings/native-resolver.ts
mahatoankitkumar: Would just logging the length be useful?

crates/superposition_core/src/experiment.rs
mahatoankitkumar: I think we should cleanup at the same time we do for the whole repo, for consistency.

crates/superposition_types/src/database/models/experimentation.rs
mahatoankitkumar: Not needed i guess now, will verify and remove

crates/superposition_provider/src/client.rs
mahatoankitkumar: Lets keep this way, it better than hardcoding with "", anyways `unwrap_or_default()` will do the same.

crates/superposition_types/src/database/models/experimentation.rs
ayushjain17: @mahatoankitkumar its needed

crates/superposition_provider/src/utils.rs
mahatoankitkumar: Two similar types exist so have imported one and explicitly qualified one just to avoid confusion

crates/superposition_types/src/database/models/experimentation.rs
mahatoankitkumar: Not needed i guess as i am explicitly converting the type by traversing the Buckets

clients/haskell/hs-exp-client/src/Main.hs
mahatoankitkumar: Kept the alignment in sync with the above lines.

crates/superposition_types/src/database/models/experimentation.rs
mahatoankitkumar: yeah sorry, its needed

crates/superposition_core/src/experiment.rs
Datron: No why are we increasing work later? This is not even needed, no one using this provider version can use the jsonlogic deployment. We should look forward and remove this now.

crates/superposition_provider/src/client.rs
ayushjain17: I am not saying to hardcode `""`, rather I am asking, should we handle this explicitly instead of defaulting to `""`

crates/superposition_core/src/experiment.rs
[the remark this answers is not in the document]
ayushjain17: @mahatoankitkumar change for this has not been made in `experimentation_client` crate

crates/superposition_provider/src/utils.rs
```diff
@@ -351,6 +353,18 @@ impl ConversionUtils {
             dimensions,
         })
     }
+    
+    fn convert_condition_document(
```
ayushjain17, line 343: there are 2 more places where this can be used

clients/javascript/bindings/native-resolver.ts
ayushjain17: this log anyways meaningless, it would log the entire data which is too huge for anyone to see and make meaning out of it

crates/superposition_provider/src/client.rs
ayushjain17: that aside, what about making the calls parallel, using `futures::join` macro ?

crates/superposition_core/src/experiment.rs
```diff
@@ -30,32 +31,150 @@ pub struct FfiExperiment {
     pub context: Condition,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, uniffi::Record)]
+pub struct FfiExperimentGroup {
```
ayushjain17, line 35: ideally these types should be unified

mahatoankitkumar: separate PR

crates/superposition_provider/src/client.rs
mahatoankitkumar: what should we default to `default`?

crates/superposition_provider/src/provider.rs
```diff
@@ -102,11 +102,7 @@ impl SuperpositionProvider {
         let dimensions_info = self.get_dimensions_info().await;
         let variant_ids = if let Some(exp_config) = &self.exp_config {
             exp_config
-                .get_applicable_variants(
-                    &dimensions_info,
-                    &context,
-                    targeting_key.unwrap_or_default().parse::<i8>().ok(),
```
ayushjain17, line 146: i think change for this has not been made in the other providers, can you please check this once

crates/superposition_provider/src/utils.rs
Datron: Use an alias for the import?

crates/superposition_provider/src/provider.rs
mahatoankitkumar: didn't find any

crates/superposition_core/src/ffi.rs
```diff
@@ -148,16 +144,13 @@ fn ffi_get_applicable_variants(
         .map_err(|err| OperationError::Unexpected(err.to_string()))?;
 
     // TODO Migrate to new bucketing procedure.
```
ayushjain17: I think this TODO can be removed now

crates/superposition_provider/src/utils.rs
```diff
@@ -351,6 +337,21 @@ impl ConversionUtils {
             dimensions,
         })
     }
+
+    fn convert_condition_document(
+        context: &::std::collections::HashMap<
+            ::std::string::String,
+            ::aws_smithy_types::Document,
+        >,
```
ayushjain17: ?
```suggestion
        context: &HashMap<String, aws_smithy_types::Document>,
```

mahatoankitkumar: make fmt did this

ayushjain17: fmt changed `String` to `::std::string::String` ?

crates/superposition_core/src/experiment.rs
ayushjain17: can we create an issue or something to track it

clients/java/openfeature-provider/src/main/kotlin/io/juspay/superposition/openfeature/EvaluationArgs.kt
```diff
@@ -84,6 +87,22 @@ internal class EvaluationArgs {
...
+            return FfiExperimentGroup(
+                er.id(),
+                serializeDocumentValues(er.context()),
+                er.trafficPercentage().toUByte(),
+                er.memberExperimentIds(),
+                when (er.groupType()) {
+                    io.juspay.superposition.model.GroupType.USER_CREATED -> GroupType.USER_CREATED
+                    io.juspay.superposition.model.GroupType.SYSTEM_GENERATED -> GroupType.SYSTEM_GENERATED
+                    else -> throw IllegalArgumentException("Unknown group type: ${er.groupType()}")
+                },
```
ayushjain17: can we have this as a separate function ?
and I dont think we need to throw an error here, defaulting to a value should be fine

just like we did for `DimensionType`

crates/superposition_core/src/experiment.rs
```diff
@@ -30,32 +31,150 @@ pub struct FfiExperiment {
...
+                if exp_group.group_type == GroupType::SystemGenerated {
+                    Some((hashed_percentage, b))
+                } else if exp_group.traffic_percentage > 0 {
+                    Some((
+                        (hashed_percentage * 100) / exp_group.traffic_percentage as usize,
+                        b,
+                    ))
+                } else {
+                    None
+                }
```
ayushjain17, line 138: the if condition change which was made here, was not updated in experimentation_client crate, please update there as well

crates/superposition_provider/src/client.rs
ayushjain17: please raise an issue or something to track it

crates/superposition_provider/src/client.rs
ayushjain17: i am not sure, but want to understand that whether `""` case is at all handled or not

I believe the logic for get applicable variants should in general not resolve to any experiment if the sent identifier is `""`, `""` is being used as the default value in all providers for the case where `get_applicable_variant` is not needed

so a change in the rust side in all the relevant places should solve it

ayushjain17: I dont think we should generate any toss value when the identifier is `""`

ayushjain17: ideally `get_applicable_buckets_from_group` function should have a check as the first line (in both the crate)

```rust
fn get_applicable_buckets_from_group(identifier: &str, ..) -> Vec<..> {
    if identifier.is_empty() {
        return Vec::new();
    }
    ... rest of the processing here
}
```

ayushjain17 (changes requested):
Pending comments

crates/superposition_provider/src/client.rs
mahatoankitkumar: issue already there right https://github.com/juspay/superposition/issues/640

crates/superposition_core/src/experiment.rs
mahatoankitkumar: https://github.com/juspay/superposition/issues/734

clients/java/openfeature-provider/src/main/kotlin/io/juspay/superposition/openfeature/EvaluationArgs.kt
mahatoankitkumar: function not needed

PR #731, opened 2025-10-21
fix: update rust provider to match standard s11n provider interface

Datron:
## Problem 
Primitives missing in the rust open feature provider that help work with superposition

## Solution
Add those primitives

### DIFF
--- a/crates/superposition_provider/src/client.rs
+++ b/crates/superposition_provider/src/client.rs
@@ -179,6 +179,7 @@ impl CacConfig {
         let sdk_config = SdkConfig::builder()
             .endpoint_url(&options.endpoint)
             .bearer_token(options.token.clone().into())
+            .behavior_version_latest()
             .build();
 
         // Create Superposition client
@@ -192,13 +193,14 @@ impl CacConfig {
             .send()
             .await
             .map_err(|e| {
-                SuperpositionError::NetworkError(format!("Failed to get config: {}", e))
+                let error = format!("Failed to get config: {}", e);
+                SuperpositionError::NetworkError(error)
             })?;
 
         // Use ConversionUtils to convert to proper Config type
         let config = ConversionUtils::convert_get_config_response(&response)?;
 
-        info!("Successfully fetched and converted config with {} contexts, {} overrides, {} default configs", 
+        info!("Successfully fetched and converted config with {} contexts, {} overrides, {} default configs",
               config.contexts.len(), config.overrides.len(), config.default_configs.len());
 
         Ok(config)
@@ -418,6 +420,7 @@ impl ExperimentationConfig {
         let sdk_config = SdkConfig::builder()
             .endpoint_url(&options.endpoint)
             .bearer_token(options.token.clone().into())
+            .behavior_version_latest()
             .build();
 
         // Create Superposition client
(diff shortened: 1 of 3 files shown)


knutties:
@Datron - are we adding the experimentation request flag changes as well here?  They also need to go into the provider right ?

crates/superposition_provider/src/client.rs
```diff
@@ -192,13 +193,14 @@ impl CacConfig {
             .send()
             .await
             .map_err(|e| {
-                SuperpositionError::NetworkError(format!("Failed to get config: {}", e))
+                let error = format!("Failed to get config: {}", e);
```
knutties, line 196: general question: is this a clippy warn or just a taste thing ?

crates/superposition_provider/src/client.rs
```diff
@@ -418,6 +420,7 @@ impl ExperimentationConfig {
         let sdk_config = SdkConfig::builder()
             .endpoint_url(&options.endpoint)
             .bearer_token(options.token.clone().into())
+            .behavior_version_latest()
```
knutties, line 423: We are doing this because we want to provider to hit the latest version always? We don't have API versioning yet right - so this should be the case always right ?

crates/superposition_provider/src/client.rs
Datron: Helps with the debugger since the SDK doesn't surface the error properly

crates/superposition_provider/src/client.rs
Datron: Yeah and the rust AWS SDK that is generated expects this flag to be set otherwise requests fail

crates/superposition_provider/src/provider.rs
```diff
@@ -91,7 +91,38 @@ impl SuperpositionProvider {
...
+                        "Failed to initialize experimentation configuration: {}",
+                        e
+                    )));
+                }
+            }
+        };
+        Ok(())
+    }
+
+    pub async fn eval_config(
```
knutties: @Datron  - For the function that is exposed for clients to call - do we need a more appropriate name - something like `get_all_config()` or `resolve_full_config`- which can just pass through to `eval_config` for now.  It would be nice to make the signature of that similar to that of the existing resolve functions for individual types.

Datron: Sure @knutties , makes sense. I will make this change

crates/superposition_provider/src/provider.rs
```diff
@@ -91,6 +91,44 @@ impl SuperpositionProvider {
         }
     }
 
+    pub async fn init(&self) -> Result<()> {
+        // Initialize CAC config
+        if let Some(cac_config) = &self.cac_config {
+            match cac_config.create_config().await {
```
ayushjain17, line 97: shouldn't we do these both in parallel ?

crates/superposition_provider/src/client.rs
ayushjain17: can someone help me with the context for this addition, unable to understand why is this added

crates/superposition_provider/src/provider.rs
Datron: Its not needed right now

ayushjain17 (approved):

PR #737, opened 2025-10-30
feat: Support multi value query params as repeated keys

ayushjain17:
## Changes
1. Add support for multi value query params as repeated keys in both frontend and backend
2. update smithy to accept these params as a List
3. smithy suggests to use `UPPERCASE` for enum variants, and there were a lot of warnings for us not defining them in the correct case, so made changes for that


crates/frontend/src/query_updater.rs
Datron, line 1: https://github.com/juspay/superposition/blob/d866d0d05c818ba9dadd5edc317e49d18af1f6ac/crates/frontend/src/query_updater.rs#L96-L104

Can we instead split on & and then on = and use a Vec<(String, String)> to iterate over? That would simplify this code.

ayushjain17: done

Datron (approved):

PR #739, opened 2025-10-30
fix: Read orgs from db in no_auth

ayushjain17:
Instead of relying on LOCAL_ORGS env for DISABLED auth, read  the `full list` of orgs from db.

### DIFF
--- a/crates/service_utils/Cargo.toml
+++ b/crates/service_utils/Cargo.toml
@@ -26,7 +26,11 @@ rs-snowflake = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum_macros = { workspace = true }
-superposition_types = { workspace = true, features = ["result", "api"] }
+superposition_types = { workspace = true, features = [
+    "result",
+    "api",
+    "diesel_derives",
+] }
 url = { workspace = true }
 urlencoding = "~2.1.2"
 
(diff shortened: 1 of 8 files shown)


makefile
```diff
@@ -312,7 +313,7 @@ smithy-api-docs: smithy-build
 smithy-updates: smithy-clients smithy-api-docs
 
 leptosfmt:
-	leptosfmt $(LEPTOS_FMT_FLAGS) crates/frontend
+	leptosfmt $(LEPTOS_FMT_FLAGS) $(LEPTOS_PACKAGES)
```
Datron, line 316: This just formats code I think, don't we need to add `--check`?

crates/service_utils/Cargo.toml
```diff
@@ -26,7 +26,11 @@ rs-snowflake = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum_macros = { workspace = true }
-superposition_types = { workspace = true, features = ["result", "api"] }
+superposition_types = { workspace = true, features = [
+    "result",
+    "api",
+    "diesel_derives",
```
Datron, line 32: given how widely service_utils is used, should we add diesel_derives? This will increase the compile time and binary size. Anyway to avoid this?

makefile
ayushjain17: `$(LEPTOS_FMT_FLAGS)` has the `--check` thing

crates/service_utils/Cargo.toml
ayushjain17: I was also skeptic of that, but then I saw `service_utils` already uses diesel stuff

Datron (approved):

mahatoankitkumar (approved):

PR #740, opened 2025-11-01
fix(cohort): Update cohort related restrictions

ayushjain17:
### Changes regarding cohort restrictions
1. position of local/remote cohort to be always less than position of based on dimension - dimension creation/updation check
2. `value-compute-function` becomes a mandatory requirement for `remote cohort`, optional for `regular` and definitely absent for `local cohort`
3. if based on dimension is provided, the dependent dimensions cannot be provided in the context - context creation check
4. value validation of dimensions in context creation is not needed anymore (due to above mentioned check)
5. `value-validation-function`, if available for any dimension, would run irrespective of type - context creation check

### For control auto population
1. `resolve api` supports an extra param to also compute the remote cohort during resolution - if extra param is not sent, then normal resolution takes place without evaluating the remote cohort values
2. nothing changes with respect to local cohort
3. for remote cohort value is always computed using `value-compute-function` in the resolve api call
4. frontend to call resolve api with the extra param for control population

### General changes:
1. during control population, if `value-compute-function` is not giving a valid response, then it is to be treated as a failure case - a valid response from `value-compute-function` for this usage is an array response with exactly one element in it - no ambiguity allowed
2. functions used in default config and dimensions are only allowed to be used, if a published version of the function exists - creation/updation check

#### Note
`value-compute-function` is same as `auto-complete-function`

### DIFF
--- a/crates/context_aware_config/src/api/context/helpers.rs
+++ b/crates/context_aware_config/src/api/context/helpers.rs
@@ -84,8 +84,10 @@ fn validate_condition_with_dependent_dimensions(
         {
             for invalid_dimension in &invalid_dimensions {
                 if dependents.contains(invalid_dimension) {
-                    error_messages
-                        .push(format!("{} depends on {}", invalid_dimension, dim_key));
+                    error_messages.push(format!(
+                        "{} can be derived from {} dimension",
+                        invalid_dimension, dim_key
+                    ));
                 }
             }
         }
@@ -93,7 +95,7 @@ fn validate_condition_with_dependent_dimensions(
 
     if !error_messages.is_empty() {
         return Err(validation_error!(
-            "Dependent dimensions are not required in context when parent dimension is provided: [{}].",
+            "Cohort Dimension(s): [ {} ] using the cohort definitions. Hence, usage of this/these dimension(s) is not allowed.",
             error_messages.join(", ")
         ));
     }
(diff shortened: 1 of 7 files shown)


crates/context_aware_config/src/api/dimension/validations.rs
```diff
@@ -0,0 +1,391 @@
...
+            log::error!("invalid position: 0 for dimension: {dimension_name}",);
+            Err(validation_error!("Oth position is reserved for variantIds"))
+        }
+        (_, d_position) if d_position as i64 > max_allowed => {
+            log::error!("position {d_position} value exceeds total number of dimensions {max_allowed}");
+            Err(validation_error!(
+                "position value exceeds total number of dimensions"
+            ))
+        }
+        _ => Ok(()),
```
sauraww, line 46: I am not sure but looks like a negative position can be passed through this condition . Can we add this: 
(_, d_position) if d_position < 0 => {
            log::error!("invalid negative position: {d_position}");
            Err(validation_error!("position must be non-negative"))
        }
No one will put negative position but it will be safer that way.

sauraww (changes requested):

crates/context_aware_config/src/api/dimension/validations.rs
ayushjain17: it is already handled 
Position is a type which does not allow negative values

sauraww: Cool

sauraww (approved):

PR #742, opened 2025-11-03
fix: Smithy fixes

ayushjain17:
## Changes
1. API consistency - backwards compatible - missed in the previous one
a. update context - old (PUT) new (PATCH)
2. Changes to support `Move` action in context bulk api (breaking)
3. Smithy types fixes regarding optional/non-optional fields
4. Add support to unset functions in default config and dimensions via Smithy by passing `"null"` string - same as i64 strings
5. Fixed types and function signature and other warnings in python provider
6. Minor frontend fixes
7. Org name allowed empty string, now restricted
8. Removed warnings and errors from server tests, removed unnecessary tests from org tests

### Note
There are 2 commits in the PR, review only the first one, second one is just smithy generations


clients/python/provider/superposition_provider/configuration_client.py
```diff
@@ -98,14 +98,7 @@ def eval(self, query_data: dict, targeting_key: Optional[str]) -> dict[str, Any]
...
             logger.error(f"Evaluation failed: {e}")
             raise
-    
 
-    def get_applicable_variants(self, context: Dict[str, str], toss: Optional[int] = None) -> list:
-        if self.exp_config:
-            # No such function in exp_config
-            return self.exp_config.get_applicable_variants(context, toss)
-        else:
-            return []
```
Datron, line 108: This is removed but not added anywhere else?

crates/superposition_provider/tests/integration_test.rs
Datron, line 1: https://github.com/juspay/superposition/blob/c87cd9b7577fd7b1ef7f2450fff8158f040190f3/crates/superposition_provider/tests/integration_test.rs#L299-L300

This is mandatory? In general the older way was easier

clients/python/provider/superposition_provider/configuration_client.py
Datron, line 1: https://github.com/juspay/superposition/blob/c87cd9b7577fd7b1ef7f2450fff8158f040190f3/clients/python/provider/superposition_provider/configuration_client.py#L169-L169

get_applicable_variants

clients/python/provider/superposition_provider/configuration_client.py
ayushjain17: this was the wrong and redundant version, correct one still exists

clients/python/provider/superposition_provider/provider.py
```diff
@@ -89,48 +78,50 @@ async def shutdown(self):
...
-        if evaluation_context is None or evaluation_context.attributes is None:
-            return ({}, None)
+        if evaluation_context is None:
+            return {}, None
+        elif evaluation_context.attributes is None:
+            return {}, evaluation_context.targeting_key
         else:
             context =  {k: json.dumps(v) for k, v in evaluation_context.attributes.items()}
-            return (context, evaluation_context.targeting_key)
+            return context, evaluation_context.targeting_key
```
mahatoankitkumar: if evaluation_context is None:
        return {}, None
    
    if evaluation_context.attributes is None:
        return {}, evaluation_context.targeting_key

    context = {k: json.dumps(v) for k, v in evaluation_context.attributes.items()}
    return context, evaluation_context.targeting_key

    
Wouldn't this be better as every block returns something
no need of chaining?

PR #746, opened 2025-11-13
feat(hs): Open feature provider for Haskell.

ShreyBana:
feat: `superposition-bindings` package for HS bindings to `superposition_core`.
- Refactored `ffi_legacy` to accept an error buffer for returning the buffer as opposed to using thread locals.

ref: Deleted openapi symlink.
feat: Added Setup.hs in hs-bindings.


crates/superposition_core/cbindgen.toml
```diff
@@ -0,0 +1,6 @@
+language = "C"
+header = "/* Headers for my-api-crate */"
```
ayushjain17: which is your api crate ?

crates/superposition_core/cbindgen.toml
ayushjain17, line 1: whats the purpose of this file ?

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -349,16 +351,16 @@ pub extern "C" fn core_test_connection() -> i32 {
...
-pub extern "C" fn core_get_error_details() -> *mut c_char {
-    ERROR_DETAILS.with(|details| match details.borrow().clone() {
-        Some(details_str) => match CString::new(details_str) {
-            Ok(c_str) => c_str.into_raw(),
-            Err(_) => ptr::null_mut(),
-        },
-        None => ptr::null_mut(),
-    })
-}
+// #[no_mangle]
```
ayushjain17: can we delete this

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -374,67 +376,72 @@ pub unsafe extern "C" fn core_free_string(s: *mut c_char) {
...
+//     })
+// }
 
+/// # Safety
+///
+/// Caller ensures that `ebuf` is a sufficiently long buffer to store the
+/// error message.
 #[no_mangle]
-pub extern "C" fn core_get_applicable_variants(
+pub unsafe extern "C" fn core_get_applicable_variants(
```
ayushjain17, line 367: why `unsafe` ?

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -374,67 +376,72 @@ pub unsafe extern "C" fn core_free_string(s: *mut c_char) {
...
+//     }
+// }
+
+// #[no_mangle]
+// pub extern "C" fn core_last_error_length() -> i32 {
+//     LAST_ERROR.with(|prev| match *prev.borrow() {
+//         Some(ref err) => err.len() as i32 + 1,
+//         None => 0,
+//     })
+// }
```
ayushjain17: can we delete these

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -44,18 +35,23 @@ fn parse_json<T: serde::de::DeserializeOwned>(s: *const c_char) -> Result<T, Str
...
 
 fn string_to_c_str(s: String) -> *mut c_char {
-    match CString::new(s) {
-        Ok(c_str) => c_str.into_raw(),
-        Err(e) => {
-            set_last_error(format!("Failed to create C string: {}", e));
-            ptr::null_mut()
-        }
-    }
+    CString::new(s).unwrap().into_raw()
```
ayushjain17, line 32: this would panic

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -12,22 +11,14 @@ use crate::experiment::{ExperimentGroups, ExperimentationArgs};
...
-fn set_last_error(err: String) {
-    log::error!("FFI core error: {}", err);
-    LAST_ERROR.with(|prev| {
-        *prev.borrow_mut() = Some(err);
-    })
-}
+// thread_local! {
+//     static LAST_ERROR: RefCell<Option<String>> = const { RefCell::new(None) };
+//     static ERROR_DETAILS: RefCell<Option<String>> = const {RefCell::new(None)};
+// }
```
ayushjain17: can we delete this

crates/superposition_core/src/ffi_legacy.rs
ayushjain17, line 1: changes in js provider is also needed

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -64,6 +60,7 @@ pub extern "C" fn core_get_resolved_config(
     merge_strategy_str: *const c_char,
     filter_prefixes_json: *const c_char,
     experimentation_json: *const c_char,
+    ebuf: *mut c_char,
```
ayushjain17, line 57: this has to be passed from the existing places of usage as well

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -374,67 +376,72 @@ pub unsafe extern "C" fn core_free_string(s: *mut c_char) {
...
 #[no_mangle]
-pub extern "C" fn core_get_applicable_variants(
+pub unsafe extern "C" fn core_get_applicable_variants(
     experiments_json: *const c_char,
     experiment_groups_json: *const c_char,
     dimensions: *const c_char,
     query_data_json: *const c_char,
     identifier: *const c_char,
     filter_prefixes_json: *const c_char,
+    ebuf: *mut c_char,
```
ayushjain17, line 374: same for this as well

ayushjain17 (changes requested):

crates/superposition_core/src/ffi_legacy.rs
ShreyBana: suggest alternative

ayushjain17: can we not default to the null pointer like the previous behaviour ?

ShreyBana: Is that better? It will certainly make things harder to debug. If we're going to do that, then at-least we should report the error. But it just doesn't make sense, this feels like trying to catch an exception for adding 2 numbers. I think it's okay to panic if such a basic operation cannot be performed.

crates/superposition_core/src/ffi_legacy.rs
ShreyBana: Because it is using unsafe functions like `copy_nonoverlapping` indirectly.

crates/superposition_core/cbindgen.toml
ShreyBana: To generate the C header file. (But I think you can just do it w/o cbindgen.)

ayushjain17: this works without needing this file as well

ayushjain17 (approved):

sauraww (approved):

crates/superposition_core/cbindgen.toml
ShreyBana: Scratch that, it doesn't.

PR #750, opened 2025-11-17
fix: let number of workers come from env

Datron:
## Problem
The actix workers are limiting the number of DB operations we can perform, and are becoming a bottleneck

## Solution
Let workers be specified by an env

ACTIX_WORKER_COUNT

### DIFF
--- a/crates/superposition/src/main.rs
+++ b/crates/superposition/src/main.rs
@@ -218,7 +218,7 @@ async fn main() -> Result<()> {
             .wrap(auth_n.clone())
     })
     .bind(("0.0.0.0", cac_port))?
-    .workers(5)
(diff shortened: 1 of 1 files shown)


crates/superposition/src/main.rs
```diff
@@ -218,7 +218,7 @@ async fn main() -> Result<()> {
             .wrap(auth_n.clone())
     })
     .bind(("0.0.0.0", cac_port))?
-    .workers(5)
+    .workers(get_from_env_or_default("ACTIX_WORKERS", 5))
```
knutties: `ACTIX_WORKER_COUNT` is a better choice for name ?

PR #752, opened 2025-11-18
fix: fetch orgs in FE and support redirect to requested url from keycloak

ayushjain17:
## Changes
1. fix fetch org list in FE
2. support redirecting to the request url once auth_n is successful from keycloak instead of taking the user to list org page every time


crates/service_utils/src/middlewares/auth_n/authentication.rs
```diff
@@ -1,47 +1,86 @@
...
+        let org_id = path.organisation_id.clone();
+
+        Box::pin(async move {
+            match user_token_future.await {
+                Ok(token) => {
+                    let cookie = Cookie::build(Login::Org.to_string(), token)
+                        .path(cookie_path)
+                        .http_only(true)
+                        .max_age(Duration::days(1))
+                        .finish();
```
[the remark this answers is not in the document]
ayushjain17: resolved

crates/service_utils/src/middlewares/auth_n/oidc.rs
```diff
@@ -251,8 +258,19 @@
...
-            log::error!("Error org_user not found in cookies");
-            Err(self.new_redirect(&Login::Org.to_string()))
+            self.generate_org_user(&request, &org_id)
+                .await
+                .and_then(|token| {
+                    let cookie = Cookie::build(Login::Org.to_string(), token)
+                        .path(self.get_cookie_path())
+                        .http_only(true)
+                        .max_age(Duration::days(1))
+                        .finish();
```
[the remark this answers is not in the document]
ayushjain17: resolved

keycloak/docker-compose.yaml
```diff
@@ -0,0 +1,45 @@
+version: "3.8"
```
Datron, line 1: Why not use the same postgres that superposition is using? Create a separate logical DB inside

crates/service_utils/src/middlewares/tenant.rs
```diff
@@ -1,7 +1,7 @@
 use std::future::{ready, Ready};
 
 use crate::{
-    extensions::ServiceRequestExt,
+    extensions::HttpRequestExt,
```
Datron, line 4: Why do we need this change?

crates/service_utils/src/middlewares/auth_n/oidc.rs
```diff
@@ -171,6 +174,7 @@ impl OIDCAuthenticator {
             .map(|cookie| {
                 Cookie::build("protection", cookie)
                     .max_age(Duration::days(7))
+                    .secure(true)
                     // .http_only(true)
                     // .same_site(SameSite::Strict)
```
Datron: lets enable same site?

crates/service_utils/src/middlewares/auth_n/oidc.rs
```diff
@@ -183,8 +187,9 @@ impl OIDCAuthenticator {
                 .cookie(p_cookie)
                 // Deletes the cookie.
                 .cookie(
-                    Cookie::build(cookie_name, "")
+                    Cookie::build(cookie_type.to_string(), "")
                         .max_age(Duration::seconds(0))
+                        .secure(true)
```
Datron, line 193: Can we enable same site?

crates/service_utils/src/middlewares/auth_n/oidc.rs
```diff
@@ -251,8 +264,20 @@ impl OIDCAuthenticator {
...
         } else {
-            log::error!("Error org_user not found in cookies");
-            Err(self.new_redirect(&Login::Org.to_string()))
+            self.generate_org_user(&request, &org_id, &login_type)
+                .await
+                .and_then(|token| {
+                    let cookie = Cookie::build(login_type.to_string(), token)
+                        .path(self.get_cookie_path())
+                        .http_only(true)
+                        .secure(true)
```
Datron, line 274: Can we enable same site checks?

crates/service_utils/src/middlewares/auth_n/authentication.rs
```diff
@@ -10,38 +15,84 @@ pub(super) struct SwitchOrgParams {
...
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum Login {
     None,
     Global,
-    Org,
+    Org(String),
 }
```
Datron, line 23: What is the difference between Global and Org level Login?

crates/service_utils/src/middlewares/tenant.rs
ayushjain17: `ServiceRequest` internally contains `HttpRequest`

the options provided inside the trait are needed in some places where `ServiceRequest` is not available but the inner data `HttpRequest` is available

so to be able have a wider coverage for the functionality, moved the implementation to the inner core type on which the functionalities inherently depend upon

crates/service_utils/src/middlewares/auth_n/authentication.rs
ayushjain17: we have every user in 2 realm, one in common users realm and one in the realm specific to the org

keycloak/docker-compose.yaml
ayushjain17: this is just for local setup

crates/service_utils/src/middlewares/auth_n/oidc.rs
ayushjain17: I think there was some reason for which we had to disable it
for now it seems to be working fine so I am adding it back, might have to remove it later if we encounter the issue

keycloak/docker-compose.yaml
Datron: Yeah asking for local setup itself, don't they both occupy the same port? This seems a waste of resources

Datron (approved):

PR #755, opened 2025-11-21
feat: expose get_satisfied_experiments and refactor code for resolution

Datron:
## Problem
Internal juspay systems require more information from the provider to provide adequate debugging information

## Solution
- refactor resolution
- expose get_satisfied_experiments
- Add variants chosen to the Resolution Details so users can validate and debug

## Future TODO
Have all s11n providers expose the same interface. This will be done in a future PR

### DIFF
--- a/crates/superposition_provider/src/provider.rs
+++ b/crates/superposition_provider/src/provider.rs
@@ -231,16 +231,19 @@ impl SuperpositionProvider {
             self.get_context_from_evaluation_context(evaluation_context);
 
         let dimensions_info = self.get_dimensions_info().await;
-        let variant_ids = if let Some(exp_config) = &self.exp_client {
-            exp_config
-                .get_applicable_variants(&dimensions_info, &context, targeting_key)
-                .await?
-        } else {
-            vec![]
-        };
+        let mut variant_ids = Vec::new();
+        if targeting_key.is_some() {
+            if let Some(exp_config) = &self.exp_client {
+                let applicable_variant_ids = exp_config
+                    .get_applicable_variants(&dimensions_info, &context, targeting_key)
+                    .await?;
 
-        context.insert("variantIds".to_string(), json!(variant_ids));
-
+                context.insert("variantIds".to_string(), json!(applicable_variant_ids));
+                variant_ids = applicable_variant_ids;
+            } else {
+                log::warn!("Targeting key is set, but experiments have not been defined in the superposition provider builder options")
+            }
+        }
         let Some(ref client) = self.cac_client else {
             return Err(SuperpositionError::ConfigError(
                 "CAC config not initialized".into(),
--- a/crates/superposition_provider/src/utils.rs
+++ b/crates/superposition_provider/src/utils.rs
@@ -115,7 +115,7 @@ impl ConversionUtils {
             dimensions,
         };
 
-        debug!("Successfully converted config with {} contexts, {} overrides, {} default configs", 
+        debug!("Successfully converted config with {} contexts, {} overrides, {} default configs",
                config.contexts.len(), config.overrides.len(), config.default_configs.len());
 
         Ok(config)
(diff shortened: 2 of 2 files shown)


crates/superposition_provider/src/provider.rs
```diff
@@ -125,40 +126,90 @@ impl SuperpositionProvider {
     pub async fn resolve_full_config(
         &self,
         evaluation_context: &EvaluationContext,
-    ) -> Result<serde_json::Map<String, Value>> {
-        self.eval_config(evaluation_context).await
+        prefix_filters: Option<Vec<String>>,
+    ) -> Result<(serde_json::Map<String, Value>, Vec<String>)> {
```
mahatoankitkumar: Should we make a type for this?

Datron: Aliasing it for now

crates/superposition_provider/src/provider.rs
```diff
@@ -125,40 +162,95 @@ impl SuperpositionProvider {
...
+                "Experimentation config not initialized".into(),
+            ));
+        };
+        let (context_map, _) = self.get_context_from_evaluation_context(context);
+        exp_client
+            .get_satisfied_experiments(&context_map, filter_prefixes)
+            .await
+    }
+
+    pub async fn get_running_experiments_from_provider(&self) -> Result<Experiments> {
```
knutties, line 186: Why call this `get_running_experiments_from_provider` while using the `get_cached_experiments` inside?  1-1 naming will be helpful here ?

crates/superposition_provider/src/client.rs
```diff
@@ -217,7 +217,7 @@ impl CacConfig {
     pub async fn evaluate_config(
         &self,
         query_data: &serde_json::Map<String, Value>,
-        prefix_filter: Option<&[String]>,
+        prefix_filter: Option<Vec<String>>,
```
ayushjain17, line 220: why not reference, if we dont need reference here then in line 233 of this file, we dont need the to_vec conversion anymore

crates/superposition_provider/src/lib.rs
```diff
@@ -31,7 +31,7 @@ mod tests {
             None,
         );
 
-        let cac_config = CacConfig::new(superposition_options, config_options);
+        let cac_config = CacClient::new(superposition_options, config_options);
```
ayushjain17, line 34:
```suggestion
        let cac_client = CacClient::new(superposition_options, config_options);
```

crates/superposition_provider/src/lib.rs
```diff
@@ -49,7 +49,7 @@ mod tests {
             OnDemandStrategy::default(),
         ));
 
-        let exp_config = ExperimentationConfig::new(superposition_options, exp_options);
+        let exp_config = ExperimentationClient::new(superposition_options, exp_options);
```
ayushjain17, line 52:
```suggestion
        let exp_client = ExperimentationClient::new(superposition_options, exp_options);
```

crates/superposition_provider/src/provider.rs
```diff
@@ -125,40 +162,98 @@ impl SuperpositionProvider {
...
+            ));
+        };
+        exp_client.get_cached_experiments().await.ok_or(
+            SuperpositionError::ProviderError(
+                "Could not retrieve running experiments".into(),
+            ),
+        )
+    }
+
+    pub async fn get_last_modified_time(&self) -> Result<DateTime<Utc>> {
```
ayushjain17, line 199: this function name says last modified, but it is returning last updated
they seem totally different and confusing though

crates/superposition_provider/src/provider.rs
```diff
@@ -91,10 +95,43 @@ impl SuperpositionProvider {
         }
     }
 
+    async fn resolve_value<T>(
+        &self,
+        flag_key: &str,
+        evaluation_context: &EvaluationContext,
+        converter: fn(&Value) -> EvaluationResult<T>,
+    ) -> EvaluationResult<ResolutionDetails<T>> {
```
ayushjain17, line 103: astonishing! open-feature already had such types

crates/superposition_provider/src/provider.rs
```diff
@@ -125,40 +162,98 @@ impl SuperpositionProvider {
...
-            None => Err(SuperpositionError::ConfigError(
-                "CAC config not initialized".into(),
-            )),
+                context.insert("variantIds".to_string(), json!(applicable_variant_ids));
+                variant_ids = applicable_variant_ids;
+            } else {
+                log::warn!("Targeting key is set, but experiments have not been defined in the superposition provider builder options")
+            }
         }
+        let Some(ref client) = self.cac_client else {
```
ayushjain17, line 247: this section feels wrong, ideally shouldn't this have been separate and not part of eval_config, rather passed in to this
or maybe having this in here is still fine but not returned from every resolve operation

PR #760, opened 2025-11-25
feat: full fledged evaluation endpoint

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Release Notes

* **New Features**
  * Added GetResolvedConfigWithIdentifier endpoint for resolving configurations with an identifier context. This complements existing config resolution capabilities with additional identifier-based resolution logic. Available across all supported SDKs (Java, JavaScript, Python, Rust, Haskell).

* **Documentation**
  * OpenAPI specification and API documentation for the new GetResolvedConfigWithIdentifier endpoint.

* **Tests**
  * Added end-to-end test coverage for identifier-based configuration resolution.

### DIFF
--- a/crates/experimentation_platform/src/api/experiments/helpers.rs
+++ b/crates/experimentation_platform/src/api/experiments/helpers.rs
@@ -451,6 +451,8 @@ pub async fn fetch_cac_config(
 ) -> superposition::Result<(Config, Option<String>)> {
     let http_client = reqwest::Client::new();
     let query_params = ConfigQuery {
+        // Forced latest version to ensure we get the most recent config from CAC.
+        // Without this, CAC falls back to the workspace's default version setting, which may cause issue.
         version: Some("latest".to_string()),
         prefix: None,
     };
--- a/crates/experimentation_platform/src/api/experiments/cac_api.rs
+++ b/crates/experimentation_platform/src/api/experiments/cac_api.rs
@@ -175,6 +175,8 @@ pub async fn get_resolved_config(
     let http_client = state.http_client.clone();
     let resolve_params = ResolveConfigQuery {
         resolve_remote: Some(true),
+        // Forced latest version to ensure we get the most recent config from CAC.
+        // Without this, CAC falls back to the workspace's default version setting, which may cause issue.
         version: Some("latest".to_string()),
         ..resolve_params
     };
--- a/tests/src/resolve_config.test.ts
+++ b/tests/src/resolve_config.test.ts
@@ -1,0 +1,281 @@
+import {
+    GetConfigCommand,
+    GetResolvedConfigWithIdentifierCommand,
+    ListVersionsCommand,
+    UpdateWorkspaceCommand,
+    CreateDimensionCommand,
+    CreateDefaultConfigCommand,
+    CreateExperimentCommand,
+    RampExperimentCommand,
+    ConcludeExperimentCommand,
+    DeleteDimensionCommand,
+    DeleteDefaultConfigCommand,
+    DeleteContextCommand,
+    ListContextsCommand,
+    VariantType,
+    ExperimentStatusType,
+    type Variant,
+    DiscardExperimentCommand,
+    type ExperimentResponse,
+    GetExperimentCommand,
+} from "@juspay/superposition-sdk";
+import { superpositionClient, ENV } from "../env.ts";
+import { describe, test, expect, beforeAll, afterAll } from "bun:test";
+import { nanoid } from "nanoid";
+
+let configVersionId: string | undefined = undefined;
+
+describe("Resolve Config API - GetResolvedConfigWithIdentifier", () => {
+    let experimentId: string | undefined;
+    let dimensionName: string;
+    let configKey: string;
+    const testClientId = "test-client-bucketing-123";
+    const testIdentifier = "test-identifier-bucketing-456";
+    const defaultValue = "default-bucketing-value";
+    const experimentalValue = "experimental-bucketing-value";
+
+    beforeAll(async () => {
+        // Create unique names
+        dimensionName = `clientId`;
+        configKey = `testKey_1234`;
+
+        try {
+            // 1. Create dimension for clientId
+            try {
+                console.log(`Creating dimension: ${dimensionName}`);
+                const createDimCmd = new CreateDimensionCommand({
+                    dimension: dimensionName,
+                    workspace_id: ENV.workspace_id,
+                    org_id: ENV.org_id,
+                    schema: { type: "string" },
+                    position: 1,
+                    change_reason: "Automated Test - Bucketing test dimension",
+                    description: "Client ID dimension for bucketing test",
+                });
+                await superpositionClient.send(createDimCmd);
+                console.log(`Dimension created: ${dimensionName}`);
+            } catch (e: any) {
+                // If dimension already exists (409 conflict), continue
+                if (e.$response?.statusCode === 409 || e.statusCode === 409) {
+                    console.log(
+                        `Dimension ${dimensionName} already exists, continuing...`
+                    );
+                } else {
+                    throw e;
+                }
+            }
+
+            // 2. Create default config
+            try {
+                console.log(`Creating default config: ${configKey}`);
+                const createDefaultCmd = new CreateDefaultConfigCommand({
+                    workspace_id: ENV.workspace_id,
+                    org_id: ENV.org_id,
+                    key: configKey,
+                    value: defaultValue,
+                    schema: { type: "string" },
+                    description: `Default config for bucketing test`,
+                    change_reason: "Automated Test - Bucketing default config",
+                });
+                await superpositionClient.send(createDefaultCmd);
+                console.log(`Default config created: ${configKey} = ${defaultValue}`);
+            } catch (e: any) {
+                // If default config already exists (409 conflict), continue
+                const statusCode = e.$response?.statusCode || e.statusCode;
+                const errorMessage = e.$response?.body || e.message || "";
+
+                if (statusCode === 409 || errorMessage.includes("already exists")) {
+                    console.log(
+                        `Default config ${configKey} already exists, continuing...`
+                    );
+                } else {
+                    console.error(
+                        "Error creating default config:",
+                        e?.$response || e.message
+                    );
+                    throw e;
+                }
+            }
+
+            // 3. Create experiment
+            console.log("Creating experiment...");
+            const experimentContext = ENV.jsonlogic_enabled
+                ? {
+                      and: [{ "==": [{ var: dimensionName }, testClientId] }],
+                  }
+                : {
+                      [dimensionName]: testClientId,
+                  };
+
+            const experimentVariants: Omit<
+                Variant,
+                "id" | "context_id" | "override_id"
+            >[] = [
+                {
+                    variant_type: VariantType.CONTROL,
+                    overrides: {
+                        [configKey]: defaultValue,
+                    },
+                },
+                {
+                    variant_type: VariantType.EXPERIMENTAL,
+                    overrides: {
+                        [configKey]: experimentalValue,
+                    },
+                },
+            ];
+
+            const createExpCmd = new CreateExperimentCommand({
+                workspace_id: ENV.workspace_id,
+                org_id: ENV.org_id,
+                name: `bucketing-test-exp-${nanoid(8)}`,
+                context: experimentContext,
+                variants: experimentVariants.map((v, index) => ({
+                    id: index === 0 ? "control" : `test${index}`,
+                    variant_type: v.variant_type,
+                    overrides: v.overrides,
+                    description: "Bucketing test variant",
+                    change_reason: "Automated Test - Bucketing variant",
+                })),
+                description: "Experiment for bucketing test",
+                change_reason: "Automated Test - Bucketing experiment",
+            });
+
+            const expResponse = await superpositionClient.send(createExpCmd);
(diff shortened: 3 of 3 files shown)


crates/superposition/Cargo.toml
```diff
@@ -10,11 +10,13 @@ actix-files = { version = "0.6" }
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 aws-sdk-kms = { workspace = true }
+cac_client = { path = "../cac_client" }
 chrono = { workspace = true }
 context_aware_config = { path = "../context_aware_config" }
 diesel = { workspace = true }
 dotenv = "0.15.0"
 env_logger = "0.8"
+experimentation_client = { path = "../experimentation_client" }
```
Datron: Lets start using superposition_core. It has both of these functions

crates/superposition/src/resolve/types.rs
```diff
@@ -0,0 +1,21 @@
...
+    custom_query::{CommaSeparatedStringQParams, QueryParam},
+    IsEmpty,
+};
+
+#[derive(Deserialize, IsEmpty, QueryParam, Default)]
+pub struct ExtendedResolveConfigQuery {
+    #[query_param(skip_if_empty)]
+    pub version: Option<String>,
+    pub show_reasoning: Option<bool>,
+    pub resolve_remote: Option<bool>,
```
Datron: instead of `Option<bool>`, You can have serde deserialize these values to false if they are not sent by the user

crates/superposition/src/resolve/handlers.rs
```diff
@@ -0,0 +1,179 @@
...
+    let show_reason = query_filters.show_reasoning.unwrap_or_default();
+
+    let experiments = fetch_all_experiments(conn, &workspace_context.schema_name)?
+        .into_iter()
+        .map(|exp| {
+            let exp_response = ExperimentResponse::from(exp);
+            let id = exp_response.id.clone();
+            (id, exp_response)
+        })
+        .collect::<HashMap<String, ExperimentResponse>>();
```
Datron: Shouldn't we fetch experiments that are created and inprogress? Why fetch all?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -931,7 +931,7 @@ async fn get_applicable_variants(
         actix_web::http::Method::POST => {
             Ok(Either::Right(Json(ListResponse::new(variants))))
         }
-        _ => Ok(Either::Left(Json(variants))),
+        _ => Ok(Either::Left(Json(variants))), // Why Left?
```
Datron: Backwards compatibility

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -451,3 +451,13 @@ pub fn fetch_experiment_group(
         .get_result::<ExperimentGroup>(conn)?;
     Ok(experiment_group)
 }
+
+pub fn fetch_all_experiments_groups(
```
Datron: `fetch_experiment_groups` and take a filter? Helps make it multi-use

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -845,3 +845,16 @@ pub async fn validate_control_overrides(
 
     Ok(())
 }
+
+// Needs to be fetched from redis if present
+pub fn fetch_all_experiments(
```
Datron: `fetch_experiments` and take a filter? Helps make it multi-use

Datron (changes requested):

crates/superposition/Cargo.toml
mahatoankitkumar: Yeah looked into it earlier, it uses FfiExperiment and FfiExperimentGroup, so better to use experimentation_client here

Datron: Hmm got it, do you need any field that is not present in FfiExperiment or FfiExperimentGroup? If we can still use superposition_core and then later refactor the types it should be fine right?

crates/superposition/src/resolve/types.rs
mahatoankitkumar: I guess better to keep it optional here, as i am also using the resolve function if targeting_key not provided and that function is also used by the resolve endpoint.
Will have to make it optional again if i do this.

mahatoankitkumar:
Yet to push the smithy changes.

crates/superposition/Cargo.toml
```diff
@@ -11,6 +11,7 @@ actix-files = { version = "0.6" }
 actix-web = { workspace = true }
 anyhow = { workspace = true }
 aws-sdk-kms = { workspace = true }
+cac_client = { path = "../cac_client" }
```
Datron: Do we need client when core is present?

crates/superposition/src/resolve/types.rs
```diff
@@ -0,0 +1,21 @@
...
+    #[query_param(skip_if_empty)]
+    pub context_id: Option<String>,
+    #[query_param(skip_if_empty, iterable)]
+    pub prefix: Option<CommaSeparatedStringQParams>,
+    #[query_param(skip_if_empty)]
+    pub targetting_key: Option<String>,
```
ayushjain17: why are we calling it `targetting_key` here ?
our term is `identifier` for it, right ?

crates/superposition/src/resolve/types.rs
```diff
@@ -0,0 +1,21 @@
+use serde::Deserialize;
+use superposition_derives::{IsEmpty, QueryParam};
+
+use superposition_types::{
+    custom_query::{CommaSeparatedStringQParams, QueryParam},
+    IsEmpty,
+};
+
+#[derive(Deserialize, IsEmpty, QueryParam, Default)]
+pub struct ExtendedResolveConfigQuery {
```
ayushjain17: why are we recreating all the fields ?
we only need the `identifier` field separately we can reuse all other fields from the existing type no ?

any reason to repeat all of these ?

crates/superposition/src/resolve/types.rs
ayushjain17, line 1: smithy not updated

crates/superposition/src/resolve/types.rs
ayushjain17: keep these optional only

crates/superposition/src/resolve/handlers.rs
```diff
@@ -0,0 +1,190 @@
...
+        resolve(
+            req,
+            body,
+            merge_strategy,
+            &mut conn,
+            dimension_params,
+            resolve_query_filters,
+            workspace_context,
+        )
+    }
```
ayushjain17: why are we doing it like this ?
it all seems way to redundant and repetitive

it could have simply been something like:
```rust
let mut context = context.clone();
if let Some(identifier) = query.identifier {
    let applicable_variants = get_applicable_variants( .. );
    context.insert("variantIds", Value::Array(applicable_variants));
}

return resolve(config, context);
```

crates/superposition/src/main.rs
```diff
@@ -1,6 +1,7 @@
 #![deny(unused_crate_dependencies)]
 mod app_state;
 mod organisation;
+mod resolve;
```
ayushjain17, line 4: I am not sure if it is correct call it simply resolve as the resource/action

crates/superposition/src/main.rs
```diff
@@ -203,10 +204,15 @@ async fn main() -> Result<()> {
                             .service(webhooks::endpoints()),
                     )
                     .service(
-                        web::scope("/variables")
+                        scope("/variables")
                             .wrap(OrgWorkspaceMiddlewareFactory::new(true, true))
                             .service(variables::endpoints())
                     )
+                    .service(
+                        scope("/resolve")
```
ayushjain17, line 215: the endpoint here too seems a bit weird based on our conventions

crates/superposition/Cargo.toml
ayushjain17: +1

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: @mahatoankitkumar can we remove the comment

crates/context_aware_config/src/api/config/helpers.rs
```diff
@@ -12,3 +36,263 @@ pub fn apply_prefix_filter_to_config(
...
+
+pub fn resolve(
+    req: HttpRequest,
+    body: Option<Json<ContextPayload>>,
+    merge_strategy: Header<MergeStrategy>,
+    conn: &mut DBConnection,
+    dimension_params: DimensionQuery<QueryMap>,
+    query_filters: ResolveConfigQuery,
+    workspace_context: WorkspaceContext,
+) -> superposition::Result<HttpResponse> {
```
ayushjain17: this being a helper function, shouldn't it simply return the resolved value itself and not a `HttpResponse`

crates/experimentation_platform/src/api/experiment_groups/types.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub group_type_filter: Option<GroupType>,
+}
+
+impl ExperimentGroupQuery {
+    pub fn with_group_type(mut self, group_type: GroupType) -> Self {
+        self.group_type_filter = Some(group_type);
+        self
+    }
+
+    pub fn fetch(
```
ayushjain17: should we keep this as a part of the type implementation ?

crates/superposition/src/resolve/handlers.rs
```diff
@@ -0,0 +1,190 @@
...
+        .fetch(conn, &workspace_context.schema_name)?
+        .into_iter()
+        .map(FfiExperiment::from)
+        .collect::<Experiments>();
+
+    let experiment_groups = ExperimentGroupQuery::default()
+        .fetch(conn, &workspace_context.schema_name)?
+        .into_iter()
+        .map(FfiExperimentGroup::from)
+        .collect::<ExperimentGroups>();
```
ayushjain17: when have put in effort to move functions for context-aware-config, why not do the same for experimentation-platform, instead of repeating all of this

crates/experimentation_platform/src/api/experiments/types.rs
```diff
@@ -1,11 +1,58 @@
...
 pub struct StartedByChangeSet {
     pub started_by: Option<String>,
     pub started_at: Option<DateTime<Utc>>,
 }
+
+#[derive(Default)]
+pub struct ExperimentQuery {
+    status_filter: Option<Vec<ExperimentStatusType>>,
+    experiment_type_filter: Option<ExperimentType>,
+}
```
ayushjain17: I dont think using builder pattern is being of much help here, when all these fields would always be needed
I feel these new types are redundant and could have been avoided with a minor amount of cleanup

crates/superposition/src/resolve/types.rs
ayushjain17: pending

crates/context_aware_config/src/api/config/helpers.rs
Datron: Agreed on this

crates/superposition/src/main.rs
Datron: `resolve` is fine I think. We have
- /config/resolve for config only resolve
- /resolve for general resolve

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -822,66 +645,25 @@ async fn get_resolved_config(
...
-    if query_filters.resolve_remote.unwrap_or_default() {
-        query_data = QueryMap::from(evaluate_remote_cohorts(
-            &config.dimensions,
-            &query_data,
-            &mut conn,
-            &workspace_context.schema_name,
-        )?);
-    }
+    let config_version =
+        get_config_version(&query_filters.version, &workspace_context, &mut conn)?;
```
ayushjain17: this is repeated, you already have get config version at the top of this funciton

crates/superposition/src/resolve/handlers.rs
```diff
@@ -0,0 +1,89 @@
...
+    let identifier_query = identifier_query.into_inner();
+    let max_created_at = get_max_created_at(&mut conn, &workspace_context.schema_name)
+        .map_err(|e| log::error!("failed to fetch max timestamp from event_log : {e}"))
+        .ok();
+
+    if identifier_query.identifier.is_none() && is_not_modified(max_created_at, &req) {
+        return Ok(HttpResponse::NotModified().finish());
+    }
+
+    let mut config = generate_config_from_version(
```
ayushjain17, line 61: the behaviour over here should also be same as config/resolve, right ?
it should support taking `version` as well

crates/superposition/src/resolve/handlers.rs
```diff
@@ -0,0 +1,89 @@
...
+        query_data,
+        merge_strategy,
+        &mut conn,
+        &query_filters,
+        &workspace_context,
+    )?;
+
+    let config_version =
+        get_config_version(&query_filters.version, &workspace_context, &mut conn)?;
+
```
ayushjain17: and this should not be needed here, rather before generate config function is called

ayushjain17 (changes requested):
Build failing
smithy changes and tests not added yet

ayushjain17 (approved):

PR #777, opened 2025-12-11
fix: Auth behaviour in python provider

ayushjain17:
## Problem
token not being passed to sdk client in python provider

## Solution
pass the token to the client by doing proper configuration

### Note
java and python smithy generators do not have complete support for auth generation yet
it requires implementing a few types which needs to be passed to the sdk client as config during setup
this was already being done correctly inside java, but the implemented types were present in the java-provider
this setup, inherently, requires all users of the java sdk to reimplement these types on their side to be able to use

to avoid this, moved this piece of code to the java sdk itself, and introduced the same change for the python sdk also

### DIFF
--- a/clients/python/sdk/superposition_sdk/auth_helpers.py
+++ b/clients/python/sdk/superposition_sdk/auth_helpers.py
@@ -1,0 +1,263 @@
+"""
+Authentication helpers for Superposition SDK.
+
+This module provides Bearer Token and Basic authentication implementations for the Smithy-generated
+Python SDK. This is a temporary workaround until smithy-python codegen adds built-in
+auth support (similar to Rust and TypeScript SDKs).
+
+Supported authentication schemes:
+    - Bearer Token: HTTP Bearer authentication using bearer tokens
+    - Basic Auth: HTTP Basic authentication using username and password
+
+TODO: Remove this once smithy-python 1.0+ provides built-in HTTP auth schemes.
+"""
+
+from dataclasses import dataclass
+from typing import Any, Dict
+
+from smithy_core.aio.interfaces.identity import IdentityResolver
+from smithy_core.interfaces.identity import Identity, IdentityProperties
+from smithy_http import Field, Fields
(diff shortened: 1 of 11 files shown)


clients/java/sdk/src/main/java/io/juspay/superposition/client/auth/AuthHelper.java
knutties, line 1: Why a AuthHelper wrapper class?  Can't we just have them under auth package itself ?

ayushjain17: auth `package` as in ?

ayushjain17: have removed the wrapper AuthHelper class

knutties, line 1: Can we split them into separate files as is typically the case in Java instead of keeping them in a AuthHelper.java files ?  Typically public classes are stored in files of the same name in Java.

clients/python/sdk/superposition_sdk/auth_helpers.py
```diff
@@ -0,0 +1,263 @@
...
+from smithy_core.interfaces.identity import Identity, IdentityProperties
+from smithy_http import Field, Fields
+from smithy_http.aio.interfaces.auth import (
+    HTTPAuthOption,
+    HTTPAuthScheme,
+    HTTPSigner,
+)
+
+from .auth import HTTPAuthParams, HTTPAuthSchemeResolver
+from .config import Config
```
knutties, line 29: It might be a good idea to keep this out in the main program and not make it dependent on Superposition Config.  The helper functions in this file can return the `auth_scheme_resolver` and the `auth_schemes` as a tuple and then the main function can construct the config using that instead of making auth dependent on Superposition Config.

PR #783, opened 2025-12-15
feat: Add helm charts for superposition deployment

Shailesh-714:
## Problem
Docker or manifest installations are not efficient for managing kubernetes resources

## Solution
Adding helm charts for superposition deployment in kubernetes clusters
The helm chart is under the `charts/superosition`

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/helm/README.md.gotmpl
+++ b/helm/README.md.gotmpl
@@ -38,22 +38,7 @@ cd superposition/charts/superposition
 kubectl create namespace superposition
 ```
 
-### Step 3 - Configure Database Connection
-
-Update the values.yaml with your database connection details:
-
-```yaml
-configs:
-  database_url: postgres://user:password@host:5432/dbname?sslmode=disable
-  db_user: postgres
-  db_host: localhost:5432
-  db_name: config
-
-secrets:
-  db_password: "your-secure-password"
-```
-
-### Step 4 - Install Superposition
+### Step 3 - Install Superposition
 
 ```bash
 helm install superposition . -n superposition
--- a/helm/README.md
+++ b/helm/README.md
@@ -33,22 +33,7 @@ cd superposition/charts/superposition
 kubectl create namespace superposition
 ```
 
-### Step 3 - Configure Database Connection
-
-Update the values.yaml with your database connection details:
-
-```yaml
-configs:
-  database_url: postgres://user:password@host:5432/dbname?sslmode=disable
-  db_user: postgres
-  db_host: localhost:5432
-  db_name: config
-
-secrets:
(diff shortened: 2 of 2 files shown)


Datron:
@Shailesh-714 thanks for the contribution - some changes that we need to merge this:
- change the commit to follow semantic versioning convention? 
- should we rename the folder to helm instead of charts/superposition?

@knutties do take a look as well

Shailesh-714:
Made the required changes @Datron

helm/Chart.yaml
```diff
@@ -0,0 +1,27 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.88.0"
```
knutties, line 24: can we make this dynamic and pick from the version of the repo?

something like https://github.com/juspay/superposition/blob/main/.github/workflows/release.yaml#L68

@Datron - any comments ?

helm/README.md.gotmpl
```diff
@@ -0,0 +1,345 @@
...
+
+```bash
+helm install superposition . -n superposition
+```
+
+## Configuration
+
+### Database Configuration
+
+Superposition requires a PostgreSQL database:
```
knutties, line 51: This is a repetition of lines 45 - 54 ?  Do we need this documented twice ?

helm/Chart.yaml
Shailesh-714: We cant't make this dynamic, but could update this using workflows, if possible you can also write a chart release workflow to package and host the helm chart using github pages and chart releaser action:
ref: [https://github.com/juspay/hyperswitch-helm/blob/main/.github/workflows/release.yml](url)

helm/README.md.gotmpl
Shailesh-714: Yes it was reductant
removed!

helm/Chart.yaml
knutties: Ok - we can do with a hard-coded version for now.  Let us update and do the chart-release workflow with the superposition dynamic release version in a separate PR.

hope that works @Shailesh-714

Shailesh-714: Sure

PR #785, opened 2025-12-17
fix: functions changes

mahatoankitkumar:
Screenshot 2025-12-17 at 10 26 11 AM

### DIFF
--- a/crates/frontend/src/components/function_form.rs
+++ b/crates/frontend/src/components/function_form.rs
@@ -153,24 +153,25 @@ pub fn function_editor(
                     };
 
                     view! {
-                        <div class="flex flex-col gap-4 w-full">
-                            <Tip
-                                message="Reference variables using"
-                                code_snippet="VARS.KEY_NAME"
-                                example="VARS.API_KEY"
-                                code_signature=function_code_signature
-                            />
-
-                            <div class="w-full min-w-[800px] p-1 bg-base-100 rounded-2xl shadow">
-                                <MonacoEditor
-                                    node_id="code_editor_fn"
-                                    data=function_code_rws.get_untracked()
-                                    on_change=move |value| function_code_rws.set(value)
-                                    classes=vec!["w-[inherit] min-h-[500px]"]
-                                    read_only=mode.with(|v| *v != Mode::Editor)
-                                    language=crate::components::monaco_editor::Languages::Javascript
+                        <div class="w-full min-w-[800px] flex flex-col gap-3">
+                            <Show when=move || mode.get() == Mode::Editor>
+                                <Tip
+                                    message="Reference variables using"
+                                    code_snippet="VARS.KEY_NAME"
+                                    example="VARS.API_KEY"
+                                    code_signature=function_code_signature
                                 />
-                            </div>
+                            </Show>
+                            <MonacoEditor
+                                node_id="code_editor_fn"
+                                data=function_code_rws.get_untracked()
+                                on_change=move |value| function_code_rws.set(value)
+                                classes=vec![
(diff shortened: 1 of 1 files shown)


crates/frontend/src/pages/function.rs
```diff
@@ -235,8 +240,25 @@ pub fn function_page() -> impl IntoView {
...
+                                    FunctionType::ContextValidation => {
+                                        CONTEXT_VALIDATION_TEMPLATE_FN
+                                    }
+                                    FunctionType::ChangeReasonValidation => {
+                                        CHANGE_REASON_VALIDATION_TEMPLATE_FN
+                                    }
+                                });
 
                             view! {
+                                <Tip
```
ayushjain17: Tip is already is there in line `250/272` this should also be there only instead of being here again

crates/frontend/src/components/function_form.rs
```diff
@@ -274,17 +284,19 @@ fn edit_form(
...
-                                    FunctionType::ValueValidation => VALUE_VALIDATION_TEMPLATE_FN,
-                                    FunctionType::ValueCompute => VALUE_COMPUTE_TEMPLATE_FN,
-                                    FunctionType::ContextValidation => {
-                                        CONTEXT_VALIDATION_TEMPLATE_FN
-                                    }
-                                    FunctionType::ChangeReasonValidation => {
-                                        CHANGE_REASON_VALIDATION_TEMPLATE_FN
-                                    }
+                                    FunctionType::ValueValidation => VALUE_VALIDATION_DEFAULT_FN,
+                                    _ => VALUE_COMPUTE_DEFAULT_FN,
```
ayushjain17: even though we removed the options, shall we still handle all the cases over here, just for correctness

mahatoankitkumar: redundant here

crates/frontend/src/pages/function.rs
```diff
@@ -246,14 +268,6 @@ pub fn function_page() -> impl IntoView {
                             }
                         }
 
-                        <Show when=move || mode_rws.get() == Mode::Editor>
```
ayushjain17, line 249: this is not supposed to be removed
revert

crates/frontend/src/components/function_form.rs
ayushjain17: its not about redundancy, its about keeping it future proof, using `_` goes against it
later when we add one more type, unless someone explicitly remembers to update here, it would be difficult for the person to make sure that this is handled in all the places as compiler wont even give an error over here that a branch is not handled over here

ideally there already should have been a function for this which should have wrapped this match case internally, as this is present in multiple places

crates/frontend/src/components/form/label.rs
```diff
@@ -1,13 +1,44 @@
...
+    fn to_class(self) -> &'static str {
+        match self {
+            TooltipPosition::Top => "bottom-full left-1/2 -translate-x-1/2 mb-1",
+            TooltipPosition::Right => "left-full top-1/2 -translate-y-1/2 ml-2",
+            TooltipPosition::Bottom => "top-full left-1/2 -translate-x-1/2 mt-1",
+            TooltipPosition::Left => "right-full top-1/2 -translate-y-1/2 mr-2",
+            TooltipPosition::TopRight => "bottom-full left-0 mb-1",
+            TooltipPosition::TopLeft => "bottom-full right-0 mb-1",
+            TooltipPosition::BottomRight => "top-full left-0 mt-1",
+            TooltipPosition::BottomLeft => "top-full right-0 mt-1",
```
ayushjain17: why is margin added to these ?

PR #787, opened 2025-12-17
chore: use self-hosted runner for builds

Datron:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflow runner configuration to enhance build consistency and reliability.
  * Refined step labels in CI workflows for improved clarity.

### DIFF
--- a/.github/workflows/ci_check_pr.yaml
+++ b/.github/workflows/ci_check_pr.yaml
@@ -44,7 +44,7 @@ env:
 jobs:
     formatting:
         name: Check formatting
-        runs-on: codebuild-superposition-${{ github.run_id }}-${{ github.run_attempt }}
+        runs-on: anton1
         steps:
             - name: Checkout repository
               uses: actions/checkout@v4
@@ -101,7 +101,7 @@ jobs:
 
     test:
         name: Testing
-        runs-on: codebuild-superposition-${{ github.run_id }}-${{ github.run_attempt }}
+        runs-on: anton1
         services:
             postgres:
                 image: public.ecr.aws/docker/library/postgres:15-alpine3.21
@@ -183,7 +183,7 @@ jobs:
 
     java-build:
         name: Java build
-        runs-on: codebuild-superposition-${{ github.run_id }}-${{ github.run_attempt }}
+        runs-on: anton1
         defaults:
             run:
                 working-directory: clients/java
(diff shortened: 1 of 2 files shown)


tests/src/config.test.ts
```diff
@@ -2,12 +2,50 @@ import {
...
-import { describe, test, expect } from "bun:test";
+import { describe, test, expect, beforeAll } from "bun:test";
 
 let configVersionId: string | undefined = undefined;
 
+beforeAll(async () => {
+    const testConfigKey = "test-prefix";
+
+    // Check if config already exists
+    const listCmd = new ListDefaultConfigsCommand({
```
ayushjain17, line 17: Instead of doing this, can we create a random key in beforeAll and drop it in afterAll

.github/workflows/ci_check_pr.yaml
```diff
@@ -44,7 +44,7 @@ env:
 jobs:
     formatting:
         name: Check formatting
-        runs-on: codebuild-superposition-${{ github.run_id }}-${{ github.run_attempt }}
+        runs-on: anton1
```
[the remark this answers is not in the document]
ayushjain17: Do we not need to address this ?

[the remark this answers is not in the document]
ayushjain17: @Datron

tests/src/config.test.ts
knutties: @ayushjain17 - subsequent tests depend on this key naming right ?

ayushjain17: Only in this test suite
Suggested to keep it random to avoid conflict with other test suites

.github/workflows/ci_check_pr.yaml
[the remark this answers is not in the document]
Datron: Will address this in the follow up PR, some dependencies need to be added for uniffi generation and smithy CLI checks

tests/src/config.test.ts
Datron: Done

.github/workflows/ci_check_pr.yaml
```diff
@@ -310,8 +310,7 @@ jobs:
               if: matrix.provider == 'kotlin'
               run: cd clients/java && ./gradlew assemble
 
-            - name: Install Node.js (for JS tests)
-              if: matrix.provider == 'js'
+            - name: Install Node.js
```
ayushjain17, line 313: why do we need Node for other languages as well ?

Datron: Npm is not available by default in the environment and some dependencies are installed with it

PR #788, opened 2025-12-17
fix: Add status in FfiExperiment type

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/clients/python/provider/superposition_provider/exp_config.py
+++ b/clients/python/provider/superposition_provider/exp_config.py
@@ -2,9 +2,11 @@ import json
 import logging
 from decimal import Decimal
 from typing import Any, Dict, Optional, TypeVar
+from unittest import case
+
 from superposition_bindings.superposition_client import FfiExperiment, FfiExperimentGroup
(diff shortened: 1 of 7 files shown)


clients/python/provider/superposition_provider/exp_config.py
```diff
@@ -75,11 +77,25 @@ def document_to_python_value(doc: Document) -> Any:
             return val
 
 
-def to_group_type(sdk_group_type: SDKGroupType) -> GroupType:
-    if sdk_group_type == SDKGroupType.USER_CREATED:
-        return GroupType.USER_CREATED
-    return GroupType.SYSTEM_GENERATED
-
+def to_group_type(sdk_group_type: str) -> GroupType:
```
mahatoankitkumar, line 80: Just a query why `sdk_group_type: str`? why not `sdk_group_type: SDKGroupType`

ayushjain17: the input was not of `SDKGroupType` type, it was a str

PR #789, opened 2025-12-18
ci: update release workflow to update version in helm chart

Shailesh-714:
## Problem

The helm chart has a static appVersion of superposition

## Solution

Update the release workflow to update the appVersion in `Chart.yaml` to the release version

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -141,12 +141,22 @@ resources: {}
 # This is to setup the liveness and readiness probes more information can be found here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
 livenessProbe:
   httpGet:
-    path: /
+    path: /health
     port: http
+  initialDelaySeconds: 30
+  periodSeconds: 10
+  timeoutSeconds: 5
+  failureThreshold: 3
+  successThreshold: 1
 readinessProbe:
   httpGet:
-    path: /
+    path: /health
     port: http
+  initialDelaySeconds: 5
+  periodSeconds: 5
+  timeoutSeconds: 3
+  failureThreshold: 3
+  successThreshold: 1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -41,6 +41,11 @@ jobs:
                   crate: cocogitto
                   version: "6.2.0"
 
+            - name: install yq
+              run: |
+                  sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
```
knutties: why sudo ?  just the commands should suffice.  can we run yq from where-ever we copy it?

knutties: Also can we download a specific version instead of the latest?  We need minimal yq features.  Don't want to download and run a dynamic binary - just to be safe.

knutties: @Shailesh-714  - some recommendations.  I think storing the binary under bin/ in `${GITHUB_WORKSPACE}` should be good for now.

Ref: https://claude.ai/share/2d967672-3f3e-4bcd-9b9d-bb210d189184

Shailesh-714: @knutties Made the required changes

.github/workflows/release.yaml
```diff
@@ -41,6 +41,13 @@ jobs:
                   crate: cocogitto
                   version: "6.2.0"
 
+            - name: install yq
+              run: |
+                  mkdir -p ${{ github.workspace }}/bin
+                  wget -qO ${{ github.workspace }}/bin/yq https://github.com/mikefarah/yq/releases/download/v4.44.2/yq_linux_amd64
```
knutties, line 47: os / architecture can be made variables if possible.

knutties: may be in a later PR.

knutties (approved):

PR #795, opened 2025-12-19
feat(auth_n): Separate Saas OIDC login from core OIDC login

ayushjain17:
## Problem
Current OIDC login setup was coupled to the SaaS setup

## Solution
Separate out SaaS OIDC login from core OIDC login into `OIDC_SAAS` and `OIDC` login types

### Difference between SimpleOIDC and SaasOIDC
In SaaS setup, each organisation has its own OIDC provider endpoints and the first issuer also acts as a global identity provider which provides authorization to the individual orgs and in Simple OIDC a global IDP is used and no org specific providers

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,13 +21,20 @@ ACTIX_KEEP_ALIVE=120
 MAX_DB_CONNECTION_POOL_SIZE=3
 TENANT_MIDDLEWARE_EXCLUSION_LIST="/health,/assets/favicon.ico,/pkg/frontend.js,/pkg,/pkg/frontend_bg.wasm,/pkg/tailwind.css,/pkg/style.css,/assets,/admin,/oidc/login,/admin/organisations,/organisations,/organisations/switch/{organisation_id},/"
 SERVICE_PREFIX=""
+### Env(s) for NO AuthN
 AUTH_PROVIDER=DISABLED
+### Env(s) for Simple OIDC AuthN
 # AUTH_PROVIDER=OIDC+http://localhost:8081/realms/users
 # OIDC_CLIENT_ID=superposition
 # OIDC_CLIENT_SECRET=superposition_secret
-# OIDC_TOKEN_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>/protocol/openid-connect/token"
-# OIDC_ISSUER_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>"
 # OIDC_REDIRECT_HOST="http://localhost:8080"
+### Env(s) for SaaS OIDC AuthN
+# AUTH_PROVIDER=OIDC_SAAS+http://localhost:8081/realms/users
+# OIDC_CLIENT_ID=superposition
+# OIDC_CLIENT_SECRET=superposition_secret
+# OIDC_REDIRECT_HOST="http://localhost:8080"
+# OIDC_ORG_TOKEN_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>/protocol/openid-connect/token"
+# OIDC_ORG_ISSUER_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>"
 WORKER_ID=1
 ENCRYPTED_KEYS="" # Used for webhook secrets, API keys, etc.
 # ENCRYPTED_KEYS="TEST_API_KEY,DASHBOARD_API_KEY"
--- a/crates/service_utils/src/middlewares/auth_n/helpers.rs
+++ b/crates/service_utils/src/middlewares/auth_n/helpers.rs
@@ -1,0 +1,44 @@
+use actix_web::{web::Data, HttpRequest};
+use diesel::{
+    query_dsl::methods::{OrderDsl, SelectDsl},
+    Connection, ExpressionMethods, RunQueryDsl,
+};
+use superposition_types::database::superposition_schema::superposition::organisations;
+
+use crate::service::types::AppState;
+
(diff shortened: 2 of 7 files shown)


.env.example
```diff
@@ -25,9 +25,13 @@ AUTH_PROVIDER=DISABLED
...
 # OIDC_CLIENT_ID=superposition
 # OIDC_CLIENT_SECRET=superposition_secret
-# OIDC_TOKEN_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>/protocol/openid-connect/token"
-# OIDC_ISSUER_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>"
 # OIDC_REDIRECT_HOST="http://localhost:8080"
+# AUTH_PROVIDER=OIDC_SAAS+http://localhost:8081/realms/users
+# OIDC_CLIENT_ID=superposition
+# OIDC_CLIENT_SECRET=superposition_secret
+# OIDC_REDIRECT_HOST="http://localhost:8080"
+# OIDC_SAAS_TOKEN_ENDPOINT_FORMAT="http://localhost:8081/realms/<organisation>/protocol/openid-connect/token"
```
knutties: @ayushjain17 - are we now supporting two types of env's one with `OIDC` and one with `OIDC_SASS` and the corresponding one will be picked based on the `AUTH_PROVIDER` env ?

If so - can we update the example to have both formats ?  What will Superposition use?  Just `OIDC_` right ?

crates/service_utils/src/middlewares/auth_n/oidc/authenticator.rs
```diff
@@ -0,0 +1,175 @@
...
+};
+use superposition_types::User;
+
+use crate::middlewares::auth_n::{
+    authentication::{Authenticator, Login},
+    helpers::fetch_orgs_from_db,
+    oidc::{
+        login,
+        utils::{try_user_from, verify_presence},
+        OIDCAuthenticatorExt,
```
knutties: Looks like we have two traits - `Authenticator` and `OIDCAuthenticatorExt` - while I can roughly make sense what each of them is meant for (the first for Superposition auth capabilities and the latter for supporting OIDC) - might be better to be more clear with the naming.  the OIDCAuthenticator type naming also feels a bit of place.  Maybe the `OIDCAuthenticatorExt` should have been called `OIDCAuthenticator` and the current `OIDCAuthenticator` should have been called `SimpleOIDC` or some such ?

Or I have completely mis-understood.

A couple of lines around the traits and functions would help first time code readers get familiar with the code.

crates/service_utils/src/middlewares/auth_n/helpers.rs
```diff
@@ -0,0 +1,42 @@
+use actix_web::{web::Data, HttpRequest};
+use diesel::{
+    query_dsl::methods::{OrderDsl, SelectDsl},
+    Connection, ExpressionMethods, RunQueryDsl,
+};
+use superposition_types::database::superposition_schema::superposition::organisations;
+
+use crate::service::types::AppState;
+
+pub(super) fn fetch_orgs_from_db(req: &HttpRequest) -> Result<Vec<String>, &'static str> {
```
knutties: is calling this `fetch_org_ids_from_db` (not the entire org details) more appropriate ?

crates/service_utils/src/middlewares/auth_n/oidc.rs
```diff
@@ -198,240 +91,10 @@ impl OIDCAuthenticator {
...
-                })
-                .map_err(|e| e.map(Into::into).unwrap_or(redirect))
-        })
-    }
 }
 
-#[get("/login")]
-async fn login(
-    data: Data<OIDCAuthenticator>,
+async fn login<T: Authenticator + OIDCAuthenticator>(
```
knutties:
```suggestion
async fn login<T: OIDCAuthenticator>(
```
as OIDCAuthenticator already extends Authenticator ?

crates/service_utils/src/middlewares/auth_n/oidc/saas_authenticator.rs
```diff
@@ -0,0 +1,391 @@
...
+    pub async fn new(
+        idp_url: String,
+        base_url: String,
+        path_prefix: String,
+        client_id: String,
+        client_secret: String,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        let issuer_endpoint_format =
+            get_from_env_unsafe::<String>("OIDC_ISSUER_ENDPOINT_FORMAT").unwrap();
+        let token_endpoint_format =
```
knutties, line 69: The env is `OIDC_SAAS_TOKEN_ENDPOINT_FORMAT` in the example but here we don't have the `_SAAS` part in the env.  Is there some magic happening elsewhere or do we need to fix something?

ayushjain17: yeah this somehow got missed

ayushjain17: this should rather be `OIDC_ORG_TOKEN_ENDPOINT_FORMAT`

crates/service_utils/src/middlewares/auth_n/oidc.rs
knutties: Should we move the login as default implementation in the OIDCAuthenticator trait itself (as the first arg is essentially self here) ?  That way we can move the routes definition to the default implementation in the OIDCAuthenticator trait itself instead of writing it in both the Simple/SaaS versions.

PR #796, opened 2025-12-21
feat: add toml parsing to superposition_core

knutties:
## Problem

There is no easy human readable/write-able format of Superposition configuration data for simple applications (without a dedicated store) to use.

## Solution

This PR adds the support to generate and parse Superposition configuration data written in as toml files.  It introduces a specific convention over the TOML format to be able to support this.

More details in the design doc in the PR.

NA

## Pre-deployment activity
NA

## Post-deployment activity
NA

## API changes
| /config      | GET | Accept header based negotiation | TOML response if Accept header is `application/toml` |

## Summary by CodeRabbit

* **New Features**
  * Added TOML configuration parsing support across multiple language bindings (Java/Kotlin, JavaScript, Python, and Haskell)
  * Enabled HTTP content negotiation for config endpoints to serve responses in TOML format in addition to JSON (via `Accept: application/toml` header)
  * Added comprehensive example application demonstrating TOML configuration parsing and evaluation
  * Introduced test suites for TOML parsing functionality across all supported language bindings

* **Documentation**
  * Added README and design documentation for TOML parsing and response format capabilities

### DIFF
--- a/clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
+++ b/clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
@@ -18,8 +18,8 @@ import Foreign.ForeignPtr (newForeignPtr, withForeignPtr)
 import Foreign.Ptr (FunPtr)
 import Foreign.Marshal (free)
 
-type BufferSize = Int 
-errorBufferSize :: BufferSize 
+type BufferSize = Int
+errorBufferSize :: BufferSize
 errorBufferSize = 2048
 
 foreign import capi "superposition_core.h core_get_resolved_config"
@@ -130,16 +130,16 @@ getResolvedConfig params = do
 --   - dimensions: object mapping dimension names to dimension info (schema, position, etc.)
 parseTomlConfig :: String -> IO (Either String Value)
 parseTomlConfig tomlContent = do
-  ebuf <- callocBytes errorBufferSize  
+  ebuf <- callocBytes errorBufferSize
   tomlStr <- newCString tomlContent
   res <- parse_toml_config tomlStr ebuf
-  errBytes <- packCString ebuf 
+  errBytes <- packCString ebuf
   let errText = fromRight mempty $ decodeUtf8' errBytes
   result <- if res /= nullPtr
-    then do 
-      resFptr <- newForeignPtr p_free_string res 
+    then do
+      resFptr <- newForeignPtr p_free_string res
       -- Registers p_free_string as the finalizer (for automatic cleanup)
-      withForeignPtr resFptr $ \ptr -> Just <$> packCString ptr
+      withForeignPtr resFptr $ fmap Just . packCString
     else pure Nothing
   free tomlStr
   free ebuf
--- a/clients/java/bindings/src/main/kotlin/uniffi/superposition_client/superposition_client.kt
+++ b/clients/java/bindings/src/main/kotlin/uniffi/superposition_client/superposition_client.kt
@@ -967,7 +967,7 @@ private fun uniffiCheckApiChecksums(lib: IntegrityCheckingUniffiLib) {
     if (lib.uniffi_superposition_core_checksum_func_ffi_get_applicable_variants() != 58234.toShort()) {
         throw RuntimeException("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
     }
-    if (lib.uniffi_superposition_core_checksum_func_ffi_parse_toml_config() != 20800.toShort()) {
+    if (lib.uniffi_superposition_core_checksum_func_ffi_parse_toml_config() != 1558.toShort()) {
         throw RuntimeException("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
     }
 }
@@ -1762,14 +1762,14 @@ public object FfiConverterMapStringTypeOverrides: FfiConverterRustBuffer<Map<kot
          *
          * # Example TOML
          * ```toml
-         * [default-config]
+         * [default-configs]
          * timeout = { value = 30, schema = { type = "integer" } }
          *
          * [dimensions]
          * os = { position = 1, schema = { type = "string" } }
          *
-         * [[context]]
-         * _condition_ = { os = "linux" }
+         * [[overrides]]
+         * _context_ = { os = "linux" }
          * timeout = 60
          * ```
          */
--- a/clients/java/bindings/src/test/kotlin/TomlFunctionsTest.kt
+++ b/clients/java/bindings/src/test/kotlin/TomlFunctionsTest.kt
@@ -19,7 +19,7 @@ class TomlFunctionsTest {
     companion object {
         // Sample TOML configuration - ride-sharing pricing example
         private const val EXAMPLE_TOML = """
-[default-config]
+[default-configs]
 per_km_rate = { "value" = 20.0, "schema" = { "type" = "number" } }
 surge_factor = { "value" = 0.0, "schema" = { "type" = "number" } }
 
@@ -28,24 +28,24 @@ city = { position = 1, schema = { "type" = "string", "enum" = ["Bangalore", "Del
 vehicle_type = { position = 2, schema = { "type" = "string", "enum" = [ "auto", "cab", "bike", ] } }
 hour_of_day = { position = 3, schema = { "type" = "integer", "minimum" = 0, "maximum" = 23 }}
 
-[[context]]
-_condition_ = { vehicle_type = "cab" }
+[[overrides]]
+_context_ = { vehicle_type = "cab" }
 per_km_rate = 25.0
 
-[[context]]
-_condition_ = { vehicle_type = "bike" }
+[[overrides]]
+_context_ = { vehicle_type = "bike" }
 per_km_rate = 15.0
 
-[[context]]
-_condition_ = { city = "Bangalore", vehicle_type = "cab" }
+[[overrides]]
+_context_ = { city = "Bangalore", vehicle_type = "cab" }
 per_km_rate = 22.0
 
-[[context]]
-_condition_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 18 }
+[[overrides]]
+_context_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 18 }
 surge_factor = 5.0
 
-[[context]]
-_condition_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 6 }
+[[overrides]]
+_context_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 6 }
 surge_factor = 5.0
 """
     }
@@ -134,7 +134,7 @@ city = { position = 1, schema = { "type" = "string" } }
         } catch (e: OperationException) {
             println("\n✓ Correctly caught error: ${e.javaClass.simpleName}")
             println("  Message: ${e.message?.take(100)}")
-            assertTrue(e.message?.contains("default-config") == true)
+            assertTrue(e.message?.contains("default-configs") == true)
         }
     }
 }
--- a/clients/javascript/bindings/native-resolver.ts
+++ b/clients/javascript/bindings/native-resolver.ts
@@ -485,21 +485,6 @@ export class NativeResolver {
             simpleLibName = "libsuperposition_core.so";
         }
 
-        const simpleLocalBuildPath = path.resolve(
-            dirname,
-            "..",
-            "..",
-            "..",
-            "..",
-            "target",
-            "release",
-            simpleLibName
-        );
-        if (this.fileExists(simpleLocalBuildPath)) {
-            console.log(`Using simple local build: ${simpleLocalBuildPath}`);
-            return simpleLocalBuildPath;
-        }
-
         // 5. Final fallback - assume it's in the system path
         console.warn(
             `Native library not found in expected locations, trying: ${filename}`
--- a/clients/python/bindings/superposition_bindings/superposition_client.py
+++ b/clients/python/bindings/superposition_bindings/superposition_client.py
@@ -501,7 +501,7 @@ def _uniffi_check_api_checksums(lib):
         raise InternalError("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
     if lib.uniffi_superposition_core_checksum_func_ffi_get_applicable_variants() != 58234:
         raise InternalError("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
-    if lib.uniffi_superposition_core_checksum_func_ffi_parse_toml_config() != 20800:
+    if lib.uniffi_superposition_core_checksum_func_ffi_parse_toml_config() != 1558:
         raise InternalError("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
 
 # A ctypes library to expose the extern-C FFI definitions.
@@ -1643,14 +1643,14 @@ def ffi_parse_toml_config(toml_content: "str") -> "Config":
 
     # Example TOML
     ```toml
-    [default-config]
+    [default-configs]
     timeout = { value = 30, schema = { type = "integer" } }
 
     [dimensions]
     os = { position = 1, schema = { type = "string" } }
 
-    [[context]]
-    _condition_ = { os = "linux" }
+    [[overrides]]
+    _context_ = { os = "linux" }
     timeout = 60
     ```
     """
--- a/clients/python/bindings/test_toml_functions.py
+++ b/clients/python/bindings/test_toml_functions.py
@@ -15,7 +15,7 @@ from superposition_bindings.superposition_client import ffi_parse_toml_config
 
 # Sample TOML configuration - ride-sharing pricing example
 EXAMPLE_TOML = """
-[default-config]
+[default-configs]
 per_km_rate = { "value" = 20.0, "schema" = { "type" = "number" } }
 surge_factor = { "value" = 0.0, "schema" = { "type" = "number" } }
 
@@ -24,24 +24,24 @@ city = { position = 1, schema = { "type" = "string", "enum" = ["Bangalore", "Del
 vehicle_type = { position = 2, schema = { "type" = "string", "enum" = [ "auto", "cab", "bike", ] } }
 hour_of_day = { position = 3, schema = { "type" = "integer", "minimum" = 0, "maximum" = 23 }}
 
-[[context]]
-_condition_ = { vehicle_type = "cab" }
+[[overrides]]
+_context_ = { vehicle_type = "cab" }
 per_km_rate = 25.0
 
-[[context]]
-_condition_ = { vehicle_type = "bike" }
+[[overrides]]
+_context_ = { vehicle_type = "bike" }
 per_km_rate = 15.0
 
-[[context]]
-_condition_ = { city = "Bangalore", vehicle_type = "cab" }
+[[overrides]]
+_context_ = { city = "Bangalore", vehicle_type = "cab" }
 per_km_rate = 22.0
 
-[[context]]
-_condition_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 18 }
+[[overrides]]
+_context_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 18 }
 surge_factor = 5.0
 
-[[context]]
-_condition_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 6 }
+[[overrides]]
+_context_ = { city = "Delhi", vehicle_type = "cab", hour_of_day = 6 }
 surge_factor = 5.0
 """
 
@@ -153,7 +153,7 @@ def test_error_handling():
         },
         {
             "name": "Missing position in dimension",
-            "toml": "[default-config]\nkey1 = { value = 10, schema = { type = \"integer\" } }\n\n[dimensions]\ncity = { schema = { \"type\" = \"string\" } }\n\n[[context]]\n_condition_= {city=\"bangalore\"}\nkey1 = 20"
+            "toml": "[default-configs]\nkey1 = { value = 10, schema = { type = \"integer\" } }\n\n[dimensions]\ncity = { schema = { \"type\" = \"string\" } }\n\n[[overrides]]\n_context_= {city=\"bangalore\"}\nkey1 = 20"
         },
     ]
 
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -19,6 +19,10 @@ use service_utils::{
     helpers::fetch_dimensions_info_map,
     service::types::{AppState, DbConnection, WorkspaceContext},
 };
+use superposition_core::{
+    helpers::{calculate_context_weight, hash},
+    serialize_to_toml,
+};
 use superposition_derives::authorized;
 #[cfg(feature = "high-performance-mode")]
 use superposition_macros::response_error;
@@ -55,10 +59,6 @@ use crate::{
 };
 
 use super::helpers::{apply_prefix_filter_to_config, resolve, setup_query_data};
-use superposition_core::{
-    helpers::{calculate_context_weight, hash},
-    serialize_to_toml,
-};
 
 #[allow(clippy::let_and_return)]
 pub fn endpoints() -> Scope {
@@ -454,7 +454,7 @@ async fn reduce_handler(
     for (key, _) in default_config {
         let contexts = config.contexts;
         let overrides = config.overrides;
-        let default_config = config.default_configs;
+        let default_config = config.default_configs.into_inner();
         config = reduce_config_key(
             &user,
             &mut conn,
@@ -462,7 +462,7 @@ async fn reduce_handler(
             overrides.clone(),
             key.as_str(),
             &dimensions_info_map,
-            (*default_config).clone(),
+            default_config.clone(),
             is_approve,
             &workspace_context,
             &state,
@@ -654,13 +654,8 @@ async fn get_toml_handler(
     let detailed_config =
         generate_detailed_cac(&mut conn, &workspace_context.schema_name)?;
 
-    let toml_str = serialize_to_toml(detailed_config).map_err(|e| {
-        log::error!("Failed to serialize config to TOML: {}", e);
-        superposition::AppError::UnexpectedError(anyhow::anyhow!(
-            "Failed to serialize config to TOML: {}",
-            e
-        ))
-    })?;
+    let toml_str = serialize_to_toml(detailed_config)
+        .map_err(|e| unexpected_error!("Failed to serialize config to TOML: {}", e))?;
 
     let mut response = HttpResponse::Ok();
     add_last_modified_to_header(max_created_at, false, &mut response);
--- a/crates/context_aware_config/src/api/context/validations.rs
+++ b/crates/context_aware_config/src/api/context/validations.rs
@@ -4,7 +4,7 @@ use diesel::{ExpressionMethods, QueryDsl, RunQueryDsl};
 use jsonschema::ValidationError;
 use serde_json::{Map, Value};
 use service_utils::service::types::SchemaName;
-use superposition_core::validations::{compile_schema, validation_err_to_str};
+use superposition_core::validations::{try_into_jsonschema, validation_err_to_str};
 use superposition_macros::{bad_argument, validation_error};
 use superposition_types::{DBConnection, DimensionInfo, database::schema, result};
 
@@ -30,7 +30,7 @@ pub fn validate_override_with_default_configs(
             .get(key)
             .ok_or(bad_argument!("failed to get schema for config key {}", key))?;
 
-        let jschema = compile_schema(schema).map_err(|e| {
+        let jschema = try_into_jsonschema(schema).map_err(|e| {
             log::error!("({key}) schema compilation error: {}", e);
             bad_argument!("Invalid JSON schema")
         })?;
@@ -69,11 +69,8 @@ pub fn validate_context_jsonschema(
     dimension_value: &Value,
     dimension_schema: &Value,
 ) -> result::Result<()> {
-    let dimension_schema = compile_schema(dimension_schema).map_err(|e| {
-        log::error!(
-            "Failed to compile as a Draft-7 JSON schema: {}",
-            e.to_string()
-        );
+    let dimension_schema = try_into_jsonschema(dimension_schema).map_err(|e| {
+        log::error!("Failed to compile as a Draft-7 JSON schema: {}", e);
         bad_argument!("Error encountered: invalid jsonschema for dimension.")
     })?;
 
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -16,7 +16,7 @@ use service_utils::{
         WorkspaceContext,
     },
 };
-use superposition_core::validations::{compile_schema, validation_err_to_str};
+use superposition_core::validations::{try_into_jsonschema, validation_err_to_str};
 use superposition_derives::authorized;
 use superposition_macros::{
     bad_argument, db_error, not_found, unexpected_error, validation_error,
@@ -107,7 +107,7 @@ async fn create_handler(
 
     let schema = Value::from(&default_config.schema);
 
-    let schema_compile_result = compile_schema(&schema);
+    let schema_compile_result = try_into_jsonschema(&schema);
     let jschema = match schema_compile_result {
         Ok(jschema) => jschema,
         Err(e) => {
@@ -234,7 +234,7 @@ async fn update_handler(
     if let Some(ref schema) = req.schema {
         let schema = Value::from(schema);
 
-        let jschema = compile_schema(&schema).map_err(|e| {
+        let jschema = try_into_jsonschema(&schema).map_err(|e| {
             log::info!("Failed to compile JSON schema: {e}");
             bad_argument!("Invalid JSON schema.")
         })?;
--- a/crates/context_aware_config/src/api/dimension/validations.rs
+++ b/crates/context_aware_config/src/api/dimension/validations.rs
@@ -1,12 +1,9 @@
 use std::collections::HashSet;
 
 use diesel::{ExpressionMethods, OptionalExtension, QueryDsl, RunQueryDsl};
-use jsonschema::{JSONSchema, ValidationError};
 use serde_json::{Map, Value};
 use service_utils::{helpers::fetch_dimensions_info_map, service::types::SchemaName};
-use superposition_core::validations::{
-    compile_schema, validate_cohort_schema_structure, validation_err_to_str,
-};
+use superposition_core::validations::validate_cohort_schema_structure;
 use superposition_macros::{unexpected_error, validation_error};
 use superposition_types::{
     DBConnection,
@@ -91,34 +88,6 @@ pub fn validate_position_wrt_dependency(
     Ok(())
 }
 
-/*
-  This step is required because an empty object
-  is also a valid JSON schema. So added required
-  validations for the input.
-*/
-// TODO: Recursive validation.
-
-pub fn validate_jsonschema(
-    validation_schema: &JSONSchema,
-    schema: &Value,
-) -> superposition::Result<()> {
-    // Compile schema first - returns specific error message
-    compile_schema(schema).map_err(|e| {
-        validation_error!("Invalid JSON schema (failed to compile): {:?}", e)
-    })?;
-
-    // Additional validation against the provided meta-schema
-    validation_schema.validate(schema).map_err(|e| {
-        let verrors = e.collect::<Vec<ValidationError>>();
-        validation_error!(
-            "schema validation failed: {}",
-            validation_err_to_str(verrors)
-                .first()
-                .unwrap_or(&String::new())
-        )
-    })
-}
-
 pub fn allow_primitive_types(schema: &Map<String, Value>) -> superposition::Result<()> {
     match schema.get("type").cloned().unwrap_or_default() {
         Value::String(type_val) if type_val != "array" && type_val != "object" => Ok(()),
@@ -134,19 +103,6 @@ pub fn allow_primitive_types(schema: &Map<String, Value>) -> superposition::Resu
             schema
         )),
     }
-}
-
-fn get_cohort_enum_options(schema: &Value) -> superposition::Result<Vec<String>> {
-    let enum_options = schema
-        .get("enum")
-        .and_then(|v| v.as_array())
-        .ok_or_else(|| {
-            validation_error!("Cohort schema must have an 'enum' field of type array")
-        })?
-        .iter()
-        .filter_map(|v| v.as_str().map(str::to_string))
-        .collect::<Vec<String>>();
-    Ok(enum_options)
 }
 
 pub fn does_dimension_exist_for_cohorting(
@@ -250,14 +206,13 @@ pub fn validate_cohort_schema(
     }
 
     // Use shared validation from superposition_core for cohort schema structure
-    validate_cohort_schema_structure(cohort_schema).map_err(|errors| {
-        validation_error!(
-            "schema validation failed: {}",
-            errors.first().unwrap_or(&String::new())
-        )
-    })?;
-
-    let enum_options = get_cohort_enum_options(cohort_schema)?;
+    let enum_options =
+        validate_cohort_schema_structure(cohort_schema).map_err(|errors| {
+            validation_error!(
+                "schema validation failed: {}",
+                errors.first().unwrap_or(&String::new())
+            )
+        })?;
 
     let cohort_schema = cohort_schema.get("definitions").ok_or(validation_error!(
         "Local cohorts require the jsonlogic rules to be written in the `definitions` field. Refer our API docs for examples",
@@ -349,20 +304,19 @@ pub fn validate_cohort_schema(
 
 #[cfg(test)]
 mod tests {
-    use crate::helpers::get_meta_schema;
+    use jsonschema::ValidationError;
     use serde_json::json;
-
-    use super::*;
+    use superposition_core::validations::get_meta_schema;
 
     #[test]
     fn test_get_meta_schema() {
-        let x = get_meta_schema();
+        let x = get_meta_schema().expect("Failed to get meta-schema");
 
         let ok_string_schema = json!({"type": "string", "pattern": ".*"});
         let ok_string_validation = x.validate(&ok_string_schema);
         assert!(ok_string_validation.is_ok());
 
-        let error_object_schema = json!({"type": "object"});
+        let error_object_schema = json!({"type": "objec"});
         let error_object_validation = x.validate(&error_object_schema).map_err(|e| {
             let verrors = e.collect::<Vec<ValidationError>>();
             format!(
@@ -370,6 +324,7 @@ mod tests {
                 verrors.as_slice()
             )
         });
+
         assert!(error_object_validation.is_err_and(|error| error.contains("Bad schema")));
 
         let ok_enum_schema = json!({"type": "string", "enum": ["ENUMVAL"]});
--- a/crates/context_aware_config/src/helpers.rs
+++ b/crates/context_aware_config/src/helpers.rs
@@ -10,13 +10,11 @@ use chrono::Utc;
 use diesel::{ExpressionMethods, QueryDsl, RunQueryDsl, SelectableHelper};
 #[cfg(feature = "high-performance-mode")]
 use fred::interfaces::KeysInterface;
-use jsonschema::JSONSchema;
 use serde_json::{Map, Value, json};
 use service_utils::{
     helpers::{fetch_dimensions_info_map, generate_snowflake_id},
     service::types::{AppState, EncryptionKey, SchemaName, WorkspaceContext},
 };
-use superposition_core::validations::compile_schema;
 use superposition_macros::{db_error, unexpected_error, validation_error};
 #[cfg(feature = "high-performance-mode")]
 use superposition_types::database::schema::event_log::dsl as event_log;
@@ -78,25 +76,10 @@ pub fn parse_headermap_safe(headermap: &HeaderMap) -> HashMap<String, String> {
     req_headers
 }
 
-pub fn get_meta_schema() -> JSONSchema {
-    let my_schema = json!({
-        "type": "object",
-        "properties": {
-            "type": {
-                "enum": ["boolean", "number", "integer", "string", "array", "null"]
-            },
-        },
-        "required": ["type"],
-    });
-
-    compile_schema(&my_schema)
-        .expect("Error encountered: Failed to compile 'context_dimension_schema_value'. Ensure it adheres to the correct format and data type.")
-}
-
-pub fn generate_cac(
+fn get_context_data(
     conn: &mut DBConnection,
     schema_name: &SchemaName,
-) -> superposition::Result<Config> {
+) -> superposition::Result<(Vec<Context>, HashMap<String, Overrides>)> {
     let contexts_vec: Vec<(String, Condition, String, Overrides)> = ctxt::contexts
         .select((ctxt::id, ctxt::value, ctxt::override_id, ctxt::override_))
         .order_by((ctxt::weight.asc(), ctxt::created_at.asc()))
@@ -107,47 +90,49 @@ pub fn generate_cac(
             db_error!(err)
         })?;
     let contexts_vec: Vec<(String, Condition, i32, String, Overrides)> = contexts_vec
-        .iter()
+        .into_iter()
         .enumerate()
         .map(|(index, (id, value, override_id, override_))| {
-            (
-                id.clone(),
-                value.clone(),
-                index as i32,
-                override_id.clone(),
-                override_.clone(),
-            )
+            (id, value, index as i32, override_id, override_)
         })
         .collect();
 
     let mut contexts = Vec::new();
     let mut overrides: HashMap<String, Overrides> = HashMap::new();
 
-    for (id, condition, weight, override_id, override_) in contexts_vec.iter() {
-        let condition = Cac::<Condition>::validate_db_data(condition.clone().into())
+    for (id, condition, weight, override_id, override_) in contexts_vec.into_iter() {
+        let condition = Cac::<Condition>::validate_db_data(condition.into())
             .map_err(|err| {
                 log::error!("generate_cac : failed to decode context from db {}", err);
                 unexpected_error!(err)
             })?
             .into_inner();
 
-        let override_ = Cac::<Overrides>::validate_db_data(override_.clone().into())
+        let override_ = Cac::<Overrides>::validate_db_data(override_.into())
             .map_err(|err| {
                 log::error!("generate_cac : failed to decode overrides from db {}", err);
                 unexpected_error!(err)
             })?
             .into_inner();
         let ctxt = Context {
-            id: id.to_owned(),
+            id,
             condition,
-            priority: weight.to_owned(),
-            weight: weight.to_owned(),
+            priority: weight,
+            weight,
             override_with_keys: OverrideWithKeys::new(override_id.to_owned()),
         };
         contexts.push(ctxt);
-        overrides.insert(override_id.to_owned(), override_);
+        overrides.insert(override_id, override_);
     }
 
+    Ok((contexts, overrides))
+}
+
+pub fn generate_cac(
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+) -> superposition::Result<Config> {
+    let (contexts, overrides) = get_context_data(conn, schema_name)?;
     let default_config_vec = def_conf::default_configs
         .select((def_conf::key, def_conf::value))
         .schema_name(schema_name)
@@ -181,62 +166,7 @@ pub fn generate_detailed_cac(
     conn: &mut DBConnection,
     schema_name: &SchemaName,
 ) -> superposition::Result<DetailedConfig> {
-    let contexts_vec: Vec<(String, Condition, String, Overrides)> = ctxt::contexts
-        .select((ctxt::id, ctxt::value, ctxt::override_id, ctxt::override_))
-        .order_by((ctxt::weight.asc(), ctxt::created_at.asc()))
-        .schema_name(schema_name)
-        .load::<(String, Condition, String, Overrides)>(conn)
-        .map_err(|err| {
-            log::error!("failed to fetch contexts with error: {}", err);
-            db_error!(err)
-        })?;
-    let contexts_vec: Vec<(String, Condition, i32, String, Overrides)> = contexts_vec
-        .iter()
-        .enumerate()
-        .map(|(index, (id, value, override_id, override_))| {
-            (
-                id.clone(),
-                value.clone(),
-                index as i32,
-                override_id.clone(),
-                override_.clone(),
-            )
-        })
-        .collect();
-
-    let mut contexts = Vec::new();
-    let mut overrides: HashMap<String, Overrides> = HashMap::new();
-
-    for (id, condition, weight, override_id, override_) in contexts_vec.iter() {
-        let condition = Cac::<Condition>::validate_db_data(condition.clone().into())
-            .map_err(|err| {
-                log::error!(
-                    "generate_detailed_cac : failed to decode context from db {}",
-                    err
-                );
-                unexpected_error!(err)
-            })?
-            .into_inner();
-
-        let override_ = Cac::<Overrides>::validate_db_data(override_.clone().into())
-            .map_err(|err| {
-                log::error!(
-                    "generate_detailed_cac : failed to decode overrides from db {}",
-                    err
-                );
-                unexpected_error!(err)
-            })?
-            .into_inner();
-        let ctxt = Context {
-            id: id.to_owned(),
-            condition,
-            priority: weight.to_owned(),
-            weight: weight.to_owned(),
-            override_with_keys: OverrideWithKeys::new(override_id.to_owned()),
-        };
-        contexts.push(ctxt);
-        overrides.insert(override_id.to_owned(), override_);
-    }
+    let (contexts, overrides) = get_context_data(conn, schema_name)?;
 
     // Fetch default_configs with both value and schema
     let default_config_vec = def_conf::default_configs
@@ -256,8 +186,7 @@ pub fn generate_detailed_cac(
         },
     );
 
-    let dimensions: HashMap<String, DimensionInfo> =
-        fetch_dimensions_info_map(conn, schema_name)?;
+    let dimensions = fetch_dimensions_info_map(conn, schema_name)?;
 
     Ok(DetailedConfig {
         contexts,
--- a/crates/superposition_core/src/ffi.rs
+++ b/crates/superposition_core/src/ffi.rs
@@ -168,14 +168,14 @@ fn ffi_get_applicable_variants(
 ///
 /// # Example TOML
 /// ```toml
-/// [default_configs]
+/// [default-configs]
 /// timeout = { value = 30, schema = { type = "integer" } }
 ///
 /// [dimensions]
 /// os = { position = 1, schema = { type = "string" } }
 ///
-/// [[contexts]]
-/// _condition_ = { os = "linux" }
+/// [[overrides]]
+/// _context_ = { os = "linux" }
 /// timeout = 60
 /// ```
 #[uniffi::export]
--- a/crates/superposition_core/src/toml.rs
+++ b/crates/superposition_core/src/toml.rs
@@ -21,8 +21,8 @@ use toml::Value as TomlValue;
 use crate::{
     helpers::{build_context, create_connections_with_dependents},
     toml::helpers::{
-        context_toml_to_condition, format_key, format_toml_value, overrides_toml_to_map,
-        toml_to_json,
+        format_key, format_toml_value, toml_to_json, try_condition_from_toml,
+        try_overrides_from_toml,
     },
     validations,
 };
@@ -32,6 +32,12 @@ use crate::{
 pub enum TomlError {
     TomlSyntaxError(String),
     InvalidDimension(String),
+    InvalidCohortDimensionPosition {
+        dimension: String,
+        dimension_position: i32,
+        cohort_dimension: String,
+        cohort_dimension_position: i32,
+    },
     UndeclaredDimension {
         dimension: String,
         context: String,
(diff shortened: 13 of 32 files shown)


clients/python/bindings/superposition_bindings/superposition_client.py
```diff
@@ -476,12 +480,34 @@ def _uniffi_load_indirect():
     if not triple:
         raise RuntimeError(f"❌ Unsupported platform: {sys.platform} / {platform.machine()}")
 
-    libname = f"libsuperposition_core-{triple}"
-    libpath = os.path.join(folder, libname)
-    if not os.path.exists(libpath):
-        raise FileNotFoundError(f"❌ Required binary not found: {libpath}")
+    # Determine simple library name (without platform triple)
```
knutties: This needs to be removed as this is an auto-generated file and needs to be handled outside potentially.

clients/python/bindings/superposition_bindings/superposition_types.py
```diff
@@ -443,12 +444,34 @@ def _uniffi_load_indirect():
     if not triple:
         raise RuntimeError(f"❌ Unsupported platform: {sys.platform} / {platform.machine()}")
 
-    libname = f"libsuperposition_core-{triple}"
-    libpath = os.path.join(folder, libname)
-    if not os.path.exists(libpath):
-        raise FileNotFoundError(f"❌ Required binary not found: {libpath}")
+    # Determine simple library name (without platform triple)
```
knutties: Remove changes to this file as it is auto-generated by uniffi.

crates/superposition_core/src/ffi.rs
```diff
@@ -156,3 +156,29 @@ fn ffi_get_applicable_variants(
...
+/// ```toml
+/// [default-config]
+/// timeout = { value = 30, schema = { type = "integer" } }
+///
+/// [dimensions]
+/// os = { schema = { type = "string" } }
+///
+/// [context]
+/// "os=linux" = { timeout = 60 }
+/// ```
```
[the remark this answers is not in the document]
knutties: this is resolved.

clients/javascript/bindings/README_TOML_TESTS.md
```diff
@@ -0,0 +1,280 @@
+# JavaScript TOML Binding Tests
+
+This directory contains JavaScript/Node.js bindings for the TOML parsing functions using the C FFI (Foreign Function Interface) implementation.
+
+> **Note**: JavaScript is not supported by uniffi, so these bindings use the `ffi_legacy` C FFI interface instead.
+
+> **⚠️ Node.js Compatibility**: The `ffi-napi` and `ref-napi` packages have compatibility issues with Node.js v24+. It's recommended to use **Node.js v18 or v20 LTS** for running these bindings. You can use a version manager like `nvm` to switch versions:
+> ```bash
+> nvm use 20  # or nvm use 18
+> ```
```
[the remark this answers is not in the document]
knutties: this is resolved.

clients/javascript/bindings/README_TOML_TESTS.md
```diff
@@ -0,0 +1,280 @@
...
+```bash
+node test.js
+```
+
+## Architecture
+
+The JavaScript bindings use:
+- **ffi-napi**: Node.js FFI (Foreign Function Interface) library for calling C functions
+- **ref-napi**: Library for creating and dereferencing pointers
+- **C FFI Functions**: `core_parse_toml_config` and `core_eval_toml_config` from `ffi_legacy.rs`
```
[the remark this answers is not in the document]
knutties: this is resolved.

clients/python/bindings/README_TOML_TESTS.md
```diff
@@ -0,0 +1,138 @@
...
+======================================================================
+  TEST SUMMARY
+======================================================================
+  ✓ Parse TOML
+  ✓ Eval TOML
+  ✓ External File
+
+  Total: 3/3 tests passed
+======================================================================
+```
```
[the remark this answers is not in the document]
knutties: resolved.

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1546 @@
...
+        Value::Object(obj) => {
+            let items: Vec<String> = obj
+                .iter()
+                .map(|(k, v)| format!("{} = {}", k, value_to_toml(v)))
+                .collect();
+            format!("{{ {} }}", items.join(", "))
+        }
+        Value::Null => "null".to_string(),
+    }
+}
```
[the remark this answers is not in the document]
knutties: resolved.

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1546 @@
...
+    match value {
+        Value::String(s) => {
+            format!("\"{}\"", s.replace('\\', "\\\\").replace('"', "\\\""))
+        }
+        Value::Number(n) => n.to_string(),
+        Value::Bool(b) => b.to_string(),
+        Value::Array(arr) => {
+            let items: Vec<String> = arr.iter().map(|v| value_to_toml(v)).collect();
+            format!("[{}]", items.join(", "))
+        }
```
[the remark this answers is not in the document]
knutties: resolved.

.gitignore
```diff
@@ -67,4 +67,9 @@ clients/generated/smithy/typescript/dist-types/
...
 
 .npmrc
-.zed
\ No newline at end of file
+.zed
+
+# Dynamic libraries copied for testing
+*.dylib
+*.so
+*.dll
```
Datron, line 75: For testing, should we link to libraries in target dir?

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -99,3 +108,26 @@ getResolvedConfig params = do
...
     (Nothing, []) -> Left "null pointer returned"
     _ -> Left err
+
+-- | Parse TOML configuration string into structured format
+-- Returns JSON with:
+--   - default_config: object with configuration key-value pairs
+--   - contexts_json: JSON string containing array of context objects
+--   - overrides_json: JSON string containing object mapping hashes to override configurations
+--   - dimensions_json: JSON string containing object mapping dimension names to dimension info
+parseTomlConfig :: String -> IO (Either String String)
```
Datron: If we're returning an object, can we parse it to Aeson.Value and return instead of String - without this the user would need to parse the config

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -99,3 +108,26 @@ getResolvedConfig params = do
...
+parseTomlConfig tomlContent = do
+  ebuf <- callocBytes 2048  -- Error buffer size matches Rust implementation
+  tomlStr <- newCString tomlContent
+  res <- parse_toml_config tomlStr ebuf
+  err <- peekCAString ebuf
```
Datron: Use peekCString which supports utf8 (for emojis)

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -99,3 +108,25 @@ getResolvedConfig params = do
...
+  let peekMaybe p | p /= nullPtr = Just <$> peekCAString p
+                  | otherwise = pure Nothing
+  result <- peekMaybe res
+  free tomlStr
+  free ebuf
+  pure $ case (result, err) of
+    (Just cfg, []) -> Right cfg
+    (Nothing, []) -> Left "null pointer returned"
+    _ -> Left err
```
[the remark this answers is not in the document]
Datron: Create a new ForeignPtr that has a finalizer, and then use withForeignPtr 
https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Foreign-ForeignPtr.html#v:withForeignPtr

clients/java/bindings/src/main/kotlin/uniffi/superposition_client/superposition_client.kt
```diff
@@ -1736,4 +1750,36 @@ public object FfiConverterMapStringTypeOverrides: FfiConverterRustBuffer<Map<kot
...
+         *
+         * [context."os=linux"]
+         * timeout = 60
+         * ```
+         */
+    @Throws(OperationException::class) fun `ffiParseTomlConfig`(`tomlContent`: kotlin.String): Config {
+            return FfiConverterTypeConfig.lift(
+    uniffiRustCallWithError(OperationException) { _status ->
+    UniffiLib.INSTANCE.uniffi_superposition_core_fn_func_ffi_parse_toml_config(
+        FfiConverterString.lower(`tomlContent`),_status)
```
Datron, line 1780: @knutties this should be generated by uniFFI, no need to manually code this

clients/java/bindings/README_TOML_TESTS.md
```diff
@@ -0,0 +1,217 @@
+# Kotlin/Java TOML Binding Tests
```
Datron, line 1: We don't need this file

clients/java/bindings/src/main/kotlin/uniffi/superposition_client/superposition_client.kt
Datron: This is a guess, based on the code comments

clients/java/bindings/src/test/kotlin/TomlFunctionsTest.kt
```diff
@@ -0,0 +1,135 @@
+package uniffi.superposition_client.test
```
Datron, line 1: Well have tests in a separate directory, should this be a part of those tests?

clients/javascript/bindings/README_TOML_TESTS.md
```diff
@@ -0,0 +1,258 @@
+# JavaScript TOML Binding Tests
```
Datron, line 1: We can probably remove this file

clients/python/bindings/README_TOML_TESTS.md
```diff
@@ -0,0 +1,138 @@
+# Python TOML Binding Tests
```
Datron, line 1: We don't need this file

clients/python/bindings/test_toml_functions.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
```
Datron, line 1: lets move this to the test suite

design-docs/2025-12-21-toml-parsing-ffi-design.md
```diff
@@ -0,0 +1,1860 @@
+# TOML Parsing FFI Interface Design
```
Datron: should we commit this?

If yes should we move it to docs so that it can be served by docusaurus

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,1850 @@
...
+        }
+        toml::Value::Table(table) => {
+            let mut map = Map::new();
+            for (k, v) in table {
+                map.insert(k, toml_value_to_serde_value(v));
+            }
+            Value::Object(map)
+        }
+    }
+}
```
Datron: You don't need this function, serde should let you move between toml and json easily. You can just use json! macro on the toml Value type and see what results that presents. I think this should do:
```
serde_json::to_value(toml_value)
```

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+/// Parse a TOML string value as JSON (for object type values stored as triple-quoted strings)
+fn parse_toml_string_as_json(toml_value: toml::Value) -> Value {
+    match toml_value {
+        toml::Value::String(s) => {
+            // Try to parse as JSON
+            serde_json::from_str(&s).unwrap_or(Value::String(s))
+        }
+        _ => toml_value_to_serde_value(toml_value),
+    }
+}
```
Datron: Can be replaced with:
```
serde_json::to_string_pretty(&toml_value)
```

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,1546 @@
...
+        Value::Object(map) => {
+            let items: Vec<String> = map
+                .iter()
+                .sorted_by_key(|(k, _)| *k)
+                .map(|(k, v)| format!("\"{}\":{}", k, json_to_sorted_string(v)))
+                .collect();
+            format!("{{{}}}", items.join(","))
+        }
+    }
+}
```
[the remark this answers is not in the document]
Datron: The JSON in our case is going to be an Object which is Map<String, Value>. Can you get consistent hashing just by sorting on Map keys?

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+) -> Result<Map<String, Value>, TomlError> {
+    let mut result = Map::new();
+
+    for pair in input.split(';') {
+        let pair = pair.trim();
+        if pair.is_empty() {
+            continue;
+        }
+
+        let parts: Vec<&str> = pair.splitn(2, '=').collect();
```
Datron: Since this is always key and value in superposition v2, should it be a tuple here?

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+            return Err(TomlError::DuplicatePosition {
+                position,
+                dimensions,
+            });
+        }
+    }
+
+    // Second pass: parse dimension types and validate schemas based on type
+    for (key, value) in section {
+        let table = value.as_table().unwrap();
```
Datron: Conversion error missing here

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+            let escaped = s
+                .replace('\\', "\\\\")
+                .replace('"', "\\\"")
+                .replace('\n', "\\n")
+                .replace('\r', "\\r")
+                .replace('\t', "\\t");
+            format!("\"{}\"", escaped)
+        }
+        Value::Number(n) => n.to_string(),
+        Value::Bool(b) => b.to_string(),
```
Datron: You don't need these functions, use the serde converters directly

Datron (changes requested):

makefile
```diff
@@ -402,3 +403,49 @@ test-kotlin-provider: provider-template
 test-rust-provider: provider-template
 	cargo test --package superposition_provider --test integration_test -- --nocapture --ignored
 	-@pkill -f $(CARGO_TARGET_DIR)/debug/superposition
+
+# Target to run all TOML bindings tests
+bindings-test: uniffi-bindings
```
ayushjain17, line 421:
```suggestion
toml-bindings-test: uniffi-bindings
```

crates/context_aware_config/Cargo.toml
```diff
@@ -22,11 +22,11 @@ itertools = { workspace = true }
 jsonlogic = { workspace = true }
 jsonschema = { workspace = true }
 log = { workspace = true }
-num-bigint = "0.4"
 serde = { workspace = true }
 serde_json = { workspace = true }
 service_utils = { workspace = true }
 superposition_derives = { workspace = true }
+superposition_core = { workspace = true }
```
ayushjain17:
```suggestion
superposition_core = { workspace = true }
superposition_derives = { workspace = true }
```

crates/superposition_core/src/config.rs
```diff
@@ -42,7 +42,7 @@ pub fn eval_config(
     )?;
 
     // Apply overrides to default config
-    let mut result_config = config.default_configs;
+    let mut result_config = (*config.default_configs).clone();
     merge_overrides_on_default_config(&mut result_config, overrides_map, &merge_strategy);
 
     Ok(result_config)
```
ayushjain17: we can avoid cloning in all of the places, we need to implement a function like this for `ExtendedMap`
```rust
impl ExtendedMap {
    pub fn into_inner(self) -> Map<String, Value> {
        self.0
    }
}
```
optionally we should add this function as well
```rust
impl ExtendedMap {
    pub fn inner(&self) -> &Map<String, Value> {
        &self.0
    }
}
```

and then the current code can become something like this
```suggestion
    let mut result_config = config.default_configs;
    merge_overrides_on_default_config(&mut result_config, overrides_map, &merge_strategy);

    Ok(result_config.into_inner())
```

we need to carry out a similar change/refactor in other places as well for the other custom types, to avoid unnecessay clones

crates/superposition_core/src/config.rs
```diff
@@ -90,7 +90,7 @@ pub fn eval_config_with_reasoning(
         Some(&mut reasoning_collector),
     )?;
 
-    let mut result_config = config.default_configs;
+    let mut result_config = (*config.default_configs).clone();
     merge_overrides_on_default_config(&mut result_config, overrides_map, &merge_strategy);
```
ayushjain17: same would be applicable here as well, revert this section to original code, and just while returning, 
```rust
   Ok(result_config.into_inner())
```

crates/superposition_types/src/config.rs
```diff
@@ -419,7 +419,7 @@ impl Config {
         Self {
             contexts: filtered_context,
             overrides: filtered_overrides,
-            default_configs: filtered_default_config,
+            default_configs: ExtendedMap(filtered_default_config),
```
ayushjain17: instead of changing the type over here, I think we should do it in `filter_default_by_prefix` fn itself

changing that fucntion to something like:
```rust
    pub fn filter_default_by_prefix(&self, prefix_list: &HashSet<String>) -> ExtendedMap {
        filter_config_keys_by_prefix(&self.default_configs, prefix_list).into()
    }
```

crates/superposition_types/src/overridden.rs
```diff
@@ -58,15 +58,15 @@ mod tests {
 
         assert_eq!(
             filter_config_keys_by_prefix(&config.default_configs, &prefix_list),
-            get_prefix_filtered_config1().default_configs
+            (*get_prefix_filtered_config1().default_configs).clone()
```
ayushjain17: these can also take leverage of `into_inner`

crates/superposition_provider/src/utils.rs
```diff
@@ -584,7 +584,7 @@ impl ConversionUtils {
         // Convert default_configs
         result.insert(
             "default_configs".to_string(),
-            Value::Object(config.default_configs.clone()),
+            Value::Object((*config.default_configs).clone()),
```
ayushjain17, line 587: this function is very weird, all the input config's fields are entirely cloned inside over here, there is no point in borrowing the ipnut, it should have taken the entire ownership itself

crates/superposition_core/src/lib.rs
```diff
@@ -12,5 +15,59 @@ pub use experiment::{
 pub use ffi_legacy::{
     core_free_string, core_get_resolved_config, core_get_resolved_config_with_reasoning,
 };
+pub use superposition_types::Config as ParsedTomlConfig;
```
ayushjain17: why this alias ?

design-docs/2025-12-21-toml-parsing-ffi-design.md
ayushjain17: same question

ayushjain17 (commented):
still reviewing

docs/api/toml-response-format.md
ayushjain17, line 1: what is this doc intended for ?
is this supposed to be published to github pages ?

because I dont think this folder is currently being published, correct me if I am wrong here

nix/rust.nix
```diff
@@ -330,6 +330,17 @@
               };
             };
           };
+          "superposition_toml_example" = {
+            crane = {
+              args = {
+                buildInputs =
+                  [
+                    pkgs.openssl
+                    pkgs.postgresql_15
```
ayushjain17, line 339: this should not have psql dependencies, no ?

uniffi/patches/python.patch
ayushjain17, line 1: can we revert the formatting changes in the file, not sure if it will impact its application

tests/bun.lockb
ayushjain17, line 1: I think this file needs to rebased correctly, this has been deleted

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -148,59 +146,22 @@ pub fn validate_context_jsonschema(
...
-                                .unwrap_or(&String::new())
-                        ))
-                    }
-                }
-            }
-        }
-        _ => dimension_schema.validate(dimension_value).map_err(|e| {
-            let verrors = e.collect::<Vec<ValidationError>>();
+    // Use shared validation from superposition_core for simple cases
+    #[cfg(not(feature = "jsonlogic"))]
```
ayushjain17: this is repeating the code unnecessarily, I dont think any change in this function was required

crates/superposition_core/src/lib.rs
```diff
@@ -12,5 +15,59 @@ pub use experiment::{
...
+///
+///     [context]
+///     "os=linux" = { timeout = 60 }
+/// "#;
+///
+/// let parsed = parse_toml_config(toml_content)?;
+/// println!("Parsed {} contexts", parsed.contexts.len());
+/// # Ok::<(), superposition_core::TomlError>(())
+/// ```
+pub fn parse_toml_config(toml_content: &str) -> Result<ParsedTomlConfig, TomlError> {
```
ayushjain17: do we need this ?
it seems like it is only adding documentation on an external function and is not actually providing any encapsulation

examples/superposition_toml_example/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "superposition_toml_example"
+description = "Example demonstrating TOML parsing with superposition_core"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+superposition_core = { path = "../../crates/superposition_core" }
```
ayushjain17, line 8: ?
```suggestion
superposition_core = { workspace = true }
```

examples/superposition_toml_example/example.toml
```diff
@@ -0,0 +1,23 @@
+[default-config]
+per_km_rate = { "value" = 20.0, "schema" = { "type" = "number" } }
+surge_factor = { "value" = 0.0, "schema" = { "type" = "number" } }
+
+[dimensions]
+city = { position = 1, schema = { "type" = "string", "enum" = ["Bangalore", "Delhi"] } }
+vehicle_type = { position = 2, schema = { "type" = "string", "enum" = [ "auto", "cab", "bike", ] } }
+hour_of_day = { position = 3, schema = { "type" = "integer", "minimum" = 0, "maximum" = 23 }}
```
ayushjain17: is this file updated ?
we had decided to keep dimension type as well, right ?
and the parses should fail if the required fields are missing, right ?

then this should be failing as of now, no ?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -612,11 +637,27 @@ async fn get_handler(
...
+    match response_format {
+        ResponseFormat::Toml => {
+            let toml_str = serialize_to_toml(&config).map_err(|e| {
+                log::error!("Failed to serialize config to TOML: {}", e);
+                superposition::AppError::UnexpectedError(anyhow::anyhow!(
+                    "Failed to serialize config to TOML: {}",
+                    e
+                ))
+            })?;
+            response.insert_header(("Content-Type", "application/toml"));
```
ayushjain17:
```suggestion
            response.insert_header((header::CONTENT_TYPE, "application/toml"));
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -612,11 +637,27 @@ async fn get_handler(
...
-    Ok(response.json(config))
+
+    match response_format {
+        ResponseFormat::Toml => {
+            let toml_str = serialize_to_toml(&config).map_err(|e| {
+                log::error!("Failed to serialize config to TOML: {}", e);
+                superposition::AppError::UnexpectedError(anyhow::anyhow!(
+                    "Failed to serialize config to TOML: {}",
+                    e
+                ))
```
ayushjain17:
```suggestion
                unexpected_error!(
                    "Failed to serialize config to TOML: {}",
                    e
                )
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -612,11 +637,27 @@ async fn get_handler(
         config = config.filter_by_dimensions(&context);
     }
 
+    let response_format = determine_response_format(&req);
```
ayushjain17: instead of calling this here like a function
can we do it the way it is done for `MergeStrategy`, moving all the type logic to the same file `superposition_types/src/api/config.rs`

following the same way, the header can used as:
```
async fn get_handler(
    ...
    response_format: Header<ConfigResponseFormat>,
    ...
```

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -57,6 +57,31 @@ use crate::{
 };
 
 use super::helpers::{apply_prefix_filter_to_config, resolve, setup_query_data};
+use superposition_core::serialize_to_toml;
+
+/// Supported response formats for get_config
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum ResponseFormat {
```
ayushjain17: can we also move it to `superosition_types/src/api/config.rs`
```suggestion
enum ConfigResponseFormat {
```

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -227,7 +229,7 @@ async fn update_handler(
             let verrors = e.collect::<Vec<ValidationError>>();
             validation_error!(
                 "Schema validation failed: {}",
-                validation_err_to_str(verrors)
+                &validation_err_to_str(verrors)
```
ayushjain17, line 246: just a question, this would have worked either ways, right ?

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,642 @@
...
+                "type": "object",
+                "not": {
+                    "required": ["otherwise"]
+                }
+            }
+        },
+        "required": ["type", "enum", "definitions"]
+    });
+
+    JSONSchema::options()
```
ayushjain17: can we use the newly created `compile_schema` fn here

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,642 @@
...
+        "type": "object",
+        "properties": {
+            "type": {
+                "enum": ["boolean", "number", "integer", "string", "array", "null", "object"]
+            },
+        },
+        "required": ["type"],
+    });
+
+    JSONSchema::options()
```
ayushjain17: can we use the newly created `compile_schema` fn here

can we do the same for the other places as well, currently, `compile_schema` seems to be used in one place only

crates/context_aware_config/src/api/dimension/validations.rs
```diff
@@ -309,7 +268,25 @@ pub fn validate_cohort_schema(
         ));
     }
 
-    let enum_options = validate_cohort_jsonschema(cohort_schema)?;
+    log::error!(
+        "##### checking validate_cohort_schema_structure {}",
```
ayushjain17: should we better format these logs, if we at all want to keep them (both this one and the following one) ?

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,642 @@
...
+    // Use the new compile function
+    compile_schema(schema).map_err(|e| vec![e])?;
+
+    // Then validate against the meta-schema
+    let meta_schema = get_meta_schema();
+    meta_schema
+        .validate(schema)
+        .map_err(|errors| errors.map(|e| e.to_string()).collect())
+}
+
```
ayushjain17, line 97: are we sure, we need these functions ?
not able to verify there usage properly

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,642 @@
...
+/// cohort schemas contain jsonlogic rules in the `definitions` field which
+/// are not valid JSON Schema syntax.
+///
+/// # Arguments
+/// * `schema` - The cohort schema to validate
+///
+/// # Returns
+/// * `Ok(())` if the schema structure is valid
+/// * `Err(Vec<String>)` containing validation error messages
+pub fn validate_cohort_schema_structure(schema: &Value) -> Result<(), Vec<String>> {
```
ayushjain17: why do we need this manual check of all the fields ?
JSONSchema validation was already doing the exact same thing

crates/superposition_core/src/toml_parser.rs
ayushjain17: +1

crates/superposition_core/src/toml.rs
ayushjain17, line 1: there are direct unwrap usage in the this file - 4 instance other than usage in tests (test ones are fine)

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -452,3 +452,73 @@ pub unsafe extern "C" fn core_get_applicable_variants(
...
+        }
+    };
+
+    // Create result with default_config as Map and others as JSON strings
+    let result = serde_json::json!({
+        "default_config": &*parsed.default_configs,
+        "contexts_json": parsed.contexts,
+        "overrides_json": parsed.overrides,
+        "dimensions_json": parsed.dimensions,
+    });
```
ayushjain17: I dont think we are doing what this comment is suggesting

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -452,3 +452,73 @@ pub unsafe extern "C" fn core_get_applicable_variants(
...
+
+    let result_str = match serde_json::to_string(&result) {
```
ayushjain17: we could have simply done this
```suggestion
    let result_str = match serde_json::to_string(&config) {
```

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+            || c == '"'
+            || c == '\''
+            || c == '.'
+    })
+}
+
+/// Detailed error type for TOML parsing and serialization
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum TomlError {
+    FileReadError(String),
```
ayushjain17: dead variant

crates/superposition_core/src/toml.rs
ayushjain17: an intermediate type would have helped a lot, serde would have handled 90% of the work, we would have needed effort only for 10%
also, the top level fields do not have the key names consistent with the get config's fields

then something like this could have been done:
```rust
    let value: Config = toml::from_str(FILE).unwrap();

    println!("{}", serde_json::to_string_pretty(&value).unwrap());
```

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+                    r#"{ type = "integer" }"#
+                } else {
+                    r#"{ type = "number" }"#
+                }
+            }
+            Value::Bool(_) => r#"{ type = "boolean" }"#,
+            Value::Array(_) => r#"{ type = "array" }"#,
+            Value::Object(_) => r#"{ type = "object" }"#,
+            Value::Null => r#"{ type = "null" }"#,
+        };
```
ayushjain17: our input to this function should have been something else
inferring types like this would lead to loss of all the info someone would have saved in their default config's schema

crates/superposition_core/src/toml.rs
ayushjain17: the intermediate Config could be something like:
```rust
// for DimensionInfo - existing one can be reused

struct DefaultConfigInfo {
    schema: ExtendedMap, // or Map<String, Value>
    value: serde_json::Value,
}

// the key for contexts map would have to be pre-processed  => (key, value) -> sort -> stringify with custom logic using `=` and `;`
struct TomlConfig {
    contexts: HashMap<String, ExtendedMap>, // or HashMap<String, Map<String, Value>>,
    default_config: HashMap<String, DefaultConfigInfo>,
    dimensions: HashMap<String, DimensionInfo>,
}
```

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,1850 @@
...
+            || c == '#'
+            || c == '['
+            || c == ']'
+            || c == '{'
+            || c == '}'
+            || c == '"'
+            || c == '\''
+            || c == '.'
+    })
+}
```
sauraww: TOML spec mentions that bare keys can only contain: 
ASCII letters: A-Z, a-z
ASCII digits: 0-9
ASCII underscore: _
ASCII hyphen: -
Reference: https://toml.io/en/v1.1.0 , keys section. 

Example of unicode edge cases : 
1. "café" - actually it should return false but the code will return true. 

2. "key＝value" - imagine using a full width = , not the ascii one 
Fullwidth ＝ (U+FF1D)
this will also return false but it should return true. 

So, instead we should return true if ANY character is NOT in the allowed set for bare keys

```
fn needs_quoting(s: &str) -> bool {
    s.chars().any(|c| {
        !c.is_ascii_alphanumeric() && c != '_' && c != '-'
    })
}
```

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,1850 @@
...
+/// Parse a TOML string value as JSON (for object type values stored as triple-quoted strings)
+fn parse_toml_string_as_json(toml_value: toml::Value) -> Value {
+    match toml_value {
+        toml::Value::String(s) => {
+            // Try to parse as JSON
+            serde_json::from_str(&s).unwrap_or(Value::String(s))
+        }
+        _ => toml_value_to_serde_value(toml_value),
+    }
+}
```
sauraww, line 136: Not logging parse failures?

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1850 @@
...
+        }
+    }
+}
+
+/// Parse a TOML string value as JSON (for object type values stored as triple-quoted strings)
+fn parse_toml_string_as_json(toml_value: toml::Value) -> Value {
+    match toml_value {
+        toml::Value::String(s) => {
+            // Try to parse as JSON
+            serde_json::from_str(&s).unwrap_or(Value::String(s))
```
sauraww:
```
serde_json::from_str(&s).map_err(|e| {
    TomlError::ConversionError(format!("Invalid JSON in triple-quoted string: {}", e))
})
```

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,1850 @@
+use std::collections::HashMap;
```
sauraww: Why two #[cfg(all(test, not(feature = "jsonlogic")))] mod serialization_tests blocks. Can we merge them to avoid confusion?

sauraww: value_to_toml() and schema_to_toml() share ~90% of their logic. The only difference is object handling (triple-quoted JSON vs inline tables). Can we consider extracting common logic?

crates/superposition_core/src/config.rs
sauraww: +1

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -99,3 +108,26 @@ getResolvedConfig params = do
     (Just cfg, []) -> Right cfg
```
sauraww, line 121: The traverse_ freeNonNull list is missing di (dimensionInfo), any specific reason ? 
Isn't this memory leak ? 
Out of scope of this PR but curious.

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -99,3 +108,26 @@ getResolvedConfig params = do
...
+
+-- | Parse TOML configuration string into structured format
+-- Returns JSON with:
+--   - default_config: object with configuration key-value pairs
+--   - contexts_json: JSON string containing array of context objects
+--   - overrides_json: JSON string containing object mapping hashes to override configurations
+--   - dimensions_json: JSON string containing object mapping dimension names to dimension info
+parseTomlConfig :: String -> IO (Either String String)
+parseTomlConfig tomlContent = do
+  ebuf <- callocBytes 2048  -- Error buffer size matches Rust implementation
```
sauraww: The error buffer is hardcoded to 2048 bytes with no validation. If the Rust code writes beyond this, it's undefined behavior , 2048 is a magic number , possible to define a constant for this ?

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -35,6 +35,15 @@ foreign import capi "superposition_core.h core_get_resolved_config"
     -- | resolved config json
```
sauraww, line 45: typo at line 23 , it should be dimension_info_json

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# OPTIONS_GHC -Wno-name-shadowing #-}
 
-module FFI.Superposition (getResolvedConfig, ResolveConfigParams (..), defaultResolveParams, MergeStrategy (..)) where
+module FFI.Superposition (getResolvedConfig, ResolveConfigParams (..), defaultResolveParams, MergeStrategy (..), parseTomlConfig) where
```
sauraww, line 5: importing prelude is not needed.

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -99,3 +108,26 @@ getResolvedConfig params = do
...
+--   - contexts_json: JSON string containing array of context objects
+--   - overrides_json: JSON string containing object mapping hashes to override configurations
+--   - dimensions_json: JSON string containing object mapping dimension names to dimension info
+parseTomlConfig :: String -> IO (Either String String)
+parseTomlConfig tomlContent = do
+  ebuf <- callocBytes 2048  -- Error buffer size matches Rust implementation
+  tomlStr <- newCString tomlContent
+  res <- parse_toml_config tomlStr ebuf
+  err <- peekCAString ebuf
+  let peekMaybe p | p /= nullPtr = Just <$> peekCAString p
```
sauraww: parseTomlConfig checks res for null after peeking, but getResolvedConfig does it correctly. Can we reorder the operations in parseTomlConfig?

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -31,30 +34,25 @@ pub fn validate_override_with_default_configs(
...
-                return Err(bad_argument!(
-                    "failed to compile ({}) config key schema",
-                    key
-                ));
-            }
-        };
-        if let Err(e) = jschema.validate(instance) {
-            let verrors = e.collect::<Vec<ValidationError>>();
+
+        let jschema = jsonschema::JSONSchema::options()
```
sauraww: Can we precompile schemas outside the loop ? It is doing a lot of wasteful recompilation. 
Precompilation can really help if let's say we have 10 unique schemas for 1000 overrides. It is like caching.

ayushjain17: Pre compiling requires Lazy static, I am not sure if Lazy static is by default thread safe or it would have to be wrapped in Arc Mutex

clients/java/bindings/src/main/kotlin/uniffi/superposition_client/superposition_client.kt
knutties: This file was auto-generated and not hand-written @Datron  - if that is what you are referring to.

crates/superposition_core/src/toml_parser.rs
knutties: Removing this function completely as we are going with inline toml table itself.

crates/superposition_core/src/toml_parser.rs
knutties: While I might not need all the cases - I will still need to match and return string versions of each serde variant.  I will explain this in the call tomorrow.

makefile
knutties: actually we should do all the bindings test here at a later point and not just toml bindings.  So prefer to keep this as is.

crates/context_aware_config/src/api/dimension/validations.rs
knutties: removing this log.

crates/superposition_core/src/validations.rs
knutties: removed `validate_against_compiled_schema` as it was not used separately anywhere else - maybe refactoring from an earlier iteration.

crates/superposition_core/src/validations.rs
knutties: this does the additional validation checking if the definitions and enums have the same keys defined and there is no mismatch either way - which was already there.  Pure schema validation does not catch that.

crates/superposition_core/src/toml_parser.rs
knutties: This is just so wrong.  Realized this as I was going over the implementation myself :)  Saw that you had noticed as well.

crates/superposition_core/src/toml.rs
knutties: This portions is gone now.  Will push them.

nix/rust.nix
knutties: Just following earlier examples of crane usage here - need to dig deeper on why we need to specify this for all workspace crates.

crates/superposition_core/src/toml.rs
knutties: This is a good one - thanks @sauraww .

crates/superposition_core/src/toml.rs
knutties: both are gone now - replaced with json_to_toml.

.gitignore
knutties: @Datron  - yes, we are linking to libraries in target dir for the `bindings-test` make target - are you saying we should or we should not?

Datron: @knutties I assumed we were copying the libs from target to root dir for testing since the comment
```
# Dynamic libraries copied for testing
```
If we are linking with binaries in target dir then we can remove this gitignore. Though this is a nitpick and need not be done

crates/superposition_core/src/toml_parser.rs
```diff
@@ -0,0 +1,1546 @@
...
+fn compute_priority(
+    context_map: &Map<String, Value>,
+    dimensions: &HashMap<String, DimensionInfo>,
+) -> i32 {
+    context_map
+        .keys()
+        .filter_map(|key| dimensions.get(key))
+        .map(|dim_info| 1 << dim_info.position)
+        .sum()
+}
```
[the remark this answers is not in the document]
knutties: resolved.

.gitignore
knutties: I think for python it copies the library instead of using the target path - so is useful for now.  Need to see if the Python code can be changed to pick from the target itself.

crates/superposition_provider/src/utils.rs
knutties: This function does not appear to be used at all - not sure why the warning does not show up.  We can clean-up later maybe ?

crates/superposition_core/src/toml.rs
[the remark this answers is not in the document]
knutties: This is addressed by using serde native functions instead of manual stringifying.

crates/context_aware_config/src/api/context/validations.rs
knutties: removed.

docs/api/toml-response-format.md
knutties: deleted the docs.

design-docs/2025-12-21-toml-parsing-ffi-design.md
sauraww: this is pending.

docs/plans/2026-01-02-toml-response-format.md
```diff
@@ -0,0 +1,908 @@
+# TOML Response Format Implementation Plan
```
sauraww, line 1: Can we remove this too ?

examples/superposition_toml_example/Cargo.toml
sauraww: this is also pending

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -623,9 +625,50 @@ async fn get_handler(
...
+    if is_not_modified(max_created_at, &req) {
+        return Ok(HttpResponse::NotModified().finish());
+    }
+
+    let detailed_config =
+        generate_detailed_cac(&mut conn, &workspace_context.schema_name)?;
+
+    let toml_str = serialize_to_toml(&detailed_config).map_err(|e| {
+        log::error!("Failed to serialize config to TOML: {}", e);
+        superposition::AppError::UnexpectedError(anyhow::anyhow!(
```
sauraww: unexpected_error!("Failed to serialize config to TOML: {}", e)

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -28,30 +29,22 @@ pub fn validate_override_with_default_configs(
...
-                ));
-            }
-        };
-        if let Err(e) = jschema.validate(instance) {
-            let verrors = e.collect::<Vec<ValidationError>>();
+
+        let jschema = compile_schema(schema).map_err(|e| {
+            log::error!("({key}) schema compilation error: {}", e);
+            bad_argument!("Invalid JSON schema")
+        })?;
```
sauraww: Can we move it out of for loop , every validation recompiles the schema which would be expensive.

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -76,16 +69,13 @@ pub fn validate_context_jsonschema(
...
-            log::error!(
-                "Failed to compile as a Draft-7 JSON schema: {}",
-                e.to_string()
-            );
-            bad_argument!("Error encountered: invalid jsonschema for dimension.")
-        })?;
+    let dimension_schema = compile_schema(dimension_schema).map_err(|e| {
+        log::error!(
+            "Failed to compile as a Draft-7 JSON schema: {}",
+            e.to_string()
```
sauraww: .to_string() not required.

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -28,30 +29,22 @@ pub fn validate_override_with_default_configs(
...
+        })?;
+
+        jschema.validate(value).map_err(|e| {
+            let verrors = e.collect::<Vec<jsonschema::ValidationError>>();
             log::error!("({key}) config key validation error: {:?}", verrors);
-            return Err(validation_error!(
+            validation_error!(
                 "schema validation failed for {key}: {}",
-                validation_err_to_str(verrors)
+                &validation_err_to_str(verrors)
```
sauraww, line 43: this is highly inefficient - allocates an empty string unnecessarily when there are no errors. returns a reference when we likely needed owned. 
We can use this : 
Moves, no extra alloc and returns String

```suggestion
    validation_err_to_str(errors)
        .into_iter()
        .next()
        .unwrap_or_default()
```
Also , this pattern is repeated in the code for 3 times.

crates/context_aware_config/src/api/context/validations.rs
knutties: That is what we need here right ? because each key in the override might have its own schema - the for loop iterates over the different keys of the override.

sauraww: @knutties I meant can we precompile this outside the loop , Precompilation can really help if let's say we have 10 unique schemas for 1000 overrides. 
[ref comment](https://github.com/juspay/superposition/pull/796#discussion_r2682349954)

crates/context_aware_config/src/api/context/validations.rs
```diff
@@ -28,30 +29,22 @@ pub fn validate_override_with_default_configs(
...
-        };
-        if let Err(e) = jschema.validate(instance) {
-            let verrors = e.collect::<Vec<ValidationError>>();
+
+        let jschema = compile_schema(schema).map_err(|e| {
+            log::error!("({key}) schema compilation error: {}", e);
+            bad_argument!("Invalid JSON schema")
+        })?;
+
+        jschema.validate(value).map_err(|e| {
```
sauraww:
```suggestion
    let compiled_schemas: HashMap<String, JSONSchema> = res
        .into_iter()
        .map(|(key, schema)| {
            let compiled = compile_schema(&schema).map_err(|e| {
                log::error!("({key}) schema compilation error: {}", e);
                bad_argument!("Invalid JSON schema for key {}", key)
            })?;
            Ok((key, compiled))
        })
        .collect::<result::Result<HashMap<_, _>>>()?;

    for (key, value) in override_.iter() {
        let jschema = compiled_schemas.get(key).ok_or_else(|| {
            bad_argument!("failed to get schema for config key {}", key)
        })?;

        jschema.validate(value).map_err(|e| {
            let verrors = e.collect::<Vec<jsonschema::ValidationError>>();
            log::error!("({key}) config key validation error: {:?}", verrors);
            validation_error!(
                "schema validation failed for {key}: {}",
                validation_err_to_str(verrors)
                    .into_iter()
                    .next()
                    .unwrap_or_default()
            )
        })?;
    }
```

crates/context_aware_config/src/api/context/validations.rs
sauraww: We can actually leave it , it will only help if we precompile and cache it for across the calls. We can maybe make an issue and tackle it later on. 
https://github.com/juspay/superposition/pull/796#discussion_r2682661713

crates/superposition_core/src/toml/helpers.rs
```diff
@@ -0,0 +1,161 @@
+use std::collections::HashMap;
+
+use serde_json::{Map, Value};
+use superposition_types::{DefaultConfigsWithSchema, DimensionInfo};
+use toml::Value as TomlValue;
+
+use crate::{validations, TomlError};
+
+/// Convert toml::Value to serde_json::Value for validation
+pub fn toml_to_json(value: TomlValue) -> Value {
```
knutties, line 8: We might be able to remove this as well with native toml deserialization.  Will do that shortly.

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,498 @@
...
+            .get(context.override_with_keys.get_key())
+            .map(|ov| ov.clone().into_iter().collect())
+            .unwrap_or_default();
+
+        Self {
+            context: TomlValue::try_from(context_map)
+                .expect("Context should not contain null values"),
+            overrides: TomlValue::try_from(overrides_map)
+                .expect("Overrides should not contain null values"),
+        }
```
ayushjain17: this is wrong, you cannot use `expect` in this, expect will crash the server, instead of propagating the error to the user
to achieve this, you should have rather converted this into a `TryFrom` implementation

also, this over here assumes that this would have failed only because there was a `null` value which is a poor assumption

ayushjain17 (changes requested):

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,498 @@
...
+    DimensionType::default().to_string()
+}
+
+impl From<DimensionInfo> for DimensionInfoToml {
+    fn from(d: DimensionInfo) -> Self {
+        Self {
+            position: d.position,
+            schema: TomlValue::try_from(d.schema.into_inner())
+                .expect("Schema should not contain null values"),
+            dimension_type: d.dimension_type.to_string(),
```
ayushjain17: here as well

crates/superposition_core/src/toml/helpers.rs
```diff
@@ -0,0 +1,95 @@
...
+            ))
+        }
+    };
+    Cac::<Condition>::try_from(map)
+        .map(|cac| cac.into_inner())
+        .map_err(|e| TomlError::ConversionError(format!("Invalid condition: {}", e)))
+}
+
+pub fn overrides_toml_to_map(overrides: &toml::Table) -> Result<Overrides, TomlError> {
+    let json = toml_to_json(TomlValue::Table(overrides.clone()));
```
ayushjain17:
```suggestion
pub fn try_overrides_from_toml(overrides: toml::Table) -> Result<Overrides, TomlError> {
    let json = toml_to_json(TomlValue::Table(overrides));
```

crates/superposition_core/src/toml/helpers.rs
```diff
@@ -0,0 +1,95 @@
...
+                .iter()
+                .map(|(k, v)| format!("{} = {}", format_key(k), format_toml_value(v)))
+                .collect();
+            format!("{{ {} }}", entries.join(", "))
+        }
+    }
+}
+
+pub fn context_toml_to_condition(ctx: &toml::Table) -> Result<Condition, TomlError> {
+    let json = toml_to_json(TomlValue::Table(ctx.clone()));
```
ayushjain17:
```suggestion
pub fn try_condition_from_toml(ctx: toml::Table) -> Result<Condition, TomlError> {
    let json = toml_to_json(TomlValue::Table(ctx));
```

crates/superposition_core/src/toml.rs
```diff
@@ -0,0 +1,518 @@
...
+        // Check for duplicate positions
+        for (position, dimensions) in position_to_dimensions {
+            if dimensions.len() > 1 {
+                return Err(TomlError::DuplicatePosition {
+                    position,
+                    dimensions,
+                });
+            }
+        }
+
```
ayushjain17, line 415:
```suggestion
            let condition = try_condition_from_toml(ctx.context)?;
            let override_vals = try_overrides_from_toml(ctx.overrides)?;
```

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,779 @@
...
+///
+/// # Returns
+/// * `Ok(())` if the schema is valid
+/// * `Err(Vec<String>)` containing validation error messages
+pub fn validate_schema(schema: &Value) -> Result<(), Vec<String>> {
+    // Use the new compile function
+    compile_schema(schema).map_err(|e| vec![e])?;
+
+    // Then validate against the meta-schema
+    let meta_schema = get_meta_schema();
```
ayushjain17: this schema should have been used from state

crates/context_aware_config/src/helpers.rs
```diff
@@ -90,46 +89,10 @@ pub fn get_meta_schema() -> JSONSchema {
         "required": ["type"],
```
ayushjain17: this function is moved to `superposition_core/src/validation.rs`

crates/superposition_core/src/toml/helpers.rs
```diff
@@ -0,0 +1,95 @@
...
+        format!("\"{}\"", key.replace('"', r#"\"#))
+    } else {
+        key.to_string()
+    }
+}
+
+/// Format a TOML value as a string for inline table usage
+pub fn format_toml_value(value: &TomlValue) -> String {
+    match value {
+        TomlValue::String(s) => format!("\"{}\"", s.replace('"', r#"\"#)),
```
ayushjain17, line 48: does toml allow usage of `'` to quote strings ?

makefile
```diff
@@ -413,3 +415,50 @@ test-kotlin-provider: provider-template
 test-rust-provider: provider-template
 	cargo test --package superposition_provider --test integration_test -- --nocapture --ignored
 	$(MAKE) kill
+	-@pkill -f $(CARGO_TARGET_DIR)/debug/superposition
+
+# Target to run all TOML bindings tests
+bindings-test: uniffi-bindings
+	@echo ""
+	@echo ""
+	@echo "========================================"
```
ayushjain17, line 424: should all of these have been in some shell script file ?

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,779 @@
...
+///
+/// # Arguments
+/// * `schema` - The cohort schema to validate
+///
+/// # Returns
+/// * `Ok(())` if the schema structure is valid
+/// * `Err(Vec<String>)` containing validation error messages
+pub fn validate_cohort_schema_structure(schema: &Value) -> Result<(), Vec<String>> {
+    // Get the cohort meta-schema
+    let cohort_meta_schema = get_cohort_meta_schema();
```
ayushjain17: this should come from app_state or somewhere, as this code can panic on runtime

crates/superposition_core/src/toml/helpers.rs
knutties: Yes - it does https://toml.io/en/v1.1.0#string:~:text=Literal%20strings%20are%20surrounded%20by%20single%20quotes.%20Like%20basic%20strings%2C%20they%20must%20appear%20on%20a%20single%20line%3A

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,779 @@
...
+            return Err(vec![format!(
+                "Cohort definition '{}' does not have a corresponding enum option",
+                key
+            )]);
+        }
+        if key == "otherwise" {
+            return Err(vec![
+                "Cohort definitions should not contain 'otherwise'".to_string()
+            ]);
+        }
```
ayushjain17: this is a not reachable case due to schema validation

crates/superposition_core/src/validations.rs
```diff
@@ -0,0 +1,779 @@
...
+        .iter()
+        .filter_map(|v| v.as_str().map(str::to_string))
+        .collect::<Vec<String>>();
+
+    // Check that "otherwise" is in the enum
+    if !enum_options.contains(&"otherwise".to_string()) {
+        return Err(vec![
+            "Cohort schema enum must contain 'otherwise' as an option".to_string(),
+        ]);
+    }
```
ayushjain17: this is a not reachable case due to schema validation

clients/python/bindings/test_toml_functions.py
ayushjain17: +1

clients/javascript/bindings/native-resolver.ts
```diff
@@ -411,7 +475,32 @@ export class NativeResolver {
             return localBuildPath;
         }
 
-        // 4. Final fallback - assume it's in the system path
+        // 4. Try simple library name format (libsuperposition_core.dylib/so/dll)
+        let simpleLibName: string;
```
ayushjain17, line 479: why is this needed now ?

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -94,8 +116,36 @@ getResolvedConfig params = do
...
+  ebuf <- callocBytes errorBufferSize  
+  tomlStr <- newCString tomlContent
+  res <- parse_toml_config tomlStr ebuf
+  errBytes <- packCString ebuf 
+  let errText = fromRight mempty $ decodeUtf8' errBytes
+  result <- if res /= nullPtr
+    then do 
+      resFptr <- newForeignPtr p_free_string res 
+      -- Registers p_free_string as the finalizer (for automatic cleanup)
+      withForeignPtr resFptr $ \ptr -> Just <$> packCString ptr
```
ayushjain17:
```suggestion
      withForeignPtr resFptr $ fmap Just . packCString
```

crates/superposition_core/src/toml/helpers.rs
knutties: Present in your diff.

crates/superposition_core/src/toml/helpers.rs
knutties: present in your diff.

crates/superposition_core/src/toml.rs
knutties: present in your diff.

crates/superposition_core/src/validations.rs
knutties: can we do this later please ?

makefile
knutties: Will move this to directory specific makefiles in my makefile refactor.

crates/superposition_core/src/validations.rs
knutties: let us do this separately - since we have injected the schema the risk is low.  I want to understand and do this and not force it in.

crates/superposition_core/src/validations.rs
knutties: there in your diff.

crates/superposition_core/src/validations.rs
knutties: there in your diff.

crates/superposition_core/src/validations.rs
ayushjain17: yeah we can even ditch the field from app state altogether as that is not needed, I was being cautious as the function internally uses `expect` if we could just change the return type to a `Result` for both the functions then that would be way better

clients/javascript/bindings/native-resolver.ts
knutties: This is for local build time testing.  Not great actually - we should have had a build step to move it to a standard path and then run this locally possibly.  This is lazy code - will fix it.

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
knutties: done

crates/superposition_core/src/validations.rs
```diff
@@ -343,6 +335,31 @@ pub fn validate_config_value(
...
+/// * `cohort_dimension_info` - Information about the parent cohort dimension
+///
+/// # Returns
+/// * `Ok(())` if the position is valid
+/// * `Err(String)` containing an error message if the position is invalid
+pub fn validate_cohort_dimension_position(
+    dimension_info: &DimensionInfo,
+    cohort_dimension_info: &DimensionInfo,
+) -> Result<(), String> {
+    if dimension_info.position > cohort_dimension_info.position {
```
ayushjain17: `>=`

crates/superposition_core/src/validations.rs
```diff
@@ -343,6 +335,31 @@ pub fn validate_config_value(
...
+        return Err(format!(
+            "Dimension position {} should be less than or equal to cohort dimension position {}",
```
ayushjain17: only `less than`

crates/superposition_core/src/toml.rs
```diff
@@ -328,6 +344,23 @@ impl TryFrom<DetailedConfigToml> for DetailedConfig {
...
+                        &dim_info,
+                        cohort_dimension_info,
+                    )
+                    .map_err(|_| {
+                        TomlError::InvalidCohortDimensionPosition {
+                            dimension: dim.clone(),
+                            dimension_position: dim_info.position,
+                            cohort_dimension: cohort_dim.to_string(),
+                            cohort_dimension_position: cohort_dimension_info.position,
+                        }
```
ayushjain17, line 361: this error type response should have been sent from the function, later when the function might be validating multiple conditions, this would require a change here as well, which can be avoided if this was already inside the function

knutties: Will leave this one for now as we do not have the dimension names inside the validation function.

PR #797, opened 2025-12-23
feat: introduce redis for caching configurations

Datron:
## Problem
Provider APIs are dependent on DB connections and DB uptime, and tend to be slower.

## Solution
Introduce redis as a cache store for get config response, get experiments list and get experiment groups list.

REDIS_URL=""
REDIS_POOL_SIZE=""
REDIS_MAX_ATTEMPTS=""
REDIS_CONN_TIMEOUT=""

## Summary by CodeRabbit

* **New Features**
  * Added Redis-backed caching for configuration, experiments, and experiment groups with automatic database fallback
  * Introduced new C APIs for core configuration resolution and variant retrieval

* **Chores**
  * Added ENCRYPTED_KEYS environment variable to configuration
  * Made Redis a required dependency across services

* **Infrastructure**
  * Enabled Redis service in Docker Compose


crates/service_utils/src/redis.rs
```diff
@@ -0,0 +1,110 @@
...
+where
+    T: Serialize + DeserializeOwned,
+{
+    if redis_pool.is_none() {
+        log::trace!("Redis pool not configured, using fallback");
+        return database_call(db_pool);
+    }
+    let pool = redis_pool.ok_or(unexpected_error!(
+        "Could not access redis pool, this message should never be seen",
+    ))?;
```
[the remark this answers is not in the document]
Datron: Not a valid comment, these lines are only unreachable if redis URL is undefined, turning off the redis feature

.env.example
```diff
@@ -38,6 +38,14 @@ AUTH_PROVIDER=DISABLED
 AUTH_Z_PROVIDER=DISABLED
 WORKER_ID=1
 # MASTER_ENCRYPTION_KEY - add this for enabling secrets in local
+ENCRYPTED_KEYS="" # Used for webhook secrets, API keys, etc.
+# ENCRYPTED_KEYS="SERVICE1_API_KEY,SERVICE2_API_KEY"
```
ayushjain17: this is deprecated
```suggestion
```

crates/superposition/src/app_state.rs
```diff
@@ -41,41 +38,45 @@ pub async fn get(
...
-        .unwrap();
+    // Initialize Redis pool only if REDIS_URL is explicitly set (not default)
+    let redis_pool = match std::env::var("REDIS_URL") {
+        Ok(redis_url) if !redis_url.is_empty() => {
+            let redis_pool_size = get_from_env_or_default("REDIS_POOL_SIZE", 10);
+            let redis_max_attempts = get_from_env_or_default("REDIS_MAX_ATTEMPTS", 10);
+            let redis_connection_timeout =
+                get_from_env_or_default("REDIS_CONN_TIMEOUT", 1000);
+            let config = RedisConfig::from_url(&redis_url).unwrap_or_else(|_| {
+                panic!("Failed to create RedisConfig from url {}", redis_url)
```
mahatoankitkumar, line 47: Should we panic here in this anywhere or fall back to DB-only mode

crates/service_utils/src/service.rs
```diff
@@ -1 +1,26 @@
...
+) -> superposition::Result<DbConnection> {
+    match db_pool.get() {
+        Ok(mut conn) => {
+            conn.set_prepared_statement_cache_size(
+                diesel::connection::CacheSize::Disabled,
+            );
+            Ok(DbConnection(conn))
+        }
+        Err(e) => {
+            log::info!("Unable to get db connection from pool, error: {e}");
```
mahatoankitkumar: log::error?

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
```diff
@@ -369,30 +425,35 @@ async fn delete_handler(
...
+            }
+            diesel::delete(experiment_groups::experiment_groups)
+                .filter(experiment_groups::id.eq(&id))
+                .schema_name(&schema_name)
+                .execute(conn)?;
+            Ok(Json(marked_group))
+        });
+    put_experiment_groups_in_redis(state.redis.clone(), &mut db_conn, &schema_name)
+        .await?;
+    result
```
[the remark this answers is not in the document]
mahatoankitkumar: This should be handled i guess

crates/context_aware_config/src/api/dimension/handlers.rs
```diff
@@ -603,14 +608,16 @@ async fn delete_handler(
...
+        if let Err(e) = put_config_in_redis(
+            version_id,
             state,
             &workspace_context.schema_name,
             &mut conn,
         )
-        .await?;
+        .await
+        {
+            log::warn!("Failed to update redis cache with new context: {}", e);
```
mahatoankitkumar: Move this log inside the put_config_in_redis

crates/context_aware_config/src/helpers.rs
```diff
@@ -240,45 +244,85 @@ pub fn add_config_version(
...
+            false,
+        )
         .await;
     if let Ok(uuid) = event_log::event_log
         .select(event_log::id)
         .filter(event_log::table_name.eq("contexts"))
+        .schema_name(schema_name)
         .order_by(event_log::timestamp.desc())
         .first::<Uuid>(db_conn)
     {
```
mahatoankitkumar: Already have a function for this is guess

.env.example
Datron: done

crates/context_aware_config/src/api/dimension/handlers.rs
Datron: Sure

crates/context_aware_config/src/helpers.rs
Datron: The function `fetch_audit_id` converts it to string, and this segment has been written in a few places. That can be refactored but this is not a blocker

crates/service_utils/src/service.rs
Datron: Done

crates/superposition/src/app_state.rs
Datron: We should panic, falling back to DB only mode would give a false sense of working to end users

crates/service_utils/src/redis.rs
```diff
@@ -0,0 +1,111 @@
...
+    match get_data_from_redis(key.clone(), client).await {
+        Ok(data) => Ok(data),
+        Err(e) => {
+            log::info!(
+                "Falling back to DB for schema {} due to Redis error: {}",
+                **schema_name,
+                e
+            );
+            let data = run_query(&db_pool, query_fn);
+            if let Ok(ref value) = data {
```
sauraww: Can we flatten the control flow , this is 3 levels deep nesting. 
We are also mixing two separate concerns ( writeback is a separate concern ) , better to split this into two methods. This also replaces map_err logging trick with straightforward match. Also ,for if let Err — no more converting errors to () just to use if let Ok.

```
pub async fn fetch_from_redis_else_writeback<T>(
    key: String,
    schema_name: &SchemaName,
    redis_pool: Option<RedisPool>,
    db_pool: PgSchemaConnectionPool,
    query_fn: impl FnOnce(&mut DBConnection) -> superposition::DieselResult<T>,
) -> superposition::Result<T>
where
    T: Serialize + DeserializeOwned,
{
    let Some(pool) = redis_pool else {
        log::trace!("Redis pool not configured, using fallback");
        return run_query(&db_pool, query_fn);
    };

    let client = pool.next_connected();

    match get_data_from_redis(key.clone(), client).await {
        Ok(data) => return Ok(data),
        Err(e) => log::info!(
            "Falling back to DB for schema {} due to Redis error: {}",
            **schema_name,
            e
        ),
    }

    let data = run_query(&db_pool, query_fn)?;
    writeback_to_redis(client, key, &data).await;
    Ok(data)
}

/// Best-effort write of a value to Redis. Errors are logged but never propagated.
async fn writeback_to_redis<T: Serialize>(client: &RedisClient, key: String, value: &T) {
    let serialized = match serde_json::to_string(value) {
        Ok(s) => s,
        Err(e) => {
            log::error!("Failed to serialize data for redis writeback: {}", e);
            return;
        }
    };

    let key_ttl: i64 = get_from_env_or_default("REDIS_KEY_TTL", 604800);
    let expiration = Some(Expiration::EX(key_ttl));

    if let Err(e) = client
        .set::<(), String, String>(key, serialized, expiration, None, false)
        .await
    {
        log::error!("Failed to write back data to redis: {}", e);
    }
}

```

clients/java/sdk/src/main/java/io/juspay/superposition/client/SuperpositionAsyncClientImpl.java
```diff
@@ -267,8 +267,8 @@
 final class SuperpositionAsyncClientImpl extends Client implements SuperpositionAsyncClient {
     private static final TypeRegistry TYPE_REGISTRY = TypeRegistry.builder()
         .putType(NotAuthorizedException.$ID, NotAuthorizedException.class, NotAuthorizedException::builder)
-        .putType(AccessDeniedException.$ID, AccessDeniedException.class, AccessDeniedException::builder)
```
sauraww, line 270: We can revert these changes.

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -62,16 +63,26 @@ use super::helpers::{apply_prefix_filter_to_config, resolve, setup_query_data};
...
+    conn: &mut DBConnection,
+    schema_name: &SchemaName,
+) -> DieselResult<String> {
+    event_log::event_log
+        .select(event_log::id)
+        .filter(event_log::table_name.eq("contexts"))
+        .order_by(event_log::timestamp.desc())
+        .schema_name(schema_name)
+        .first::<Uuid>(conn)
+        .map(|uuid| uuid.to_string())
```
sauraww:
```suggestion
        .map(String::from)
```

crates/superposition/Cargo.toml
```diff
@@ -17,7 +17,8 @@ context_aware_config = { path = "../context_aware_config" }
 diesel = { workspace = true }
 dotenv = "0.15.0"
 experimentation_platform = { path = "../experimentation_platform" }
-fred = { workspace = true, optional = true }
+superposition_core = { workspace = true }
+fred = { workspace = true }
```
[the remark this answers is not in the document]
sauraww: this doesn't change the compiled behaviour , superposition already imports all the sibling crates since it is a server binary. wherever it is important (i.e service utils, etc), metrics feature is already present. 
If we want , we can add it purely for consistency purpose.

crates/superposition/src/workspace/handlers.rs
```diff
@@ -211,10 +218,44 @@ async fn update_handler(
...
         })?;
+
+    put_workspace_in_redis(updated_workspace.clone(), &app_state, &schema_name.0).await;
+
     let response = WorkspaceResponse::from(updated_workspace);
     Ok(Json(response))
 }
 
+async fn put_workspace_in_redis<T>(
+    workspace: T,
```
sauraww: Why do we need this to be generic ? Can we rather make it reusable ? workspace's type is already known

crates/superposition/src/workspace/handlers.rs
```diff
@@ -393,6 +438,10 @@ pub async fn rotate_encryption_key_handler(
             )
         })?;
 
+    // Refetch workspace after transaction to get updated data
```
sauraww, line 436: Can we rather change the return type above to return workspace_settings and use that instead of call get_workspace again ?

crates/context_aware_config/src/api/config/helpers.rs
```diff
@@ -35,28 +36,41 @@ pub fn apply_prefix_filter_to_config(
...
+            None => fetch_from_redis_else_writeback::<i64>(
+                format!(
+                    "{}{CONFIG_VERSION_KEY_SUFFIX}",
+                    *workspace_context.schema_name
+                ),
+                &workspace_context.schema_name,
+                state.redis.clone(),
+                state.db_pool.clone(),
+                |conn| {
+                    config_versions::config_versions
```
sauraww, line 60: Can we reuse this function here: generate_config_from_version

Datron: This is getting the config version ID, not the config. `generate_config_from_version` gives the whole config and is used in other places where the config is cached

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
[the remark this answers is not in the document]
Datron: handled

crates/service_utils/src/redis.rs
Datron: This function is a bit obtuse, let me see if I can flatten it a bit

crates/superposition/src/workspace/handlers.rs
Datron: Removed generics 👍

crates/superposition/src/workspace/handlers.rs
Datron: Its better to fetch it again, since if another transaction has made a change prior/during/after the above transaction those changes will be cached too and we avoid a dirty read problem

crates/service_utils/src/redis.rs
```diff
@@ -0,0 +1,111 @@
...
+    redis_pool: Option<RedisPool>,
+    db_pool: PgSchemaConnectionPool,
+    query_fn: impl FnOnce(&mut DBConnection) -> superposition::DieselResult<T>,
+) -> superposition::Result<T>
+where
+    T: Serialize + DeserializeOwned,
+{
+    let Some(pool) = redis_pool else {
+        log::trace!("Redis pool not configured, using fallback");
+        return run_query(&db_pool, query_fn);
```
ayushjain17: ideally run_query should not be used here, it should have been kept entirely localised to the query
this is what the expectation was from while writing the function

having run_query here, opens the problem wide for granular connection usage

crates/superposition/Cargo.toml
```diff
@@ -43,11 +44,7 @@ tracing-actix-web = { workspace = true }
 json-subscriber = { version = "0.2.7", features = ["tracing-log"] }
 
 [features]
```
ayushjain17:
```suggestion
```

docker-compose.yaml
```diff
@@ -23,14 +23,14 @@ services:
...
-#     command: redis-server
-#     network_mode: bridge
+  redis:
+    image: redis:7
+    container_name: superposition_redis
+    restart: on-failure
+    ports:
+      - 6379:6379
+    command: redis-server
+    network_mode: bridge
```
ayushjain17, line 21: we are enabling redis in local by default, then better we keep the envs for it in `.env.example` non-commented

crates/superposition_core/include/superposition_core.h
ayushjain17, line 1: I am slightly confused, where is this coming from ?
why is this here ?

crates/superposition/Cargo.toml
```diff
@@ -17,7 +17,8 @@ context_aware_config = { path = "../context_aware_config" }
 diesel = { workspace = true }
 dotenv = "0.15.0"
 experimentation_platform = { path = "../experimentation_platform" }
-fred = { workspace = true, optional = true }
+superposition_core = { workspace = true }
```
ayushjain17: can we have it sorted

crates/service_utils/src/service.rs
```diff
@@ -1 +1,26 @@
+use crate::db::PgSchemaConnectionPool;
+use diesel::Connection;
+use superposition_macros::unexpected_error;
+use superposition_types::result as superposition;
+use types::DbConnection;
+
 pub mod types;
+
+pub fn get_db_connection(
```
ayushjain17: we already have this, and this is not meant to be exposed, its there in db.rs file
and where this is having a need to be used, ideally run_query should have been directly used

this feels like we are going round in circles

docker-compose.yaml
Datron: Done, uncommented this in .env.example

crates/superposition/src/workspace/handlers.rs
```diff
@@ -174,6 +179,7 @@ async fn create_handler(
 async fn update_handler(
     workspace_name: web::Path<String>,
     request: Json<UpdateWorkspaceRequest>,
+    app_state: web::Data<AppState>,
```
ayushjain17: we already have Data imported in the file
```suggestion
    app_state: Data<AppState>,
```

crates/superposition_core/include/superposition_core.h
Datron: Unsure, removing it now

crates/superposition/src/workspace/handlers.rs
```diff
@@ -211,10 +217,42 @@ async fn update_handler(
...
         })?;
+
+    put_workspace_in_redis(updated_workspace.clone(), &app_state, &schema_name.0).await;
+
     let response = WorkspaceResponse::from(updated_workspace);
     Ok(Json(response))
 }
 
+async fn put_workspace_in_redis(
+    workspace: Workspace,
```
ayushjain17: this can simply be
```suggestion
    workspace: &Workspace,
```

the only consumption of this variable in this function is also as a reference
all the clones done while using this function can be entirely avoided

crates/superposition/src/resolve/types.rs
```diff
@@ -2,6 +2,8 @@ use serde::Deserialize;
 use superposition_derives::{IsEmpty, QueryParam};
 use superposition_types::{IsEmpty, custom_query::QueryParam};
 
+/// Query param for targeting key or the identifier to be used in experiments.
+/// Also known as toss
```
ayushjain17, line 6: not exactly, we dont have it at all exposed as toss anymore, right?

crates/service_utils/Cargo.toml
```diff
@@ -31,18 +31,18 @@ secrecy = { workspace = true }
...
     "result",
     "api",
     "diesel_derives",
 ] }
+superposition_macros = { workspace = true }
 url = { workspace = true }
 urlencoding = "~2.1.2"
 tracing-actix-web = { workspace = true }
 
 [features]
```
ayushjain17:
```suggestion
```

crates/service_utils/Cargo.toml
```diff
@@ -31,18 +31,18 @@ secrecy = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum_macros = { workspace = true }
-superposition_macros = { workspace = true }
 superposition_types = { workspace = true, features = [
     "result",
     "api",
     "diesel_derives",
 ] }
+superposition_macros = { workspace = true }
```
ayushjain17: was ordered better previously

crates/service_utils/src/service/types.rs
```diff
@@ -189,18 +188,12 @@ impl FromRequest for DbConnection {
...
-                    diesel::connection::CacheSize::Disabled,
-                );
-                Ok(DbConnection(conn))
-            }
-            Err(e) => {
-                log::info!("Unable to get db connection from pool, error: {e}");
-                Err(error::ErrorInternalServerError(""))
-            }
-        };
+        let result = super::get_db_connection(app_state.db_pool.clone()).map_err(|e| {
```
ayushjain17: we dont need the change here

crates/superposition/src/resolve/types.rs
Datron: Its only for colloquial use, so that juspay folks understand the code

crates/service_utils/src/service.rs
Datron: Sure will resolve

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -435,6 +443,54 @@ pub async fn fetch_cac_config(
     }
 }
 
+pub async fn fetch_webhook_by_event(
```
ayushjain17: this is not used anymore, we dont need this
you can delete this function

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
```diff
@@ -447,6 +523,12 @@ async fn backfill_handler(
             }
             Ok(results)
         })?;
+    let _ = put_experiment_groups_in_redis(
+        state.redis.clone(),
+        &mut conn,
+        &workspace_context.schema_name,
+    )
```
sauraww, line 530: Can we handle the failure here ?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -986,15 +1029,30 @@ async fn list_handler(
...
+                .select(diesel::dsl::max(event_log::timestamp))
+                .schema_name(&workspace_context.schema_name)
+                .first(&mut conn)
+                .map_err(|e| {
+                    log::error!("failed to fetch max timestamp from event_log: {e}");
+                    unexpected_error!("failed to fetch max timestamp from event_log: {e}")
+                })
+        },
+    )
+    .await?;
```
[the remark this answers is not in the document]
sauraww: Can we look into this ?

crates/context_aware_config/src/helpers.rs
```diff
@@ -226,46 +230,102 @@ pub fn add_config_version(
     Ok(version_id)
 }
 
-#[cfg(feature = "high-performance-mode")]
+pub fn get_workspace(
```
sauraww: this is redundant , this isn't even used anywhere.

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
Datron: I didn't want to fail the whole request because redis write failed since the DB would have the change, doing this in the transaction breaks granularity and write_back function can handle the failure case + invalidation, the failure is logged in the function currently

Datron:
@sauraww at the end of handlers for PUT/POST requests, you have functions that update a specific redis key. That makes sure that stale data isn't served if a write happens to DB.

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -141,15 +138,13 @@ async fn create_handler(
         })?;
 
     let DbConnection(mut conn) = db_conn;
-
-    #[cfg(feature = "high-performance-mode")]
-    put_config_in_redis(
+    let _ = put_config_in_redis(
```
ayushjain17, line 141: why are we ignoring the entire result ?
do we not want to even handle or propagate the error ?

previously it was handling the error properly

crates/superposition/src/workspace/handlers.rs
```diff
@@ -211,10 +216,42 @@ async fn update_handler(
...
+            log::debug!("Redis not configured, skipping workspace cache update");
+            return;
+        }
+    };
+
+    let key_ttl: i64 =
+        service_utils::helpers::get_from_env_or_default("REDIS_KEY_TTL", 604800);
+    let expiration = Some(Expiration::EX(key_ttl));
+
+    if let Ok(serialized) = serde_json::to_string(&workspace) {
```
ayushjain17: this becomes ref of ref, not sure if the serialization would work correctly or not, ideally should work fine, but still just to be sure
```suggestion
    if let Ok(serialized) = serde_json::to_string(workspace) {
```

crates/superposition/src/workspace/handlers.rs
```diff
@@ -211,10 +216,42 @@ async fn update_handler(
...
+    let redis_pool = match &state.redis {
+        Some(pool) => pool,
+        None => {
+            log::debug!("Redis not configured, skipping workspace cache update");
+            return;
+        }
+    };
+
+    let key_ttl: i64 =
+        service_utils::helpers::get_from_env_or_default("REDIS_KEY_TTL", 604800);
```
ayushjain17, line 239: I dont think we should read from env in runtime, our current practice does not follow something like this
should our redis data in app_state have been something like a struct with these fields in it, on that we could have the required functions derefed

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -795,3 +851,51 @@ pub async fn fetch_and_validate_change_reason_with_function(
         }
     }
 }
+
+pub async fn put_experiments_in_redis(
+    redis_pool: Option<RedisPool>,
```
ayushjain17: we dont need ownership, we can avoid clones
```suggestion
    redis_pool: &Option<RedisPool>,
```

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -452,3 +457,44 @@ pub fn fetch_experiment_group(
         .get_result::<ExperimentGroup>(conn)?;
     Ok(experiment_group)
 }
+
+pub async fn put_experiment_groups_in_redis(
+    redis_pool: Option<RedisPool>,
```
ayushjain17: same here
```suggestion
    redis_pool: &Option<RedisPool>,
```

crates/service_utils/src/redis.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Fetch data from Redis if available, else fall back to database call and write back to Redis
+/// if redis is disabled read from the database directly
+/// the fallback function is expected to return Result<T, String>
+/// You can use move closures to capture variables in the database_call
+pub async fn fetch_from_redis_else_writeback<T>(
+    key: String,
+    schema_name: &SchemaName,
+    redis_pool: &Option<RedisPool>,
+    db_pool: &PgSchemaConnectionPool,
+    query_fn: impl FnOnce(&mut DBConnection) -> superposition::DieselResult<T>,
```
ayushjain17: ?
```suggestion
    fallback_query: impl FnOnce(&mut DBConnection) -> superposition::DieselResult<T>,
```

crates/service_utils/src/redis.rs
```diff
@@ -0,0 +1,111 @@
...
+pub const CONFIG_KEY_SUFFIX: &str = "::cac_config";
+pub const EXPERIMENTS_LIST_KEY_SUFFIX: &str = "::experiments_list";
+pub const EXPERIMENTS_LAST_MODIFIED_KEY_SUFFIX: &str = "::experiments::last_modified_at";
+pub const EXPERIMENT_GROUPS_LIST_KEY_SUFFIX: &str = "::experiment_groups_list";
+
+/// Fetch data from Redis if available, else fall back to database call and write back to Redis
+/// if redis is disabled read from the database directly
+/// the fallback function is expected to return Result<T, String>
+/// You can use move closures to capture variables in the database_call
+pub async fn fetch_from_redis_else_writeback<T>(
```
ayushjain17: writeback ?
I do not write into DB when redis read fails, I read from DB
I am getting confused with this `writeback`

crates/context_aware_config/src/api/context/handlers.rs
Datron: Couple of things:
- Redis/elasticache failures are unlikely once a connection is made (if you look at the error enum most are connection related)
- the failure is logged inside the function, so not hidden
- Failing the request after the DB transaction seems wrong, would confuse the user as well. Doing it inside the transaction violates granularity
- I'm thinking of following this PR up with an invalidation function that unsets the key, still thinking of a way around without going to redis

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -966,7 +996,56 @@ async fn list_handler(
     if max_event_timestamp.is_some() && max_event_timestamp < last_modified {
         return Ok(HttpResponse::NotModified().finish());
     };
+    let show_all = pagination_params.all.unwrap_or_default();
+    let read_from_redis = show_all
+        && filters
+            .status
+            .clone()
+            .is_some_and(|v| *v == ExperimentStatusType::active_list())
+        && dimension_params.is_empty();
```
ayushjain17, line 1034: should we rather have the inverse logic here checking for fail fast ?

crates/service_utils/src/redis.rs
sauraww: Can we rather name it **read_through_cache**?

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -795,3 +851,51 @@ pub async fn fetch_and_validate_change_reason_with_function(
...
+
+    let paginated_response = PaginatedResponse::all(experiment_responses);
+
+    let serialized = serde_json::to_string(&paginated_response).map_err(|e| {
+        log::error!("Failed to serialize experiments for redis: {}", e);
+        unexpected_error!("Failed to serialize experiments for redis: {}", e)
+    })?;
+
+    let key = format!("{}{EXPERIMENTS_LIST_KEY_SUFFIX}", **schema_name);
+    let key_ttl: i64 = get_from_env_or_default("REDIS_KEY_TTL", 604800);
```
ayushjain17, line 840: same here for reading from env during runtime

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
```diff
@@ -452,3 +457,44 @@ pub fn fetch_experiment_group(
...
+
+    let paginated_response = PaginatedResponse::all(experiment_group_list);
+
+    let serialized = serde_json::to_string(&paginated_response).map_err(|e| {
+        log::error!("Failed to serialize experiment groups for redis: {}", e);
+        unexpected_error!("Failed to serialize experiment groups for redis: {}", e)
+    })?;
+
+    let key = format!("{}{EXPERIMENT_GROUPS_LIST_KEY_SUFFIX}", **schema_name);
+    let key_ttl: i64 = get_from_env_or_default("REDIS_KEY_TTL", 604800);
```
ayushjain17, line 488: here as well

crates/superposition/src/resolve/handlers.rs
```diff
@@ -32,65 +53,168 @@ async fn resolve_with_exp_handler(
...
+        format!("{}{AUDIT_ID_KEY_SUFFIX}", schema_name.0),
+        &schema_name,
+        &state.redis,
+        &state.db_pool,
+        |conn| fetch_audit_id(conn, &workspace_context.schema_name),
+    )
+    .await
+    {
+        resp.insert_header((AppHeader::XAuditId.to_string(), audit_id));
+    }
```
ayushjain17: can we rather delete this (`x-audit-id`) altogether, our config version header is the correct means for this exact purpose
we dont need this anymore and is of no help

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -62,16 +60,26 @@ use super::helpers::{apply_prefix_filter_to_config, resolve, setup_query_data};
 
 #[allow(clippy::let_and_return)]
```
ayushjain17: we dont this anymore after the current change

ayushjain17: we dont this anymore after the current change

crates/context_aware_config/src/helpers.rs
```diff
@@ -226,46 +227,92 @@ pub fn add_config_version(
...
     }
-    let _ = state
-        .redis
-        .set::<(), String, i64>(config_version_key, version_id, None, None, false)
-        .await;
+    redis_pool
+        .set::<(), String, i64>(
+            config_version_key,
+            version_id,
+            expiration.clone(),
```
ayushjain17: 🫠
```suggestion
            expiration
```

crates/superposition/src/resolve/handlers.rs
Datron: Outside the scope of this PR

crates/context_aware_config/src/api/context/handlers.rs
ayushjain17: just thinking should we rather handle the same way we handle webhook failure with a custom error code ?

crates/service_utils/src/redis.rs
Datron: > when redis read fails, I read from DB

writeback here is just whats happening, "write back" to redis and serve the data
The strategy is Cache aside @sauraww

crates/context_aware_config/src/api/context/handlers.rs
ayushjain17: for some reason if we are unable to make a connection with redis itself then we wont have a mechanism to receive the alerts for these failures and this can go unnoticed for long

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: Hmm none of these are time consuming checks, they all fail fast

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -966,7 +996,56 @@ async fn list_handler(
...
+                        pagination_params.clone(),
+                        filters.clone(),
+                        dimension_params.clone(),
+                        conn,
+                        &workspace_context,
+                    )
+                },
+            )
+            .await?;
+        Ok(HttpResponse::Ok().json(response))
```
ayushjain17, line 1055: just like how CF adds a header stating that whether it was a Cache Miss or hit, we should also add a similar header for 3 APIs

Datron: Nice idea, can be done in a later PR

crates/superposition/src/resolve/handlers.rs
```diff
@@ -32,65 +53,168 @@ async fn resolve_with_exp_handler(
...
+                        (id, ffi_exp)
+                    })
+                    .collect())
+            })?
+        } else {
+            HashMap::new()
+        };
+
+        let applicable_variants =
+            get_applicable_variants_from_group_response(&exps, &context, &buckets);
```
ayushjain17: why are doing all of this manually ?
we have a function/handler for this thing which was being used for this exact purpose

if the difference is usage of redis then ideally, that original function/handler should also read from redis itself internally - as airborne uses that heavily and they would prefer being read from redis than from DB

crates/superposition/src/resolve/handlers.rs
```diff
@@ -32,65 +53,168 @@ async fn resolve_with_exp_handler(
...
+        &state.db_pool,
+        |conn| {
+            generate_config_from_version(
+                &mut Some(config_version),
+                conn,
+                &workspace_context.schema_name,
+            )
+            .map_err(|err| {
+                log::error!("failed to generate config from version with error: {}", err);
+                // can't throw the AppError from here because fetch_from_redis_else_writeback
```
sauraww, line 83: this same comment is being repeated at multiple places.

crates/service_utils/src/redis.rs
ayushjain17: @Datron I am not very sure if the meaning is getting conveyed properly from this, at least I got confused with this

crates/superposition/src/workspace/handlers.rs
Datron: > our current practice does not follow something like this

When did we set this rule up? The env read for `REDIS_KEY_TTL` isn't a performance hit, it doesn't suit app state since redis is optional.

crates/superposition/src/resolve/handlers.rs
Datron: > if the difference is usage of redis then ideally, that original function/handler should also read from redis itself internally - as airborne uses that heavily and they would prefer being read from redis than from DB

This is only for providers, thats why `all` and filters are checked

> why are doing all of this manually ? we have a function/handler for this thing which was being used for this exact purpose

If its there, I'll make the change

crates/service_utils/src/redis.rs
Datron: We can rename it if it's confusing

crates/service_utils/src/redis.rs
Datron: Either works, this is a nitpick so will see if change is needed

crates/context_aware_config/src/api/context/handlers.rs
Datron: The end user will have alerts on redis setup, we do have it for ourselves. We can get on the call to discuss it more

crates/experimentation_platform/src/api/experiment_groups/helpers.rs
Datron: > our current practice does not follow something like this

When did we set this rule up? The env read for `REDIS_KEY_TTL` isn't a performance hit, it doesn't suit app state since redis is optional.

crates/superposition/src/workspace/handlers.rs
Datron: It still derefs, wasn't even a clippy lint

crates/superposition/src/resolve/handlers.rs
ayushjain17: `all` ?
didnt get it

crates/superposition/src/workspace/handlers.rs
ayushjain17: no I am not saying to keep it as a separate field
I was suggesting something like:
```rust
struct Redis {
    pool: PoolDataType,
    pub key_ttl: u32,
    ..other such fields
}

impl Redis {
    pub fn get_client(&self) -> RedisClient {
        self.pool.next_connected()
    }
}

struct AppState {
    pub redis: Option<Redis>,
    ..
}

//usage
let client = state.redis.get_client();
let expiration = Some(Expiration::EX(state.redis.key_ttl));
```

crates/experimentation_platform/src/api/experiments/handlers.rs
ayushjain17: can we add a issue for this to track it

crates/superposition/src/resolve/handlers.rs
ayushjain17: its just deletion of something which is entirely unused, removing this would help prevent adding of unnecessary code

crates/service_utils/src/redis.rs
sauraww: Definition: In a read-through cache, data is loaded into the cache on demand, typically when a read request occurs for data that is not already in the cache.
Process:
When a read request is made, the cache first checks if the data is available (cache hit).
If the data is not in the cache (cache miss), the cache system reads the data from the primary storage, stores it in the cache, and then returns the data to the client.
Subsequent read requests for the same data will be served directly from the cache until the data expires or is evicted.
Source: https://www.designgurus.io/answers/detail/what-is-read-through-vs-write-through-cache

ayushjain17: yeah that's what I wanted to maybe call it something else
maybe only `fetch_from_redis` or maybe `read_through_cache` as @sauraww is suggesting with the fallback function param called as `fallback_fn` might be better

crates/superposition/src/resolve/handlers.rs
Datron: Yeah its needed till this is fixed

crates/superposition/src/resolve/handlers.rs
Datron: This is the query params that providers send to fetch data

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: Done

crates/superposition/src/resolve/handlers.rs
Datron: I think the plan for this was that the provider can switch between different versions on the user side, hence I want to know the direction we go before removing it

crates/superposition/src/workspace/handlers.rs
```diff
@@ -352,6 +389,10 @@ async fn migrate_schema_handler(
         Ok(())
     })?;
 
+    // Refetch workspace after transaction to get updated data
+    let workspace = get_workspace(&schema_name, &mut conn)?;
+    put_workspace_in_redis(&workspace, &state, &schema_name.0).await;
```
ayushjain17, line 393: why do we have this separate behaviour only for the case of secrets ?
in other places although we have a transaction, we return workspace data from it and that is put into redis

why not do the same and fetch the workspace from DB in the transaction itself and return it to be saved in redis ?

crates/service_utils/src/redis.rs
Datron: updated

crates/superposition/src/resolve/types.rs
sauraww: @Datron but this has a wider scope.

crates/superposition/src/workspace/handlers.rs
Datron: I think we can go with reading the key each time, it doesn't have a KMS decrypt step either. It simplifies the AppState struct as well

crates/superposition/src/resolve/handlers.rs
ayushjain17: this is something which was added even before snapshots was a concept
and was never used even when introduced

crates/superposition/src/resolve/handlers.rs
Datron: Will make this change

.env.example
```diff
@@ -38,6 +38,12 @@ AUTH_PROVIDER=DISABLED
 AUTH_Z_PROVIDER=DISABLED
 WORKER_ID=1
 # MASTER_ENCRYPTION_KEY - add this for enabling secrets in local
+# REDIS_URL="http://localhost:6379"
```
ayushjain17, line 41: are we going ahead without tests for redis then ?

Datron: For now, I'll setup the build pipeline for this. I haven't added it to github builds

PR #800, opened 2025-12-23
build: fix query in test

ayushjain17:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

* **Chores**
  * Updated database migration to streamline data insertion by removing an unused field from the initialization process.

### DIFF
--- a/scripts/create-tenant.sh
+++ b/scripts/create-tenant.sh
@@ -55,6 +55,6 @@ function create_schema() {
 }
 
 create_schema "superposition_types" $TENANT_SCHEMA
-psql "$DB_URL" -c "INSERT INTO $TENANT_SCHEMA.dimensions (dimension, priority, created_at, created_by, schema, function_name) VALUES ('variantIds', 1, CURRENT_TIMESTAMP, 'user@example.com', '{\"type\": \"string\",\"pattern\": \".*\"}'::json, null);"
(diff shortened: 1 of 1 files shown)


knutties (approved):

scripts/create-tenant.sh
```diff
@@ -55,6 +55,6 @@ function create_schema() {
 }
 
 create_schema "superposition_types" $TENANT_SCHEMA
-psql "$DB_URL" -c "INSERT INTO $TENANT_SCHEMA.dimensions (dimension, priority, created_at, created_by, schema, function_name) VALUES ('variantIds', 1, CURRENT_TIMESTAMP, 'user@example.com', '{\"type\": \"string\",\"pattern\": \".*\"}'::json, null);"
+psql "$DB_URL" -c "INSERT INTO $TENANT_SCHEMA.dimensions (dimension, priority, created_at, created_by, schema) VALUES ('variantIds', 1, CURRENT_TIMESTAMP, 'user@example.com', '{\"type\": \"string\",\"pattern\": \".*\"}'::json);"
```
knutties, line 58: Forget the context for this.  Are we removing value validation functions for dimensions completely now?  They can still be there right?

@mahatoankitkumar @ayushjain17  ?

ayushjain17: this was a dummy entry I suppose, used only in the test flow
even this entire script can be omitted now, we have a better alternate

in this query here, outdated column name was being used that to entry `null` value which anyways is the default value for the column

PR #803, opened 2025-12-25
build: clean setup for test

ayushjain17:
Using `superposition.sql` to setup db for tests instead of using the old setup of creating tenant specific migration file by going through all the queries from `migrations` folder - the tenant created in this process was not even used

the setup was not clean and was do unnecessary extra steps
`superposition.sql` file is idempotent


knutties:
@ayushjain17 - so we are deprecating all manual ways of creating workspace and moving to the API which ends up using the sql file.  is that right ?

ayushjain17:
> @ayushjain17 - so we are deprecating all manual ways of creating workspace and moving to the API which ends up using the sql file. is that right ?

@knutties in a way yes, the correctness of the script was not properly verified and maintained as well
the one way which we guarantee and use is via the workspace_template.sql file which in turn is used in the API

the script way should have been long deprecated

PR #804, opened 2025-12-26
fix: functions payload structure

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

* **New Features**
  * Added runtime version support for function execution to enable improved version compatibility and management.

* **Refactor**
  * Updated function invocation to use a unified payload-based pattern.

* **Chores**
  * Updated database migrations and function templates to align with new invocation pattern.


crates/superposition_types/src/database/models/cac.rs
```diff
@@ -258,6 +230,52 @@ pub struct Function {
...
+            .as_str()
+            .try_into()
+            .map_err(|e| format!("Invalid FunctionRuntimeVersion: {e}"))
+    }
+}
+
+impl Display for FunctionRuntimeVersion {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::V1 => write!(f, "1.0.0"),
```
knutties: Minor nit: `1.0.0` ?  `1.0` should suffice ?

ayushjain17: Either works
We anyways need run a backfill in prod to correct the values as in Prod we do have very old values which have it as 1.0 and the new values as 1.0.0

crates/context_aware_config/src/api/functions/types.rs
```diff
@@ -0,0 +1,13 @@
+use diesel::{prelude::Queryable, Selectable};
+use superposition_types::database::{
+    models::cac::{FunctionCode, FunctionRuntimeVersion},
+    schema::functions,
+};
+
+#[derive(Clone, Selectable, Queryable)]
+#[diesel(table_name = functions)]
+pub struct FunctionInfo {
```
knutties, line 9: This is better called `PublishedFunctionInfo` as it deals with only published functions ?

knutties: Assuming purpose of this type is to hold data that is relevant for actual execution?  Is that right?

crates/superposition_types/src/database/models/cac.rs
knutties: we can keep it 1.0.0 itself - the rationale might be semantic versioning.

ayushjain17: sure, reverting the change then

crates/context_aware_config/src/api/functions/types.rs
ayushjain17: yeah, fair ask
but not keeping it specific gave the type room to expand freely

crates/superposition_types/src/database/models/cac.rs
ayushjain17: updated decision - going ahead with `1.0`

workspace_template.sql
```diff
@@ -330,18 +330,32 @@ EXCEPTION
...
-
 DO $$ BEGIN
-    ALTER TABLE {replaceme}.default_configs ADD FOREIGN KEY(function_name) REFERENCES {replaceme}.functions(function_name);
+    IF NOT EXISTS (
+        SELECT 1 FROM information_schema.columns 
+        WHERE table_schema = '{replaceme}' 
+        AND table_name = 'default_configs' 
+        AND column_name = 'value_validation_function_name'
+    ) THEN
+        ALTER TABLE {replaceme}.default_configs ADD COLUMN IF NOT EXISTS function_name text NULL;
```
knutties, line 355: This column is `function_name` or `value_validation_function_name` ?  Or we just maintaining the older one.

ayushjain17: this is for `function_name` the existing column, which we are renaming later to `value_validation_function_name`

crates/context_aware_config/src/api/functions/types.rs
knutties: expand meaning - you can add more fields (even the draft ones) to this struct later ?

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -113,73 +114,52 @@ const CODE_GENERATION_SNIPPET: &str = r#"
     }
     "#;
 
-fn type_check(code_str: &FunctionCode, function_name: &str) -> String {
+fn type_check(code_str: &FunctionCode) -> String {
     FUNCTION_TYPE_CHECK_SNIPPET
-        .replace(FUNCTION_NAME_TOKEN, function_name)
+        .replace(FUNCTION_NAME_TOKEN, "execute")
```
knutties: Do we have a consts file to keep this string?  and name the const `DEFAULT_FUNCTION_NAME`.  Don't want it buried in here.

ayushjain17: yeah its in this file itself
will add it there as a CONST

crates/frontend/src/components/function_form.rs
```diff
@@ -32,56 +32,74 @@ use crate::utils::use_url_base;
...
+pub const VALUE_VALIDATION_TEMPLATE_FN: &str = r#"// Payload structure:
+// {
+//   version: "1.0",
+//   value_validate: { key, value, type, environment }
+// }
 // returns: boolean
-async function validate_value(key, value, type, environment) {
+async function execute(payload) {
+    const { value_validate } = payload;
+    const { key, value, type, environment } = value_validate;
```
knutties, line 46: We can leave a commented section like the following to show where the validation logic goes for better clarity.
```javascript
const { value_validate } = payload;
const { key, value, type, environment } = value_validate;

// validation logic goes here

return true;
```

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -273,43 +283,42 @@ fn get_functions_map(
...
+                .get(&function_name)
+                .cloned()
+                .map(|func_info| (key, func_info))
         })
         .collect();
+
     Ok(function_to_primitives_map)
 }
 
 pub fn validate_value_with_function(
```
knutties: This is used for both validate_value and validate_context now right ?  basically it is the single entry point for all validation functions it appears.

Might be a good idea to call it just `validation_function_executor` or something like that.

crates/context_aware_config/src/api/functions/types.rs
ayushjain17: yeah

crates/context_aware_config/src/api/functions/handlers.rs
```diff
@@ -228,14 +228,19 @@ async fn delete_function(
...
-    {
-        log::error!("Attempted to delete reserved function: {}", f_name);
-        return Err(bad_argument!(
-            "Cannot delete function {}: This function is reserved and cannot be deleted.",
-            f_name
-        ));
+    let function = fetch_function(&f_name, &mut conn, &schema_name)?;
+    match function.function_type {
+        FunctionType::ContextValidation | FunctionType::ChangeReasonValidation => {
+            log::error!(
```
knutties, line 234: Minor nit: There are a bunch of places where there is an error log with a message followed by Err struct that is returned.

Should we unify the error strings here?

crates/context_aware_config/src/api/functions/handlers.rs
```diff
@@ -273,27 +278,30 @@ async fn test(
...
-                return Err(bad_argument!(
-                    "Function test failed as function not published yet"
-                ));
+    let (code, version) = match path_params.stage {
+        Stage::Draft => (function.draft_code, function.draft_runtime_version),
+        Stage::Published => {
+            match (function.published_code, function.published_runtime_version) {
+                (Some(code), Some(version)) => (code, version),
+                _ => {
+                    log::error!("Function test failed: function not published yet");
```
knutties, line 287: Same comment as above.

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -113,73 +116,54 @@ const CODE_GENERATION_SNIPPET: &str = r#"
...
-        } => (
-            FunctionType::ChangeReasonValidation
-                .get_fn_signature()
-                .replace(
-                    "{change_reason}",
-                    &serde_json::to_string(&change_reason).unwrap_or_default(),
-                ),
-            "output!=true",
-        ),
+    let payload = match runtime_version {
```
knutties, line 137: This is good simplification because of the payload approach - nice !!!

knutties:
LGTM now @ayushjain17  - left a minor comment.  But that unification of error logs and return strings can be a separate effort.

knutties (approved):

PR #814, opened 2026-01-02
fix: workspace settings usage for change reason validation

ayushjain17:
## Problem
For change reason validation, workspace settings was not being checked consistently

## Changes
1. Added proper checks for change reason validation
2. Moved workspace fetch to middleware - for all workspace level requests the workspace settings would get auto fetched before any other operation in the handle 
3. this doubles down as a validation for the org and workspace input in the request, previously it used to lead to 500 when an invalid combination was provided

## Summary by CodeRabbit

* **New Features**
  * Added tracking of last modification time and user for functions.
  * Type template updates now support change reason field.

* **Bug Fixes**
  * Change reason validation is now consistently enforced across all operations.

* **Refactor**
  * Workspace configuration defaults updated: change reason validation and context validation now default to disabled.
  * Workspace settings now provided directly to validation operations for improved efficiency.

### DIFF
--- a/crates/context_aware_config/src/api/variables/handlers.rs
+++ b/crates/context_aware_config/src/api/variables/handlers.rs
@@ -28,7 +28,7 @@ pub fn endpoints() -> Scope {
 #[authorized]
 #[get("")]
 async fn list_handler(
-    workspace_request: WorkspaceContext,
+    workspace_context: WorkspaceContext,
     db_conn: DbConnection,
     pagination: Query<PaginationParams>,
     filters: Query<VariableFilters>,
@@ -39,7 +39,7 @@ async fn list_handler(
 
     let query_builder = |filters: &VariableFilters| {
         let mut builder = variables
-            .schema_name(&workspace_request.schema_name)
+            .schema_name(&workspace_context.schema_name)
             .into_boxed();
 
         if let Some(ref var_names) = filters.name {
@@ -98,7 +98,7 @@ async fn list_handler(
 #[authorized]
 #[post("")]
 async fn create_handler(
-    workspace_request: WorkspaceContext,
+    workspace_context: WorkspaceContext,
     req: web::Json<CreateVariableRequest>,
     user: User,
     db_conn: DbConnection,
@@ -106,7 +106,7 @@ async fn create_handler(
     let DbConnection(mut conn) = db_conn;
     let req = req.into_inner();
 
-    validate_change_reason(&workspace_request, &req.change_reason, &mut conn)?;
+    validate_change_reason(&workspace_context, &req.change_reason, &mut conn)?;
 
     let now = chrono::Utc::now();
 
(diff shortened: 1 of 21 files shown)


crates/superposition_types/src/api/workspace.rs
```diff
@@ -67,10 +67,14 @@ pub struct CreateWorkspaceRequest {
...
-    pub enable_context_validation: Option<bool>,
-    pub enable_change_reason_validation: Option<bool>,
+    #[serde(default)]
+    pub allow_experiment_self_approval: bool,
+    #[serde(default = "default_true")]
+    pub auto_populate_control: bool,
+    #[serde(default)]
+    pub enable_context_validation: bool,
+    #[serde(default)]
+    pub enable_change_reason_validation: bool,
```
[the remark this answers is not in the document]
ayushjain17: its expected to be TRUE by default

ayushjain17:
@coderabbitai help

crates/context_aware_config/src/api/variables/handlers.rs
```diff
@@ -156,6 +160,9 @@ async fn update_handler(
     let DbConnection(mut conn) = db_conn;
     let var_name = path.into_inner();
 
+    // TODO: if ever workspace settings is fetched in this request lifecycle, pass it here to avoid extra db call.
+    validate_change_reason(None, &req.change_reason, &mut conn, &schema_name)?;
```
mahatoankitkumar: We’re passing None here and then re-fetching it inside validate_change_reason. It would be cleaner to fetch the value here itself, avoiding dependencies on future changes. We can also do the check here and call the function only when needed, keeping the function purely responsible for validation.

ayushjain17: That in fact is a very bad practice
The exact reason why this PR was needed, as the check was separate responsibility which was missed out in most of the places causing the problem

You dont keep core inter-related logic decoupled like this

ayushjain17:
@coderabbitai full review

crates/context_aware_config/src/api/variables/handlers.rs
ayushjain17: @mahatoankitkumar have made the changes to pass the settings to the function and removed the fetch from inside

crates/superposition_types/src/database/models.rs
```diff
@@ -282,6 +284,27 @@ pub struct Workspace {
...
+        let result = req.extensions().get::<Self>().cloned().ok_or_else(|| {
+            crate::result::AppError::UnexpectedError(
+                anyhow::anyhow!("Please check that the organisation id and workspace id are being properly sent")
+            ).into()
+        });
+
+        ready(result)
+    }
+}
+
```
Datron: Is this needed? You aren't inserting Workspace anywhere explicitly, we are inserting this data into `WorkspaceContext::settings` field

ayushjain17: yeah its not needed, I too wanted to remove it, left it for the consistency part
if we agree then yeah good to remove it

Datron: Can you remove it and push, I'll approve

ayushjain17: yeah dead code, pushed by mistake

knutties (approved):

PR #815, opened 2026-01-03
feat(UI): use browser locale time

ayushjain17:
Show time related data in browser local time zone

### Note
In SSR, due to lack of user's timezone info, the data is returned as UTC time itself in the pre-rendered page in ISO 8601 format, which is immediately converted to user locale `time` or `date only` on page load via JS which triggers on `DOMContentLoaded` event
when CSR kicks in, it updates it to the same value again - no impact for end user

## Summary by CodeRabbit

* **New Features**
  * Locale-aware datetime component for browser/server that displays dates/times per your regional settings.
  * Date input controls improved with better defaults and more robust parsing.

* **UI Improvements**
  * Consistent, localized timestamps across audit logs, lists, details, and tables.
  * Reduced display inconsistencies during page hydration for more stable date/time rendering.

### DIFF
--- a/crates/frontend/src/app.rs
+++ b/crates/frontend/src/app.rs
@@ -3,6 +3,7 @@ use leptos_meta::*;
 use leptos_router::*;
 use serde_json::json;
 
+use crate::components::datetime::DatetimeConversionScript;
 use crate::hoc::layout::{use_org, CommonLayout, Layout, Providers};
 use crate::pages::compare_overrides::CompareOverrides;
 use crate::pages::config_version::ConfigVersion;
@@ -57,6 +58,7 @@ pub fn app(app_envs: Envs) -> impl IntoView {
                     type_="text/javascript"
                     src="https://cdn.jsdelivr.net/npm/sortablejs@1.15.6/Sortable.min.js"
                 />
+                <DatetimeConversionScript />
                 {move || {
                     if base.is_empty() {
                         view! {}.into_view()
--- a/crates/frontend/src/components/datetime.rs
+++ b/crates/frontend/src/components/datetime.rs
@@ -1,0 +1,200 @@
+#[cfg(target_arch = "wasm32")]
+use chrono::Datelike;
+use chrono::{offset::Utc, DateTime, NaiveDate};
+#[cfg(target_arch = "wasm32")]
+use js_sys::{Date, Object};
+use leptos::*;
+use leptos_meta::*;
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen::JsValue;
+
+use crate::providers::csr_provider::use_client_side_ready;
+
+#[cfg(target_arch = "wasm32")]
+fn browser_locale() -> String {
+    web_sys::window()
+        .and_then(|w| w.navigator().language())
+        .unwrap_or_else(|| "en-GB".to_string())
+}
+
+#[cfg(target_arch = "wasm32")]
+fn format_local_datetime(time: &DateTime<Utc>) -> String {
+    let millis = time.timestamp_millis();
+    let date = Date::new(&JsValue::from_f64(millis as f64));
+    let locale = browser_locale();
+
+    date.to_locale_string(&locale, &Object::new()).into()
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+fn format_local_datetime(time: &DateTime<Utc>) -> String {
+    time.to_rfc3339()
+}
+
+fn format_local_datetime_from_str(time: &str) -> String {
+    match DateTime::parse_from_rfc3339(time) {
+        Ok(dt) => format_local_datetime(&dt.with_timezone(&Utc)),
+        Err(_) => time.to_string(),
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+pub fn format_local_date(time: &DateTime<Utc>) -> String {
+    let millis = time.timestamp_millis();
+    let date = Date::new(&JsValue::from_f64(millis as f64));
+
+    let year = date.get_full_year();
+    let month = date.get_month() + 1; // JS months are 0-based
+    let day = date.get_date();
+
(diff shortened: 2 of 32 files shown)


crates/frontend/src/components/datetime.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// Component to display a datetime in local format
+/// # Arguments
+/// * `datetime` - A DateTime<Utc> object
+/// # Returns
+/// A view displaying the formatted datetime based on the user's locale
+/// For SSR, it will display the datetime as `%v %T UTC` format
+#[component]
+pub fn datetime(datetime: DateTime<Utc>) -> impl IntoView {
+    move || view! { <span>{use_format_local_datetime(&datetime)}</span> }
```
knutties: Would you need to support a class or some such to do targetted styling for this component?

ayushjain17: it was not needed right now, but yeah can add the option
didn't add it now as we can add it whenever we would have need for it

but will update

knutties: Ok - we can add when the need comes.  Not urgent.

crates/frontend/src/components/datetime.rs
```diff
@@ -0,0 +1,133 @@
...
+/// # Arguments
+/// * `datetime` - A DateTime<Utc> object
+/// * `class` - An optional CSS class for styling the span element
+/// # Returns
+/// A view displaying the formatted datetime based on the user's locale
+/// For SSR, it will display the datetime as `%v %T UTC` format
+#[component]
+pub fn datetime(
+    datetime: DateTime<Utc>,
+    #[prop(into, optional)] class: Option<String>,
```
knutties: 👍

crates/frontend/src/app.rs
```diff
@@ -57,6 +58,7 @@ pub fn app(app_envs: Envs) -> impl IntoView {
                     type_="text/javascript"
                     src="https://cdn.jsdelivr.net/npm/sortablejs@1.15.6/Sortable.min.js"
                 />
+                <DatetimeConversionScript />
```
knutties, line 61: It is hard to find this component - looks like leptops component macro derives it from `datetime_conversion_script`.  Can we use the same names for easier finding / navigation?

This can be a separate discussion.

crates/frontend/src/components/datetime.rs
knutties, line 1: Might be a good idea to have unit tests added to this file :)

ayushjain17: All the actual logic is in JS which is wasm specific
In rust there is barely any logic to add test for

crates/frontend/src/app.rs
ayushjain17: Currently all our components follow this pattern itself
The functions are defined as per rust conventions of snake case and are then converted to title case as per JSX style by the macro

crates/frontend/src/components/datetime.rs
knutties: Agree there is not a lot of logic in there (but some places like month index adjustments and parse patterns) are possibly worth checking for.

knutties: My point was more about making this a habit in rust files where there is rust logic.

crates/frontend/src/components/datetime.rs
```diff
@@ -0,0 +1,166 @@
...
+
+/// Script to fetch the inner content of both the span with class 'datetime' and span with class 'datetime_date' to
+/// convert it to local datetime and local date, respectively, from date strings in ISO 8601 format
+/// to avoid hydration mismatches between server and client rendering, leading to a flicker effect.
+#[component]
+pub fn datetime_conversion_script() -> impl IntoView {
+    view! {
+        <Script type_="text/javascript">
+            r#"
+            function convertDateTimeElements() {
```
knutties:
```suggestion
const convertDates = () => {
  // 1. Combine selectors and use querySelectorAll for a static NodeList
  const elements = document.querySelectorAll('.datetime, .datetime_date');

  elements.forEach(elem => {
    const date = new Date(elem.textContent);
    
    // 2. Early return if date is invalid
    if (isNaN(date.getTime())) return;

    // 3. Use a simple conditional for the formatting logic
    if (elem.classList.contains('datetime_date')) {
      // Formats as YYYY-MM-DD using ISO string for consistency
      elem.textContent = date.toISOString().split('T')[0];
    } else {
      elem.textContent = date.toLocaleString(navigator.language);
    }
  });
};

window.addEventListener('DOMContentLoaded', convertDates);
```

ayushjain17: we can't use `date.toISOString().split('T')[0]` though for date conversion
it does not convert the UTC date to local date

ayushjain17: we would need something like this for proper conversion
```javascript
            const dateToString = (date) => {
                const year = date.getFullYear();
                const month = String(date.getMonth() + 1).padStart(2, '0');
                const day = String(date.getDate()).padStart(2, '0');

                return `${year}-${month}-${day}`;
            }
```

crates/frontend/src/components/datetime.rs
```diff
@@ -0,0 +1,200 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+pub fn try_utc_date_from_locale_date(locale_date: &str) -> Option<DateTime<Utc>> {
+    use chrono::NaiveTime;
+
+    let date = NaiveDate::parse_from_str(locale_date, "%Y-%m-%d").ok()?;
+    let date_time =
+        DateTime::from_naive_utc_and_offset(date.and_time(NaiveTime::default()), Utc);
+
+    Some(date_time)
+}
```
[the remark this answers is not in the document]
ayushjain17: this is intended and has been handled separately

PR #817, opened 2026-01-04
feat: Add workspace locking using lease metadata inside workspace to serialize write operations

knutties:
## Problem
Currently, changes within workspace are not serialized leading to potential race conditions in version_state generation.  We need to serialize them.

## Solution
Serialize all write operations per workspace using a **lease-based lock** stored in the `workspaces` table itself.

When a write request comes in (POST, PUT, PATCH, DELETE), the `WorkspaceWritePermit` extractor:
1. Checks out a DB connection
2. Tries to `UPDATE workspaces SET lock_columns WHERE lock is NULL or expired`
3. If `updated_rows == 1` → we got the lock, proceed with the write
4. If `updated_rows == 0` → someone else holds the lock, return **409 Conflict** with lock details
5. When the request completes, the permit is dropped and the lock columns are set back to NULL

Read operations (GET) are **not** locked — they proceed as before.

| Variable | Default | Description |
|---|---|---|
| `WORKSPACE_LOCK_DEFAULT_TTL_MS` | 60000 (60s) | Lock TTL for normal write operations |
| `WORKSPACE_LOCK_BATCH_TTL_MS` | 1200000 (20min) | Lock TTL for long-running batch operations |

## Pre-deployment activity

``` 
Run the migration: 2026-04-29-000001_workspace_table_lease_lock 
```

## Post-deployment activity
NA

## API changes
All workspace write endpoints now return **409 Conflict** if the workspace is already locked by another write:

```json
{
  "message": "Workspace is busy with another write request; retry later",
  "lock": {
    "lock_id": "uuid",
    "operation": "PUT /default-config/my-key",
    "locked_by": "user@example.com",
    "acquired_at": "2026-04-30T10:00:00Z",
    "expires_at": "2026-04-30T10:01:00Z"
  }
}
```

The `GET /workspaces/{name}` response now includes an optional `workspace_lock` field showing the current lock state (omitted when no lock is active).

### Why `lock_id`?

`lock_id` prevents a stale owner from clearing someone else'''s lock. Without it:

1. Request A acquires lock (lock_id = X)
2. Request A takes too long, lock expires
3. Request B acquires lock (lock_id = Y) on the same workspace
4. Request A finishes and releases — **it would clear B'''s lock**

The `release_workspace_lease` function filters on `workspace_lock_id = lease.lock_id`, so Request A'''s release would match 0 rows and not interfere with B'''s lock.

### How is the lock acquisition atomic?

The entire check-and-set happens in a single SQL `UPDATE` statement:

```sql
UPDATE workspaces
SET workspace_lock_id = $1, workspace_lock_operation = $2, ...
WHERE organisation_id = $3
  AND workspace_name = $4
  AND (workspace_lock_expires_at IS NULL OR workspace_lock_expires_at <= now())
```

If the lock is active (not expired), `updated_rows` will be `0` and we return `LockContended`. If the lock is absent or expired, `updated_rows` will be `1` and we own the lock. There is no SELECT-then-UPDATE race condition because PostgreSQL evaluates the WHERE clause and applies the SET atomically within a single statement.

### Why have `WorkspaceLock` as a separate struct in the API?

The 5 lock columns (`workspace_lock_id`, `workspace_lock_operation`, etc.) are internal DB details. `WorkspaceLock` groups them into a clean API response type that:
- Returns `None` when no lock is active (instead of 5 nullable fields)
- Filters out expired locks (`from_workspace` returns `None` if `expires_at <= now()`)
- Is reused in the 409 Conflict error body so clients get meaningful feedback


crates/service_utils/src/middlewares/workspace_lock.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    fn call(&self, req: ServiceRequest) -> Self::Future {
+        let srv = self.service.clone();
+
+        Box::pin(async move {
+            // Only lock for write operations
+            let is_write_operation = matches!(
+                req.method(),
+                &Method::POST | &Method::PUT | &Method::DELETE | &Method::PATCH
+            );
```
ShubhranshuSanjeev: Don't remember completely, but we have read operations which are `POST`. 
@ayushjain17 @Datron

crates/service_utils/src/middlewares/workspace_lock.rs
```diff
@@ -0,0 +1,292 @@
...
+                }
+                log::debug!(
+                    "released advisory lock for workspace (org_key: {}, workspace_key: {})",
+                    org_key, workspace_key
+                );
+            }
+
+            // Return the result
+            result.map(|r| r.map_into_left_body())
+        })
```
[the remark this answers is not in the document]
knutties: resolved.

crates/service_utils/src/middlewares/workspace_lock.rs
```diff
@@ -0,0 +1,292 @@
...
+        if attempt < MAX_RETRIES - 1 {
+            let backoff_ms = std::cmp::min(
+                INITIAL_BACKOFF_MS * 2_u64.pow(attempt),
+                MAX_BACKOFF_MS
+            );
+            log::debug!(
+                "lock contention detected, retrying in {}ms (attempt {}/{}, org_key: {}, workspace_key: {})",
+                backoff_ms, attempt + 1, MAX_RETRIES, org_key, workspace_key
+            );
+            std::thread::sleep(std::time::Duration::from_millis(backoff_ms));
```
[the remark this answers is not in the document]
knutties: resolved.

crates/service_utils/src/middlewares/workspace_lock.rs
```diff
@@ -0,0 +1,317 @@
...
+                    );
+                    return Err(error::ErrorInternalServerError(
+                        "Failed to acquire workspace lock",
+                    ));
+                }
+                log::debug!(
+                    "acquired advisory lock for workspace (org_key: {}, workspace_key: {})",
+                    org_key, workspace_key
+                );
+                Some(AdvisoryLockGuard::new(&mut db_conn, org_key, workspace_key))
```
ShubhranshuSanjeev: The guard can be returned by the `acquire` fn

crates/service_utils/src/middlewares/workspace_lock.rs
```diff
@@ -0,0 +1,317 @@
...
+        if attempt < MAX_RETRIES - 1 {
+            let backoff_ms = std::cmp::min(
+                INITIAL_BACKOFF_MS * 2_u64.pow(attempt),
+                MAX_BACKOFF_MS
+            );
+            log::debug!(
+                "lock contention detected, retrying in {}ms (attempt {}/{}, org_key: {}, workspace_key: {})",
+                backoff_ms, attempt + 1, MAX_RETRIES, org_key, workspace_key
+            );
+            actix_web::rt::time::sleep(std::time::Duration::from_millis(backoff_ms)).await;
```
ShubhranshuSanjeev: We can use Conditional Variables to awake the sleeping tasks. The awake call can go in `drop` of LockGuard.

ShubhranshuSanjeev (dismissed):

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -569,13 +587,16 @@ async fn delete_handler(
...
+        {
+            let conn = write_permit.checkout();
+            let _ = put_config_in_redis(
+                &config_version,
+                &state,
+                &workspace_context.schema_name,
+                conn,
+            )
+            .await;
+        }
```
mahatoankitkumar: Block not needed, and taking mut ref and dropping is unnecessary.

crates/service_utils/src/service/types.rs
```diff
@@ -229,18 +239,138 @@ impl FromRequest for DbConnection {
...
+/// Use `checkout()` to get a mutable reference to the underlying connection,
+/// or `transaction()` to run a transaction on the locked connection.
+pub struct WorkspaceWritePermit {
+    conn: superposition_types::DBConnection,
+    lease: Option<WorkspaceLockLease>,
+}
+
+impl WorkspaceWritePermit {
+    /// Returns a mutable reference to the database connection.
+    /// The workspace lock remains held while the connection is in use.
```
mahatoankitkumar: misleading comment

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -136,9 +128,19 @@ async fn create_handler(
...
+    let conn = write_permit.checkout();
+
+    validate_change_reason(
+        &workspace_context,
+        &req.change_reason,
+        conn,
+        &state.master_encryption_key,
+    )
+    .await?;
+
```
mahatoankitkumar: why moved?
should be where it was at the starting.

docker-compose/postgres/db_init.sql
```diff
@@ -1771,4 +1771,11 @@ ALTER TABLE superposition.workspaces
 ADD COLUMN IF NOT EXISTS encryption_key TEXT NOT NULL DEFAULT '',
 ADD COLUMN IF NOT EXISTS key_rotated_at TIMESTAMPTZ;
 
+ALTER TABLE superposition.workspaces
+ADD COLUMN IF NOT EXISTS workspace_lock_id UUID,
```
Datron, line 1775: What is the use of lock ID?

package.json
```diff
@@ -15,6 +15,7 @@
   ],
   "devDependencies": {
     "@tsconfig/node18": "^18.2.4",
+    "axios": "^1.11.0",
```
Datron: remove this

crates/superposition/src/workspace/handlers.rs
```diff
@@ -455,6 +461,15 @@ async fn migrate_schema_handler(
...
         }
+
+        diesel::update(workspaces::table)
+            .filter(workspaces::organisation_id.eq(&org_id.0))
+            .filter(workspaces::workspace_name.eq(&workspace_name))
+            .set((
+                workspaces::last_modified_by.eq(&user_email),
+                workspaces::last_modified_at.eq(Utc::now()),
+            ))
+            .execute(transaction_conn)?;
```
Datron: This seems redundant. Its done above anyway?

crates/superposition/src/workspace/handlers.rs
```diff
@@ -231,7 +236,7 @@ async fn update_handler(
         }
     }
 
-    if let Some(I64Update::Add(version)) = request.config_version {
+    if let Some(I64Update::Add(version)) = request.config_version.clone() {
```
Datron: Why the clone, I don't see any other changes in this function

crates/service_utils/Cargo.toml
```diff
@@ -19,7 +19,7 @@ chrono = { workspace = true }
 derive_more = { workspace = true }
 diesel = { workspace = true }
 diesel-adapter = { version = "1.2.0" }
-fred = { workspace = true, features = ["metrics"] }
+fred = { workspace = true, features = ["metrics", "i-scripts"] }
```
Datron: Is this needed?

mahatoankitkumar (commented):
May be in the subsequent PR lets have a workspaceReadPermit just for consistency how a db conn is being acquired throughout the whole repo.
Right now its workspaceWritePermit which has dbconnection and reads have direct dbconnection access.

crates/service_utils/src/db.rs
```diff
@@ -8,9 +8,10 @@ pub mod utils;
 
 pub type PgSchemaConnectionPool = Pool<ConnectionManager<PgConnection>>;
 
-// this should not be made public, instead we should have helper functions that use
-// this internally to run queries/transactions with proper error handling and connection management
-fn get_connection(db_pool: &PgSchemaConnectionPool) -> result::Result<DBConnection> {
+/// Checks out a connection from the pool with statement caching disabled.
+pub fn checkout_connection(
```
mahatoankitkumar: why pub

Datron: get was nicer

crates/service_utils/src/workspace_lock.rs
```diff
@@ -0,0 +1,214 @@
...
+    api::workspace::WorkspaceLock,
+    database::{models::Workspace, superposition_schema::superposition::workspaces::dsl},
+    result::{self as superposition},
+};
+use uuid::Uuid;
+
+/// Creates a unique workspace key from organisation and workspace IDs.
+fn workspace_key(organisation_id: &str, workspace_id: &str) -> String {
+    format!("{}/{}", organisation_id, workspace_id)
+}
```
Datron: Use the workspace schema name as the key

crates/superposition/src/workspace/handlers.rs
mahatoankitkumar: why clone?

crates/context_aware_config/src/api/default_config/handlers.rs
sauraww: you would not even need to take a mutable reference to the connection if it fails early

crates/service_utils/src/workspace_lock.rs
```diff
@@ -0,0 +1,214 @@
...
+        AcquireLockError::LockContended(lock) => WorkspaceLockConflict::new(lock).into(),
+        AcquireLockError::Diesel(diesel::result::Error::NotFound) => {
+            actix_web::error::ErrorNotFound("Workspace not found")
+        }
+        AcquireLockError::Diesel(e) => {
+            log::error!("failed to acquire workspace lock: {}", e);
+            actix_web::error::ErrorInternalServerError("Failed to acquire workspace lock")
+        }
+    }
+}
```
Datron: This can be part of our error enum, would make '?' simpler

crates/service_utils/src/workspace_lock.rs
```diff
@@ -0,0 +1,214 @@
...
+        )
+        .set((
+            dsl::workspace_lock_id.eq(lock_id), // TODO: getting used in case of experimentation handler
+            dsl::workspace_lock_operation.eq(Some(request.operation)),
+            dsl::workspace_locked_by.eq(Some(request.locked_by)),
+            dsl::workspace_lock_acquired_at.eq(Some(now)),
+            dsl::workspace_lock_expires_at.eq(Some(expires_at)),
+        ))
+        .execute(conn)?;
+
```
Datron, line 168: Should lock be checked first before being updated? This basically erases the previous lock owner

crates/service_utils/src/service/types.rs
```diff
@@ -229,18 +239,124 @@ impl FromRequest for DbConnection {
...
-                    diesel::connection::CacheSize::Disabled,
-                );
-                Ok(DbConnection(conn))
+        let result = checkout_db_connection(app_state).map(DbConnection);
+
+        ready(result)
+    }
+}
+
+fn checkout_db_connection(
```
mahatoankitkumar: remove

crates/superposition/src/workspace/handlers.rs
```diff
@@ -203,9 +208,9 @@ async fn update_handler(
 ) -> superposition::Result<Json<WorkspaceResponse>> {
     let request = request.into_inner();
     let workspace_name = workspace_name.into_inner();
+    let DbConnection(mut conn) = db_conn;
```
mahatoankitkumar: unnecessary changes in this file

Datron (changes requested):
Merge the error types and check the locking mechanism

crates/service_utils/src/workspace_lock.rs
sauraww: It is atomic - wouldn't erase the previous lock owner.

sauraww:
```
If the lock is active -> updated_rows == 0 → no row got touched  → LockContended
If the lock is expired or absent → updated_rows == 1 → we overwrite stale data, which is correct
```

sauraww: This is an atomic compare and swap.

Datron: How is this atomic? There are no DB semantics that indicate this behaviour.

sauraww: This is atomic because the compare condition and write happen in one UPDATE statement. Postgres row-locks the row for UPDATE;

Datron: @sauraww but the next request will call acquire lock, which calls this function and the lock is then allocated to the new request even though the existing one is still running. I am not talking about two requests simultaneously trying to acquire the lock, I am talking about one after another where the atomic operation would erase the previous lock and give it to the new acquirer

crates/service_utils/src/workspace_lock.rs
sauraww: this is workspace update @Datron . 
In the workspace handlers - also we are doing in the same way. 
Though we don't need any keep , i can just pass org and workspace. 

Here is the reference from the handler 

```
            diesel::update(workspaces::table)
                .filter(workspaces::organisation_id.eq(&org_id.0))
                .filter(workspaces::workspace_name.eq(&workspace_name))
 ```

crates/service_utils/src/workspace_lock.rs
sauraww: @Datron Yes, a later request can acquire the lock if the previous lock has expired by TTL while the previous handler is still running. That is intentional stale lock recovery.

Datron: Got it

mahatoankitkumar (approved):

Datron (approved):

PR #822, opened 2026-01-06
refactor(frontend): Convert all Leptos component function names to PascalCase

### LINKED ISSUE #821: Refactor leptos component function names
Use `PascalCase` for leptos component definition, instead of `snake_case`

knutties:
- Renamed 121 component functions from snake_case to PascalCase
- Improves consistency between function definitions and component usage in JSX
- Makes component functions easier to search and identify
- All component definitions now match their usage in view! macros
- Build verified successfully with no errors

Examples of changes:
- experiment_list -> ExperimentList
- button -> Button
- side_nav -> SideNav
- experiment_form -> ExperimentForm

fixes #821

## Summary by CodeRabbit

* **Refactor**
  * Standardized component naming conventions across the frontend to improve code consistency and maintainability.
  * Enhanced several components with additional configuration options for improved flexibility.

* **Chores**
  * Added automated validation to enforce component naming standards.
  * Updated build system to include new validation checks.

### DIFF
--- a/makefile
+++ b/makefile
@@ -9,6 +9,7 @@ CARGO_FLAGS := --color always --no-default-features
 EXCLUDE_PACKAGES := experimentation_client_integration_example superposition_sdk
 FMT_EXCLUDE_PACKAGES_REGEX := $(shell echo "$(EXCLUDE_PACKAGES)" | sed "s/ /|/g")
 LINT_FLAGS := --workspace --all-targets --all-features $(addprefix --exclude ,$(EXCLUDE_PACKAGES)) --no-deps
+COMPONENT_NAME_FLAGS := 
 CARGO_TARGET_DIR := $(shell cargo metadata --no-deps --format-version 1 | jq -r .target_directory)
 
 # Get all workspace members and filter out excluded ones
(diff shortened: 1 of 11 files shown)


ayushjain17:
Should we consider adding some lint check to force PascalCase for functions with component macro?

makefile
```diff
@@ -49,31 +49,7 @@ DB_UP = $(shell $(call check-container,$(DB_CONTAINER_NAME)))
...
-	smithy-api-docs
-	smithy-updates
-	validate-aws-connection
-	validate-psql-connection
-	uniffi-bindings
-	test-js-provider
-	test-py-provider
-	test-kotlin-provider
-	test-rust-provider
+.PHONY: cac check-component-names ci-test clients db-init grafana-local kill local-docs-view node-dependencies run schema-file setup setup-clients smithy-clean smithy-build smithy-clean-build smithy-api-docs smithy-updates validate-aws-connection validate-psql-connection uniffi-bindings test-js-provider test-py-provider test-kotlin-provider test-rust-provider
```
ayushjain17: we can keep this multi-lined as well by adding `\` at the end of each line
that will allow us to keep it clean as well as correct

makefile
```diff
@@ -319,7 +295,10 @@ lint-fix: lint
 check: FMT_FLAGS += --check
 check: LEPTOS_FMT_FLAGS += --check
 check: LINT_FLAGS += -- -Dwarnings
-check: fmt leptosfmt lint
+check: fmt leptosfmt lint check-component-names
+
+check-component-names:
+	@./scripts/check_component_names.sh
```
ayushjain17: I dont think this is working as intended, because there are 6 more instances in code which are not converted to PascalCase and there is one behind doc comments as well
would have preferred a linter way of achieving this

scripts/check_component_names.sh
```diff
@@ -0,0 +1,67 @@
...
+                ERRORS=$((ERRORS + 1))
+            fi
+        done < <(grep -n "#\[component\]" "$file" | while IFS=: read -r line_num _; do
+            # Look at next few lines for the function definition (pub or private)
+            # Only match top-level function definitions (not nested helpers)
+            for offset in 1 2 3 4; do
+                next_line=$((line_num + offset))
+                line_content=$(sed -n "${next_line}p" "$file")
+                # Match fn at start of line or after 'pub ', not indented helper functions
+                if echo "$line_content" | grep -q "^pub fn [a-z_][a-z0-9_]*\|^fn [a-z_][a-z0-9_]*"; then
```
ayushjain17: this grep is only looking for `pub` functions or lines which start with `fn`  only
it is not covering all the cases

ayushjain17: if we have to regex way only, then I think this would be a better regex, covering both the lines
```regex
#\[component\]\n.*fn (.*)\(
```
without any start or end delimiters

and then what whatever group value is captured, just need to check that the first character is not a lower case alphabet and the entire capture does not include any occurrences of `_`
which could be represented by a regex like this
```regex
^[A-Z][A-Za-z0-9]*$
```

PR #823, opened 2026-01-06
test: add test cleanup script for organizations and workspaces

knutties:
## Problem
In local dev setup - `bun test` leaves behind organisations used for testing.  It is a pain to cleanup manually.

## Solution
Add conditional cleanup mechanism for test organizations and workspaces that preserves data on test failures for debugging.

- Add cleanup.sql: SQL script to delete organizations with IDs starting with 'org%' and their associated workspace schemas
- Add run-cleanup.ts: TypeScript script to execute cleanup with detailed logging via PostgreSQL NOTICE messages
- Add test-with-cleanup.sh: Shell wrapper that runs tests and triggers cleanup only on successful test execution
- Add npm scripts: test (normal), test:clean (with cleanup), cleanup (manual)
- Add pg dependency for PostgreSQL connectivity

Usage:
  bun test          - Run tests preserving data for debugging
  npm run test:clean - Run tests with cleanup on success
  npm run cleanup   - Manually cleanup test organizations

No

## Pre-deployment activity
No

## Post-deployment activity
No

## Summary by CodeRabbit

* **Tests**
  * Added test infrastructure for database cleanup and test automation, including a SQL cleanup script and shell automation to manage test data between test runs.
  
* **Chores**
  * Added PostgreSQL client dependency to support test database operations.


tests/package.json
```diff
@@ -9,6 +9,13 @@
...
   "dependencies": {
-    "@juspay/superposition-sdk": "file:../clients/javascript/sdk"
+    "@juspay/superposition-sdk": "file:../clients/javascript/sdk",
+    "dotenv": "^17.2.3",
+    "pg": "^8.16.3"
+  },
+  "scripts": {
+    "test": "bun test",
+    "test:clean": "./test-with-cleanup.sh",
+    "cleanup": "bun run-cleanup.ts"
```
ayushjain17: whats the expected flow of running the test now ?
when does these execute ?

would i not use `make test` for these ?

I am sort of expecting that by default when I run tests, the cleanup should execute
and if I find that there are issues and I want to debug, I can re-run the tests with cleanup disabled so that I get the data to test

tests/test-with-cleanup.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Running tests..."
+bun test
+TEST_EXIT_CODE=$?
+
+if [ $TEST_EXIT_CODE -eq 0 ]; then
+    echo ""
+    echo "========================================"
+    echo "Tests passed. Running cleanup..."
+    echo "========================================"
+    bun run-cleanup.ts
```
ayushjain17: do we want to do all these things one after another ?
we can simply have a bash file which executes the psql file
which can be specified in `tests/package.json` as well

We wont need `run-cleanup.ts` then and other dep addition in tests' package.json

tests/package.json
knutties: We can update the make test flow to use bun run test:clean - it will automatically skip the clean-up if tests fail.  But you can force run cleanup if you want.

tests/test-with-cleanup.sh
knutties: This is a good point - we can maybe just run a shell that execute - but then parsing env file and sending it to psql might involve work which this does rather neatly by reading the env file.

tests/package.json
knutties: I don't want to necessarily re-run to reproduce - I might want to look at what happened in the broken run as well.  So not cleaning up always has a use-case.

tests/test-with-cleanup.sh
ayushjain17: you can do something like this
```sh
if [ -f .env ]; then
    set -a
    . .env
    set +a
fi
```

ayushjain17: and then you get all the envs

knutties: psql command line does not take it in that format right ?  Need to parse and send it via command line opts right ?

ayushjain17: but its only the DB Url right which you are extracting from ENV, right ?
your script could be something like:

```sh
if [ -f .env ]; then
    set -a
    . .env
    set +a
fi

# we have the full url as an ENV already
psql "$DATABASE_URL" -f cleanup.sql
```

tests/package.json
ayushjain17: yeah we could do that, to avoid cleanup if the tests fails
didn't suggest it upfront as I thought it might quire some complex changes

tests/package.json
```diff
@@ -10,5 +10,10 @@
   },
   "dependencies": {
     "@juspay/superposition-sdk": "file:../clients/javascript/sdk"
+  },
+  "scripts": {
+    "test": "bun test",
+    "test:clean": "./test-with-cleanup.sh",
+    "cleanup": "psql postgres://postgres:docker@localhost:5432/config -f cleanup.sql"
```
ayushjain17: just to confirm, this is just separate command for future use only right ?
its not being used as of now right ?

ayushjain17: can we do something like this ?
would need to install `cross-env` node dependency for this (as psql tries to read DATABASE_URL from env)
```suggestion
    "cleanup": "cross-env psql -f cleanup.sql"
```
or we can do something like this, but then ENV auto reading wont work in Windows
```suggestion
    "cleanup": "psql -f cleanup.sql"
```

tests/bun.lockb
ayushjain17, line 1: is the bun lock correct ?
there are currently no dependency changes in the PR

tests/test-with-cleanup.sh
knutties: Makes a lot of sense.  Thanks for the feedback.  This is addressed now.

tests/package.json
knutties: `cross-env` is a good suggestion (man traded off one dependency for the other)

ayushjain17: we can skip this, we remove this command
as we are not using this command in the cleanup shell script also

tests/bun.lockb
knutties: bun.lockb is deprecated it appears - will regenerate a bun.lock and add it to the PR.

PR #825, opened 2026-01-07
fix(webhook): encrypted key injection

ayushjain17:
## Problem
Encrypted keys not being injected in webhook

## Summary by CodeRabbit

* **New Features**
  * Added encrypted value support for custom webhook headers. Webhook headers can now securely reference encrypted secrets that are automatically decrypted when webhooks are transmitted. The system intelligently handles both encrypted and non-encrypted values, preserving variable substitution functionality for standard headers to maintain full backward compatibility.

### DIFF
--- a/crates/service_utils/src/helpers.rs
+++ b/crates/service_utils/src/helpers.rs
@@ -485,8 +485,17 @@ where
         &workspace_request.workspace_id,
     );
 
-    webhook.custom_headers.iter().for_each(|(key, value_str)| {
-        let rendered = substitute_variables(value_str, &variables_map);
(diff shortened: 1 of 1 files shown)


crates/service_utils/src/helpers.rs
```diff
@@ -485,8 +485,17 @@ where
         &workspace_request.workspace_id,
     );
 
-    webhook.custom_headers.iter().for_each(|(key, value_str)| {
-        let rendered = substitute_variables(value_str, &variables_map);
+    webhook.custom_headers.iter().for_each(|(key, value)| {
```
ayushjain17, line 488: NOTE: this logic has to be cleaned up when we introduce secrets

PR #831, opened 2026-01-09
fix: exp and config partial apply

ayushjain17:
## Changes
1. fixed local cohort resolution behaviour in case of filtering - local cohort value should not be set if the cohorting dimension's value is not provided
2. `get_satisfied_experiments` in `superposition_core` is wrongly calling `partial_apply` inside it instead of `apply`
3. added corresponding `get_filtered_satisfied_experiments` which is supposed to use `partial_apply` inside it
4. added support for `OR` behaviour in case of partial apply - in full resolution only `variantIds` is allowed to be an array value in the resolve context, in partial apply all dimensions in the filter context are allowed to be array value

Support in UI for `OR` based filtering is not yet added

### DIFF
--- a/crates/superposition_types/src/contextual.rs
+++ b/crates/superposition_types/src/contextual.rs
@@ -36,10 +36,10 @@ pub trait Contextual: Clone {
         contexts
             .into_iter()
             .filter(|context| {
-                let variables: Map<String, Value> = context.get_condition().into();
-                dimension_keys
-                    .iter()
-                    .all(|dimension| variables.contains_key(dimension))
+                !context
+                    .get_condition()
+                    .keys()
+                    .all(|key| !dimension_keys.contains(key))
             })
             .collect()
     }
@@ -97,7 +97,7 @@ mod tests {
                 config.contexts.clone(),
                 &get_dimension_data2().keys().cloned().collect::<Vec<_>>()
             ),
-            get_dimension_filtered_contexts2()
+            get_dimension_filtered_contexts1()
         );
 
         assert_eq!(
@@ -105,7 +105,7 @@ mod tests {
                 config.contexts,
                 &get_dimension_data3().keys().cloned().collect::<Vec<_>>()
             ),
-            Vec::new()
+            get_dimension_filtered_contexts2()
         );
     }
 }
--- a/crates/context_aware_config/src/api/config/handlers.rs
+++ b/crates/context_aware_config/src/api/config/handlers.rs
@@ -15,7 +15,10 @@ use itertools::Itertools;
 use serde_json::{Map, Value, json};
 #[cfg(feature = "high-performance-mode")]
 use service_utils::service::types::{AppHeader, AppState};
-use service_utils::service::types::{DbConnection, WorkspaceContext};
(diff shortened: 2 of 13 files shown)


ayushjain17:
@coderabbitai full review

crates/superposition_types/src/logic.rs
```diff
@@ -37,15 +38,42 @@ fn apply_logic(
...
     context: &Map<String, Value>,
 ) -> bool {
-    apply_logic(condition, context, true)
+    for (dimension, value) in condition {
+        let Some(context_value) = context.get(dimension) else {
+            continue; // Skip dimensions not in context (partial matching)
+        };
+
+        // For array context values, check containment
+        if let Value::Array(context_values) = context_value {
```
knutties, line 64: @ayushjain17  - the input in the handler is a JSON stringified array or CSV ?  You showed JSON stringified array.

In some other places we do repeated values of the same key and collect them in the handler.

Just asking if it is all consistent in how we get multi-value input in the handler?

ayushjain17: Need to verify once which all spec are supported

Input spec is not changed via this change-log

Here, updating the internal logic only

ayushjain17: the spec for these APIs are different, they are un-typed json
we cant have that other behaviour here, it has to be JSON stringified only

repeating values with same key is only possible for types which have type `CommaSeparatedQParams`

crates/superposition_core/src/experiment.rs
```diff
@@ -208,6 +208,33 @@ pub fn get_satisfied_experiments(
     Ok(running_experiments)
 }
 
+pub fn get_filtered_satisfied_experiments(
```
Datron: Function needs a better name - it implies that `get_statisfied_experiments` doesn't filter. This function is filtering contexts by partially application, and can be merged into `get_satisfied_experiments`. Take an argument called `partial_apply_filter`: bool which does filtering via contexts

Datron (changes requested):

crates/superposition_core/src/experiment.rs
ayushjain17: @Datron this is what it was called in the exp_client crate
This was missed in the process of creating core crate

So i added this in this PR

crates/superposition_types/src/contextual.rs
```diff
@@ -97,15 +97,15 @@ mod tests {
                 config.contexts.clone(),
                 &get_dimension_data2().keys().cloned().collect::<Vec<_>>()
             ),
-            get_dimension_filtered_contexts2()
+            get_dimension_filtered_contexts1()
```
Datron, line 100: I guess the comment is obvious, what is `get_dimension_filtered_contexts1` and `get_dimension_filtered_contexts2`? Can we document differences at least in doc comments?

ayushjain17: This is for test
Just static values to compare with

To keep the tests clean these were moved out as separate entity
No as such meaning behind them

PR #832, opened 2026-01-09
fix: function attachment logic for dimension and default config

ayushjain17:
## Problem
No validation on the type of function given in input for validation or compute function attachment to default config and dimension

## Summary by CodeRabbit

* **Refactor**
  * Unified how function kinds are passed and honored across backend validation and retrieval paths for more consistent behavior.
  * Frontend form code refactored to use a shared helper for populating function dropdowns, simplifying state handling.

* **Bug Fixes / Validation**
  * Enforced runtime checks to ensure retrieved functions match their declared kind, surfacing mismatches as validation errors.

### DIFF
--- a/crates/context_aware_config/src/api/context/helpers.rs
+++ b/crates/context_aware_config/src/api/context/helpers.rs
@@ -147,8 +147,12 @@ pub fn validate_condition_with_functions(
         }
     }
 
-    let dimension_functions_map =
-        get_functions_map(conn, new_keys_function_array, schema_name)?;
+    let dimension_functions_map = get_functions_map(
+        conn,
+        new_keys_function_array,
+        FunctionType::ValueValidation,
+        schema_name,
+    )?;
     for (key, value) in context_map.iter() {
         if let Some(functions_map) = dimension_functions_map.get(key) {
             if let (function_name, Some(function_code), Some(published_runtime_version)) = (
@@ -192,8 +196,12 @@ pub fn validate_override_with_functions(
         .filter_map(|(key_, f_name)| f_name.map(|func| (key_, func)))
         .collect();
 
-    let default_config_functions_map =
-        get_functions_map(conn, new_keys_function_array, schema_name)?;
+    let default_config_functions_map = get_functions_map(
+        conn,
+        new_keys_function_array,
+        FunctionType::ValueValidation,
+        schema_name,
+    )?;
     for (key, value) in override_.iter() {
         if let Some(functions_map) = default_config_functions_map.get(key) {
             if let (function_name, Some(function_code), Some(published_runtime_version)) = (
@@ -226,6 +234,7 @@ pub fn validate_override_with_functions(
 fn get_functions_map(
     conn: &mut DBConnection,
     keys_function_array: Vec<(String, String)>,
+    function_type: FunctionType,
     schema_name: &SchemaName,
 ) -> superposition::Result<HashMap<String, FunctionInfo>> {
     let functions_map: HashMap<String, FunctionInfo> = get_published_functions_by_names(
@@ -234,6 +243,7 @@ fn get_functions_map(
             .iter()
             .map(|(_, f_name)| f_name.clone())
             .collect(),
+        function_type,
         schema_name,
     )?
     .into_iter()
--- a/crates/context_aware_config/src/api/default_config/handlers.rs
+++ b/crates/context_aware_config/src/api/default_config/handlers.rs
@@ -31,7 +31,7 @@ use superposition_types::{
     custom_query::PaginationParams,
     database::{
         models::{
-            cac::{self as models, Context, DefaultConfig},
+            cac::{self as models, Context, DefaultConfig, FunctionType},
             Description,
         },
         schema::{self, contexts::dsl::contexts, default_configs::dsl},
@@ -44,7 +44,10 @@ use crate::helpers::put_config_in_redis;
 use crate::{
     api::{
         context::helpers::validation_function_executor,
-        functions::{helpers::get_published_function_code, types::FunctionInfo},
+        functions::{
+            helpers::{check_fn_published, get_published_function_code},
+            types::FunctionInfo,
+        },
     },
     helpers::{add_config_version, validate_change_reason},
 };
@@ -131,6 +134,7 @@ async fn create_handler(
 
     validate_fn_published(
         &default_config.value_compute_function_name,
+        FunctionType::ValueCompute,
         &mut conn,
         &workspace_context.schema_name,
     )?;
@@ -252,6 +256,7 @@ async fn update_handler(
     if let Some(ref value_compute_function_name) = req.value_compute_function_name {
         validate_fn_published(
             value_compute_function_name,
+            FunctionType::ValueCompute,
             &mut conn,
             &workspace_context.schema_name,
         )?;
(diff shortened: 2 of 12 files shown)


crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -282,6 +290,7 @@ pub fn validate_override_with_functions(
 fn get_functions_map(
```
knutties, line 234: We can now call this `get_functions_map_by_type` to be more specific?

crates/context_aware_config/src/api/dimension/validations.rs
```diff
@@ -242,11 +242,12 @@ pub fn validate_cohort_position(
 
 fn check_fn_published(
     fn_name: &str,
+    fn_type: FunctionType,
```
knutties: Why is a type needed when the name is unique by itself for this get use-case?  In this case they seem to be mutually exclusive.

This is actually indicative of a deeper problem - are you doing this because we can change function types later (I don't think we allow it).

Once associated with a function with a name - the type is not needed right or is there any case I am missing?

crates/context_aware_config/src/api/functions/helpers.rs
```diff
@@ -24,10 +26,15 @@ pub fn fetch_function(
 pub fn get_published_function_code(
```
knutties, line 25: Same comment as the `check_fn_published` change.

crates/frontend/src/components/dimension_form.rs
```diff
@@ -525,23 +525,26 @@ pub fn dimension_form(
...
-                                .iter()
-                                .for_each(|ele| {
-                                    if ele.function_type == FunctionType::ValueValidation {
-                                        validation_function_names.push(ele.function_name.clone());
-                                    } else {
-                                        value_compute_function_names
-                                            .push(ele.function_name.clone());
-                                    }
-                                });
+                            let validation_function_names = std::iter::once("None".to_string())
```
knutties: Can we DRY getting this list for the select box in a common function across the default-config and dimension forms ?

crates/context_aware_config/src/api/dimension/validations.rs
ayushjain17: I intended to add this check to prevent devs from adding a get function call for a fn of type `x`, where the intended type of the fn get was `y`

it is a redundant check by all means, but just to ensure that an intended change is not pushed and is easily caught in dev testing step itself

crates/context_aware_config/src/api/context/helpers.rs
ayushjain17: this still fetches by name and only validates by type

knutties: It fetches the name (not by name) - but only by type - this is like getElementsByClassName right ?

ayushjain17: This was previously 
Fetch all functions by name (input : array of names)
Now this is
Fetch all functions by name of this particular type (common for all)

crates/context_aware_config/src/api/dimension/validations.rs
knutties: our typing feels broken a bit.  Do we have a Function type that we can use around instead of passing name and type in separate args?  From the code I see we don't have - except for the db models.  We probably need one type in the Domain model - to pass around to make this tight.

knutties: Maybe worth introducing the domain type at this point with something like:

```rust
pub enum FunctionType {
    #[default]
    ValueValidation,
    ValueCompute,
    ContextValidation,
    ChangeReasonValidation
}

// The container
pub struct Function {
    pub name: String,
    pub kind: FunctionType,
    // .... more fields that are relevant to domain
}
```

ayushjain17: The type which you suggested is the exact setup of which which we already have

ayushjain17: But why do we need extra set of domain types?
Didn't understand the disconnect

crates/context_aware_config/src/api/dimension/validations.rs
knutties: If we are not classifying the errors cleanly - we need to add to this error messaging saying:

```
Function {} doesn't exist / function code not published yet / function type is mis-matched
```

which starts to look a bit fishy :)

crates/context_aware_config/src/api/dimension/validations.rs
knutties: This particular place is a not a domain type use-case.  Let us ignore this.

crates/context_aware_config/src/api/dimension/validations.rs
knutties: Just to confirm understanding - this error should now just have not published wordings.  Function will exist and of the same type if we reach here right (because of the short return).

crates/context_aware_config/src/api/context/helpers.rs
knutties: what I see differently is that - it is:

> Fetch all functions with name (not by name) of this particular type

We have already told it is a map to give the dev a clue of what this function will return - if we go into what type of map it is - we are possibly getting too verbose.

Being a bit OCD about this - to ensure code-base is in shape for humans and machines :)

knutties, line 234:
```suggestion
fn get_functions_map_by_type(
```

crates/context_aware_config/src/api/dimension/validations.rs
knutties:
```suggestion
            "Function {}'s published code does not exist.",
```

ayushjain17: Yeah

crates/context_aware_config/src/api/default_config/handlers.rs
```diff
@@ -294,17 +296,18 @@ async fn update_handler(
 
 fn validate_fn_published(
```
mahatoankitkumar, line 300: I guess similar function exists in dimension.validation called check_fn_published. i think both can be unified.

crates/context_aware_config/src/api/functions/helpers.rs
```diff
@@ -24,19 +25,29 @@ pub fn fetch_function(
...
+    f_type: FunctionType,
     schema_name: &SchemaName,
 ) -> superposition::Result<FunctionInfo> {
     let function = functions
         .filter(schema::functions::function_name.eq(f_name))
         .select(FunctionInfo::as_select())
         .schema_name(schema_name)
         .first(conn)?;
+
+    if function.function_type != f_type {
```
mahatoankitkumar, line 37: Can we not write this with the filter itself and throw an error saying
`Function '{}' with type {:?} not found`
Rather than filtering from database and again checking

crates/context_aware_config/src/api/functions/helpers.rs
```diff
@@ -24,19 +25,29 @@ pub fn fetch_function(
...
+        ));
+    }
+
     Ok(function)
 }
 
 pub fn get_published_functions_by_names(
     conn: &mut DBConnection,
     function_names: Vec<String>,
+    f_type: FunctionType,
```
mahatoankitkumar, line 50: I think the name misleading, how function_type is being used. This being a util function, sending an array of names and just one function_type. Either the name should be changed or the function logic.

crates/frontend/src/utils.rs
```diff
@@ -417,3 +420,17 @@ pub fn to_title_case(input: &str) -> String {
         .collect::<Vec<String>>()
         .join(" ")
 }
+
+pub fn get_fn_names_by_type(
```
mahatoankitkumar, line 424: better function name using filter and maybe specify its adding none or better add none in the caller function.
Keep the function just for filtering.

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/functions/helpers.rs
ayushjain17: It does not fetch by function type, it fetches by function name and only validates by function type

crates/context_aware_config/src/api/functions/helpers.rs
ayushjain17: It does not fetch by function type, it fetches by function name and only validates by function type

crates/frontend/src/utils.rs
ayushjain17: Certain behaviours are very context specific
If we start putting everything in function name then we will start having endlessly long function names itself

Moving None outside and keeping only filter here is counter productive

If you still dont find things valid enough
Then please suggest appropriate name

crates/context_aware_config/src/api/default_config/handlers.rs
ayushjain17: I think they had difference in the logic
Need to verify once

PR #833, opened 2026-01-13
fix: Haskell provider breeze

Datron:
## Problem
the status returned by the haskell provider is wrong

## Solution
Get the correct status by checking the config data stored in the polling thread

## Release Notes

* **New Features**
  * Added full configuration resolution capability.
  * Added optional prefix filtering parameter for configuration lookups.
  * Added variant lookup functionality for querying applicable variants.
  * Improved asynchronous configuration fetching and retrieval.

### DIFF
--- a/clients/haskell/open-feature-provider/superposition-open-feature-provider.cabal
+++ b/clients/haskell/open-feature-provider/superposition-open-feature-provider.cabal
@@ -1,4 +1,4 @@
-cabal-version:   3.4
+cabal-version:   3.0
 name:            superposition-open-feature-provider
 version:         0.1.0.0
 license:         BSD-3-Clause
@@ -32,14 +32,22 @@ library
                       SuperpositionSDK,
                       superposition-bindings
     hs-source-dirs:   lib
-    default-language: GHC2021
+    default-language: Haskell2010
     default-extensions: TypeFamilies,
                         DisambiguateRecordFields,
                         RecordWildCards,
                         LambdaCase,
                         OverloadedStrings,
                         ExistentialQuantification,
-                        RankNTypes
+                        RankNTypes,
+                        MultiParamTypeClasses,
+                        FlexibleInstances,
+                        TupleSections,
+                        TypeApplications,
+                        DeriveAnyClass,
+                        DeriveGeneric,
+                        StandaloneDeriving,
+                        ImportQualifiedPost
 
 -- test-suite superposition-open-feature-provider-test
 --     import:           warnings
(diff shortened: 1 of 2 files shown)


clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -94,8 +128,40 @@ getResolvedConfig params = do
...
+        qry
+        tkey
+        pfltr
+        ebuf
+  err <- peekCAString ebuf
+  traverse_ freeNonNull [exps, expGrps, dims, qry, tkey, pfltr, ebuf]
+  pure $ case (res, err) of
+    (Just vars, []) -> Right vars
+    (Nothing, []) -> Left "null pointer returned"
+    _ -> Left err
```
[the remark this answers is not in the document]
knutties: @Datron - this came in my toml PR as well - should we just address this ?

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
```diff
@@ -35,6 +34,20 @@ foreign import capi "superposition_core.h core_get_resolved_config"
     -- | resolved config json
     IO CString
 
+foreign import capi "superposition_core.h core_free_string"
+  free_string :: CString -> IO ()
+
+foreign import capi "superposition_core.h core_get_applicable_variants"
```
knutties, line 40: @Datron - why do we rename the generated core.h file to `superposition_core.h` in our builds ?  Can't we use it as it is ?

clients/haskell/open-feature-provider/lib/Data/OpenFeature/SuperpositionProvider.hs
```diff
@@ -140,15 +150,24 @@ resolveValue p@(SuperpositionProvider {..}) key ec = do
       runLogger $ Log.logError ("An error occured while resolving a key." Log.:# lctx)
       pure $ Left $ e
 
+resolveFullConfig ::
```
knutties, line 153: Why two functions with exactly same signature @Datron  - `resolveFullConfig` and `evalConfig` - or am I missing some Haskell basics here?

Datron: @knutties keeping the naming convention the same as the spec - its just that resolveFullConfig and evalConfig does the same thing. This is the same as the rust provider PR I have raised

clients/haskell/superposition-bindings/lib/FFI/Superposition.hs
Datron: @knutties some languages expect the binary name and the header file name to be the same - that is why we do the rename

clients/haskell/open-feature-provider/lib/Data/OpenFeature/SuperpositionProvider.hs
knutties: You mean this one https://github.com/juspay/superposition/pull/755 ?

PR #835, opened 2026-01-15
feat(UI): allow variant ids as array for resolve/filter

ayushjain17:
## Change log
1. Removed jsonlogic related dead code form frontend
2. Added support to send array of variantIds from frontend for all endpoints which perform any type of resolve operation (be it partial or full)

### DIFF
--- a/crates/frontend/src/components/cohort_schema.rs
+++ b/crates/frontend/src/components/cohort_schema.rs
@@ -516,18 +516,16 @@ fn CohortForm(
                                 let schema_type = schema_type.clone();
                                 move || {
                                     jsonlogic::condition::try_from_condition_json(&value_rws.get())
-                                        .unwrap_or(
-                                            Condition::new_with_default_expression(
-                                                dimension.clone(),
-                                                schema_type.clone(),
-                                            ),
-                                        )
+                                        .unwrap_or_else(|_| Condition {
+                                            variable: dimension.clone(),
+                                            value: schema_type.default_value(),
+                                        })
                                 }
                             });
                             let input_type = InputType::from((schema_type.clone(), enum_variants));
                             let fn_environment = Memo::new(move |_| {
                                 FunctionEnvironment {
-                                    context: Conditions(vec![condition.get()]).as_context_json(),
+                                    context: Conditions(vec![condition.get()]).into(),
                                     overrides: Map::new(),
                                 }
                             });
@@ -554,19 +552,18 @@ fn CohortForm(
                                         allow_remove=false
                                         condition=condition.get()
                                         input_type
-                                        schema_type=schema_type.clone()
+                                        schema_type
                                         value_compute_callbacks
                                         on_remove=|_| ()
                                         on_value_change=move |new_expr| {
                                             value_rws
                                                 .set(
-                                                    jsonlogic::expression::to_expression_json(
+                                                    jsonlogic::expression::to_condition_json(
                                                         &new_expr,
                                                         &condition.get().variable,
                                                     ),
                                                 );
                                         }
-                                        on_operator_change=|_| ()
                                     />
                                 </div>
                             }
@@ -585,7 +582,7 @@ fn CohortForm(
                                     <NumberArrayInput
                                         unique=true
                                         options=value_rws
-                                            .with(|v| {
+                                            .with_untracked(|v| {
                                                 v.as_object()
                                                     .and_then(|obj| obj.get("in"))
                                                     .and_then(Value::as_array)
@@ -626,7 +623,7 @@ fn CohortForm(
                                     <StringArrayInput
                                         unique=true
                                         options=value_rws
-                                            .with(|v| {
+                                            .with_untracked(|v| {
                                                 v.as_object()
                                                     .and_then(|obj| obj.get("in"))
                                                     .and_then(Value::as_array)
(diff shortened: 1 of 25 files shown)


crates/frontend/src/components/input.rs
```diff
@@ -600,43 +600,51 @@ fn array_input<T: Clone + PartialEq + 'static + Display + FromStr>(
...
+                                                AlertType::Error,
+                                                5000,
+                                            );
+                                            return;
+                                        };
+                                        if !unique || !options_rws.get().contains(&value_t) {
+                                            value_rws.set(value_t.to_string());
+                                            options_rws.update(|opts| opts.push(value_t));
+                                            on_change.call(options_rws.get());
+                                        } else {
```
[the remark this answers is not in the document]
ayushjain17: this is not the correct fix, check the latest code in this commit, to understand the correct way of doing it

crates/frontend/src/components/condition_pills.rs
```diff
@@ -61,8 +62,14 @@ pub fn condition_expression(
                 .with_value(|v| {
                     (
                         v.variable.clone(),
-                        Operator::from(&v.expression),
-                        v.expression.to_value().html_display(),
+                        if v.variable == "variantIds" && resolve_summary {
```
knutties:
```suggestion
match v.variable.as_str() {
    "variantIds" if resolve_summary => "in",
    "variantIds" => "has",
    _ => "==",
},
```

Could we use match with guards here for idiomatic Rust?

crates/superposition_types/src/config.rs
```diff
@@ -179,6 +179,10 @@ impl Condition {
         }
         Ok(true)
     }
+
+    pub fn into_inner(self) -> Map<String, Value> {
+        self.0
+    }
```
Datron, line 190: Deref is implemented on this right? Why do we need into_inner?

ayushjain17: `Deref` borrows and this does not work for types which do not implement `Copy`
And `into_inner` is a common practice to take ownership

crates/frontend/src/components/cohort_schema.rs
```diff
@@ -516,18 +516,16 @@ fn cohort_form(
                                 let schema_type = schema_type.clone();
                                 move || {
                                     jsonlogic::condition::try_from_condition_json(&value_rws.get())
-                                        .unwrap_or(
-                                            Condition::new_with_default_expression(
-                                                dimension.clone(),
-                                                schema_type.clone(),
-                                            ),
-                                        )
+                                        .unwrap_or_else(|_| Condition {
```
mahatoankitkumar, line 519: Should we add a log in this else part?

ayushjain17:
@mahatoankitkumar 
this frontend change, we dont have test for frontend
the alignment is not broken, that is the default alignment which we are using

crates/frontend/src/components/context_form.rs
```diff
@@ -196,15 +246,23 @@ pub fn ContextForm(
...
                 logging::log!(
-                    "Inserting dimension {:?} with schema {:?}",
+                    "Inserting dimension {} with schema {:?}",
                     dimension.dimension,
                     dimension.schema
                 );
-                context.push(Condition::new_with_default_expression(
-                    dimension.dimension.clone(),
+                let default_value = if dimension.dimension == "variantIds" && resolve_mode
+                {
```
sauraww: The && resolve_mode check is redundant if variantIds is already filtered out in non-resolve mode. It is a dead code.
We are already filtering it out above: 
`
.filter(|dimension| dimension.dimension != "variantIds")
`

crates/frontend/src/components/condition_pills.rs
```diff
@@ -61,8 +62,12 @@ pub fn ConditionExpression(
                 .with_value(|v| {
                     (
                         v.variable.clone(),
-                        Operator::from(&v.expression),
-                        v.expression.to_value().html_display(),
+                        match v.variable.as_str() {
+                            "variantIds" if resolve_summary => "in",
+                            "variantIds" => "has",
```
sauraww, line 67: Can we add a comment here ?

crates/frontend/src/logic.rs
```diff
@@ -103,78 +20,41 @@ impl Condition {
 pub struct Conditions(pub Vec<Condition>);
 
 impl Conditions {
-    pub fn includes(&self, variable: &String) -> bool {
+    pub fn includes(&self, variable: &str) -> bool {
```
sauraww, line 23: The *variable dereference is unnecessary since String implements PartialEq<&str>
Can we remove it ?

sauraww (changes requested):

mahatoankitkumar (approved):

crates/frontend/src/components/condition_pills.rs
ayushjain17: for what ?

crates/frontend/src/components/context_form.rs
```diff
@@ -26,6 +27,30 @@ pub enum TooltipType {
     None,
 }
 
+#[derive(Clone, Copy)]
+enum ResolveOperator {
+    Is,
+    In,
+}
```
Datron: nit: you can use derive_more::Display here instead of implementing the trait

sauraww (approved):

PR #839, opened 2026-01-23
chore: remove redundant files

Datron:
## Problem
Lot of redundant files present in the repo

## Solution
Remove them

## Summary by CodeRabbit

* **Chores**
  * Removed internal development documentation files and editor configuration to streamline repository structure and reduce maintenance overhead. No changes to application code or public interfaces.


Datron:
@knutties Agents don't use these files when working. The docs are also enough, and given that we will mostly update the docs I would not want an agent to execute anything based on these files

PR #840, opened 2026-01-23
feat: move from nodeJS to rustyscript

Datron:
## Problem
We are using a nodeJS runtime via the shell to execute JS code within superposition. This setup has limitations on the size of responses, type validations and is fragile when making modifications

## Solution
Replace the node runtime with rustyscript, a crate that embeds a deno runtime within rust. This makes it easy to run JS code through rust - with type validations, getting the right response and allows us to properly add on libraries like crypto and fetch

## Post-deployment activity
Any JS functions using axios should be moved to use fetch

## Release Notes

* **New Features**
  * Enhanced JavaScript validation execution with improved runtime support.

* **Bug Fixes**
  * Improved transaction handling for bulk context operations ensuring better data consistency.

* **Chores**
  * Updated core dependencies for improved stability and compatibility.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,8 @@ once_cell = { version = "1.18.0" }
 regex = "1.9.1"
 reqwest = { version = "0.11.18", features = ["json"] }
 rs-snowflake = "0.6.0"
-serde = { version = "^1", features = ["derive"] }
+rustyscript = { version = "0.12.3", default-features = false, features = ["safe_extensions", "web", "worker"] }
+serde = { version = "=1.0.221", features = ["derive"] }
 serde_json = { version = "1.0.140" }
 secrecy = "0.10"
 strum = "0.25"
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -1,4 +1,4 @@
-use std::{cmp::min, collections::HashSet};
+use std::{cmp::min, collections::HashMap, collections::HashSet};
 
 use actix_web::{
     Either, HttpResponse, Scope, delete, get, post, put, routes,
@@ -26,8 +26,8 @@ use superposition_core::helpers::{calculate_context_weight, hash};
 use superposition_derives::{authorized, declare_resource};
 use superposition_macros::{bad_argument, db_error, unexpected_error};
 use superposition_types::{
-    Contextual, DBConnection, InternalUserContext, ListResponse, Overridden, Overrides,
-    PaginatedResponse, Resource, SortBy, User,
+    Contextual, DBConnection, DimensionInfo, InternalUserContext, ListResponse,
+    Overridden, Overrides, PaginatedResponse, Resource, SortBy, User,
     api::{
         DimensionMatchStrategy,
         context::{
@@ -43,7 +43,7 @@ use superposition_types::{
     },
     database::{
         models::{ChangeReason, Description, cac::Context, others::WebhookEvent},
-        schema::contexts::{self, id},
+        schema::contexts::{self, dsl, id},
     },
     logic::evaluate_local_cohorts_skip_unresolved,
     result::{self as superposition, AppError},
@@ -51,7 +51,10 @@ use superposition_types::{
 
 use crate::{
     api::context::{
-        helpers::{changed_keys, query_description, validate_ctx},
+        helpers::{
+            changed_keys, create_ctx_from_put_req, query_description, validate_ctx,
+            validate_override_with_functions,
+        },
         operations,
     },
     helpers::{add_config_version, put_config_in_redis, validate_change_reason},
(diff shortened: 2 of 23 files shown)


Cargo.toml
```diff
@@ -72,7 +72,8 @@ once_cell = { version = "1.18.0" }
 regex = "1.9.1"
 reqwest = { version = "0.11.18", features = ["json"] }
 rs-snowflake = "0.6.0"
-serde = { version = "^1", features = ["derive"] }
+rustyscript = { version = "0.12.3", default-features = false, features = ["safe_extensions", "web", "worker"] }
+serde = { version = "=1.0.221", features = ["derive"] }
```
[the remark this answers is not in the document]
sauraww: @Datron This seems to be legit.

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -12,166 +14,117 @@ use superposition_types::{
...
-) -> String {
-    let payload = match runtime_version {
-        FunctionRuntimeVersion::V1 => FunctionPayload {
-            version: runtime_version,
-            payload: function_args.clone(),
-        },
-    };
+) -> Result<FunctionExecutionResponse, (String, Option<String>)> {
+    let wrapped_code = generate_wrapped_code(&code.0);
+    log::trace!("Running function code: {:?}", wrapped_code);
```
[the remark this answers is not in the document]
sauraww: @Datron This needs to addressed as well.

crates/context_aware_config/src/api/functions/handlers.rs
```diff
@@ -142,16 +150,22 @@ async fn update_handler(
...
     // Function Linter Check
-    if let Some(function) = &req.draft_code {
-        compile_fn(function)?;
+    if let Some(fc) = &req.draft_code {
+        let handle = rustyscript::tokio::runtime::Handle::current();
+        let function = fc.clone();
+        handle
+            .spawn_blocking(move || compile_fn(&function))
+            .await
+            .map_err(|err| bad_argument!("Invalid function code: {}", err))??;
```
sauraww, line 159: Can we handle the error in the similar way as we did it below.

crates/context_aware_config/src/api/context/helpers.rs
```diff
@@ -282,14 +287,24 @@ pub fn validation_function_executor(
...
         master_encryption_key,
-    ) {
+    )?;
+
+    let args_owned = args.clone();
+    let handle = rustyscript::tokio::runtime::Handle::current();
+    let result = handle
+        .spawn_blocking(move || run_js_function(code, args_owned, runtime_version))
+        .await
+        .map_err(|e| {
```
sauraww, line 302: Here as well , we have inconsistent error handling pattern, Can we unify it ?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -777,51 +848,143 @@ async fn bulk_operations_handler(
...
 
-    let mut all_change_reasons = Vec::new();
     let mut webhook_actions: Vec<Action> = Vec::new();
     let mut webhook_contexts: Vec<Context> = Vec::new();
 
     let tags = parse_config_tags(custom_headers.config_tags)?;
+    // ── Phase 1: async validation & preparation ──
+    let mut prepared_ops =
+        Vec::with_capacity(if ops.len() > 100 { 100 } else { ops.len() });
+
```
sauraww, line 858: Why 100 ? Looks like a magic number - Can we rather create constant?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -240,16 +252,43 @@ async fn update_handler(
...
+    let r_override = req.override_.clone().into_inner();
+
+    validate_override_with_functions(
+        &workspace_context,
+        &mut conn,
+        &r_override,
+        &context,
+        &state.master_encryption_key,
+    )
+    .await?;
```
sauraww, line 282: One slight problem would be there if we move this validation function outside the transaction: 

```
Phase 1: validate_override_with_functions reads function code & runs JS  ← outside txn
              ↕  (another request could ALTER the function or default config here)
Phase 2: operations::update writes the override                          ← inside txn
```

crates/context_aware_config/src/api/context/handlers.rs
Datron: Its more random, just an upper bound on the size of the vec allocated so we don't have runaway memory allocation. I don't think it warrants a const, doesn't really have any meaning

crates/context_aware_config/src/api/context/helpers.rs
Datron: I did not get the inconsistency here, or is the comment resolved?

crates/context_aware_config/src/api/functions/handlers.rs
Datron: Can you elaborate on this?

crates/context_aware_config/src/api/context/handlers.rs
Datron: This is a possibility, a very very small probability of it occurring and also a transaction won't solve it, since when the transaction starts it ignores all updates to the tables in the DB until a COMMIT or SAVEPOINT.

sauraww: Yes , it is a TOCTOU , there is a very small possibility and also the changes to validation function might not be as much and as frequent I am assuming . just wanted to add it as a nit pick.

sauraww (approved):

PR #842, opened 2026-01-28
feat: better logging for API services

Datron:
## Problem
logging for our saas version is not adequate

## Solution
Provider json logs with proper format for indexing in elastic/opensearch/other indexes. some points:

- non JSON responses like HTML pages or wasm are ignored
- request and response info logs are separate
- Response logs are called `GoldenSignal` and and request logs are labelled as `Request`
- `body`, `latency

```
{
  "level": "INFO",
  "timestamp": "2026-01-28T13:29:08.015545Z",
  "body": "<body>",
  "headers": "{\"cache-control\": \"no-store\", \"x-server-version\": \"v0.1.0\", \"content-type\": \"application/json\"}",
  "latency": 23,
  "message": "GoldenSignal",
  "http.client_ip": "127.0.0.1",
  "http.flavor": "1.1",
  "http.host": "localhost:8080",
  "http.method": "GET",
  "http.route": "/types",
  "http.scheme": "http",
  "http.status_code": 200,
  "http.target": "/types?count=10&page=1",
  "http.user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:145.0) Gecko/20100101 Firefox/145.0",
  "method": "GET",
  "name": "HTTP request",
  "org": "orgid162145664241766405",
  "otel.kind": "server",
  "otel.name": "GET /types",
  "otel.status_code": "OK",
  "path": "/types",
  "request_id": "bd81dcd7-63a9-45f1-92a7-3295334c4d00",
  "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:145.0) Gecko/20100101 Firefox/145.0",
  "workspace": "mjos"
}
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced request/response logging with structured JSON format for improved diagnostics.
  * Added latency measurement tracking for request handling.
  * Improved observability with custom tracing spans including workspace and organization context.

* **Chores**
  * Updated core dependencies for improved stability and feature support.
  * Restructured workspace architecture with new modular components.

### DIFF
--- a/crates/service_utils/src/middlewares/request_response_logging.rs
+++ b/crates/service_utils/src/middlewares/request_response_logging.rs
@@ -8,7 +8,6 @@ use actix_web::web::Bytes;
 use actix_web::{Error, error::PayloadError, http::header};
 use futures_util::future::{Ready, ok};
 use futures_util::stream::{StreamExt, once};
-use serde_json::Value;
 use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
@@ -152,42 +151,21 @@ where
                 }
             }
 
-            // Try to parse as JSON and log as structured data
-            let request_body_value = if !body_bytes.is_empty() {
-                serde_json::from_slice::<Value>(&body_bytes).ok()
+            let request_body = if body_bytes.is_empty() {
+                String::from("(empty)")
             } else {
-                None
+                String::from_utf8_lossy(&body_bytes).into_owned()
             };
 
-            if let Some(ref json) = request_body_value {
-                // Successfully parsed as JSON - log as structured
-                trace!(
-                    query = %if query_string.is_empty() {
-                        "(none)".to_string()
-                    } else {
-                        query_string.clone()
-                    },
-                    body = ?json,
-                    "RequestSignal"
-                );
-            } else {
-                // Not JSON - log as string
-                let request_body = if body_bytes.is_empty() {
-                    String::from("(empty)")
+            trace!(
+                query = %if query_string.is_empty() {
+                    "(none)".to_string()
                 } else {
-                    String::from_utf8_lossy(&body_bytes).into_owned()
-                };
-
-                trace!(
-                    query = %if query_string.is_empty() {
-                        "(none)".to_string()
-                    } else {
-                        query_string.clone()
-                    },
-                    body = %request_body,
-                    "RequestSignal"
-                );
-            }
+                    query_string.clone()
+                },
+                body = %request_body,
+                "RequestSignal"
+            );
 
             let new_payload = if body_bytes.is_empty() {
                 Payload::None


crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -85,6 +117,18 @@ where
     }
 }
 
+/// Check if the response has a JSON content type
+fn is_json_response(res: &ServiceResponse<impl MessageBody>) -> bool {
+    res.headers()
+        .get(header::CONTENT_TYPE)
+        .and_then(|ct| ct.to_str().ok())
+        .map(|ct| {
+            ct.starts_with("application/json")
```
knutties: Might want to use the `mime` crate here for the content-type strings list.

ayushjain17: Mime crate would not be needed
Actix has constants for such values
ContentType::APPLICATION_JSON or something similar is there
We are using it already in code

crates/superposition/src/main.rs
```diff
@@ -273,8 +289,8 @@ async fn main() -> Result<()> {
                 Compress::default(),
             ))
             // Conditionally add request/response logging middleware for development
-            .wrap(Condition::new(log_enabled!(Level::Trace), RequestResponseLogger))
-            .wrap(Logger::default())
+            .wrap(RequestResponseLogger)
```
ayushjain17, line 304: we cant log frontend data, local server just dies responding to the requests

Datron: It doesn't log frontend data

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +187,114 @@ where
...
+                let bytes = Bytes::from(body_bytes);
+                let stream = once(async move { Ok::<Bytes, PayloadError>(bytes) });
+                Payload::from(Box::pin(stream)
+                    as Pin<
+                        Box<dyn futures_util::Stream<Item = Result<Bytes, PayloadError>>>,
+                    >)
+            };
+            let new_req = ServiceRequest::from_parts(http_req, new_payload);
+            let start_time = Instant::now();
+            // let _guard = span.enter();
```
mahatoankitkumar: Guess this can be removed

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -1,37 +1,70 @@
...
+use std::time::Instant;
+use tracing::{info, warn};
+
+const SENSITIVE_HEADERS: &[&str] = &[
+    "authorization",
+    "cookie",
+    "x-api-key",
+    "x-auth-token",
+    "set-cookie",
+];
```
mahatoankitkumar: SENSITIVE_HEADERS can be used as enum instead of a string array, this would provide better type safety

Datron: Why do we need type safety? :)

crates/superposition/src/log_span.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn on_request_start(request: &ServiceRequest) -> Span {
+        let headers = request.headers();
+        let extractor_header = |headers: &HeaderMap, key: &str| {
+            headers
+                .get(key)
+                .and_then(|v| HeaderValue::to_str(v).ok())
+                .map(String::from)
+        };
+        let workspace = extractor_header(headers, "x-workspace")
+            .unwrap_or_else(|| "no-workspace-header".to_string());
```
sauraww, line 29: .unwrap_or("no-workspace-header".to_string());

crates/superposition/src/log_span.rs
```diff
@@ -0,0 +1,41 @@
...
+        let org = extractor_header(headers, "x-org-id")
+            .unwrap_or_else(|| "no-org-header".to_string());
```
sauraww, line 31: .unwrap_or("no-org-header".to_string());

crates/superposition/src/log_span.rs
```diff
@@ -0,0 +1,41 @@
...
+        let user_agent = extractor_header(headers, "user-agent")
+            .unwrap_or_else(|| "no-user-agent".to_string());
```
sauraww: .unwrap_or("no-user-agent".to_string());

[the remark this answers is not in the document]
sauraww: At the least , We should sanitize control characters for security.

crates/service_utils/src/middlewares/request_response_logging.rs
mahatoankitkumar: 🤨

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +187,113 @@ where
...
-                trace!(
-                    "REQUEST: {} {} query={} headers=[{}] body={}",
-                    method,
-                    uri,
-                    if query_string.is_empty() {
-                        "(none)"
+            if let Some(ref json) = request_body_value {
+                // Successfully parsed as JSON - log as structured
+                info!(
+                    query = %if query_string.is_empty() {
```
knutties: We can DRY the info log line by just creating two versions of the body to avoid some repetition.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -1,37 +1,70 @@
...
+    let name_lower = name.to_lowercase();
+    let is_sensitive = SENSITIVE_HEADERS
+        .iter()
+        .any(|sensitive| name_lower == *sensitive);
+
+    if is_sensitive {
+        (name.to_string(), "[REDACTED]".to_string())
+    } else {
+        (name.to_string(), value)
+    }
```
knutties:
```suggestion
    let value = if SENSITIVE_HEADERS.iter().any(|sensitive| name_lower == *sensitive) {
        "[REDACTED]".to_string()
    } else {
        value
    };
    (name.to_string(), value)
```

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,21 +95,42 @@ where
...
-                        "RESPONSE: {} headers=[{}] body={}",
-                        this.status,
-                        this.headers.join(", "),
-                        response_body
-                    );
+                        // Non-JSON response, skip body entirely
+                        info!(
+                            headers = ?this.headers,
+                            latency = latency_ms,
+                            "GoldenSignal - No Body"
```
knutties: This is also GoldenSignal.  The `No Body` can be in the message like the previous case.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,21 +95,42 @@ where
...
+                    if this.log_body_as_json {
+                        // Parse and log as structured JSON
+                        match serde_json::from_slice::<Value>(&this.body_bytes) {
+                            Ok(json_value) => {
+                                info!(
+                                    headers = ?this.headers,
+                                    body = %json_value,
+                                    latency = latency_ms,
+                                    "GoldenSignal"
+                                );
```
knutties: This could be moved out after the match to avoid repetition.  One final log is possibly good enough.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +187,113 @@ where
...
 
-                trace!(
-                    "REQUEST: {} {} query={} headers=[{}] body={}",
-                    method,
-                    uri,
-                    if query_string.is_empty() {
-                        "(none)"
+            if let Some(ref json) = request_body_value {
+                // Successfully parsed as JSON - log as structured
+                info!(
```
knutties: Let us use trace for the `Request` line log.  Ideally we just need the GoldenSignal in the most common case.  We can enable trace if we have issues and we need to debug.  This will maintain log hygiene as well.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +187,113 @@ where
...
+                .filter_map(|(name, value)| {
+                    value
+                        .to_str()
+                        .ok()
+                        .map(|v| redact_header(name.as_str(), v.to_string()))
+                })
+                .collect();
+
+            // Only log body as JSON for JSON content types
+            let log_body_as_json = is_json_response(&res);
```
knutties: This check can be completely moved into the `LoggingBody` implementation itself as the headers are passed into and we can possibly do away with the `log_body_as_json` member.  Will keep all logging concerns in the `LoggingBody` implementation itself.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -1,37 +1,70 @@
...
+use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
 use std::rc::Rc;
 use std::task::{Context, Poll};
+use std::time::Instant;
+use tracing::{info, warn};
+
+const SENSITIVE_HEADERS: &[&str] = &[
+    "authorization",
```
knutties: Let us do a whitelist here rather than a blacklist - so that inadvertently added headers that might security don't leak.  If we need to expand - we can make a code change - not a big deal in general - but definitely secure.

crates/superposition/src/log_span.rs
```diff
@@ -0,0 +1,41 @@
...
+};
+use tracing::Span;
+use tracing_actix_web::{DefaultRootSpanBuilder, RootSpanBuilder};
+
+pub struct CustomRootSpanBuilder;
+
+impl RootSpanBuilder for CustomRootSpanBuilder {
+    fn on_request_start(request: &ServiceRequest) -> Span {
+        let headers = request.headers();
+        let extractor_header = |headers: &HeaderMap, key: &str| {
```
knutties: nit: `header_extractor` is a little more readable I feel.  subject-verb vs verb-subject.

crates/service_utils/src/middlewares/request_response_logging.rs
Datron: @mahatoankitkumar what would type safety achieve here? Why do I need it for this segment of the code? Finally the header key is a string, and I need to check if it exists in this array. Should there be anything more? Even if I write an enum, I will have to stringify it. 

That is why I asked, why do we need type safety here?

crates/superposition/src/log_span.rs
Datron: I want lazy evaluation of the string

crates/superposition/src/log_span.rs
Datron: I want lazy evaluation of the string

crates/superposition/src/log_span.rs
Datron: I want lazy evaluation of the string

crates/service_utils/src/middlewares/request_response_logging.rs
Datron: @knutties I want to request to be logged, in case the worker thread is killed in between I want to be sure the request was received by the server

crates/service_utils/src/middlewares/request_response_logging.rs
Datron: @knutties I'm wondering, if workspace and org are part of the log line, what other headers would we need to care about? Should we just not have headers as part of the log line GoldenSignal?

crates/service_utils/src/middlewares/request_response_logging.rs
Datron: This way is better for the debugger. You can see the value of `is_sensitive`

Cargo.toml
```diff
@@ -74,8 +74,12 @@ secrecy = "0.10"
 strum = "0.25"
 strum_macros = "0.25"
 toml = { version = "0.8.8", features = ["preserve_order"] }
+actix-http = "3.3.1"
+tracing = "0.1.44"
+tracing-subscriber = { version = "0.3.22", features = ["env-filter", "fmt", "json"] }
+tracing-actix-web = "0.7.21"
```
ayushjain17, line 80: can we please add them in alphabetical order

Cargo.toml
```diff
@@ -73,8 +73,12 @@ serde_json = { version = "1.0.140" }
...
```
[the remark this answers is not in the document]
ayushjain17: @Datron can we verify this

crates/superposition/Cargo.toml
```diff
@@ -37,6 +36,11 @@ superposition_types = { workspace = true, features = [
     "diesel_derives",
     "server",
 ] }
+actix-http = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+tracing-actix-web = { workspace = true }
+json-subscriber = { version = "0.2.7", features = ["tracing-log"] }
```
ayushjain17, line 43: can we sort these

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -1,37 +1,63 @@
...
-use log::{trace, warn};
+use serde_json::Value;
+use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
 use std::rc::Rc;
 use std::task::{Context, Poll};
+use std::time::Instant;
+use tracing::{info, warn};
+
```
ayushjain17, line 18: can we option to blacklist request and response for certain paths
we do not want to log the request body for `secret's create` under any circumstance as that includes the un-encrypted secret value in it

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -1,37 +1,63 @@
...
+}
 
 #[derive(Default)]
 pub struct RequestResponseLogger;
 
-// Custom body wrapper for logging response bodies
+// Custom body wrapper for logging request/response bodies
 pub struct LoggingBody<B> {
     inner: B,
-    status: StatusCode,
```
ayushjain17, line 21: why no status code ?

crates/superposition/src/main.rs
```diff
@@ -273,8 +289,8 @@ async fn main() -> Result<()> {
...
+            .wrap(TracingLogger::<CustomRootSpanBuilder>::new())
     })
```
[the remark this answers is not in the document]
ayushjain17: @Datron can we address this

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -1,37 +1,63 @@
...
+    "x-auth-token",
+    "set-cookie",
+];
+
+fn redact_header(name: &str, value: String) -> (String, String) {
+    let name_lower = name.to_lowercase();
+    let is_sensitive = SENSITIVE_HEADERS
+        .iter()
+        .any(|sensitive| name_lower == *sensitive);
+
```
ayushjain17: as far as I am aware, name is already processed as lowercase by default by actix, we dont need to re-convert it
```suggestion
    let is_sensitive = SENSITIVE_HEADERS.contains(&name);

```

crates/service_utils/src/middlewares/request_response_logging.rs
ayushjain17: @Datron helps reduce cases of typos though

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,20 +88,38 @@ where
...
                     this.consumed = true;
-                    // Log the complete response body
-                    let response_body = if this.body_bytes.is_empty() {
-                        String::from("(empty)")
+                    let latency_ms = this.start_time.elapsed().as_millis() as u64;
+                    let is_json_response = this
+                        .headers
+                        .get(header::CONTENT_TYPE.as_str())
+                        .map(|ct| ct.starts_with(&ContentType::json().to_string()))
+                        .unwrap_or(false);
```
ayushjain17, line 74:
```suggestion
                        .unwrap_or_default();
```

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,20 +88,38 @@ where
...
+                    let response_body = if is_json_response {
```
ayushjain17, line 75: what's the harm in logging the value simply as a string ?
instead of converting the bytes to json, what if we convert it string simply containing the characters `{`, `}` and other stuff
in the log line it is any ways to sit to as characters only with no special difference, right ?

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,20 +88,38 @@ where
...
+                        serde_json::from_slice::<Value>(&this.body_bytes).unwrap_or_else(
```
ayushjain17: if we still need to parse it as value only, then wont it be better to do this then
```suggestion
                        serde_json::from_slice::<Map<String, Value>>(&this.body_bytes)
                            .map(Value::Object)
                            .unwrap_or_else(
```

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,20 +88,38 @@ where
...
                     };
-
-                    trace!(
-                        "RESPONSE: {} headers=[{}] body={}",
-                        this.status,
-                        this.headers.join(", "),
-                        response_body
+                    info!(
+                        headers = ?this.headers,
+                        body = %response_body,
```
ayushjain17, line 81: we are converting the bytes to json, and then using its display trait, converting it back to string here using `%`
we could have simply converted the bytest to string - irrespective of the parsing response - I dont find any value in keeping it as `Value` type, this way, just seems to be additional processing steps only

crates/service_utils/src/middlewares/request_response_logging.rs
knutties: >  Should we just not have headers as part of the log line GoldenSignal?

Yeah - that is fine.  In some edge cases we might need it - but we can pass for now and skip headers for now.

crates/service_utils/src/middlewares/request_response_logging.rs
knutties: @Datron - could we handle the worker kills in some way or have actix log the event that worker got killed ?  Basically I would like to focus on us controlling things more than relying on catch all stuff.

A good first cut would be just clean GoldenSignal logs for status of every unit of work that the server does.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +168,109 @@ where
...
-                let request_body = if body_bytes.is_empty() {
-                    String::from("(empty)")
-                } else {
-                    String::from_utf8(body_bytes.clone()).unwrap_or_else(|_| {
-                        format!("(binary data, {} bytes)", body_bytes.len())
-                    })
-                };
+            // Try to parse as JSON and log as structured data
+            let request_body_value = if !body_bytes.is_empty() {
+                serde_json::from_slice::<Value>(&body_bytes).ok()
```
ayushjain17: same here for converting to value

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +168,109 @@ where
...
-                        "(none)"
+            if let Some(ref json) = request_body_value {
+                // Successfully parsed as JSON - log as structured
+                info!(
+                    query = %if query_string.is_empty() {
+                        "(none)".to_string()
                     } else {
-                        &query_string
+                        query_string.clone()
                     },
```
ayushjain17: query would be present in the path itself, right ?

crates/service_utils/src/middlewares/request_response_logging.rs
knutties: fair point.  Felt that the conciseness was idiomatic though.  Not a big deal.  If we drop headers - then this is moot as well.

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -124,107 +168,109 @@ where
...
+                        Box<dyn futures_util::Stream<Item = Result<Bytes, PayloadError>>>,
+                    >)
+            };
+            let new_req = ServiceRequest::from_parts(http_req, new_payload);
+            let start_time = Instant::now();
+            res = service.call(new_req).await?;
+
+            if let Some(Ok(request_id)) = request_id {
+                res.headers_mut()
+                    .insert(header::HeaderName::from_static("x-request-id"), request_id);
```
ayushjain17, line 186: wondering how this spans to other logs as well which are not request/response logs

Datron: https://docs.rs/tracing/latest/tracing/span/index.html

crates/service_utils/src/middlewares/request_response_logging.rs
knutties: @ayushjain17 - you cannot copy paste and see the nesting anywhere if the structure is complex.  so json structure is useful I feel.

ayushjain17: @knutties the content is converted to strings itself, which still is in json shape
nothing equivalent to JSON.stringify is happening here

crates/service_utils/src/middlewares/request_response_logging.rs
```diff
@@ -62,21 +66,22 @@ where
...
-                        })
+                        "(non-JSON response body omitted)".to_string()
                     };
-
                     trace!(
-                        "RESPONSE: {} headers=[{}] body={}",
-                        this.status,
-                        this.headers.join(", "),
-                        response_body
+                        body = %response_body,
```
ayushjain17, line 81:
```suggestion
                        body = response_body,
```

crates/superposition/src/main.rs
```diff
@@ -285,8 +301,8 @@ async fn main() -> Result<()> {
                 Compress::default(),
             ))
             // Conditionally add request/response logging middleware for development
```
knutties, line 303: nit: we can remove this comment now.
```suggestion
```

PR #848, opened 2026-01-29
fix: redundant cargo upgrades

knutties:
## Problem

https://github.com/juspay/superposition/blob/8725d5985b1276091979b23c5e000ab10bc86f94/.github/workflows/release.yaml#L61-L67
https://github.com/juspay/superposition/blob/8725d5985b1276091979b23c5e000ab10bc86f94/.github/workflows/release.yaml#L611-L618

Cargo commands are repeated here - the latter should just publish right - not really upgrade which is completed in the tag-releases step ?  The first one has the commit post the cargo upgrades.

## Solution
Re-use the earlier commit by checking out at that version.

## Summary by CodeRabbit

* **Chores**
  * Improved release automation with enhanced handling of version tags across the complete build pipeline
  * Refined build process by removing automatic version bump operations, providing better control over version management
  * Updated package documentation description to reflect more accurate and reliable characterization

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,6 +15,7 @@ jobs:
         if: github.ref == 'refs/heads/main'
         outputs:
             version: ${{ steps.git_tag.outputs.version }}
+            updated_ref: ${{ steps.push_main.outputs.updated_ref }}
         steps:
             - name: Checkout repository
               uses: actions/checkout@v4
@@ -83,6 +84,7 @@ jobs:
                   git commit --amend --no-edit
 
             - name: Push code to main
+              id: push_main
               shell: bash
               run: |
                   git pull --rebase origin main


ayushjain17 (approved):

.github/workflows/release.yaml
```diff
@@ -88,6 +88,7 @@ jobs:
                   git pull --rebase origin main
                   git push origin main
                   git push origin --tags
+                  echo "updated_ref=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
```
ayushjain17, line 93: we will need to add this in outputs section after line 17 in this file
```
updated_ref: ${{ steps.git_tag.outputs.updated_ref }}
```

.github/workflows/release.yaml
```diff
@@ -15,6 +15,7 @@ jobs:
         if: github.ref == 'refs/heads/main'
         outputs:
             version: ${{ steps.git_tag.outputs.version }}
+            updated_ref: ${{ steps.git_tag.outputs.updated_ref }}
```
Datron:
```suggestion
            updated_ref: ${{ steps.push_main.outputs.updated_ref }}
```
You have to update the step ID

ayushjain17 (approved):

PR #859, opened 2026-02-03
Feat/change rs ws

sauraww:
## Problem
Change reason was missing in the workspace table 

## Solution
Added change reason in the workspace table

#847


Datron:
Outdated and needs to be rebased

PR #928, opened 2026-03-24
feat: improve java Provider performance

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

### DIFF
--- a/clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
+++ b/clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
@@ -5,8 +5,10 @@ import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.ref.WeakReference;
 import java.util.Optional;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 interface RefreshJob<T> {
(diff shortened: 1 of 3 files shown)


clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionOpenFeatureProvider.java
```diff
@@ -314,4 +341,38 @@ private ExperimentationArgs getExperimentationArgs(EvaluationContext ctx) {
         }
         return null;
     }
+
+    private void reinitConfigCache() {
```
Datron, line 345: why reinit? Similar question for experiments. Should we just call it refresh? What is the name used in other providers

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionOpenFeatureProvider.java
```diff
@@ -282,9 +299,19 @@ private EvaluationArgs getEvaluationArgs(EvaluationContext ctx) throws Exception
...
     private Map<String, String> evaluateConfigInternal(EvaluationContext ctx) throws Exception {
-        EvaluationArgs args = getEvaluationArgs(ctx);
+        // Block until cache.initConfig has been called (completed inside reinitConfigCache).
+        // This guarantees cache.evalConfig never runs on an uninitialized cache.
+        if (!cacheReady.isDone()) {
+            try {
+                cacheReady.get(configTimeout, TimeUnit.MILLISECONDS);
+            } catch (TimeoutException e) {
+                throw new Exception("Config cache not initialized within timeout (" + configTimeout + "ms).");
+            }
```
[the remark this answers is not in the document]
Datron: Ankit can you check this?

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
```diff
@@ -19,46 +21,79 @@ interface RefreshJob<T> {
...
+                    if (self == null) {
+                        log.debug("Poll referent GC'd — self-cancelling polling task.");
+                        ScheduledFuture<?> t = taskRef.get();
+                        if (t != null) t.cancel(false);
+                        return;
+                    }
+                    var o = RefreshJob.runRefreshWithTimeout(self.action, self.config.timeout);
                     if (o != null) {
-                        output.complete(o);
+                        boolean changed = !o.equals(self.latestOutput);
```
Datron, line 52: Generally the equals method provided by Object.class just does a referencial check (== or just checks if the pointer is the same), not a value check unless you define a custom implementation. Given T is generic, we should use some other way of checking if data has changed here

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
```diff
@@ -19,46 +21,79 @@ interface RefreshJob<T> {
     final class Poll<T> implements RefreshJob<T> {
         private final RefreshStrategy.Polling config;
         private final Supplier<CompletableFuture<T>> action;
-        private final CompletableFuture<T> output;
+        private final Runnable onChange;
+        private final CompletableFuture<T> firstOutput;
+        private volatile T latestOutput = null;
```
Datron, line 26: Why do we need both of these?

Datron (commented):
Can you check if there is a better way to check equality between output and latestOutput?

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionOpenFeatureProvider.java
mahatoankitkumar: this reinitializes the config cache in rust
refresh are there for fetching it from the server for configs and exps
will make it confusing to have refresh here, and yes its reinit everywhere else.

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
mahatoankitkumar: So latestOutput is the steady-state store and firstOutput is purely the one-time synchronization primitive for that cold-start window. We could replace both with a single AtomicReference<CompletableFuture<T>> (replacing it on each poll with an already-completed future), but that would allocate a new CompletableFuture on every poll cycle. The current change is more efficient.

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
mahatoankitkumar: it's always a Map<String, Object>), so this works here.

Datron: Can we scope it in such a way then if possible?

Datron (approved):

clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/RefreshJob.java
mahatoankitkumar: pushed the change with custom equals()

sauraww (approved):

PR #952, opened 2026-04-01
feat: haskell provider perf improvement

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Release Notes

* **New Features**
  * Added provider cache lifecycle management with explicit shutdown support, enabling better resource control and cleanup.
  * Introduced native caching layer with automatic reinitialization for configuration and experiment data.
  * Enhanced polling mechanism to use garbage-collection-based lifecycle tracking for improved reliability.

* **Bug Fixes**
  * Fixed callback handling in experimentation polling to properly detect instance liveness and prevent stale invocations.

### DIFF
--- a/clients/javascript/open-feature-provider/experimentation-client.ts
+++ b/clients/javascript/open-feature-provider/experimentation-client.ts
@@ -52,18 +52,12 @@ export class ExperimentationClient {
     private lastUpdated: Date | null = null;
     private evaluationCache: Map<string, any> = new Map();
     private pollingInterval?: ReturnType<typeof setTimeout>;
-    private onExperimentsChange: (
-        experiments: Experiment[],
-        experimentGroups: ExperimentGroup[],
-    ) => void;
+    private onExperimentsChange?: () => void;
 
     constructor(
         private superpositionOptions: SuperpositionOptions,
         experimentOptions: ExperimentationOptions,
-        onExperimentsChange: (
-            experiments: Experiment[],
-            experimentGroups: ExperimentGroup[],
-        ) => void,
+        onExperimentsChange?: () => void
     ) {
         this.options = experimentOptions;
         this.onExperimentsChange = onExperimentsChange;
@@ -81,7 +75,7 @@ export class ExperimentationClient {
         if (experiments && experimentgroups) {
             this.cachedExperiments = experiments;
             this.cachedExperimentGroups = experimentgroups;
-            this.onExperimentsChange(experiments, experimentgroups);
+            this.onExperimentsChange?.();
             this.lastUpdated = new Date();
             console.log(
                 "Experiments and Experiment Groups fetched successfully.",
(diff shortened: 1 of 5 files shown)


crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -2,13 +2,258 @@
...
+    TomlFormat,
+};
+
+pub struct ProviderCache {
+    data: Mutex<CacheData>,
+}
+
+impl Drop for ProviderCache {
+    fn drop(&mut self) {
+        eprintln!("[Rust] ProviderCache dropped — native memory freed");
```
Datron: can we use log, trace or println rather than eprintln which prints to stderr

crates/superposition_core/src/ffi.rs
```diff
@@ -220,15 +220,21 @@ fn ffi_parse_json_config(json_content: String) -> Result<Config, OperationError>
...
 }
 
 #[derive(uniffi::Object)]
 pub struct ProviderCache {
     data: Mutex<CacheData>,
 }
 
+impl Drop for ProviderCache {
+    fn drop(&mut self) {
+        eprintln!("[Rust] ProviderCache dropped — native memory freed");
```
Datron: lets use log, trace or println! here instead of eprintln which prints to stderr

clients/javascript/open-feature-provider/experimentation-client.ts
```diff
@@ -175,7 +186,7 @@ export class ExperimentationClient {
                     id: exp.id,
                     context: this.normalizeToStringRecord(exp.context),
                     variants: variants,
-                    traffic_percentage: exp.traffic_percentage || 100,
+                    traffic_percentage: exp.traffic_percentage ?? 100,
```
Datron, line 197: If there is no traffic percentage better to set it to 0 because some error is happening

clients/javascript/open-feature-provider/experimentation-client.ts
```diff
@@ -218,7 +229,7 @@ export class ExperimentationClient {
                 experimentGroups.push({
                     id: exp_group.id,
                     context: this.normalizeToStringRecord(exp_group.context),
-                    traffic_percentage: exp_group.traffic_percentage || 100,
+                    traffic_percentage: exp_group.traffic_percentage ?? 100,
```
Datron, line 240: If there is no traffic percentage better to set it to 0 because some error is happening

crates/superposition_core/src/ffi_legacy.rs
```diff
@@ -2,13 +2,258 @@
...
+    }
+}
+
+/// # Safety
+///
+/// `handle` must be a valid pointer from `core_provider_cache_new`. All non-null pointer
+/// arguments must be valid null-terminated C strings. `ebuf` must be a sufficiently
+/// large buffer for error messages. The returned pointer must be freed with `core_free_string`.
+#[no_mangle]
+pub unsafe extern "C" fn core_provider_cache_eval_config(
```
Datron, line 143: Just confirm if this function is used instead of eval_config

Datron (changes requested):

Datron (approved):

sauraww (approved):

PR #970, opened 2026-04-14
feat: add breadcrumbs and superposition platform links

Datron:
## Problem
Navigation backwards is hard in superposition if deeply nested

## Solution
Add breadcrumbs so users can see where they are and jump to any previous screen they had visited along the way to the current page

https://github.com/user-attachments/assets/1d46ad90-933a-405a-aa34-37c4b017d35a

## Summary by CodeRabbit

* **New Features**
  * Introduced breadcrumb navigation that dynamically displays your current location within the app and provides convenient shortcuts to navigate to parent pages and sections, making it easier to explore different areas of the platform.

* **Improvements**
  * Enhanced routing and redirect patterns throughout the application to provide a smoother and more intuitive navigation experience when moving between different sections.

### DIFF
--- a/crates/frontend/src/components/breadcrumbs.rs
+++ b/crates/frontend/src/components/breadcrumbs.rs
@@ -1,0 +1,171 @@
+use leptos::*;
+use leptos_router::{A, use_location};
+
+use crate::types::{BreadcrumbSegment, OrganisationId, Workspace};
+use crate::utils::use_url_base;
+
+const SKIP_SEGMENTS: [&str; 4] = ["admin", "action", "authz", "org-authz"];
+
+/// Maps a URL path segment to a human-readable label.
+/// Returns None for segments that should be skipped (like "action").
+fn segment_to_label(segment: &str) -> Option<String> {
+    let title_segments = [
+        "types",
+        "audit-log",
+        "default-config",
+        "experiment-groups",
+        "compare",
+        "config",
+        "dimensions",
+        "experiments",
+        "function",
+        "resolve",
+        "secrets",
+        "variables",
+        "versions",
+        "webhooks",
+        "workspaces",
+        "overrides",
+        "create",
+        "edit",
+    ];
+    match segment {
+        x if !title_segments.contains(&x) => Some(segment.to_string()),
+        x => Some(
+            x.split('-')
+                .map(|s| {
+                    let mut chars = s.chars();
+                    match chars.next() {
+                        Some(first) => {
+                            first.to_uppercase().collect::<String>() + chars.as_str()
+                        }
+                        None => String::new(),
+                    }
+                })
+                .collect::<Vec<String>>()
+                .join(" "),
+        ),
+    }
+}
+
+/// Builds the breadcrumb trail from the current URL path.
+fn build_breadcrumbs(path: &str, base: &str) -> Vec<BreadcrumbSegment> {
+    // Strip the base prefix if present
+    let path = path.replace(base, "");
+    let segments: Vec<&str> = path.split('/').filter(|s| !s.is_empty()).collect();
+
+    let mut previous_segments = String::new();
+    let mut breadcrumbs = Vec::new();
+    for (i, segment) in segments.iter().enumerate() {
+        let is_current = i == segments.len() - 1;
+        if !SKIP_SEGMENTS.contains(segment) {
+            if let Some(label) = segment_to_label(segment) {
+                breadcrumbs.push(BreadcrumbSegment {
+                    label,
+                    href: if is_current {
+                        String::new()
+                    } else {
+                        let current = if previous_segments == "/admin" {
+                            "organisations"
+                        } else if previous_segments
+                            .split("/")
+                            .filter(|s| !s.is_empty())
+                            .collect::<Vec<&str>>()
+                            .len()
+                            == 2
+                        {
+                            "workspaces"
+                        } else {
+                            segment
+                        };
+                        format!("{}{}/{}", base, previous_segments, current)
+                    },
+                    is_current,
+                });
+            }
+        }
+        previous_segments.push('/');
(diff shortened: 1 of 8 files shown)


crates/frontend/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,171 @@
...
+        "function",
+        "resolve",
+        "secrets",
+        "variables",
+        "versions",
+        "webhooks",
+        "workspaces",
+        "overrides",
+        "create",
+        "edit",
```
ayushjain17, line 30: do we need this list separately ?

Datron: @ayushjain17 looked into unifying this array string and Resource enum, there are a couple of differences:
- Resource uses snake_case but all our url segments are kebab-case. I can't change resource I guess because of casbin DB
- Resource doesn't have some of the title segments I am looking for like Create and Edit, resolve
- Resource is mostly singular, but all our title segments are plural, eg: `Dimension` in Resource and dimensions in the URL

Writing code to sync these two would be more work and still fragile. I am thinking of changing app.rs to instead use a different Enum for the path names like `default-config`, `dimensions`, etc which can then convert to `Resource` type if needed. What do you think?

mahatoankitkumar: yes @Datron sounds better, enum RouteSegment or something similar

Datron: @mahatoankitkumar done

crates/frontend/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,183 @@
...
+const SKIP_SEGMENTS: [RouteSegment; 4] = [
+    RouteSegment::Admin,
+    RouteSegment::Action,
+    RouteSegment::Authz,
+    RouteSegment::OrgAuthz,
+];
+
+/// Maps a URL path segment to a human-readable label.
+/// Returns None for segments that should be skipped (like "action").
+fn segment_to_label(segment: &str) -> Option<String> {
```
sauraww: What is the None case here ? 
If this is optional , I am supposed to hit the None case too but this looks like returning Some always.

sauraww (changes requested):

crates/frontend/src/components/breadcrumbs.rs
Datron: Yeah this is weirdly written. Simplified it and fixed

sauraww (approved):

mahatoankitkumar (changes requested):
There's a small issue here, not sure how should we solve it
Like clicking on workspace name, takes me to the list of workspaces in that org
which ideally should be going to that workspace
similarly for org, clicking on org takes us to the list of orgs, not inside that org having the list of workspaces.

Its going a level above, not intuitive for a user
I guess we are doing this to improve user's experience

Datron:
@mahatoankitkumar 
>There's a small issue here, not sure how should we solve it
Like clicking on workspace name, takes me to the list of workspaces in that org
which ideally should be going to that workspace
similarly for org, clicking on org takes us to the list of orgs, not inside that org having the list of workspaces. Its going a level above, not intuitive for a user
I guess we are doing this to improve user's experience

Basically you go upto that level, that is the intuition

Datron:
Screenshot 2026-05-05 at 1 22 16 PM
<br>
Screenshot 2026-05-05 at 1 33 00 PM
<br>
Screenshot 2026-05-05 at 1 33 08 PM
<br>
Screenshot 2026-05-05 at 1 33 14 PM
<br>
Screenshot 2026-05-05 at 1 33 26 PM
<br>
Screenshot 2026-05-05 at 1 35 39 PM

Updated

mahatoankitkumar:
https://github.com/user-attachments/assets/24cb115b-1e4f-4520-846e-0d9a47323495

crates/frontend/src/types.rs
```diff
@@ -284,3 +298,119 @@ impl DropdownOption for DimensionTypeOptions {
...
+    Eq,
+    Hash,
+    strum_macros::Display,
+    strum_macros::EnumString,
+    strum_macros::EnumIter,
+)]
+pub enum RouteSegment {
+    #[strum(serialize = "types")]
+    Types,
+    #[strum(serialize = "audit-log")]
```
ayushjain17, line 316: We dont need to specify this individually 
We can put "kebab-case" at the top enum level

crates/frontend/src/types.rs
```diff
@@ -284,3 +298,119 @@ impl DropdownOption for DimensionTypeOptions {
...
+        if let Ok(seg) = RouteSegment::from_str(value) {
+            Self::Static(seg)
+        } else {
+            Self::Dynamic(value.trim_start_matches(':'))
+        }
+    }
+}
+
+/// Join static and dynamic route parts with `/`.
+pub fn join_route_parts<I, T>(parts: I) -> String
```
ayushjain17, line 406: Could this have been some trait implementation?

crates/frontend/src/components/side_nav.rs
```diff
@@ -341,7 +341,7 @@ pub fn NavComponent(
             >
                 <div class="h-[84px] px-4 py-2 flex items-center justify-center gap-8">
                     <A
-                        href=format!("{base}/admin")
+                        href=format!("{base}/admin/organisations")
```
ayushjain17, line 344: Should we rather make this conditional ?
If we are inside a workspace this would take me to list workspaces and if I am on list workspace it takes me to list orgs ?
Essentially like a back button ?

If this behaviour is slightly confusing we can consider making this non clickable and add a back button which does this itself (mentioned above) - the entire thing along with back button can also be clickable

mahatoankitkumar (approved):

sauraww (approved):

PR #982, opened 2026-04-24
feat: experiment idempotency

mahatoankitkumar:
This pull request introduces idempotency support for experiment creation in the experimentation platform. The main change is the addition of an optional `idempotency_key` field to the `experiments` table and related application logic, which ensures that duplicate experiment creation requests with the same key are handled safely and return the existing experiment instead of creating a new one. There are also supporting changes to database schema, HTTP request handling, and transaction code cleanup.

**Experiment Creation Idempotency:**

* Added an optional `idempotency_key` field to the `Experiment` model and database schema, including migrations and index creation to enforce uniqueness. [[1]](diffhunk://#diff-ba968ba8d543249dbaf43b7a7075a406abf90d66b3a29bbc85eb7fd69e1cf22aR1-R3) [[2]](diffhunk://#diff-85ae756a60c47ceadc492838efa29e1ba9b800dd5f2b136ece67541bd9bcf0a6R1-R3) [[3]](diffhunk://#diff-da9cddb6268a36f1adca603b1d04ef76a99b1c3a7027002df9fee19564f00eb3R660) [[4]](diffhunk://#diff-7072d3ac2d83ae48cabf707d1e59ac7f9981327233bca38e546ec2a53f10c132R315) [[5]](diffhunk://#diff-aacd0d65d574dc633f12aac913ed70d12c6ffdd02ca943db39a47e0420fe6590R1774-R1778) [[6]](diffhunk://#diff-74ba43f0cf5afc5c7834dec4101a60b6bad14f9814d4501efea46bc0c79e17c9R1000-R1002)
* Updated the experiment creation handler to check for an existing experiment with the same `idempotency_key` and return it if found, preventing duplicate experiments. [[1]](diffhunk://#diff-9b0eb9dd1fded65609adb97a7800ef3d57cd2bb7d4973a87f36ae5a0f0d4cbedL159-R177) [[2]](diffhunk://#diff-9b0eb9dd1fded65609adb97a7800ef3d57cd2bb7d4973a87f36ae5a0f0d4cbedR387-R392)

**HTTP Request and Header Handling:**

* Extended the `CustomHeaders` struct and its extraction logic to support reading the `Idempotency-Key` from incoming HTTP requests, making the key available to handlers. [[1]](diffhunk://#diff-2b473062e42a049440f3385238e0068c04ec7b323ef86431620b6ab6e6663045R251) [[2]](diffhunk://#diff-2b473062e42a049440f3385238e0068c04ec7b323ef86431620b6ab6e6663045R266-R268)

**Code Quality and Consistency:**

* Minor cleanups to transaction code in context-aware config handlers for consistency and readability. [[1]](diffhunk://#diff-d134f3fb88f77342e5f42aaac3b57af7631de3f0926a1c0cd0e6b702d0e97c36L284-R285) [[2]](diffhunk://#diff-d134f3fb88f77342e5f42aaac3b57af7631de3f0926a1c0cd0e6b702d0e97c36L426-R427)
* Added missing import for `OptionalExtension` to support optional query results.

## Summary by CodeRabbit

* **New Features**
  * Added idempotency support for experiment creation: requests with an idempotency key will return the existing experiment if one was already created, enabling safe retry behavior without duplicates.


knutties:
@mahatoankitkumar - we need this change in Smithy as well ?

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -156,12 +156,25 @@ async fn create_handler(
...
+        let existing: Option<Experiment> = dsl::experiments
+            .filter(dsl::idempotency_key.eq(key))
+            .select(Experiment::as_select())
+            .schema_name(&workspace_context.schema_name)
+            .first(&mut conn)
+            .optional()?;
+        if let Some(exp) = existing {
+            return Ok(HttpResponse::Ok().json(ExperimentResponse::from(exp)));
+        }
+    }
```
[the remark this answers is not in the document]
sauraww: This will be resolved post the introduction of advisory lock - writes will be serialized. There won't be any race conditions.

crates/service_utils/src/service/types.rs
```diff
@@ -262,6 +263,9 @@ impl FromRequest for CustomHeaders {
             config_tags: header_val.get("x-config-tags").and_then(|header_val| {
                 header_val.to_str().map_or(None, |v| Some(v.to_string()))
             }),
+            idempotency_key: header_val
+                .get("Idempotency-Key")
```
knutties: adding to Co-pilots suggestion here.  We should match against lower-case of the headers and ensure it is lower-cased by the time they reach the application.  Does that happen automatically ?  Could you check @mahatoankitkumar ?

mahatoankitkumar: Yes its happens automatically, but i'll change it just for consistency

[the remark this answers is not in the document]
mahatoankitkumar: http crate's HeaderMap which normalizes all header names to lowercase internally

workspace_template.sql
```diff
@@ -997,3 +997,6 @@ DO $$ BEGIN
 EXCEPTION
     WHEN duplicate_object THEN null;
 END $$;
+
+ALTER TABLE {replaceme}.experiments ADD COLUMN IF NOT EXISTS idempotency_key TEXT;
+CREATE UNIQUE INDEX IF NOT EXISTS experiments_idempotency_key_idx ON {replaceme}.experiments(idempotency_key);
```
Datron: Lets use partial index here since idempotency_key is nullable. From google:
```
If you frequently query for non-null values and your table has many NULLs, a partial index can significantly reduce index size and maintenance overhead by excluding the NULL entries. 
sql

-- Index only the actual data, ignoring NULLs
CREATE INDEX idx_not_null_only ON your_table (column_name) 
WHERE column_name IS NOT NULL;

Conversely, if you only care about finding the NULL values (e.g., finding "incomplete" records), you can index only those: 
sql

CREATE INDEX idx_nulls_only ON your_table (column_name) 
WHERE column_name IS NULL;

```

crates/superposition_types/migrations/2026-04-24-000000_experiment_idempotency_key/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+ALTER TABLE experiments ADD COLUMN IF NOT EXISTS idempotency_key TEXT;
+CREATE UNIQUE INDEX IF NOT EXISTS experiments_idempotency_key_idx ON experiments(idempotency_key);
```
Datron: Partial index would be better suited here:
```
If you frequently query for non-null values and your table has many NULLs, a partial index can significantly reduce index size and maintenance overhead by excluding the NULL entries. 
sql

-- Index only the actual data, ignoring NULLs
CREATE INDEX idx_not_null_only ON your_table (column_name) 
WHERE column_name IS NOT NULL;

Conversely, if you only care about finding the NULL values (e.g., finding "incomplete" records), you can index only those: 
sql

CREATE INDEX idx_nulls_only ON your_table (column_name) 
WHERE column_name IS NULL;

```

crates/superposition_types/migrations/2026-04-24-000000_experiment_idempotency_key/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+ALTER TABLE experiments ADD COLUMN IF NOT EXISTS idempotency_key TEXT;
```
Datron: Isn't the convention to put public.experiments as the table name?

sauraww (approved):

makefile
```diff
@@ -308,7 +308,8 @@ smithy-clients: smithy-build
 	done
 	git restore clients/python/sdk/README.md
 	git restore LICENSE
-	git apply smithy/patches/*.patch
+	git apply --ignore-whitespace smithy/patches/*.patch
+	perl -pi -e 's/^            ShapeID\("smithy\.api#httpBasicAuth"\)$$/           ShapeID("smithy.api#httpBasicAuth"),/' clients/python/sdk/superposition_sdk/models.py
```
Datron: perl might be an issue, add this as a patch in python?

smithy/patches/python.patch
```diff
@@ -14,781 +14,11 @@ index 7c295b28..b7a7bd7e 100644
...
+ 
 +[tool.hatch.version]
 +source = "env"
 +variable = "VERSION"
 +
  [tool.pyright]
  typeCheckingMode = "strict"
  reportPrivateUsage = false
-
-diff --git a/clients/python/sdk/superposition_sdk/models.py b/clients/python/sdk/superposition_sdk/models.py
```
Datron, line 26: why remove this?

Datron (commented):
small comments

smithy/patches/python.patch
mahatoankitkumar: perl does the same job

Datron: If it does why not keep the diff?

Datron (approved):

PR #990, opened 2026-05-04
fix: update experiment groups when experiments change

Datron:
## Problem
Experiment allocation is wrong because experiment groups aren't updated in redis when experiments change

## Solution
On every experiment action update experiment groups

## Release Notes

* **Bug Fixes**
  * Improved cache synchronization for experiment state changes (create, conclude, discard, update, pause, resume) to ensure consistency across all related data.
  * Enhanced variant allocation to use authenticated configuration retrieval for more reliable and current variant assignment.

### DIFF
--- a/crates/experimentation_platform/src/api/experiments/handlers.rs
+++ b/crates/experimentation_platform/src/api/experiments/handlers.rs
@@ -84,7 +84,8 @@ use superposition_types::{
 use crate::api::{
     experiment_groups::helpers::{
         add_members, create_system_generated_experiment_group,
-        detach_experiment_from_group, update_experiment_group_buckets,
+        detach_experiment_from_group, put_experiment_groups_in_redis,
+        update_experiment_group_buckets,
     },
     experiments::{
         helpers::{
@@ -479,6 +480,13 @@ async fn conclude_handler(
         put_experiments_in_redis(&state.redis, &mut conn, &workspace_context.schema_name)
             .await;
 
+    let _ = put_experiment_groups_in_redis(
+        &state.redis,
+        &mut conn,
+        &workspace_context.schema_name,
+    )
+    .await;
+
     let experiment_response = ExperimentResponse::from(response);
 
     let data = WebhookData {
@@ -754,6 +762,12 @@ async fn discard_handler(
         put_experiments_in_redis(&state.redis, &mut conn, &workspace_context.schema_name)
             .await;
 
+    let _ = put_experiment_groups_in_redis(
+        &state.redis,
+        &mut conn,
+        &workspace_context.schema_name,
+    )
+    .await;
     let experiment_response = ExperimentResponse::from(response);
 
     let data = WebhookData {
@@ -1484,6 +1498,13 @@ async fn ramp_handler(
     let _ =
         put_experiments_in_redis(&state.redis, &mut conn, &workspace_context.schema_name)
             .await;
+
+    let _ = put_experiment_groups_in_redis(
+        &state.redis,
+        &mut conn,
+        &workspace_context.schema_name,
+    )
+    .await;
 
     let webhook_event = if matches!(experiment.status, ExperimentStatusType::CREATED) {
         WebhookEvent::ExperimentStarted


crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -995,10 +1017,35 @@ async fn get_applicable_variants_handler(
         );
         (context, identifier)
     };
-
```
mahatoankitkumar, line 998: Why is this changed?

Datron: Airborne is hitting this API and during peak traffic they are being restricted by the DB connection

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -995,10 +1017,35 @@ async fn get_applicable_variants_handler(
...
-    let di = run_query(&state.db_pool, |conn| {
-        fetch_dimensions_info_map(conn, &workspace_context.schema_name)
+    let http_client = reqwest::Client::new();
+    let url = state.cac_host.clone() + "/config";
```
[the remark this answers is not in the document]
Datron: latest is assumed when not specified, and is the default

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1994,6 +2061,13 @@ async fn resume_handler(
         put_experiments_in_redis(&state.redis, &mut conn, &workspace_context.schema_name)
             .await;
 
+    let _ = put_experiment_groups_in_redis(
```
ayushjain17: @Datron we dont touch exp-groups in `resume`, `pause`, `update` and `create`

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -995,10 +1017,35 @@ async fn get_applicable_variants_handler(
...
+    let response = http_client
+        .get(&url)
+        .headers(headers_map.into())
+        .header(
+            header::AUTHORIZATION,
+            format!("Internal {}", state.superposition_token),
+        )
+        .send()
+        .await;
+    let config = process_cac_http_response::<Config>(response).await?;
```
ayushjain17: can we rather do a direct read from redis on fallback, do a run_query as the old code - can we keep it simple here

ayushjain17 (changes requested):
added comments

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: The fallbacks need to access config tables, versions and lead to duplication of code or I have to import context-aware-config. Either way I used the same way we're taking to context-aware-config today, this is the simplest method

crates/experimentation_platform/src/api/experiments/handlers.rs
Datron: @ayushjain17 given how closely experiments and experiment groups are tied it would be better to update it in every case right? If you're sure about this separation, I can remove it. Let me know

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -995,10 +1011,35 @@ async fn get_applicable_variants_handler(
         );
         (context, identifier)
     };
-
-    let di = run_query(&state.db_pool, |conn| {
-        fetch_dimensions_info_map(conn, &workspace_context.schema_name)
+    let http_client = reqwest::Client::new();
+    let url = state.cac_host.clone() + "/config";
```
knutties: Would this break with a prefix set ?

Datron: The API for applicable variants doesn't take prefixes as an argument, just context and identifier

ayushjain17: but why not make a direct redis call and use db query for fallback, why not keep it simple ?

Datron: @ayushjain17 why do say this is not the simplest method? It still hits redis right, just via API. The other way requires a lot of imports since I have to generate cac and there are other checks we do. Besides that, we would violate the abstract boundary we have between context-aware-config and experiments

knutties: > The API for applicable variants doesn't take prefixes as an argument, just context and identifier

Why so ?  that seems like an assumption.  Prefix should be common to all APIs right ?

Datron: I'm speaking looking at the type currently defined. Also this is to get Applicable variants, would prefix help?

Datron: This is the current request type defined
```
#[derive(Debug, Serialize, Deserialize)]
pub struct ApplicableVariantsRequest {
    // TODO: remove this once services start using newer sdk
    #[serde(alias = "toss", deserialize_with = "deserialize_identifier")]
    pub identifier: String,
}
```

ayushjain17 (approved):

sauraww (approved):

sauraww (approved):
Formatting changes are pending

PR #998, opened 2026-05-11
feat(observability): OpenTelemetry golden-signals middleware

knutties:
## Summary

Adds an Actix middleware and supporting subsystem that emits Google SRE golden signals (latency, traffic, errors, saturation) for every HTTP route on the main API, exposed via Prometheus scrape on a dedicated port (default `:9091`) and optional OTLP push. Built on the OpenTelemetry Rust SDK plus the upstream `opentelemetry-instrumentation-actix-web` crate, so users can scrape into Prometheus / VictoriaMetrics / Mimir or push to any OTel-native backend without code changes.

- **Spec:** [`docs/superpowers/specs/2026-05-10-otel-golden-signals-middleware-design.md`](docs/superpowers/specs/2026-05-10-otel-golden-signals-middleware-design.md)
- **Plan:** [`docs/superpowers/plans/2026-05-10-otel-golden-signals-middleware.md`](docs/superpowers/plans/2026-05-10-otel-golden-signals-middleware.md)

The spec was rewritten in-place after implementation so its body reads as the as-built design; §0 has a compact "Changes from initial design" summary table. The plan is preserved as historical record of original intent — its "Post-implementation deviations" section maps the 22 original tasks to what actually shipped (dropped / replaced / edited).

## What ships

**HTTP golden signals** — the upstream `opentelemetry-instrumentation-actix-web` `RequestMetrics` middleware, wrapped with project-specific customisations.
- `http.server.request.duration` — histogram (s), default semconv buckets, labels: `http.request.method`, `http.route`, `http.response.status_code`, `error.type` (4xx/5xx), `sp.org_id`*, `sp.workspace_id`*. Traffic + error rate are derived from `_count` in PromQL; load-average via `rate(_sum) / num_workers`.
- `http.server.active_requests` — UpDownCounter, labels: `http.request.method`, `http.route`, `sp.org_id`*, `sp.workspace_id`*.

\* Both tenant labels are env-disable-able for very-high-cardinality deployments via a process-wide `LabelConfig` (the crate's attribute hook is a `fn` pointer, so config lives in a `OnceLock`).

**Customisations layered on `RequestMetrics`:**
- `CardinalityBoundedFormatter` (`RouteFormatter` impl) collapses unmatched paths to `__not_found__` and `/pkg|/assets|/favicon` prefixes to `__static__`, keeping `http.route` cardinality bounded.
- `metrics_attrs_from_req` replaces the crate's default attribute set (which would emit `server.address`, `server.port`, `url.scheme`, `network.protocol.version`) with the minimal four above. Normalizes HTTP method to `_OTHER` for unknown verbs so a probing client can't inflate cardinality.
- `Aggregation::Drop` SDK View suppresses the body-size histograms the crate emits unconditionally.

**Saturation signals**
- `db.client.connections.{usage,max}` — passive r2d2 callbacks, no instrumentation at `pool.get()` call sites.
- `redis.client.{connections.connected,commands.in_flight}` — via a tolerant `RedisStats` trait over fred's `MetricsInterface`.
- `runtime.tokio.workers`, `runtime.tokio.global_queue.depth` — instantaneous gauges read directly from `tokio::runtime::Handle::metrics()` inside each observable callback. No background sampler, no atomics snapshot.
- `runtime.tokio.workers.busy.time` — monotonic counter (s), sum of `worker_total_busy_duration(i)` across workers. Prometheus computes saturation at query time: `rate(runtime_tokio_workers_busy_time_seconds_total[1m]) / runtime_tokio_workers`. Gated on `target_has_atomic = "64"` the same way tokio gates the underlying API.

**Endpoints**
- `GET /metrics` on `SUPERPOSITION_METRICS_PORT` (default `9091`) — Prometheus text exposition.
- The pre-existing `GET /health` on the main port is untouched.

**Resilience**
- Bad OTLP config logs and degrades to Prometheus-pull only, doesn't crash the binary.
- Metrics server lifetime is detached from the main API server, so a failure on one doesn't drag the other down.
- Middleware binds to an explicit `SdkMeterProvider` (not the OTel global), so multiple observability stacks — e.g. parallel integration tests — stay isolated.
- `__not_found__` and `__static__` route sentinels bound `http.route` cardinality.

## Configuration

Standard OTel env vars are honored (`OTEL_SERVICE_NAME`, `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_RESOURCE_ATTRIBUTES`, `OTEL_EXPORTER_OTLP_PROTOCOL`). Plus:

| Var | Default | Purpose |
|---|---|---|
| `SUPERPOSITION_METRICS_ENABLED` | `true` | Master switch. |
| `SUPERPOSITION_METRICS_PORT` | `9091` | Scrape port. |
| `SUPERPOSITION_METRICS_BIND` | `0.0.0.0` | Bind address (v6-safe; `::` for dual-stack). |
| `SUPERPOSITION_METRICS_LABEL_ORG` | `true` | Include `sp.org_id`. |
| `SUPERPOSITION_METRICS_LABEL_WORKSPACE` | `true` | Include `sp.workspace_id`. |
| `SUPERPOSITION_METRICS_COLLECT_INTERVAL` | `10s` | OTLP periodic-reader push interval. |

`OTEL_EXPORTER_OTLP_PROTOCOL` other than `http/protobuf` produces a warning at init — this binary is built with the HTTP transport only.

## Test Plan

- [x] `cargo test --workspace` — all green
- [x] 11 unit tests across `service_utils::observability` (config, instrumentation, metrics_server, init)
- [x] 3 integration tests in `crates/service_utils/tests/observability_integration.rs`:
  - `metrics_appear_after_requests` — wraps an Actix App with `build_request_metrics_middleware`, sends 200/201/500/404, asserts on scraped Prometheus exposition (per-route counts, 5xx series, `__not_found__` sentinel, body-size histograms suppressed, `active_requests` returns to 0).
  - `cardinality_stays_within_budget` — fails CI if a 3-route × 1-method scenario exceeds 200 series.
  - `runtime_tokio_metrics_appear_after_register_observers` — registers saturation observers under a real tokio runtime and asserts the three `runtime.tokio.*` series appear in the scrape (with `runtime_tokio_workers >= 1` to verify the value actually comes from `Handle::metrics()`).
- [x] `cargo check -p superposition` — clean wiring.
- [ ] Live deployment smoke (deferred to staging rollout per spec §13).

## Rollout

Phased per spec §13:
1. Land with code default `SUPERPOSITION_METRICS_ENABLED=true` but disabled in prod env files.
2. Enable in staging, watch VM ingest 48h.
3. Enable in prod with `SUPERPOSITION_METRICS_LABEL_WORKSPACE=false`.
4. Flip workspace label on after a week of clean VM headroom.

## Future work (explicitly out of scope, tracked in spec §14)

- `/healthz` `/livez` `/readyz` kubernetes-conventional probes (dropped from this PR; pre-existing `/health` covers the up-check role; the live/ready split lands in a focused follow-up with real DB/Redis readiness checks).
- Trace correlation via exemplars (requires `tracing-opentelemetry` bridge).
- Per-tenant separate histogram if global cardinality budget tightens.
- Grafana dashboards JSON + VM/Prometheus alert rules (separate PR).
- Typed pool wrapper to unlock `db.client.connection.wait.duration` + `pending_requests`.
- gRPC OTLP transport (currently `http/protobuf` only).


crates/service_utils/src/observability/saturation/db_pool.rs
```diff
@@ -0,0 +1,62 @@
...
+    let pool_for_max = pool.clone();
+    let max_pool_name = KeyValue::new("pool.name", pool_name);
+    meter
+        .u64_observable_gauge("db.client.connections.max")
+        .with_description("Configured maximum size of the DB connection pool.")
+        .with_callback(move |observer| {
+            observer.observe(pool_for_max.max_size() as u64, &[max_pool_name.clone()]);
+        })
+        .build();
+}
```
Datron, line 48: Should we just read the env DB_MAX_CONNECTION_POOL_SIZE and publish, getting the size of the pool array may not be accurate for larger pools

crates/service_utils/src/observability/saturation/db_pool.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[cfg(test)]
+mod tests {
+    // Constructing a real r2d2 pool requires a database. We assert the function
+    // signature compiles and that calling `register` does not panic with a
+    // synthetic in-memory pool; this is exercised via the integration test in
+    // Task 19 instead.
+}
```
Datron: Should be just remove this?

crates/service_utils/src/observability/saturation/redis_pool.rs
```diff
@@ -0,0 +1,128 @@
...
+        None
+    }
+
+    fn used_connections(&self) -> Option<u64> {
+        // TODO(observability): fred 9.2.1 does not expose active/used
+        // connection counts. The internal `Counters.in_flight` field tracks
+        // per-connection in-flight commands, but it is private to
+        // `fred::protocol::connection`. Return None until exposed publicly.
+        None
+    }
```
Datron: You can get the number of used clients via
```
let connected_count = pool_clone.clients().iter().filter(|c| c.is_connected()).count();
```

crates/service_utils/src/observability/saturation/redis_pool.rs
```diff
@@ -0,0 +1,128 @@
...
+pub struct FredPoolStats(pub RedisPool);
+
+impl RedisStats for FredPoolStats {
+    fn idle_connections(&self) -> Option<u64> {
+        // TODO(observability): fred 9.2.1 does not expose idle-connection
+        // counts via any public API. The pool's `RedisPoolInner.clients` are
+        // all async and there is no `is_idle()` method on `RedisClient`.
+        // Return None until fred provides such an API.
+        None
+    }
```
Datron: We can just remove this

crates/service_utils/src/observability/saturation/tokio_runtime.rs
```diff
@@ -0,0 +1,93 @@
+//! Tokio runtime saturation, gated on `cfg(tokio_unstable)`.
+//!
+//! Unlike DB/Redis, `tokio_metrics::RuntimeMonitor` is delta-based: each
+//! `.intervals()` call returns stats since the last call. So we run a
+//! background task that samples every `interval` and stores derived values
+//! in atomics that observable-gauge callbacks read.
+
+#[cfg(not(tokio_unstable))]
+#[allow(dead_code)] // called from saturation.rs only under cfg(tokio_unstable)
```
Datron: Do we need this if cfg is at the top?

crates/service_utils/src/observability/saturation/tokio_runtime.rs
```diff
@@ -0,0 +1,93 @@
...
+pub fn spawn(_meter: &opentelemetry::metrics::Meter, _interval: std::time::Duration) {}
```
Datron: dummy function

crates/service_utils/src/observability/saturation/tokio_runtime.rs
```diff
@@ -0,0 +1,93 @@
...
+    use std::sync::atomic::{AtomicU64, Ordering};
+    use std::time::Duration;
+
+    use opentelemetry::metrics::Meter;
+    use tokio_metrics::RuntimeMonitor;
+
+    #[derive(Default)]
+    struct Snapshot {
+        workers: AtomicU64,
+        global_queue_depth: AtomicU64,
```
Datron: This can be simplified a bit, just a function that reads metrics
```
// 1. Global Queue Depth: Tasks waiting to be picked up by an Actix worker
    let h_clone = handle.clone();
    meter.u64_observable_gauge("tokio.runtime.global_queue_depth")
        .with_description("Tasks stuck in the global queue waiting for a worker thread")
        .with_callback(move |observer| {
            let metrics = h_clone.metrics();
            observer.observe(metrics.global_queue_depth() as u64, &[]);
        })
        .init();

    // 2. Active Tasks: Total tasks currently alive in the runtime
    let h_clone2 = handle.clone();
    meter.u64_observable_gauge("tokio.runtime.active_tasks")
        .with_description("Total number of alive tasks (running or idle)")
        .with_callback(move |observer| {
            let metrics = h_clone2.metrics();
            observer.observe(metrics.active_tasks_count() as u64, &[]);
        })
        .init();
        
    // 3. Worker Count: How many OS threads Tokio is using for workers
    let h_clone3 = handle.clone();
    meter.u64_observable_gauge("tokio.runtime.workers_count")
        .with_description("Number of worker threads")
        .with_callback(move |observer| {
            let metrics = h_clone3.metrics();
            observer.observe(metrics.num_workers() as u64, &[]);
        })
        .init();
```

crates/service_utils/src/observability/config.rs
```diff
@@ -0,0 +1,180 @@
...
+fn hostname_or_none() -> Option<String> {
+    // Avoid pulling in a hostname crate; read /etc/hostname on Linux/macOS.
+    std::fs::read_to_string("/etc/hostname")
+        .ok()
+        .map(|s| s.trim().to_owned())
+        .filter(|s| !s.is_empty())
+}
+
+#[cfg(test)]
+mod tests {
```
Datron, line 138: Don't know if we need tests

crates/service_utils/src/observability/config.rs
```diff
@@ -0,0 +1,180 @@
+//! Configuration for the observability subsystem, parsed from env vars.
+
+use std::{net::IpAddr, str::FromStr, time::Duration};
+
+#[derive(Debug, Clone)]
+pub struct ObservabilityConfig {
```
Datron, line 6: Should this be moved to app state?

Datron (changes requested):
Overall review
1. Some crates can be removed, they are not used. Just imported
```
opentelemetry-otlp = { version = "0.27", default-features = false, features = ["metrics", "http-proto", "reqwest-client"] }
opentelemetry-semantic-conventions = { version = "0.27" }
```
2. We can use https://crates.io/crates/opentelemetry-instrumentation-actix-web which comes from open-telemetry so that we can directly monitor requests, latency and errors
3. Simplify the observability module, there are too many files with just 30 lines. Could we group everything in fewer files and use internal modules if we want namespacing
4. Is it possible to add some CPU and memory monitoring on the box with sysinfo crate or we'll leave that to a deployment runtime like ECS

crates/service_utils/src/observability/config.rs
knutties: The full `ObservabilityConfig` is only used during startup wiring (init, observer registration, metrics-server bind). The one piece the main-app middleware needs at request time — `LabelConfig` — is already captured directly into the `MetricsMiddleware` closure (`main.rs:218`). The `/metrics` endpoint runs on a separate `HttpServer` whose `app_data` is the `Registry`, not the config. Adding it to `AppState` would be an unused field.

crates/service_utils/src/observability/config.rs
knutties: I guess it is useful to ensure defaults are not modified and are typed correctly ?

knutties:
> Overall review
> 
> 1. Some crates can be removed, they are not used. Just imported
> 
> ```
> opentelemetry-otlp = { version = "0.27", default-features = false, features = ["metrics", "http-proto", "reqwest-client"] }
> opentelemetry-semantic-conventions = { version = "0.27" }
> ```
`opentelemetry-otlp` is needed for push flow of metrics instead of pull

`opentelemetry-semantic-conventions` - this is removed.

> 2. We can use https://crates.io/crates/opentelemetry-instrumentation-actix-web which comes from open-telemetry so that we can directly monitor requests, latency and errors

Will respond to this separately.

> 3. Simplify the observability module, there are too many files with just 30 lines. Could we group everything in fewer files and use internal modules if we want namespacing

the namespacing is clean based on the component that is being instrumented right.

> 4. Is it possible to add some CPU and memory monitoring on the box with sysinfo crate or we'll leave that to a deployment runtime like ECS

This we will do outside of the application.

crates/service_utils/src/observability/saturation/db_pool.rs
knutties: @Datron - max_size is a single field read right - not a walk through the array to calculate the pool length.  Not sure why you say it might not be accurate.

knutties:
> We can use https://crates.io/crates/opentelemetry-instrumentation-actix-web which comes from open-telemetry so that we can directly monitor requests, latency and errors

knutties:
> We can use https://crates.io/crates/opentelemetry-instrumentation-actix-web which comes from open-telemetry so that we can directly monitor requests, latency and errors

@Datron - this is now done.

docs/superpowers/plans/2026-05-10-otel-golden-signals-middleware.md
```diff
@@ -0,0 +1,2387 @@
+# OpenTelemetry Golden-Signals Middleware Implementation Plan
```
Datron, line 1: Lets not commit this?

crates/service_utils/src/observability/saturation/db_pool.rs
Datron: max_size gives you the size of the vector allocated, not the connections that are in the vector

Datron: You can also avoid the clone

crates/service_utils/src/observability/config.rs
Datron: The compiler would check if its typed correctly, and don't know if we should enforce defaults via tests.

crates/service_utils/src/observability/metrics_server.rs
```diff
@@ -0,0 +1,63 @@
...
+    registry: Arc<Registry>,
+    bind: SocketAddr,
+) -> std::io::Result<Server> {
+    let registry_data = web::Data::new(registry);
+    Ok(HttpServer::new(move || {
+        App::new()
+            .app_data(registry_data.clone())
+            .route("/metrics", web::get().to(scrape))
+    })
+    .workers(1)
```
Datron, line 21: nitpick: workers as an ENV, like METRICS_WORKER_COUNT

Datron (approved):
mostly nitpicks and things we can change later.

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
```diff
@@ -64,9 +64,9 @@ use crate::api::{
...
         cac_api::validate_context,
         helpers::{
-            validate_change_reason_with_function, hash,
-            validate_and_add_experiment_group_id,
+            hash, validate_and_add_experiment_group_id,
             validate_and_remove_experiment_group_id,
+            validate_change_reason_with_function,
         },
     },
 };
```
sauraww, line 72: these changes are not needed

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -89,9 +89,9 @@ use crate::api::{
     },
     experiments::{
         helpers::{
-            validate_change_reason_with_function,
             get_control_overrides_from_exp_id, put_experiments_in_redis,
-            validate_control_overrides, validate_delete_experiment_variants,
+            validate_change_reason_with_function, validate_control_overrides,
+            validate_delete_experiment_variants,
         },
```
sauraww, line 95: these too.

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -2004,7 +2004,7 @@ async fn resume_handler(
         &workspace_context,
         &req.change_reason,
         &state,
-        &user
+        &user,
     )
     .await?;
```
sauraww, line 2010: unnecessary changes in this file too

crates/experimentation_platform/src/api/experiments/helpers.rs
```diff
@@ -790,7 +790,8 @@ pub async fn validate_change_reason_with_function(
         change_reason: change_reason.clone(),
     };
 
-    let headers_map = construct_header_map(workspace_context, vec![("x-user", user_str)])?;
+    let headers_map =
+        construct_header_map(workspace_context, vec![("x-user", user_str)])?;
```
sauraww, line 794: Here as well

crates/service_utils/src/observability/saturation/db_pool.rs
sauraww: The first clone is still needed @Datron . We can remove the second clone. 
Also, I don't think we need to wrap the type with Arc here.

sauraww (changes requested):

docs/superpowers/plans/2026-05-10-otel-golden-signals-middleware.md
sauraww: +1

sauraww (changes requested):

docs/superpowers/plans/2026-05-10-otel-golden-signals-middleware.md
knutties: Why folks - it does have useful information in my view?

What are the arguments to not commit ?

crates/experimentation_platform/src/api/experiment_groups/handlers.rs
knutties: Addressed in 5dd55b87 — restored the three experimentation_platform files to their merge-base state. The rustfmt commit had touched them in a way that conflicts with the heavy churn on main since (#982, #1008, #1044). The PR diff no longer includes these files.

crates/experimentation_platform/src/api/experiments/handlers.rs
knutties: Addressed in 5dd55b87 along with the other rustfmt churn in this crate.

crates/experimentation_platform/src/api/experiments/handlers.rs
knutties: Addressed in 5dd55b87.

crates/experimentation_platform/src/api/experiments/helpers.rs
knutties: Addressed in 5dd55b87.

knutties: Addressed in 5dd55b87.

crates/service_utils/src/observability/saturation/db_pool.rs
knutties: Addressed in b85542bb — `DbPoolHandle` now aliases the bare `diesel::r2d2::Pool<...>` (r2d2 already has an internal Arc, so the outer wrap was redundant). The second clone is gone too: the first closure clones once, the second consumes the parameter.

crates/service_utils/src/observability/saturation/redis_pool.rs
```diff
@@ -0,0 +1,91 @@
...
+        )
+        .with_callback(move |observer| {
+            if let Some(n) = c.connected_connections() {
+                observer.observe(n, std::slice::from_ref(&label));
+            }
+        })
+        .build();
+
+    let c = client.clone();
+    let label = pool_label.clone();
```
sauraww:
```
    let c = client;
    let label = pool_label;
```

crates/service_utils/src/observability.rs
```diff
@@ -0,0 +1,263 @@
...
+                    builder = SdkMeterProvider::builder()
+                        .with_reader(prom_exporter)
+                        .with_resource(resource)
+                        .with_view(drop_body_size_view);
+                }
+            }
+        }
+
+        let provider = builder.build();
+        opentelemetry::global::set_meter_provider(provider.clone());
```
sauraww: Is this required ? We aren't using this.

sauraww (changes requested):

sauraww (approved):

Datron (approved):

PR #1003, opened 2026-05-13
improve show_reasoning 

sauraww:
## Problem
We don't have a way to show cascading behavior of resolve api

## Solution
Adding layers and per_key_history in resolve , if include_trace is true, you will get back the entire trace.


Datron:
Closing due to another PR #1040

PR #1008, opened 2026-05-19
feat: add AnyMatch filter to get contexts based on query permutations

sauraww:
Added a new AnyMatch dimension matching strategy.

AnyMatch works as:

partial value match using existing partial apply eval logic.
key match succeeds when at least one context dimension overlaps with the requested query dimensions or is reachable through the dependency graph

```
http://localhost:8080/context?dimension_match_strategy=any_match&{query_params}

### CODE REVIEW

File: crates/context_aware_config/src/api/context/handlers.rs [hunk 1]
```diff
@@ -1287,3 +1309,70 @@ async fn validate_handler(
     log::debug!("Context {:?} is valid", ctx_condition);
     Ok(HttpResponse::Ok().finish())
 }
+
+#[cfg(test)]
```
mahatoankitkumar: can we move this test to somewhere else

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -619,22 +619,44 @@ async fn list_handler(
         }
         let dimensions_info =
             fetch_dimensions_info_map(&mut conn, &workspace_context.schema_name)?;
+
         let original_req_keys = dimension_params.keys().collect::<Vec<_>>();
-        let dimension_params =
+        let query_keys = dimension_params.keys().cloned().collect::<HashSet<_>>();
```
Datron: You could do this, check if you need to clone
```suggestion
        let query_keys = HashSet::from_iter(original_req_keys);
```

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1244,28 +1244,43 @@ fn list_experiments_db(
...
+                    Experiment::filter_by_eval(all_experiments, &dimension_params)
+                }
+            };
 
-            Experiment::filter_by_dimension(
-                dimension_filtered_experiments,
-                &original_req_keys,
-                &dimensions_info,
-            )
+            match strategy {
```
mahatoankitkumar, line 1270: Write the empty check at the beginning of all the matches

crates/experimentation_platform/src/api/experiments/handlers.rs
```diff
@@ -1248,24 +1248,37 @@ fn list_experiments_db(
...
+                DimensionMatchStrategy::AnyMatch => {
+                    Experiment::filter_by_context_any_match(
+                        dimension_filtered_experiments,
+                        &query_keys,
+                        &dimensions_info,
+                    )
+                }
```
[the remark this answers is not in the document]
sauraww: @coderabbitai This is already handled now . Can you check ?

crates/context_aware_config/src/api/context/handlers.rs
```diff
@@ -617,24 +617,43 @@ async fn list_handler(
...
+                    eval_filtered,
+                    &query_keys,
+                    &dimensions_info,
+                ),
+                _ => Context::filter_by_dimension(
+                    eval_filtered,
+                    &original_req_keys,
+                    &dimensions_info,
+                ),
+            }
```
Datron, line 655: nitpick: if else would be better here

Datron (commented):
Small nitpicks

crates/superposition_types/src/contextual.rs
```diff
@@ -57,10 +57,37 @@ pub trait Contextual: Clone {
...
 
 #[cfg(test)]
 mod tests {
+    use std::collections::{HashMap, HashSet};
+
+    use serde_json::{Map, Value};
+
     use crate::config::tests::map::{
         get_dimension_filtered_contexts1, get_dimension_filtered_contexts2,
         with_dimensions::{
```
Datron, line 93: Can we simplify this or write it better? A for loop for the first and second layer would help readability

crates/superposition_types/src/contextual.rs
```diff
@@ -57,10 +57,37 @@ pub trait Contextual: Clone {
...
+            .collect()
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::collections::{HashMap, HashSet};
+
+    use serde_json::{Map, Value};
+
```
Datron, line 90: Why are we checking dependency graph for a match? Would be better to resolve cohorts and then run this?

sauraww: following the same pattern as we did in the filter_by_dimension - just doing it in the reverse order

crates/superposition_types/src/contextual.rs
sauraww: This is tests

smithy/models/common.smithy
```diff
@@ -50,4 +50,7 @@ enum DimensionMatchStrategy {
 
     @documentation("Match the overrides which have the given context as subset")
     SUBSET = "subset"
+
+    @documentation("Match overrides whose context has at least one supplied query dimension, with partial value matching and dependency-graph awareness")
+    ANY_MATCH = "any_match"
```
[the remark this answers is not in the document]
sauraww: This is done now

mahatoankitkumar (approved):

Datron (approved):

PR #1014, opened 2026-05-21
feat: add first class support for dimension params in smithy for clients sdk

sauraww:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

* **New Features**
  * The `list-contexts` API now accepts an optional `dimension_params` parameter, allowing additional dimension filter query parameters when retrieving contexts. Parameter keys should use full API query parameter names (e.g., `dimension[country]`), each mapped to a string value.

[Review Change Stack](https://app.coderabbit.ai/change-stack/juspay/superposition/pull/1014?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


smithy/models/common.smithy
```diff
@@ -31,6 +31,12 @@ list StringList {
     member: String
 }
 
+@documentation("Additional dimension filter query parameters. Keys must be full query parameter names accepted by the API, for example `dimension[country]`.")
+map DimensionQueryParams {
+    key: String
+    value: String
```
ayushjain17, line 37: Value should not be of String type
It should be Document

smithy/models/context.smithy
```diff
@@ -231,6 +231,10 @@ operation ListContexts {
         @httpQuery("dimension_match_strategy")
         @notProperty
         dimension_match_strategy: DimensionMatchStrategy
+
+        @httpQueryParams
+        @notProperty
+        dimension_params: DimensionQueryParams
```
ayushjain17, line 237: Is this present in exp and exp group list endpoints ?

ayushjain17 (dismissed):

smithy/models/common.smithy
sauraww: https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait

sauraww: Also tested locally , it is working.

Datron (approved):

mahatoankitkumar (approved):

PR #1032, opened 2026-05-28
fix(openfeature): block evaluation until experiments cache is ready

knutties:
## Summary
- Fix a race in the Java/Kotlin OpenFeature provider where `evaluateConfigInternal` only awaits the config cache (`cacheReady`) before calling `cache.evalConfig`, while the experiments cache is initialized by an independent refresh job.
- In the window between `cache.initConfig` and `cache.initExperiments`, an eval applies an empty experiment set and silently returns the default config for users that should have been bucketed into a variant.
- This is the root cause of intermittent Kotlin provider Test 9 failures (`Bangalore customer` returning `10000` — the default — instead of `7000`/`8000`).

## Fix
In `SuperpositionOpenFeatureProvider.java`:
- Add `experimentsReady` (a parallel `CompletableFuture<Void>` to `cacheReady`).
- Initialize a new `experimentationTimeout` field from `options.experimentationOptions.refreshStrategy.getTimeout()`. When `experimentationOptions` is `null`, complete `experimentsReady` eagerly so the eval path never blocks.
- Complete `experimentsReady` inside `reinitExperimentsCache` immediately after `cache.initExperiments(...)` succeeds.
- In `evaluateConfigInternal`, after waiting on `cacheReady`, also wait on `experimentsReady` (bounded by `experimentationTimeout`) before calling `cache.evalConfig`. On timeout, throw — caller maps that to `PROVIDER_NOT_READY`, matching the existing config-cache timeout behavior.

The pattern mirrors the existing `cacheReady` plumbing 1:1.

## Test plan
- [ ] CI: `Provider Tests (kotlin)` — Test 9 (`Experimentation - Bangalore customer`) should now be deterministic
- [ ] CI: existing JUnit `ProviderTest` suite passes
- [ ] Manual: configure provider without `experimentationOptions` — eval path should not block

https://claude.ai/code/session_01JNFhzR5UjoXwcRqSDxUvgE

---
_Generated by [Claude Code](https://claude.ai/code/session_01JNFhzR5UjoXwcRqSDxUvgE)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Experimentation data now waits for proper initialization before use
  * Added timeout protection to prevent indefinite evaluation delays
  * Graceful error handling when experiment initialization doesn't complete within the timeout window

[Review Change Stack](https://app.coderabbit.ai/change-stack/juspay/superposition/pull/1032?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionOpenFeatureProvider.java
+++ b/clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionOpenFeatureProvider.java
@@ -77,7 +77,9 @@ public class SuperpositionOpenFeatureProvider implements FeatureProvider {
     private Optional<EvaluationContext> defaultCtx;
     private final Optional<EvaluationArgs> fallbackArgs;
     private final CompletableFuture<Void> cacheReady = new CompletableFuture<>();
+    private volatile CompletableFuture<Void> experimentsReady = new CompletableFuture<>();
     private final int configTimeout;
+    private final int experimentationTimeout;
 
     public SuperpositionOpenFeatureProvider(@NonNull SuperpositionProviderOptions options) {
         if (options.fallbackConfig != null) {
@@ -107,6 +109,8 @@ public class SuperpositionOpenFeatureProvider implements FeatureProvider {
         );
 
         if (options.experimentationOptions != null) {
+            this.experimentationTimeout =
+                options.experimentationOptions.refreshStrategy.getTimeout();
             var listExpInput = ListExperimentInput.builder()
                 .orgId(options.orgId)
                 .workspaceId(options.workspaceId)
@@ -145,6 +149,9 @@ public class SuperpositionOpenFeatureProvider implements FeatureProvider {
         } else {
             this.expRefresh = Optional.empty();
             this.expGroupRefresh = Optional.empty();
+            this.experimentationTimeout = 0;
+            // No experimentation configured: unblock the evaluation path immediately.
+            this.experimentsReady.complete(null);
         }
     }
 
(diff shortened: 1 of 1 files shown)


clients/java/openfeature-provider/src/main/java/io/juspay/superposition/openfeature/SuperpositionOpenFeatureProvider.java
```diff
@@ -311,6 +318,16 @@ private Map<String, String> evaluateConfigInternal(EvaluationContext ctx) throws
                 throw new Exception("Config cache not initialized within timeout (" + configTimeout + "ms).");
             }
         }
+        // When experimentation is configured, also block until cache.initExperiments has
+        // been called, otherwise evalConfig may apply an empty experiment set and silently
+        // fall back to the default config for users that should be bucketed into a variant.
+        if (!experimentsReady.isDone()) {
+            try {
+                experimentsReady.get(experimentationTimeout, TimeUnit.MILLISECONDS);
+            } catch (TimeoutException e) {
```
sauraww, line 327: If cache.initExperiments() fails in reinitExperimentsCache, the future stays PENDING forever, between the window of a failed poll and the next successful poll, every evaluation will block here for the full experimentationTimeout , then throw TimeoutException ( if timeout is smaller than polling frequency ) , which gets caught in getEvaluation → and returns defaultValue


Whether users get defaultValue depends on how the caller configures refreshStrategy - if experimentationTimeout is smaller than the polling interval, users in the failed-poll window will get defaultValue. We cannot guarantee this won't happen

sauraww (dismissed):

Datron (approved):

mahatoankitkumar (approved):

PR #1034, opened 2026-06-03
feat(supertoml): import a full config from TOML/JSON

knutties:
Adds first-class **config import** to complement the existing TOML/JSON export.

## Endpoints

Import has its own endpoints, so the export operations stay cleanly read-only:

| Format | Export             | Import                     |
| ------ | ------------------ | -------------------------- |
| TOML   | `GET /config/toml` | `POST /config/toml/import` |
| JSON   | `GET /config/json` | `POST /config/json/import` |

POST the file as the request body. It is parsed and **fully validated** (schemas,
dimension positions, cohort relationships, and every context condition/override)
before anything is written, then persisted in a **single transaction**:
dimensions and default-configs first, then the contexts that reference them.

## Options (request headers)

- `x-import-mode`: `merge` (default, upsert) | `replace` (mirror the file —
  deletes dimensions/default-configs/contexts that are absent from it)
- `x-import-overwrite`: when `false`, existing entities are skipped instead of
  updated (defaults to `true`)
- `x-import-on-error`: `abort` (default) | `continue` (apply everything valid and
  report per-entity errors via savepoints)
- `x-import-dry-run`: validate and summarise **without writing** (rolls back)
- `x-import-value-merge`: deep-merge object-valued default configs with the
  existing value
- `x-config-tags`: recorded against the config version the import creates

The response is a JSON summary of created/updated/skipped/deleted counts per
entity type (plus the new config version, and per-entity errors in `continue`
mode).

## Smithy / SDKs

The import is modelled in Smithy as the `ImportConfigToml` / `ImportConfigJson`
operations (with `ImportConfigOutput`, `ImportEntityReport`, and the `ImportMode`
/ `ImportOnError` enums). The generated SDKs and OpenAPI/API docs are regenerated
accordingly, so import is available as a typed command in every client (e.g.
`ImportConfigTomlCommand` in the JS SDK).

## Other changes

- Raises the raw payload limit to 10MB so large workspaces can be imported.
- Integration tests (JS SDK) covering merge/replace, overwrite, dry-run,
  value-merge, validation failures and a TOML round-trip.
- Hardened the `make test` health-check to wait for the server to finish
  building and booting before running the integration tests (it previously raced
  a cold rebuild and failed CI).
- Docs for the import/export workflow.

https://claude.ai/code/session_014YMnhBNrfY5mfr4qLqSpRt


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -69,6 +71,8 @@ pub fn endpoints() -> Scope {
         .service(get_toml_handler)
         .service(get_json_handler)
         .service(detailed_resolve_handler)
+        .service(import_toml_handler)
+        .service(import_json_handler)
         .service(resolve_handler)
```
Datron: Can we have one handler that takes the file format as a url param

Datron: When we support more file formats - this would just grow to more functions

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,843 @@
...
+}
+
+#[derive(Clone, Copy)]
+pub struct ImportOptions {
+    pub strategy: ImportStrategy,
+    pub on_error: OnError,
+    pub dry_run: bool,
+}
+
+fn header_bool(req: &HttpRequest, name: &str, default: bool) -> bool {
```
Datron: single use function, can be inlined

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,843 @@
...
+    fn from(e: diesel::result::Error) -> Self {
+        TxError::App(db_error!(e))
+    }
+}
+
+/// Run `f` within a SAVEPOINT so that a failure rolls back only the work done
+/// by `f` (leaving the surrounding transaction usable). This is what makes the
+/// `continue-on-error` option safe in the face of Postgres aborting the
+/// current transaction on a database error.
+fn with_savepoint<T>(
```
Datron: I don't think we should do savepoints, for example if dimensions fail - then there is no point importing context & overrides. It would be better if this was just a transaction. Either it imports or it doesn't

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,843 @@
...
+/// Whether an entity that already exists should be left untouched.
+fn should_skip(exists: bool, strategy: ImportStrategy) -> bool {
+    exists && strategy == ImportStrategy::CreateOnly
+}
+
+fn dimension_position(
+    name: &str,
+    info: &DimensionInfo,
+) -> superposition::Result<Position> {
+    Position::try_from(info.position)
```
Datron: nitpick: would be better to inline

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,843 @@
...
+
+    let summary =
+        import_config::<F>(body_str, opts, tags, user, workspace_context, state, conn)
+            .await?;
+
+    Ok(HttpResponse::Ok().json(summary))
+}
+
+#[cfg(test)]
+mod tests {
```
Datron, line 622: Lot of the tests seem pointless, can be simplified. Some obvious tests and asserts are being checked, this takes time in the CI but doens't really validate anything

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,843 @@
...
+    Description::try_from("Config imported via TOML/JSON import".to_string())
+        .unwrap_or_default()
+}
+
+// ---------------------------------------------------------------------------
+// Pure builders / decisions (no database access — unit tested below)
+// ---------------------------------------------------------------------------
+
+/// Whether an entity that already exists should be left untouched.
+fn should_skip(exists: bool, strategy: ImportStrategy) -> bool {
```
Datron: nitpick: would be better to inline

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,843 @@
...
+                    user,
+                    &email,
+                )
+            });
+            apply_outcome(&mut summary.contexts, opts.on_error, &ctx.id, res)?;
+        }
+
+        // Replace strategy: delete anything in the workspace that is not
+        // present in the imported file. Contexts first, then the entities they
+        // can reference.
```
Datron: Replace seems to be the wrong word here - it doesn't imply the deletion of resources that couldn't be replaced. Could confuse users

crates/frontend/src/utils.rs
```diff
@@ -304,6 +304,59 @@ where
     request_with_skip_error(url, method, body, headers, &[]).await
 }
 
+pub async fn request_raw_body(
```
Datron, line 441: nit: this seems to be another kronos use case. We can migrate post merging the kronos implementation PR

crates/frontend/src/api.rs
```diff
@@ -561,6 +563,185 @@ pub async fn fetch_config(
     parse_json_response(response).await
 }
 
+pub mod config_import {
+    use super::*;
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
Datron: The types seem duplicated, can we unify them in superposition_types?

crates/superposition/src/main.rs
```diff
@@ -235,6 +235,9 @@ async fn main() -> Result<()> {
             .app_data(app_state.clone())
             .app_data(PathConfig::default().error_handler(|err, _| bad_argument!(err).into()))
             .app_data(QueryConfig::default().error_handler(|err, _| bad_argument!(err).into()))
+            // Raise the raw-body limit (default 256KB) so config imports posted
```
Datron: Instead of raw body, the handler should take the file as a file upload (multipart/form-data)

Datron (changes requested):
few comments

crates/context_aware_config/src/api/config/handlers.rs
sauraww: Sure , makes sense. I was going to refactor this.

mahatoankitkumar: +1

crates/frontend/src/utils.rs
mahatoankitkumar: needs to be changed now i guess

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -659,6 +667,55 @@ async fn get_json_handler(
...
+    req: HttpRequest,
+    body: Bytes,
+    user: User,
+    custom_headers: CustomHeaders,
+    db_conn: DbConnection,
+    workspace_context: WorkspaceContext,
+    state: Data<AppState>,
+    format: Path<MarkupFormat>,
+) -> superposition::Result<HttpResponse> {
+    let DbConnection(mut conn) = db_conn;
```
Datron: This should use WritePermit no?

crates/context_aware_config/src/api/config/import.rs
```diff
@@ -0,0 +1,645 @@
...
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ImportStrategy {
+    CreateOnly,
+    Upsert,
+    Replace,
+}
+
+#[derive(Clone, Copy)]
```
Datron, line 40: Should these types be in superposition_types rather than here?

Datron (changes requested):
needs to be rebased, left a few comments

sauraww:
Changes moved to a different branch.

PR #1040, opened 2026-06-08
Add detailed config resolution endpoint

sauraww:
Adds richer config-resolution and explain API.

Added GetDetailedResolvedConfig at:GET /config/resolve/detailed
POST /config/resolve/detailed

Added detailed resolved-config responses that include:resolved value
default-config description
default-config schema/type metadata

Added GetResolvedConfigExplanation for explaining how a single config key resolves.
path : /config/resolve/explain/{key}


crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -651,6 +654,84 @@ async fn get_json_handler(
...
+        resolve_detailed(
+            &config,
+            query_data,
+            merge_strategy,
+            &mut conn,
+            &query_filters,
+            &workspace_context,
+            &state.master_encryption_key,
+        )?
+    };
```
Datron, line 710: Next PR: this seems to be the only thing different between resolve and resolve detailed. Maybe explain can be moved as a query param to resolve and we can unify the endpoints?

crates/context_aware_config/src/api/config/helpers.rs
```diff
@@ -163,29 +178,495 @@ pub fn resolve(
...
+        eval_cac(config, query_data, merge_strategy).map_err(|err| {
             log::error!("failed to eval cac with err: {}", err);
             unexpected_error!("cac eval failed")
         })
-    }?;
+    }
+}
 
-    Ok(response)
+fn explain_mode(
```
Datron: Should be an impl From on ExplainMode

Datron (approved):
approving, please raise a follow up PR if needed

mahatoankitkumar:
Expecting the key name in prefix is misleading.

mahatoankitkumar:
No until winner and all needed for explain
Both are same, it can be just a boolean

crates/context_aware_config/src/api/config/handlers.rs
mahatoankitkumar: Agreed

mahatoankitkumar:
Rather than adding explain as boolean and a separate keyName field
Can consider explain - keyName

mahatoankitkumar (changes requested):

crates/context_aware_config/src/api/config/handlers.rs
sauraww: Changing response structure based on query params is not a good practice. I don't think we should do it.

sauraww:
@mahatoankitkumar @Datron 
I am moving explain into its own API because it does not have a direct relationship with detailed resolve. 
Also, since explain can be used independently, exposing it as a separate API makes the contract cleaner and easier to understand.
CC: @knutties

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -666,50 +674,142 @@ async fn resolve_handler(
...
+#[routes]
+#[get("/resolve/explain")]
+#[post("/resolve/explain")]
+async fn explain_resolve_handler(
+    req: HttpRequest,
+    body: Option<Json<ContextPayload>>,
+    merge_strategy: Header<MergeStrategy>,
+    dimension_params: DimensionQuery<QueryMap>,
+    query_filters: superposition_query::Query<ResolveConfigQuery>,
+    explain_key: superposition_query::Query<ExplainKeyQuery>,
```
mahatoankitkumar: Take it as url param

mahatoankitkumar (approved):

PR #1050, opened 2026-06-15
feat(provider): add change-aware file refresh

sauraww:
Add an explicit change-aware refresh path for file-backed local providers.

The new `refresh_if_changed` API lets callers check file metadata before reloading provider cache, which is useful for EFS-backed config files where inotify/watch events are unavailable or unreliable. 
Existing refresh behavior is preserved: normal file refresh still reads fresh from disk, while the new opt-in path returns `NotModified` when the file has not changed.

### DIFF
--- a/crates/superposition_provider/src/data_source.rs
+++ b/crates/superposition_provider/src/data_source.rs
@@ -107,6 +107,18 @@ pub trait SuperpositionDataSource: Send + Sync {
             .await
     }
 
+    /// Fetch the full resolved configuration only if the source changed.
+    ///
+    /// The default implementation delegates to `fetch_config`; data sources
+    /// with cheaper change detection can override this without changing normal
+    /// refresh behavior.
+    async fn fetch_config_if_modified(
+        &self,
+        if_modified_since: Option<DateTime<Utc>>,
+    ) -> Result<FetchResponse<ConfigData>> {
+        self.fetch_config(if_modified_since).await
+    }
+
     /// Fetch a resolved configuration filtered by the given context and key prefixes.
     async fn fetch_filtered_config(
         &self,
--- a/crates/superposition_provider/src/data_source/file.rs
+++ b/crates/superposition_provider/src/data_source/file.rs
@@ -47,20 +47,36 @@ impl FileDataSource {
             watcher: Mutex::new(None),
         })
     }
-}
 
-#[async_trait]
-impl SuperpositionDataSource for FileDataSource {
-    async fn fetch_filtered_config(
+    async fn last_modified_at(&self) -> Result<DateTime<Utc>> {
+        let metadata = tokio::fs::metadata(&self.file_path).await.map_err(|e| {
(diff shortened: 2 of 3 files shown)


crates/superposition_provider/src/data_source.rs
```diff
@@ -107,6 +107,18 @@ pub trait SuperpositionDataSource: Send + Sync {
             .await
     }
 
+    /// Fetch the full resolved configuration only if the source changed.
+    ///
+    /// The default implementation delegates to `fetch_config`; data sources
+    /// with cheaper change detection can override this without changing normal
+    /// refresh behavior.
+    async fn fetch_config_if_modified(
```
ayushjain17, line 115: normal fetch config also did the exact same thing

crates/superposition_provider/src/data_source.rs
```diff
@@ -107,6 +107,18 @@ pub trait SuperpositionDataSource: Send + Sync {
             .await
     }
 
+    /// Fetch the full resolved configuration only if the source changed.
```
ayushjain17, line 110: this is definitely not resolved config

crates/superposition_provider/src/data_source/file.rs
```diff
@@ -47,20 +47,36 @@ impl FileDataSource {
...
+                self.file_path, e
+            ))
+        })
+    }
+
+    fn is_not_modified(
+        last_modified_at: DateTime<Utc>,
+        if_modified_since: Option<DateTime<Utc>>,
+    ) -> bool {
+        if_modified_since.is_some_and(|modified_since| last_modified_at <= modified_since)
```
ayushjain17, line 71: the names really dont seem to match with what is happening and based on the context of the adjoining code

crates/superposition_provider/src/data_source/file.rs
```diff
@@ -47,20 +47,36 @@ impl FileDataSource {
...
+        metadata.modified().map(DateTime::<Utc>::from).map_err(|e| {
+            SuperpositionError::ConfigError(format!(
+                "Failed to read modified time for config file {:?}: {}",
+                self.file_path, e
+            ))
+        })
+    }
+
+    fn is_not_modified(
+        last_modified_at: DateTime<Utc>,
```
ayushjain17, line 68: why decouple this ?
simply take `&self`

crates/superposition_provider/src/local_provider.rs
```diff
@@ -209,6 +209,10 @@ impl LocalResolutionProvider {
         self.do_refresh().await
     }
 
+    pub async fn refresh_if_changed(&self) -> Result<()> {
```
ayushjain17, line 212: what is the purpose of this function ? who is supposed to use this ? shouldn't this be a part of watch itself and controlled via some props of watch

sauraww:
Similar changes are already done in the provider PRs

Datron:
PR is:#1094

PR #1063, opened 2026-06-22
chore(docs): start pushing docs to S3 as well as github pages

Datron:
## Problem
We want to show documentation on superposition.juspay.io but can't reverse proxy from github without compromises

## Solution
Use our own hosting service which is s3

## Summary by CodeRabbit

* **Chores**
  * Enhanced documentation deployment process to support multiple deployment targets with environment-specific configurations.


sauraww:
Declined due to different approach

PR #1070, opened 2026-07-01
feat: harden function code by validating return values

Datron:
## Problem
Superposition does not validate function return types

## Solution
During compilation and checks, simplify and strengthen checks by running the `execute` function. This does the following:
- ensure `execute` exists
- ensure `execute` is a function
- ensure `execute` returns the correct type based on the function type

## Post-deployment activity
Maybe update blog post

Review these two files, the other changes were generated when I ran `make check`
- crates/context_aware_config/src/api/functions/handlers.rs
- crates/context_aware_config/src/validation_functions.rs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved function validation so checks now account for the function’s type, reducing incorrect validation results.
  * Updated function editing checks to validate against the current saved function state as well as the draft version.
  * Background compilation now uses the correct function context, which should make validation more reliable during create and update flows.

### DIFF
--- a/crates/context_aware_config/src/validation_functions.rs
+++ b/crates/context_aware_config/src/validation_functions.rs
@@ -7,8 +7,14 @@ use superposition_macros::{bad_argument, validation_error};
 use service_utils::service::types::{EncryptionKey, WorkspaceContext};
 use superposition_types::{
     DBConnection,
-    api::functions::{FunctionExecutionRequest, FunctionExecutionResponse},
-    database::models::cac::{FunctionCode, FunctionRuntimeVersion, FunctionType},
+    api::functions::{
+        ContextValidationTrigger, FunctionEnvironment, FunctionExecutionRequest,
+        FunctionExecutionResponse, KeyType,
+    },
+    database::models::{
+        ChangeReason,
+        cac::{FunctionCode, FunctionRuntimeVersion, FunctionType},
+    },
     result as superposition,
 };
 
@@ -73,11 +79,12 @@ pub fn run_js_function(
     args: FunctionExecutionRequest,
     runtime_version: FunctionRuntimeVersion,
 ) -> Result<FunctionExecutionResponse, (String, Option<String>)> {
-    let wrapped_code = generate_wrapped_code(&code.0);
+    let wrapped_code = generate_wrapped_code(&code);
     let module = Module::new("function.js", &wrapped_code);
 
     let runtime_options = RuntimeOptions {
         timeout: Duration::from_millis(1500),
+        // import_provider: Some(Box::new(DenyFileImportProvider::default())),
         ..Default::default()
     };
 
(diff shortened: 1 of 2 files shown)


crates/superposition/src/app_state.rs
sauraww, line 176: This change is not required.

crates/service_utils/tests/observability_integration.rs
sauraww, line 171: This file changes also looks only formatting change, Can we revert ?

sauraww (changes requested):

crates/context_aware_config/src/validation_functions.rs
```diff
@@ -73,11 +79,12 @@ pub fn run_js_function(
     args: FunctionExecutionRequest,
     runtime_version: FunctionRuntimeVersion,
 ) -> Result<FunctionExecutionResponse, (String, Option<String>)> {
-    let wrapped_code = generate_wrapped_code(&code.0);
+    let wrapped_code = generate_wrapped_code(&code);
     let module = Module::new("function.js", &wrapped_code);
 
     let runtime_options = RuntimeOptions {
         timeout: Duration::from_millis(1500),
+        // import_provider: Some(Box::new(DenyFileImportProvider::default())),
```
sauraww, line 87: can we remove this comment ?

crates/superposition_core/benches/resolve.rs
```diff
@@ -25,12 +25,12 @@ use std::cell::Cell;
...
-use criterion::{Criterion, black_box, criterion_group, criterion_main};
-use serde_json::{Map, Value, json};
-use superposition_core::{Config, MergeStrategy, eval_config};
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use serde_json::{json, Map, Value};
+use superposition_core::{eval_config, Config, MergeStrategy};
 use superposition_types::{
-    Cac, Condition, Context, DimensionInfo, ExtendedMap, OverrideWithKeys, Overrides,
     database::models::cac::{DependencyGraph, DimensionType},
+    Cac, Condition, Context, DimensionInfo, ExtendedMap, OverrideWithKeys, Overrides,
```
sauraww, line 33: These changes are unrelated right ? Looks like only formatting

sauraww:
Could we avoid executing `execute` with empty/default inputs as part of create/update validation?

A valid function might legitimately depend on fields such as `environment.context.country`, `key`, or `value`. With the current synthetic payload, such a function can throw even though it works correctly during real execution.

For example:

```javascript
function execute(payload) {
  return payload.environment.context.country.toUpperCase() === "IN";
}
```

This is valid for real inputs containing `country`, but validation currently passes an empty context, so it fails on `undefined.toUpperCase()` and rejects the function.

The return-type check itself is useful, but runtime failure against an invented empty payload should not necessarily make the function invalid. Could we either:

* validate the return type using a user-provided test payload, or
* avoid hard-failing when execution fails specifically because of the synthetic input?

Otherwise we may reject valid functions at create/update time.


This bug can be reproduced in the local by hitting this: 

```
curl -i \
  -X POST http://localhost:8081/function \
  -H 'Content-Type: application/json' \
  -H 'x-org-id: localorg' \
  -H 'x-workspace: dev' \
  --data-binary @- <<'JSON'
{
  "function_name": "country_required_repro",
  "function_type": "VALUE_VALIDATION",
  "runtime_version": "1.0",
  "description": "Reproduce context-dependent validation rejection",
  "change_reason": "Local bug reproduction",
  "function": "async function execute(payload) {\n  const { environment } = payload.value_validate;\n  return environment.context.country.toUpperCase() === \"IN\";\n}"
}
JSON
```
This is what it will return: 

`{"message":"Function validation failed: TypeError: Cannot read properties of undefined (reading 'toUpperCase')\n    at Module.execute (file:///Users/saurav.suman/sdk/visualize/type_check.js:41:38)"}%   `

sauraww (changes requested):

sauraww (approved):

mahatoankitkumar (approved):

PR #1079, opened 2026-07-04
perf(core/provider): reduce resolve-time cloning

sauraww:
## Problem

Warm config resolution was spending a large amount of time cloning data instead of evaluating it. In the provider/core resolve path, the cached `Config`, full `contexts` list, and `overrides` map could be cloned during every resolve, even when no prefix filtering was requested. This becomes expensive for large workspaces with tens or hundreds of thousands of context overrides.

## Solution

This PR removes clone-heavy work from the common warm resolve path.

- Added a borrowed fast path in `superposition_core::eval_config` and `eval_config_with_reasoning` when `filter_prefixes` is `None` or empty.
- Avoided constructing a temporary owned `Config` for the common non-prefix-filter path.
- Kept the existing owned `Config::filter_by_prefix` path for prefix-filtered resolves to preserve existing semantics.
- Changed override merging to merge by reference instead of cloning override maps into temporary `Value::Object`s.
- Cloned matched `Context` values only when a callback actually consumes them.
- Added `CacConfig::get_dimensions()` so the provider can clone only dimensions instead of the full cached config.
- Avoided fetching/cloning dimensions in provider evaluation when experimentation is disabled.
- Added a Criterion benchmark for the resolve hot path with:
  - 468,006 generated contexts
  - 18 regular dimensions
  - 18 derived local-cohort dimensions
  - 3 default configs
  - 100 sampled query contexts
- Added `PERF_ANALYSIS_RESOLUTION.md` documenting root cause, benchmark scope, results, and follow-up opportunities.

Local benchmark result on this machine:

```text
optimized_borrowed:      ~63 ms
pre_optimization_cloned: ~1.56 s
speedup:                 ~25x

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Performance Improvements**
  * Faster config resolution with less cloning and a direct fast path when no filtering is needed.
  * Reduced unnecessary work when experimentation is disabled.
  * Improved access to configuration metadata for lighter-weight reads.

* **Documentation**
  * Added performance analysis notes and benchmark results, including before/after comparisons and remaining optimization ideas.

* **Tests**
  * Added a benchmark to track resolution performance on production-like workloads.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/crates/superposition_core/benches/resolve.rs
+++ b/crates/superposition_core/benches/resolve.rs
@@ -1,0 +1,234 @@
+//! Resolution benchmark for the config evaluation hot path.
+//!
+//! This benchmark reproduces the shape of the workload reported from
+//! production-scale data (hundreds of thousands of context override rules,
+//! a handful of default configs, ~18 dimensions) and measures a single
+//! "resolve all config" call — the same operation the OpenFeature provider
+//! performs on every flag evaluation via `superposition_core::eval_config`.
+//!
+//! Two variants are compared:
+//!   * `optimized_borrowed`     — the current implementation, which resolves
+//!     directly against the borrowed context/override set.
+//!   * `pre_optimization_cloned` — reproduces the pre-optimization overhead,
+//!     where every resolve deep-cloned the entire context set (once to read
+//!     the dimensions, once more to build a throwaway `Config`).
+//!
+//! Run with:
+//!     cargo bench -p superposition_core --bench resolve
+//!
+//! The dataset size defaults to the production figure (468,006 contexts) and
+//! can be overridden for quicker local runs:
+//!     BENCH_CONTEXTS=50000 cargo bench -p superposition_core --bench resolve
+
+use std::cell::Cell;
+use std::collections::HashMap;
+use std::time::{Duration, Instant};
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use serde_json::{json, Map, Value};
+use superposition_core::{eval_config, Config, MergeStrategy};
+use superposition_types::{
+    Cac, Condition, Context, DimensionInfo, OverrideWithKeys, Overrides,
+};
+
+/// Number of dimensions in the synthetic dataset (mirrors the reported data).
+const NUM_DIMENSIONS: usize = 18;
+/// Default number of contexts / override rules (reported production figure).
+const DEFAULT_NUM_CONTEXTS: usize = 468_006;
+/// Number of sampled query contexts to resolve against.
+const NUM_QUERIES: usize = 100;
+/// Per-dimension value cardinality; controls how many contexts a query matches.
+const DIMENSION_CARDINALITY: usize = 12;
+/// The (small) set of default config keys.
+const DEFAULT_CONFIG_KEYS: [&str; 3] = ["config.alpha", "config.beta", "config.gamma"];
+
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: docs/misc/PERF_ANALYSIS_RESOLUTION.md
Comment (knutties) [inline docs/misc/PERF_ANALYSIS_RESOLUTION.md:1]: Can we find a better path for this under docs?

`docs/misc/` or some such to hold some adhoc MD files till we find a way to organize them. Eventually this can become a blog post pointing to this commit and the doc can be removed.

File: crates/superposition_core/benches/resolve.rs
Comment (knutties) [inline crates/superposition_core/benches/resolve.rs:1]: This does not add cohort computation as part of the benchmark yet I think.  It is ok - just clarifying since you said so on Slack.

**Comment by ayushjain17** [comment]
Can we take the previous PR into consideration altogether?

**Comment by sauraww** [comment]
@ayushjain17 what are you referring to ? What do you mean by previous here ?

File: crates/superposition_core/benches/resolve.rs
Comment (sauraww) [reply crates/superposition_core/benches/resolve.rs:1]: the benchmark does include local cohort evaluation as part of the measured eval_config path, because eval_config calls evaluate_local_cohorts before scanning contexts.

When I said local cohorts were not taken into consideration earlier - what I meant was initially, the benchmark did not meaningfully exercise local-cohort computation because dimensions were empty/no-op.

File: docs/misc/PERF_ANALYSIS_RESOLUTION.md
Comment (sauraww) [reply docs/misc/PERF_ANALYSIS_RESOLUTION.md:1]: Moved it.

File: crates/superposition_core/src/lib.rs [hunk 1]
```diff
@@ -1,4 +1,7 @@
 #![deny(unused_crate_dependencies)]
+#[cfg(test)]
+use criterion as _;
```
Datron, line 3: We're not using this?

crates/superposition_core/src/config.rs
```diff
@@ -97,70 +135,53 @@ pub fn merge(doc: &mut Value, patch: &Value) {
...
     overrides: &HashMap<String, Overrides>,
     merge_strategy: &MergeStrategy,
     mut on_override_select: Option<&mut dyn FnMut(Context)>,
 ) -> Result<Map<String, Value>, String> {
-    let mut required_overrides: Value = json!({});
-    let mut on_override_select = |context: Context| {
+    let mut required_overrides = Map::new();
+    // Only pay for a `Context` clone when a caller actually consumes it; the
+    // borrow keeps the common (callback-less) resolution path allocation-free.
+    let mut on_override_select = |context: &Context| {
```
Datron: this closure  shadows the name passed in the function args, its called only once and checks if it is a function and executes the closure. We could remove this and just add the code directly to the bottom where on_override_select is called, reducing one operation

crates/superposition_core/src/lib.rs
sauraww: It is used by the benches/resolve.rs benchmark. However, because superposition_core has #![deny(unused_crate_dependencies)], the lib test target still flags the dev-dependency as unused unless it is referenced from the crate root. This #[cfg(test)] use criterion as _; is just a lint marker for test builds, 
it doesn't affect normal library builds.

Datron (approved):

mahatoankitkumar (approved):

PR #1080, opened 2026-07-06
fix: treat same-value experiment ramp as a no-op instead of a 400

avi892nash:
## Problem

Ramping an experiment to its current traffic percentage returned `400` ("The traffic percentage is same as provided") — re-submitting the same value failed instead of being a no-op.

## Solution

- **Backend** (`PATCH /experiments/{id}/ramp`): if the new value equals the current non-zero value, return `200` with the experiment unchanged (no DB write, Redis update, or webhook).
- **Frontend**: show a warning toast "Ramp is already set to the same value" instead of a success message.

None.

## Pre-deployment activity

None.

## Post-deployment activity

None.

## API changes

No schema change — only the status code for the same-value case.

| Endpoint | Method | Request body | Response Body |
| `/experiments/{id}/ramp` | PATCH | `{ "traffic_percentage": <0-100>, "change_reason": "<string>" }` | `ExperimentResponse` — same value now returns `200` no-op instead of `400`. |

Clients that relied on the `400` to detect "no change" now get `200`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented redundant ramp updates when the traffic percentage is already set to the requested non-zero value.
  * The ramp form now warns the user and skips the request instead of showing a confusing in-progress state.
  * Repeated ramp submissions with no actual change now return the current experiment data immediately, avoiding unnecessary update processing.

### DIFF
--- a/crates/frontend/src/components/experiment_ramp_form.rs
+++ b/crates/frontend/src/components/experiment_ramp_form.rs
@@ -46,13 +46,27 @@ where
             .await;
             req_inprogress_ws.set(false);
             match result {
-                Ok(_) => {
-                    handle_submit_clone();
-                    enqueue_alert(
-                        String::from("Experiment ramped successfully!"),
-                        AlertType::Success,
-                        5000,
-                    );
+                Ok(experiment) => {
+                    // The API returns the (possibly unchanged) experiment. If its
+                    // traffic percentage still matches the pre-ramp non-zero value,
+                    // the backend treated this as a no-op: just warn and keep the
+                    // popup open (don't call handle_submit, which closes it).
+                    if *experiment.traffic_percentage
+                            == *experiment_clone.traffic_percentage
+                    {
+                        enqueue_alert(
+                            String::from("Ramp is already set to the same value"),
+                            AlertType::Warning,
+                            5000,
+                        );
+                    } else {
+                        handle_submit_clone();
+                        enqueue_alert(
+                            String::from("Experiment ramped successfully!"),
+                            AlertType::Success,
+                            5000,
+                        );
+                    }
                 }
                 Err(e) => {
                     enqueue_alert(e, AlertType::Error, 5000);
--- a/crates/superposition_types/src/database/models/experimentation.rs
+++ b/crates/superposition_types/src/database/models/experimentation.rs
@@ -208,11 +208,16 @@ impl TrafficPercentage {
         Ok(())
     }
 
-    pub fn compare_old(&self, old: &Self) -> Result<(), String> {
-        if self.0 != 0 && self.0 == old.0 {
-            return Err("The traffic percentage is same as provided")?;
-        }
-        Ok(())
+    /// Returns true when ramping to `self` would leave the traffic percentage
+    /// unchanged and non-zero, i.e. the ramp is a redundant no-op that can be
+    /// skipped.
+    ///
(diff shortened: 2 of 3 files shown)


crates/frontend/src/components/experiment_ramp_form.rs
```diff
@@ -36,6 +36,12 @@ where
             let workspace = workspace.get_untracked().0;
             let org = org.get_untracked().0;
             let traffic_value = traffic.get_untracked();
+            // Ramping to the same non-zero value is a no-op on the backend: the
+            // request still succeeds, so surface it as a warning instead of a
+            // success once it returns.
+            let is_noop =
```
mahatoankitkumar: I don't think we should keep this check in frontend too, backend check is enough

crates/superposition_types/src/database/models/experimentation.rs
```diff
@@ -208,11 +208,10 @@ impl TrafficPercentage {
...
 
-    pub fn compare_old(&self, old: &Self) -> Result<(), String> {
-        if self.0 != 0 && self.0 == old.0 {
-            return Err("The traffic percentage is same as provided")?;
-        }
-        Ok(())
+    /// Returns true when ramping to `self` would leave the traffic percentage
+    /// unchanged and non-zero, i.e. the ramp is a no-op that can be skipped.
+    pub fn is_noop_ramp(&self, old: &Self) -> bool {
+        self.0 != 0 && self.0 == old.0
```
mahatoankitkumar: should we keep  self.0 != 0, as if old is 0 and the new is also 0 then it should be a noop

mahatoankitkumar (dismissed):

crates/frontend/src/components/experiment_ramp_form.rs
avi892nash: yes, updating the PR, I meant to show proper UI for this.

crates/superposition_types/src/database/models/experimentation.rs
avi892nash: AI is saying that with 0 -> 0, we should keep behaviour of  CREATED → INPROGRESS.
I am not sure in which API can create Release with CREATED status, otherwise all Release is start with 50%.

So, I think we should keep this .

mahatoankitkumar: In that case we should check the state too for CREATED → INPROGRESS with 0 ramp.
Maybe not a part of this PR, would be helpful if you consider it

crates/frontend/src/components/experiment_ramp_form.rs
```diff
@@ -46,13 +46,28 @@ where
...
             req_inprogress_ws.set(false);
             match result {
-                Ok(_) => {
-                    handle_submit_clone();
-                    enqueue_alert(
-                        String::from("Experiment ramped successfully!"),
-                        AlertType::Success,
-                        5000,
-                    );
+                Ok(experiment) => {
```
mahatoankitkumar, line 49: I don't think any change needed in frontend, just show whatever came from backend for consistency

avi892nash: We have to show different popup on this, 
we were showing error popup, I changed this to show warning popup

Datron (approved):

sauraww (approved):

PR #1086, opened 2026-07-08
fix: prevent 413 errors because of large org auth cookies

Datron:
## Problem
For every organisation visited, we add an auth cookie for the realm. Eventually you get a 413 error due to a large cookie header value.

## Solution
Remove the oldest org cookie visited

## Summary by CodeRabbit

* **New Features**
  * Improved organisation sign-in handling by supporting multiple saved organisation cookies with automatic eviction of the oldest entries when the limit is reached.
  * Added better tracking of organisation selection across sessions for smoother switching.

* **Bug Fixes**
  * Updated organisation login redirects to keep behavior consistent while managing cookies more reliably.
  * Prevents cookie buildup by removing excess organisation cookies automatically.

### DIFF
--- a/crates/service_utils/src/middlewares/auth_n/authentication.rs
+++ b/crates/service_utils/src/middlewares/auth_n/authentication.rs
@@ -1,7 +1,7 @@
 use std::fmt::Display;
 
 use actix_web::{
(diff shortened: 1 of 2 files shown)


sauraww:
@Datron Can we decline this PR?

PR #1094, opened 2026-07-14
fix(openfeature): Rust provider improvements

ayushjain17:
## Change log
Improvements and fixes in rust provider to make it more consistent with openfeature spec and other languages in parallel

## Summary by CodeRabbit

* **New Features**
  * Added typed array flag resolution for OpenFeature integrations.
  * Added millisecond-based polling and on-demand refresh configuration.
  * Providers now track refresh health and recover status after successful refreshes.

* **Bug Fixes**
  * Null fields in object-valued flags are ignored instead of causing failures.
  * Invalid top-level flag types now return clearer type-mismatch errors.
  * Missing targeting keys no longer prevent applicable-variant evaluation.

* **Documentation**
  * Improved guidance for typed flag-resolution errors.

### DIFF
--- a/crates/superposition_provider/src/data_source/file.rs
+++ b/crates/superposition_provider/src/data_source/file.rs
@@ -48,6 +48,30 @@ impl FileDataSource {
             watcher: Mutex::new(None),
         })
     }
+
+    async fn last_modified_at(&self) -> Result<DateTime<Utc>> {
+        tokio::fs::metadata(&self.file_path)
+            .await
+            .map_err(|e| {
+                SuperpositionError::DataSourceError(format!(
+                    "Failed to read metadata for config file {:?}: {}",
+                    self.file_path, e
+                ))
+            })?
+            .modified()
+            .map(DateTime::<Utc>::from)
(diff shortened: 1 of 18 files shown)


crates/superposition_provider/src/data_source/file.rs
```diff
@@ -48,6 +48,30 @@ impl FileDataSource {
             watcher: Mutex::new(None),
         })
     }
+
+    async fn last_modified_at(&self) -> Result<DateTime<Utc>> {
```
Datron, line 52: Should these functions `last_modified_at` and `is_not_modified` be implemented in http data source as well to keep the interfaces similar? Just a simple implementation that returns last_modified

ayushjain17: Data source does not have a last modified, it's the local provider which has the required thing
This was a specific requirement for FileDataSource so that it matches spec with HttpDataSource

sauraww (approved):

Datron (approved):

sauraww (approved):

PR #1100, opened 2026-07-17
feat: Introduce Job Manager to track long running tasks

Datron:
## Problem
https://github.com/juspay/superposition/discussions/1093

## Solution
Introduce Background Job Manager

## Summary by CodeRabbit

* **New Features**
  * Added background job management for webhook delivery, priority recomputation, and configuration reduction.
  * Added APIs to submit, list, view, monitor, and cancel jobs.
  * Job details now include status, progress, execution information, and logs.
  * Configuration reduction and priority recomputation now run asynchronously, returning a job identifier.
* **Bug Fixes**
  * Improved dispatch authentication and workspace job tracking.
* **Documentation**
  * Updated service contracts to include background job operations.


crates/superposition_types/src/database/schema.rs
```diff
@@ -100,539 +100,6 @@ diesel::table! {
     }
 }
```
sauraww, line 101: why did we delete these partitions ?

crates/service_utils/src/kronos_dispatch.rs
```diff
@@ -125,6 +137,45 @@ pub async fn setup_dispatcher(
     }
 }
 
+pub fn job_dispatcher_endpoint_spec(dispatcher_url: &str) -> serde_json::Value {
```
sauraww, line 140: Do we really need a function for this ? Can this be inlined ? Also this is being used at one place.

crates/superposition/src/dispatch/handlers.rs
```diff
@@ -43,26 +54,100 @@ async fn dispatch_handler(
...
+    } = body.into_inner();
+
+    let conn = write_permit.connection();
+    update_job_status(conn, job_id, BackgroundJobStatus::Inprogress)
+        .map_err(|e| unexpected_error!("Failed to update job status: {}", e))?;
+    let key = append_job_logs(conn, job_id, "Job started", None)
+        .map_err(|e| unexpected_error!("Failed to append logs: {}", e))?;
+
+    let result = match &job_request {
+        JobRequest::Webhook(req) => {
```
sauraww, line 85: This is unused currently. Also architecturally we have two paths to webhook execution. Currently , execute_webhook_dispatch() is being used.

crates/superposition/src/dispatch/handlers.rs
```diff
@@ -43,26 +54,100 @@ async fn dispatch_handler(
...
-        fetch_webhook(&webhook_name, &workspace_context.schema_name, &mut conn)?;
+
+    execute_webhook_dispatch(&workspace_context, &state, &mut conn, &webhook_name, &data)
+        .await?;
+    Ok(HttpResponse::Ok().finish())
+}
+
+#[authorized]
+#[post("/job")]
+async fn dispatch_job_handler(
```
sauraww, line 66: This handler acquires WorkspaceWritePermit without configuring WorkspaceLockTtlPolicy::Batch, so long-running reduce/recompute jobs use the 60-second default lease. Can we apply the batch policy to this route or renew the lease during execution?

crates/service_utils/src/kronos_dispatch.rs
```diff
@@ -197,3 +248,275 @@ pub async fn submit_webhook_job<T: Serialize>(
...
+        workspace_schema: workspace.clone(),
+        created_at: Utc::now(),
+        logs: json!({}),
+    };
+
+    diesel::insert_into(job_manager_dsl::job_manager)
+        .values(&bjm_entry)
+        .execute(conn)?;
+
+    let job_request_value = serde_json::to_value(job_request)?;
```
sauraww: Can we use a type here ? 
KronosJobRequest ?

sauraww (changes requested):

crates/superposition/src/dispatch/handlers.rs
```diff
@@ -43,26 +54,101 @@ async fn dispatch_handler(
...
+        Err(e) => {
+            let error_msg = format!("{e}");
+            log::error!("Job {job_id} failed: {error_msg}");
+            update_job_status(&mut conn, job_id, BackgroundJobStatus::Failed)
+                .map_err(|e| unexpected_error!("Failed to update job status: {}", e))?;
+            append_job_logs(&mut conn, job_id, &format!("Job failed: {error_msg}"), Some(key))
+                .map_err(|e| unexpected_error!("Failed to append logs: {}", e))?;
+            Err(unexpected_error!("Job {job_id} failed: {error_msg}"))
+        }
+    }
```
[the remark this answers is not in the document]
sauraww: Verified with a serialize/deserialize round-trip using the project’s Serde version. The flattened adjacently tagged JobRequest successfully deserializes from {job_id, job_type, job_data}, so the callback reaches the enum match. Could we add this round-trip as a unit test and resolve this finding?
It would be worth adding the round trip as it permanently proves the kronos callback contract.

crates/service_utils/src/kronos_dispatch.rs
Datron: Makes it easy to quickly check and analyze kronos endpoint config, in case we want to validate some behaviour in the future

crates/service_utils/src/kronos_dispatch.rs
Datron: Let me check

crates/superposition/src/dispatch/handlers.rs
Datron: Hmm our proxy timeout is much lower than this - so when the connection terminates the job does as well. This should be handled via async jobs/long running ones. Thanks for pointing this out, will implement the lease renewal logic for async jobs

crates/superposition/src/dispatch/handlers.rs
Datron: I've kept `execute_webhook_dispatch` for backwards compat, we can deprecate it in the following PRs

crates/service_utils/src/kronos_dispatch.rs
Datron: Done, thanks for bringing this up

crates/service_utils/src/kronos_dispatch.rs
sauraww: Can we inline it for now? 
If there are more use cases , we can take it out.

Datron: This is done for readability of kronos endpoint config. Its not about re-usability.

sauraww:
The user-facing API is asynchronous, but /dispatch/job executes the complete long-running operation synchronously while holding an HTTP request, DB connection, and workspace lease. 
How are callback timeouts, duplicate retries, cancellation, and maximum execution duration handled?

crates/context_aware_config/src/api/config/handlers.rs
```diff
@@ -469,24 +463,58 @@ async fn reduce_handler(
...
         if is_approve {
             config = generate_cac(conn, &workspace_context.schema_name)?;
         }
     }
+    Ok(())
+}
+
+#[authorized]
+#[put("/reduce")]
+async fn reduce_handler(
```
sauraww, line 487: Should we consider locking while doing reduce ?

crates/service_utils/src/kronos_dispatch.rs
```diff
@@ -197,3 +248,274 @@ pub async fn submit_webhook_job<T: Serialize>(
...
+            job_manager_dsl::job_manager.filter(job_manager_dsl::id.eq(job_id)),
+        )
+        .set(job_manager_dsl::progress.eq(progress))
+        .execute(conn)?;
+    }
+    Ok(())
+}
+
+pub fn append_job_logs(
+    conn: &mut DBConnection,
```
sauraww, line 498: This function uses a non-atomic read-modify-write and reusing the same key overwrites the start entry. 
Can this be changed to an atomic JSONB append/update?

crates/service_utils/src/kronos_dispatch.rs
```diff
@@ -197,3 +248,274 @@ pub async fn submit_webhook_job<T: Serialize>(
...
+    job_id: i64,
+    log_line: &str,
+    key: Option<String>,
+) -> anyhow::Result<String> {
```
sauraww, line 502: Also if we want to really append logs , this key should not be passed rather created inside the function.

crates/service_utils/src/kronos_dispatch.rs
```diff
@@ -197,3 +248,274 @@ pub async fn submit_webhook_job<T: Serialize>(
...
+        diesel::update(
+            job_manager_dsl::job_manager.filter(job_manager_dsl::id.eq(job_id)),
+        )
+        .set(job_manager_dsl::progress.eq(progress))
+        .execute(conn)?;
+    }
+    Ok(())
+}
+
+pub fn append_job_logs(
```
sauraww, line 497: The function name append_job_logs is misleading. 
It is actually upserting.

sauraww (changes requested):

crates/superposition/src/dispatch/handlers.rs
```diff
@@ -43,26 +54,100 @@ async fn dispatch_handler(
...
+        }
+    };
+
+    match result {
+        Ok(()) => {
+            update_job_status(conn, job_id, BackgroundJobStatus::Completed)
+                .map_err(|e| unexpected_error!("Failed to update job status: {}", e))?;
+            update_job_progress(conn, job_id, 100)
+                .map_err(|e| unexpected_error!("Failed to update job progress: {}", e))?;
+            append_job_logs(conn, job_id, "Job completed", Some(key))
```
sauraww, line 109: This is the core bug , the timestamp for job completed is wrong. It is that of job started.

sauraww (changes requested):

crates/service_utils/src/kronos_dispatch.rs
Datron: It does both, if you don't pass a key e.g for the first log line, then it generates one. You can use that key for the rest of the execution so you can see all logs for an execution that started at that time.

crates/service_utils/src/kronos_dispatch.rs
Datron: Why is it non-atomic?

crates/service_utils/src/kronos_dispatch.rs
Datron: append to the JSONB column was my meaning

crates/superposition/src/dispatch/handlers.rs
Datron: I want to group log lines by the time the job started. You can use that key for the rest of the execution so you can see all logs for an execution that started at that time.

sauraww: @Datron Can you run and show the logs ? Because the above code is actually replacing Job started.

Datron: Ah got the issue @sauraww , fixing

Datron: @sauraww this is fixed, here is a same DB entry
```
-[ RECORD 2 ]----+-------------------------------------------------------------------
id               | 7492553124826189824
kronos_job_id    | 7a0819d2-e248-41d2-882a-cedf8a64c61f
description      | Priority recompute job
job_type         | PRIORITY_RECOMPUTE
status           | COMPLETED
name             | priority_recompute
progress         | 100
workspace_schema | orgid162145664241766405_kronostest
created_at       | 2026-08-10 12:11:09.864395+00
logs             | {"2026-08-10T12:11:10.588189+00:00": "Job started\nJob completed"}
```

crates/superposition_types/migrations/2026-07-16-000001_job_manager/up.sql
```diff
@@ -0,0 +1,52 @@
+DO $$ BEGIN
```
sauraww: Do we need this? 
Can we keep following the existing convention: 
CREATE TYPE public.background_job_type AS ENUM (...);
Also , can we fix down.sql ?

crates/superposition_types/src/database/schema.rs
sauraww: @Datron Why did we delete ?

sauraww (changes requested):

crates/superposition_types/src/database/schema.rs
Datron: @sauraww these are redundant - to manage the partitions we directly manage the table

crates/superposition_types/src/api/jobs.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct JobDispatchRequest {
+    #[serde(with = "crate::database::models::i64_formatter")]
+    pub job_id: i64,
+    #[serde(flatten)]
+    pub job_request: JobRequest,
+}
```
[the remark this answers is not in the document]
Datron: This is not required

PR #1114, opened 2026-07-21
docs(kronos): add deployment documentation

sauraww:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guidance for configuring Kronos durable webhook delivery.
  * Documented embedded and standalone service deployment modes, required settings, credentials, encryption keys, and database configuration.
  * Added Kronos environment variables and production secret-handling details.
  * Added the Kronos setup guide to the self-hosting documentation navigation.

### DIFF
--- a/docs/docs/self-hosting/kronos.md
+++ b/docs/docs/self-hosting/kronos.md
@@ -1,0 +1,82 @@
+---
+title: Kronos
+description: Choose embedded library mode or a standalone Kronos service for durable webhook delivery.
+---
+
+# Kronos
+
+Superposition uses [Kronos](https://github.com/juspay/kronos) to deliver
+webhooks durably and retry failed deliveries. You can run it as an embedded
+library or as a separate service.
+
+| Mode | Best for | How it runs |
+| --- | --- | --- |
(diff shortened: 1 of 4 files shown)


docs/docs/self-hosting/kronos.md
```diff
@@ -0,0 +1,81 @@
...
+KRONOS_TABLE_PREFIX=kronos_
+```
+
+Generate the encryption key with `openssl rand -hex 32`. Keep this key stable
+and secret, and use the same value on every Superposition replica. The pool
+size is per replica; its default is `1`.
+
+The embedded worker uses the existing Superposition database connection
+settings. Kronos tables are created in each workspace schema with the
+`kronos_` prefix by default.
```
knutties: Do we need to mention this is configurable ?

sauraww: added

knutties (approved):

Datron (approved):

PR #1115, opened 2026-07-21
feat: add exclude-prefix filter for resolve api

sauraww:
## Problem

The existing `prefix` filter acts only as an allow-list. Callers cannot retrieve resolved configuration while excluding keys belonging to specific prefixes.

This affects both the Resolve API and the Detailed Resolve API.

## Solution

Added an optional, repeatable `exclude_prefix` query parameter.

- Without `prefix`, all resolved keys are returned except those matching `exclude_prefix`.
- When both filters are provided, the result is the `prefix` allow-list minus keys matching `ignore_prefix`.
- Matching default and override keys are removed.
- Contexts whose overrides become empty after filtering are pruned.
- Existing behavior remains unchanged when `ignore_prefix` is absent.

The Smithy model, OpenAPI documentation, and generated Rust, TypeScript, Python, Java, and Haskell SDKs have also been updated.

None.

## Pre-deployment activity

None.

## Post-deployment activity

None.

Consumers wanting to use this feature should upgrade to an SDK version containing the new `ignore_prefix` request field.

## API changes

| Endpoint | Method | Request | Response |
|---|---|---|---|
| `/config/resolve` | GET / POST | Added optional repeatable `exclude_prefix` query parameter | Existing resolved-config response with matching keys excluded |
| `/config/resolve/detailed` | GET / POST | Added optional repeatable `exclude_prefix` query parameter | Existing detailed resolved-config response with matching keys excluded |

### Examples

Exclude multiple prefixes:

    /config/resolve?ignore_prefix=internal.&ignore_prefix=debug.

Combine it with the existing allow-list:

    /config/resolve?prefix=payments.&ignore_prefix=payments.private.

This returns keys beginning with `payments.`, excluding keys beginning with `payments.private.`.

The request body and response schemas remain unchanged.

## Validation

- Added tests for multiple ignored prefixes.
- Added tests for non-matching and empty filter lists.
- Added tests for combined allow-list and ignore-prefix behavior.
- Verified the server and dependent Rust crates compile.
- Verified the generated Rust and TypeScript SDKs build.
- Verified the generated Python SDK syntax.
- Successfully generated the Smithy model and OpenAPI documentation.


smithy/models/config.smithy
```diff
@@ -295,6 +300,11 @@ operation GetDetailedResolvedConfig {
         @notProperty
         prefix: StringList
 
+        @documentation("Excludes configuration keys that start with any of the supplied prefixes. When combined with prefix, exclusion is applied to the allow-listed keys.")
+        @httpQuery("ignore_prefix")
+        @notProperty
+        ignore_prefix: StringList
```
ayushjain17: if we are adding something like this, then it's an incomplete feature if we dont expand it the plain get config, list experiments, get experiment config, list default configs, list contexts

features on key level filtering specially like prefix should expand to all the usecases

sauraww: And also to providers , i was thinking to do that in the second phase. 
I will close the api coverage in this PR , we can do the provider changes in the next PR. 
Didn't want to make this PR big.

ayushjain17 (approved):

Datron (approved):

PR #1121, opened 2026-07-29
feat: add retry for lock in frontend

sauraww:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of temporarily locked workspaces by automatically retrying affected requests in the browser.
  * Added progressive delays between retry attempts to improve request reliability.
  * Requests now provide a clear error when all retry attempts are exhausted.
  * Preserved existing error alerts and handling for other request failures.

### DIFF
--- a/crates/frontend/src/utils.rs
+++ b/crates/frontend/src/utils.rs
@@ -1,13 +1,14 @@
 #[cfg(feature = "ssr")]
 use std::env;
-use std::str::FromStr;
+use std::{str::FromStr, time::Duration};
 
+use leptos::leptos_dom::is_browser;
 use leptos::*;
 use reqwest::{
     StatusCode,
     header::{HeaderMap, HeaderName, HeaderValue},
 };
-use serde::de::DeserializeOwned;
+use serde::{Deserialize, de::DeserializeOwned};
 use superposition_types::{
     api::functions::{FunctionEnvironment, KeyType},
     database::models::cac::{Function, FunctionType},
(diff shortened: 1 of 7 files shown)


crates/frontend/src/utils.rs
```diff
@@ -199,6 +200,63 @@ pub fn check_url_and_return_val(s: String) -> String {
...
+}
+
+#[derive(Deserialize)]
+struct RequestErrorResponse {
+    message: String,
+    #[serde(default)]
+    lock: Option<serde_json::Value>,
+}
+
+async fn sleep(duration: Duration) {
```
Datron, line 249: Can't a native JS or tokio module do this?

crates/frontend/src/utils.rs
```diff
@@ -301,7 +400,19 @@ pub async fn request<T>(
...
+    url: String,
+    method: reqwest::Method,
+    body: Option<T>,
+    headers: HeaderMap,
+) -> Result<reqwest::Response, String>
+where
+    T: serde::Serialize,
+{
+    request_with_policy(url, method, body, headers, &[], RetryPolicy::WorkspaceLock).await
 }
```
Datron, line 416: Do we need these wrapper functions? Can they be made generic somehow - it is just different ways to call another function

crates/frontend/src/utils.rs
```diff
@@ -199,6 +200,63 @@ pub fn check_url_and_return_val(s: String) -> String {
...
+        match self {
+            Self::WorkspaceLock => WORKSPACE_LOCK_RETRY_BASE_DELAY
+                .saturating_mul(2_u32.saturating_pow(retry)),
+            Self::None => Duration::ZERO,
+        }
+    }
+}
+
+#[derive(Deserialize)]
+struct RequestErrorResponse {
```
Datron: Would WorkspaceLockedErrorResponse be better for this?

Datron (commented):
Can you attach a video of a demo

crates/frontend/src/utils.rs
sauraww: Making callers pass RetryPolicy directly would save a few wrapper lines but add policy arguments everywhere and expose internal behavior.

sauraww: request() → normal one shot request.
request_with_skip_error() → one shot with accepted error statuses.
request_with_workspace_lock_retry() → explicit lock retry.

mahatoankitkumar (approved):

Datron (approved):

PR #1136, opened 2026-08-25
Adding filtering logic for specific keys within override

vraghunandhan:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

- **New Features**
  - Context overrides can now be searched by matching keys and values.
  - Search filtering is case-insensitive and ignores surrounding whitespace.
  - Blank searches continue to display all available overrides.

- **Documentation**
  - Updated search guidance to clarify that both override keys and values are matched.

### DIFF
--- a/crates/frontend/src/pages/context_override.rs
+++ b/crates/frontend/src/pages/context_override.rs
@@ -25,7 +25,10 @@ use superposition_types::{
         experimentation::ExperimentType,
     },
 };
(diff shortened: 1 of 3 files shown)


ayushjain17 (changes requested):
This was deliberately not allowed previously

PR #1140, opened 2026-08-31
feat: merge strategy on context put

mahatoankitkumar:
## Problem

`PUT /context` on a context that already exists always deep-merges the incoming
overrides into the stored ones. There is no way to replace a key wholesale.

## Solution

`PUT /context` and bulk PUT now accept **`x-merge-strategy`** — the same header and
enum resolve already uses. Defaults to `MERGE`, so existing behaviour is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added support for choosing how existing context overrides are handled during create and bulk operations.
  * Use the merge strategy header to either deep-merge new values with existing overrides or replace them entirely.
  * Merging is the default behavior; the setting has no effect when creating a new context.

* **Bug Fixes**
  * Existing override values are now preserved when merging, while supplied values are updated and validated.

### DIFF
--- a/crates/context_aware_config/src/api/context/handlers.rs
+++ b/crates/context_aware_config/src/api/context/handlers.rs
@@ -2,7 +2,7 @@ use std::{cmp::min, collections::HashMap};
 
 use actix_web::{
     Either, HttpResponse, Scope, delete, get, post, put, routes,
-    web::{Data, Json, Path},
+    web::{Data, Header, Json, Path},
 };
 use bigdecimal::BigDecimal;
 use chrono::Utc;
@@ -31,6 +31,7 @@ use superposition_types::{
     Overridden, Overrides, PaginatedResponse, PrefixList, Resource, SortBy, User,
     api::{
         DimensionMatchStrategy,
+        config::MergeStrategy,
         context::{
             BulkOperation, BulkOperationResponse, ContextAction, ContextBulkResponse,
             ContextListFilters, ContextValidationRequest, Identifier, MoveRequest,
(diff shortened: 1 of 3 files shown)


ayushjain17 (commented):
@mahatoankitkumar  why are we doing this?
This behaviour was deliberately kept like this

When some is calling PUT /context
There is high chance that the person is not aware about existing overrides
Providing an option to take a blanket call like this is unsafe
People will be unknowingly deleting a lot of overrides

If someone wants to update a value they can very much call PUT /context that is fine
But they want explicit behaviour to deleting keys from an override 
They should firstly do it via delete via experiment if still not then, they should absolutely use /context/overrides where they are explicitly giving their intention that they are aware what the current contents are and are willing to update that set of keys specifically
Be is updating values or deleting values

mahatoankitkumar:
@ayushjain17 it's not about root level keys here
It's about deep merging or not, the same way we follow for resolve, I know the header name can confuse someone, maybe we can name it better.
Rather right now the behaviour is silent about deep merging keys
The user doesn't know about it

ayushjain17:
@mahatoankitkumar 
More than the header name I got confused with the description of the PR
This is a good call
But instead of taking a header for this, shouldn't it be `REPLACE` as the only behaviour?

mahatoankitkumar:
Not sure about replace being the only choice
Maybe some user may expect it to deep merge?

ayushjain17:
I think the actual problem might be that the end users are not even aware that for resolve also, by default it is deep merge

mahatoankitkumar:
We rather expose it to our superposition dashboard users at least
And make every merge strategy default to replace, replace sounds more intuitive in both the resolve and context put use cases
What do you think @juspay/sdk-backend

PR #1141, opened 2026-08-31
fix: free Rust-owned FFI strings to stop native memory leak

gyash1512:
## Problem
The JavaScript client's native resolver leaks native (Rust-owned) memory on **every** FFI call that returns a string. The five string-returning functions were declared with a `char*` return type, so koffi auto-decodes the return into a JS string (a copy) and discards the original pointer. That makes `typeof result === "string"` always true, which means the `core_free_string(result)` calls are **dead code** and the Rust `CString` (`CString::into_raw`) is never freed.

Every evaluation leaks the full returned string (for `core_get_resolved_config`, that's the entire resolved config, ~KBs). In a long-running, high-throughput consumer this grows the process's native heap linearly until it OOMs. Confirmed with jemalloc heap profiling (`jeprof` diff): `alloc::ffi::c_str::CString::_from_vec_unchecked`, reached via `koffi → libsuperposition_core`, accounted for 90%+ of native heap growth. Present in every published version through the latest.

## Solution
Use koffi's **disposable type** — the documented idiom for freeing heap-allocated strings returned from C. Register a named type `OwnedStr` derived from `str` with `core_free_string` as its free callback, then declare the five owned-string functions to return `OwnedStr`:

```js
this.lib.core_free_string = this.lib.func("void core_free_string(char*)");
if (!ownedStrTypeRegistered) {
    koffi.disposable("OwnedStr", "str", this.lib.core_free_string);
    ownedStrTypeRegistered = true;
}
this.lib.core_get_resolved_config = this.lib.func("OwnedStr core_get_resolved_config(...)");
```

koffi converts the returned `char*` to a JS string and then automatically calls `core_free_string` on the **original** Rust pointer, so the native allocation is freed instead of leaked.

Changes in `clients/javascript/bindings/native-resolver.ts`:
- Declare `core_get_resolved_config`, `core_get_applicable_variants`, `core_parse_toml_config`, `core_parse_json_config`, `core_provider_cache_eval_config` to return `OwnedStr` (was `char*`).
- Register the `OwnedStr` disposable type once via a module-level `ownedStrTypeRegistered` guard (koffi type names are process-global, so registering per-instance throws `Duplicate type name`).
- Remove the now-unnecessary manual decode/free branches in the function bodies — the return value is a plain string.
- `core_free_string`'s own `char*` **argument** type is unchanged (it's an input pointer, correct as-is).

No behavior change to return values — callers still receive the same string/parsed object; the pointer is now freed instead of leaked.

None.

## Pre-deployment activity
- Rebuild the JS client workspaces so `dist/` reflects the source change: `npm run build -w bindings -w sdk -w open-feature-provider`.
- Publish a new `superposition-provider` version (this ships in `dist/index.esm.js` / `dist/index.js`).
- Consumers must bump to the new version (or apply it via `pnpm patch` until they can upgrade).

## Post-deployment activity
- On a consumer running the native path, watch RSS/native heap over a few hours — it should stay flat instead of climbing to the memory limit; no OOM/restarts.
- (Optional) Verify with a jemalloc `jeprof` diff that `CString::_from_vec_unchecked` no longer dominates growth.

## API changes
None — no public API, request, or response shape changes. (FFI return types are internal to the binding.)

| Endpoint | Method | Request body | Response Body |
| N/A | N/A | N/A | N/A |

- **koffi dependency:** the fix relies on koffi's `disposable` type calling the free callback on the original pointer after C→JS conversion. Pin/verify koffi behavior on major upgrades.
- **Ownership contract:** the disposable calls `core_free_string` on every returned pointer, which assumes the Rust side always returns an owned `CString::into_raw` pointer. If a function is ever changed to return a borrowed/static pointer, this would become an invalid free — keep the `OwnedStr` return paired with `into_raw`-style ownership.
- **Debug logging (pre-existing, not addressed here):** these methods still `console.log` per call (including result/param dumps). Not a leak, but noisy/costly at high volume — worth removing in a follow-up.

### DIFF
--- a/clients/javascript/bindings/native-resolver.ts
+++ b/clients/javascript/bindings/native-resolver.ts
@@ -10,6 +10,7 @@ const ERROR_BUFFER_SIZE = 2048;
 export interface ResolverLogger {
     debug(...args: unknown[]): void;
 }
+let ownedStrTypeRegistered = false;
 
 export class NativeResolver {
     private lib: any;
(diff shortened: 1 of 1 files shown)


ayushjain17:
@gyash1512 tests are failing with segmentation fault

```
> superposition-bindings@0.0.1 build
> tsc


======================================================================
  SUPERPOSITION TOML PARSING - JAVASCRIPT/TYPESCRIPT BINDING TESTS
======================================================================

======================================================================
  TEST 1: Parse TOML Configuration
======================================================================
Using native library from package: /home/runner/_work/superposition/superposition/clients/javascript/bindings/dist/native-lib/libsuperposition_core-x86_64-unknown-linux-gnu.so
bash: line 1: 43035 Segmentation fault      (core dumped) node dist/test-toml.js
make[1]: *** [makefile:435: bindings-test] Error 139
make[1]: Leaving directory '/home/runner/_work/superposition/superposition'
make: *** [makefile:242: test] Error 2
```

ayushjain17 (approved):

ayushjain17:
can you please update the description with the latest changes

sauraww:
Possible to resolve the conflicts ?

gyash1512:
> Possible to resolve the conflicts ?

Resolved @sauraww

sauraww (approved):

PR #1144, opened 2026-09-01
Feat/merge strategy workspace

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes


superposition.sql
```diff
@@ -144,6 +144,15 @@ ADD COLUMN IF NOT EXISTS enable_change_reason_validation BOOLEAN DEFAULT FALSE;
...
 
+DO $$ BEGIN
+    CREATE TYPE superposition.merge_strategy AS ENUM ('MERGE', 'REPLACE');
+EXCEPTION
+    WHEN duplicate_object THEN null;
+END $$;
+
+ALTER TABLE superposition.workspaces
+ADD COLUMN IF NOT EXISTS merge_strategy superposition.MERGE_STRATEGY NOT NULL DEFAULT 'MERGE';
+
```
Datron: nit: can we add this at the end, makes it obvious these are new queries

PR #1147, opened 2026-09-10
feat: add evaluation cache support in all providers

Datron:
## Problem
Queries with the same context, prefix, or args passed to the resolution function cause un-necessary computations when a query is the same and a previous result for computed/resolved.

## Solution

Add a LRU cache called evaluation cache that is cleared every time the config in the provider is refreshed. The cache by default is turned off and to enable it you pass an argument mentioning the max entries you want cached.

The remote provider would benefit a lot from evaluation cache, but today their is no way to no when the config is refreshed on the server. The way we could do this in the future:
- send the header if-modified-since to the server on a parallel thread and if the server responds 304, do not clear the evaluation cache for a particular key
- if the server responds 200, clear all keys so that the new value can be stored

## Summary by CodeRabbit

- **New Features**
  - Added an optional in-process LRU evaluation cache for local providers across supported language SDKs.
  - Configure caching with a maximum entry count; unset or non-positive values disable it.
  - Cached evaluations are reused for identical resolution inputs and cleared when configuration or experiment data reloads.
  - Added provider APIs and examples for enabling evaluation caching.

- **Documentation**
  - Updated provider documentation with configuration guidance, limitations, and examples.

- **Breaking Changes**
  - Replaced legacy `ttl` and `size` settings with `maxEntries`/`max_entries`.

### DIFF
--- a/clients/javascript/bindings/native-resolver.ts
+++ b/clients/javascript/bindings/native-resolver.ts
@@ -37,6 +37,9 @@ export class NativeResolver {
             this.lib.core_provider_cache_new = this.lib.func(
                 "void* core_provider_cache_new()",
             );
+            this.lib.core_provider_cache_new_with_eval_cache = this.lib.func(
+                "void* core_provider_cache_new_with_eval_cache(uint64)",
+            );
             this.lib.core_provider_cache_free = this.lib.func(
                 "void core_provider_cache_free(void*)",
             );
@@ -541,11 +544,16 @@ export class NativeResolver {
         }
     }
 
-    createProviderCache() {
+    createProviderCache(evaluationCacheSizeMb?: number) {
         if (!this.isAvailable) {
             throw new Error("Native resolver is not available.");
         }
-        const handle = this.lib.core_provider_cache_new();
+        const handle =
+            evaluationCacheSizeMb && evaluationCacheSizeMb > 0
+                ? this.lib.core_provider_cache_new_with_eval_cache(
+                      evaluationCacheSizeMb,
+                  )
+                : this.lib.core_provider_cache_new();
         if (!handle) {
             throw new Error("core_provider_cache_new returned null");
         }
--- a/clients/javascript/bindings/native-resolver.ts
+++ b/clients/javascript/bindings/native-resolver.ts
@@ -544,14 +544,14 @@ export class NativeResolver {
         }
     }
 
-    createProviderCache(evaluationCacheSizeMb?: number) {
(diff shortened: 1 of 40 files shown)


sauraww:
Can we add a test for proving that a 304/NotModified refresh preserves evaluation-cache entries ?

crates/superposition_core/src/ffi.rs
```diff
@@ -250,13 +251,30 @@ fn ffi_parse_config_file_with_filters(
...
+    /// caching.
+    pub(crate) evaluation_cache: Option<EvalCache>,
 }
 
+/// FFI-owned provider state: the config and experiment data loaded from
+/// superposition that resolutions read from, plus a memoization cache for
+/// repeated `eval_config` queries.
 #[derive(uniffi::Object, Default)]
 pub struct ProviderCache {
     pub(crate) data: Mutex<CacheData>,
```
sauraww: The evaluation cache shares the same mutex as config and experiment data. This means concurrent evaluations can block each other, even on cache hits. Can we use a separate lock for the evaluation cache to avoid slowing down the hot path?
Basically it should not contend with experiment and config.

sauraww (changes requested):

sauraww:
There are few coderabbit comments as well. @Datron Can you address them ?

sauraww:
@coderabbitai full review

crates/superposition_core/src/ffi.rs
Datron: Great point, will replace this with RwLock instead

clients/javascript/bindings/native-resolver.ts
```diff
@@ -541,11 +544,16 @@ export class NativeResolver {
...
         if (!this.isAvailable) {
             throw new Error("Native resolver is not available.");
         }
-        const handle = this.lib.core_provider_cache_new();
+        const handle =
+            maxEntries && maxEntries > 0
+                ? this.lib.core_provider_cache_new_with_eval_cache(
+                      maxEntries,
+                  )
+                : this.lib.core_provider_cache_new();
```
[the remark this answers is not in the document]
Datron: This is an issue when a `maxEntries` is beyond Number.MAX_SAFE_INTEGER (9007,199,254,740,991). That is a very large cache, I expect most users to stay below 1 trillion.

Datron:
>Can we add a test for proving that a 304/NotModified refresh preserves evaluation-cache entries ?

I think we shouldn't test functionality that doesn't have the code to execute said functionality

PR #1150, opened 2026-09-14
fix: add last selected context info in detailed resolve response

mahatoankitkumar:
## Problem

## Solution

## Pre-deployment activity

## Post-deployment activity

## API changes

## Summary by CodeRabbit

- **New Features**
  - Detailed resolved configurations now show the condition and context ID responsible for each overridden value.
  - Source details are displayed only when available, making configuration resolution easier to understand and troubleshoot.


Datron (approved):

crates/superposition_types/src/api/config.rs
```diff
@@ -62,6 +62,12 @@ pub struct DetailedResolvedConfigValue {
     pub description: String,
     pub schema: Option<Value>,
     pub value: Value,
+    /// Condition of the last matching context that set this key. `None` means the value
+    /// came from the default config.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub context: Option<Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub context_id: Option<String>,
 }
```
ayushjain17, line 71: I dont think both should be here, either should be there
which ever is more relevant should be kept
I feel ideally context_id should be the one to be kept

crates/superposition_types/src/api/config.rs
```diff
@@ -62,6 +62,12 @@ pub struct DetailedResolvedConfigValue {
     pub description: String,
     pub schema: Option<Value>,
     pub value: Value,
+    /// Condition of the last matching context that set this key. `None` means the value
+    /// came from the default config.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub context: Option<Value>,
```
ayushjain17: if we keep this, then use this, the type is context which is already a Map, lets not lose out on the type info
```suggestion
    pub context: Option<Condition>,
```

ayushjain17 (approved):

PR #1152, opened 2026-09-17
fix: support anyOf/type-less JSON schemas in key/config creation UI

knutties:
## Problem

Creating a Default Config key (the single key creation UI) was **impossible** when the type schema had no top-level `type` field — e.g. a JSON Schema combinator (`anyOf`/`oneOf`/`allOf`/`$ref`) or a bare `const`.

Example schema from the report:
```json
{ "anyOf": [ { "type": "string" }, { "const": "default_str_ignore_this" } ] }
```

`SchemaType::try_from` errored with **`type not defined in schema`**, and the Default Value input was rendered `disabled`, so no value could be entered and the key could not be saved. The same gap affected the context, override, and variant forms and the default-config detail page.

## Root cause

`SchemaType::try_from` (`crates/frontend/src/schema.rs`) hard-required a top-level `type`. That `SchemaType` is only used to choose which input widget to render — the value itself is validated against the full JSON Schema on the backend (`jsonschema`). So a missing top-level `type` should not be a hard failure.

## Fix

Add a `SchemaType::Any` variant, returned by `try_from` when the schema object has no `type`. It:
- maps to a free-form **Monaco/JSON editor** (`InputType::Monaco`)
- parses input as arbitrary JSON (`serde_json::from_str::<Value>`)

Genuinely malformed schemas (present-but-invalid `type`, e.g. `"type": "foobar"`) still error as before. Because the match arms on `SchemaType` are exhaustive/compiler-enforced, this fixes every form that consumes it in one place.

### Files
- `crates/frontend/src/schema.rs` — new `Any` variant; `try_from` returns `Ok(Any)` on missing `type`; `default_value` arm; unit tests
- `crates/frontend/src/components/input.rs` — `InputType::from` → Monaco for `Any`; `parse_input` parses free-form JSON; unit tests
- `crates/frontend/src/components/override_form.rs` — `TypeBadge` renders an `any` badge
- `crates/frontend/src/components/cohort_schema.rs` — `Any` excluded from string/number-array cohort filters

## Note on UX

With this schema the Default Value is now a free-form JSON editor: values are entered as JSON (a string must be quoted, e.g. `"default_str_ignore_this"`). Unquoted text shows a `not a valid JSON value` hint; the backend validates against the full schema on submit.

## Verification

- `cargo test -p frontend --lib` — 6 new unit tests pass (schema resolution, `Any`→Monaco mapping, JSON parsing, invalid-type still errors)
- `cargo check -p frontend --lib` on host **and** `--target wasm32-unknown-unknown` compile clean; `cargo clippy` clean
- **End-to-end in the running app:** an `anyOf` key now renders an editable JSON value editor (no error), and creation persists via `POST /default-config → 200`

## Summary by CodeRabbit

- **New Features**
  - Added support for schema fields without a top-level type, including composed, referenced, and constant-value schemas.
  - These fields now use a JSON editor and accept arbitrary valid JSON values.
  - Added an “any” type badge to clarify the expected input format.

- **Bug Fixes**
  - String and number array input options are no longer shown when they are incompatible with an unrestricted schema.
  - Invalid JSON values now display a clear validation message.


crates/frontend/src/schema.rs
```diff
@@ -83,6 +83,10 @@ impl JsonSchemaType {
 pub enum SchemaType {
     Multiple(Vec<JsonSchemaType>),
     Single(JsonSchemaType),
+    /// Schema has no resolvable top-level `type` (e.g. `anyOf`/`oneOf`/`allOf`/`$ref`).
+    /// The value is accepted as free-form JSON and validated against the full schema
+    /// on the backend.
```
ayushjain17: can we please remove the inline comments in both the places

crates/frontend/src/schema.rs
```diff
@@ -106,6 +110,7 @@ impl SchemaType {
             SchemaType::Single(JsonSchemaType::Null) => {
                 Value::String(String::from("null"))
             }
+            SchemaType::Any => Value::Null,
```
ayushjain17: did we test this ?
I think this does not work correctly for some reason
that's why we use
`Value::String(String::default())` in other places and `Value::String(String::from("null"))` for Null
so I think we should use either of these only over here as well

crates/frontend/src/schema.rs
knutties: Done — removed both the doc-comment on the `Any` variant and the comment in `try_from` (commit 968ac810).

crates/frontend/src/schema.rs
knutties: Good catch on the default. It does render/edit correctly (verified in the browser — the Default Value became an editable Monaco editor showing `null`), and the other Monaco cases actually use real JSON values (`Object -> {}`, `Array -> []`), so `Value::Null` wasn't broken per se. But you're right that it's a poor *default*: `null` fails validation for basically every real anyOf/oneOf/const schema, so the user hits a validation error unless they edit it. Changed it to `Value::String(String::default())` (commit 968ac810) — a valid editable default for the common `anyOf: [string, ...]` case and consistent with the `Multiple` convention. Skipped `"null"` since that seeds the literal string.

ayushjain17 (approved):

crates/frontend/src/schema.rs
```diff
@@ -106,6 +107,7 @@ impl SchemaType {
             SchemaType::Single(JsonSchemaType::Null) => {
                 Value::String(String::from("null"))
             }
+            SchemaType::Any => Value::String(String::default()),
```
sauraww, line 110: Any doesn't necessarily mean string is a valid type. For example, {"anyOf":[{"type":"number"},{"type":"boolean"}]} would resolve to Any, but "" would be an invalid default. Should Any avoid providing a default value altogether (e.g. Option<Value>) and let the Monaco input/user provide the value?

Not a blocker for this PR since backend schema validation still protects correctness. Good to merge; we can handle this as a follow-up.

sauraww (approved):

PR #1153, opened 2026-09-18
feat(frontend): link compare-page values to their source override (+ compare page polish)

knutties:
## What

Enhances the **Compare** page so each resolved value tells you where it came from, and cleans up the page's layout/form.

Builds on #1150, which added per-key `context`/`context_id` to the detailed resolve output.

### Override linking
- New `resolve_config_detailed` frontend API wrapper (`GET /config/resolve/detailed`).
- The compare page now resolves via the detailed endpoint and, for each value, stashes the contributing `context_id` + condition on the row.
- Overridden cells get a hover popover showing **the contributing override's own condition** (distinct from the column-header query) and a **"View override →"** link to `/overrides/{context_id}`. Default-sourced values render plain.
- `DetailedResolvedConfigValue` now derives `Deserialize` so the frontend can parse the response.

### Hardening (surfaced by the above)
- Text interpolation is **not** HTML-escaped in this SSR path, so config values that are themselves markup (e.g. an HTML banner string) injected raw DOM into the table and, when truncated mid-tag, swallowed adjacent cell content — including the new link. `url_or_string` now escapes HTML metacharacters and renders via `inner_html`, so such values display as inert literal text.

### Compare page polish
- The results table now claims the remaining viewport height (`flex-1`), and "Add Contexts" collapses by default when a comparison already exists, so the table is no longer starved of space / pushed low.
- "Add Context" on the compare form is now a subtle left-aligned button instead of a full-width slate panel (opt-in `compact` prop on `ContextForm`; other callers unaffected).
- Context-form hint relabeled "Query your configs" → "Resolve your configs" on the compare and resolve pages.

## Verification

Ran the app locally and drove the compare page in headless Chrome against a workspace with 2062 contexts / 535 defaults:
- All 34 overridden cells link to the correct override page (HTTP 200); default-sourced keys have no link.
- Popover shows the override's own condition (`shopId == 023456-2`), not the column query.
- HTML-valued config values now render as literal text (0 raw `<div>` in SSR, was injecting a live banner); row count restored to the full 535.
- Layout: table fills ~75% of viewport (was ~48%) and starts near the top when a comparison exists.
- Empty and populated compact-form states verified.

## Follow-up (not in this PR)

Text nodes not being HTML-escaped in the frontend's SSR path is likely a broader latent surface anywhere user/config strings are rendered as element text — worth an app-wide audit.

Screenshot of the hover:

Screenshot 2026-09-18 at 6 16 17 PM

## Summary by CodeRabbit

- **New Features**
  - Compare configuration values with visibility into which override set each value.
  - View override conditions and open the associated override directly from comparison results.
  - Added a compact context-entry layout for space-constrained views.

- **UI Improvements**
  - Comparison context controls now start collapsed when results are available.
  - Updated configuration resolution wording for clarity.

- **Bug Fixes**
  - Config values containing HTML characters are now safely displayed without being interpreted as markup.


crates/frontend/src/utils.rs
```diff
@@ -181,7 +181,17 @@ pub fn url_or_string(s: &str) -> impl IntoView {
...
+        // config value that is itself markup (e.g. an HTML banner string) would otherwise
+        // inject raw DOM into the page. Escape the HTML metacharacters ourselves and hand
+        // the pre-escaped string to `inner_html`, so the browser shows it as literal text.
+        Err(_) => {
+            let escaped = s
+                .replace('&', "&amp;")
+                .replace('<', "&lt;")
+                .replace('>', "&gt;");
+            view! { <span inner_html=escaped></span> }.into_view()
+        }
```
ayushjain17, line 194: using something like this
force people to not be able to copy their original data back - this has been modified
rather a different view needs to be used where I can paste the content as is
if, inner_html is used without escapes does that still not work ?

knutties:
Screenshot 2026-09-21 at 10 47 02

@ayushjain17 @Datron - please take a look at the updated screenshot.  Includes feedback from @ayushjain17 on Slack and also some colouring to show we have overridden the value and it is not falling back for a visual cue.

crates/frontend/src/utils.rs
knutties: Good to dig into this — I checked, and copy fidelity is actually preserved. The escaping only changes the HTML *source*; the browser decodes the entities back on render, so the DOM text node (and therefore anything you select/copy) is the original value. I verified `textContent` on an HTML-valued cell equals the original `<div style='…'>` string, and no live element is injected.

A couple of related notes:

- This cell is **display-only** — there's no input to paste into on the compare page — so the "paste the content as is" case doesn't really arise on this surface. The only limitation is the pre-existing 100-char truncation, which the "…more" toggle expands.
- On "`inner_html` without escapes": that's exactly what reintroduces the problem. The raw value renders as **live DOM** (e.g. an actual full-width banner), and when the cell text is truncated mid-tag it swallows the following markup — the override link — and drops rows. It's also a stored-XSS vector. So the escaping is what keeps it safe.

Root cause is that leptos text interpolations aren't HTML-escaped in this SSR path (attributes are). I filed #1154 to fix that at the source, after which we can render values as normal escaped text and drop this `inner_html` workaround entirely.

